PR #8, opened 2018-07-27
Refinements to Mine Record

garywong-bc:
Incorporating feedback from [Kristin](https://bcmines.atlassian.net/browse/MDS-118?focusedCommentId=10030&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-10030)

Barring any major errors, let's treat this as good-enough-for-Sprint2018.2 ; we want a stable DB structure for the devs/testers to work from.

### DIFF
--- a/migrations/0001.mine-record.sql
+++ b/migrations/0001.mine-record.sql
@@ -1,4 +1,4 @@
-CREATE TABLE mine_identifier (
+CREATE TABLE mine_identity (
   mine_guid        uuid PRIMARY KEY DEFAULT gen_random_uuid(),
   create_user      character varying(60) NOT NULL DEFAULT 'DUMMY_USER',
   create_timestamp timestamp with time zone NOT NULL DEFAULT current_timestamp,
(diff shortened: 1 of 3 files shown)


migrations/0001.mine-record.sql
```diff
@@ -1,27 +1,33 @@
...
-  mine_guid uuid,
+CREATE TABLE mine_detail (
+  mine_detail_guid uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+  mine_guid uuid NOT NULL,
   mine_no   character varying(010) NOT NULL,
   mine_name character varying(100) NOT NULL,
+
+  effective_date date NOT NULL DEFAULT now(),
+  expiry_date    date     NULL DEFAULT NULL,
   create_user      character varying(60) NOT NULL DEFAULT 'DUMMY_USER',
```
samfresh: I know its not part of this PR, but i don't think we want a default 'DUMMY_USER' here? we should just leave it as NOT NULL, i will handle defaults for tests and such at the API level.

samfresh: @garywong-bc

PR #42, opened 2018-08-15
Feature/mds 34 frontend Access Control

FreshGyan:
Hides edit controls from users with no edit access


frontend/src/components/mine/CreateMineForm.js
```diff
@@ -40,32 +43,42 @@ export class CreateMineForm extends Component {
...
+                <Input type="text" ref={ref => this.mineName = ref} placeholder="Mine Name"></Input>
+                <Button type="primary" htmlType="submit">
+                  Create Mine
+                </Button>
+              </FormItem>
+            </Form>
+          </Card>
+        </div>
+      );
+    } else {
```
mmumby: can this work without the else?

FreshGyan: Seems to work! I'll remove the else 💃

PR #96, opened 2018-09-07
Feature/backend refactor model route

samfresh:
1. Added Base for models
2. Reduced items returned for maps


python-backend/app/__init__.py
```diff
@@ -106,14 +106,21 @@ def create_data(num):
         db.session.bulk_save_objects(mine_identity_list)
         db.session.bulk_save_objects(mine_detail_list)
         db.session.bulk_save_objects(mine_location_list)
-        db.session.commit()
-
-        click.echo(f'Created {num} random mines.')
+        try:
+            db.session.commit()
+            click.echo(f'Created {num} random mines.')
+        except:
```
countable: Since you're using this pattern a lot, it's probably worth specifying the exception type or re-raising it to prevent it being suppressed, or over time you may accumulate and have to handle unknown errors without being able to see them. These won't happen often, but when they do they'll likely go unresolved for a long time since they're not diagnosable due to the suppressed exception. Naked except statements are a pretty well accepted anti-pattern which will make debugging more difficult over time. See: https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements

You probably want this one http://docs.sqlalchemy.org/en/latest/core/exceptions.html#sqlalchemy.exc.DBAPIError

ie ```
from sqlalchemy.exc import DBAPIError
import sys

try:
    db.session.commit()
except DBAPIError:
    db.session.rollback()
    logger.log(sys.exc_info())
```

However, I do note that the official SQLAlchemy docs do have naked except blocks such as here:

http://docs.sqlalchemy.org/en/latest/orm/session_transaction.html

with the rationale that in the case of ANY failure you want to roll back, in which case, reraise.
```
try:
    db.session.commit()
except: # noqa
    db.session.rollback()
    raise
```

To keep things DRY you may also want to centralize this behaviour and save time :)

```

def _commit():
    try:
        db.session.commit()
    except:
        db.session.rollback()
        raise
```

Or, better yet, move this to your Base model class since it's the same for all children.

PR #106, opened 2018-09-13
Fix/local-postgres-build

samfresh:
Fix for local postgres image, had the wrong filename specified in Dockerfile.dev. Also added missing revert file for mine_location. @FreshGyan I skipped the second afterMigration file to grant user access, can you confirm that is not required for local env?


FreshGyan:
@samfresh I updated the dockerfile to support multiple aftermigrate files. Just so in future if we add more files, we don't have to keep adding them to the dockerfile.dev 💯

PR #131, opened 2018-09-27
Feature/mds 343

samfresh:
Added SQL for permit and permit status code.
Added AfterMigration script to create permit status code.
Created Permit Resource and Models
Created Permit Status Code Models
Created appropriate tests for permit and permit status code.
Updated create data to create permits for mines and also tenure for mines on a random basis between (1-5)  for permits and (1-3) for tenures


migrations/sql/afterMigrate__3__permit_status_code.sql
```diff
@@ -0,0 +1,13 @@
+INSERT INTO permit_status_code (
```
FreshGyan: Can you make this as INSERT INTO ..... ON Conflict DO NOTHING kind of syntax.
Coz the afterMigrate scrips run after each deploy. So the second time they run, it'll throw a Primary Key unique constraint error.
More deets here on implementation:
https://stackoverflow.com/questions/17267417/how-to-upsert-merge-insert-on-duplicate-update-in-postgresql

PR #180, opened 2018-10-22
Changed api routes to plurals

samfresh:
This was a change previously discussed where api routes will
take its plural form and not split into singular and plural.
It will make the api routes more consistent. The frontend constants
file was changed to follow the changes.


python-backend/app/api/location/resources/location.py
```diff
@@ -7,14 +7,11 @@
...
-class MineLocationListResource(Resource):
-    @jwt.requires_roles(["mds-mine-view"])
-    def get(self):
-        return {'mines': list(map(lambda x: x.json_by_location(), MineIdentity.query.all()))}
+    def get(self, mine_location_guid=None):
+        if mine_location_guid:
+            mine_location = MineLocation.find_by_mine_location_guid(mine_location_guid)
+            if mine_location:
+                return mine_location.json()
+            return {'message': 'Mine Location not found'}, 404
```
FreshGyan: Can you add the errorMixin to this? Just to keep the error reporting on endpoints consistent 👍

PR #183, opened 2018-10-23
Add tabnames to url route support

FreshGyan:


frontend/src/components/mine/MineDashboard.js
```diff
@@ -55,7 +66,9 @@ export class MineDashboard extends Component {
             </div>
             <div className="dashboard__content">
               <Tabs
-                defaultActiveKey='#summary'
+                activeKey={this.state.activeTab}
+                defaultActiveKey="#summary"
```
mmumby: if you aren't using the # in the url I think you can rename the keys to be the same as the url so `#summary` should be `summary`

PR #184, opened 2018-10-23
Added a new landing page for Unauthorized visitors

JustinMacaulay:
added logic to AuthGuard and NullScreens to display a new page if the user is unauthorized.


frontend/src/HOC/AuthGuard.js
```diff
@@ -49,11 +50,15 @@ export const AuthGuard = (WrappedComponent) => {
 
     render() {
       if (this.props.keycloak) {
-        if (this.props.isAuthenticated) {
+        if (this.props.isAuthenticated && this.props.userAccessData.includes("mds-mine-view")) {
```
mmumby: instead of "mds-mine-view" 

use: 
```
import { USER_ROLES } from '@/constants/environment';
this.props.userAccessData.includes(USER_ROLES.role_view)
```

PR #188, opened 2018-10-25
Added a new view mine manager history

JustinMacaulay:
Added a new table to display the mine manager history for a mine.


frontend/src/components/mine/ContactInfo/MineContactInfo.js
```diff
@@ -90,6 +90,12 @@ const defaultProps = {
     });
   }
 
+  toggleMineManagerHistory = () => {
```
mmumby: this logic would be more appropriate in the ViewMineManager component, the parent (MineContactInfo) doesn't need to know if the history is viable or not and once the permittee history logic is added MineContactInfo will get a little cluttered, I would suggest moving:

mineManagerHistoryVisible
tiggleMineManagerHistory()
into ViewMineManager.js

and at this point, mineManagerHistoryVisible can be renamed to isHistoryVisible or something shorter, since it is located in the appropiate file you will no longer need to specify if it is the mineManager or the permittee.

frontend/src/constants/assets.js
```diff
@@ -11,4 +11,5 @@ export { default as TENURE } from '../assets/images/Tenure.svg';
 export { default as EMAIL } from '../assets/images/Email.svg';
 export { default as PHONE } from '../assets/images/Phone.svg';
 export { default as PERMIT } from '../assets/images/Permit.svg';
-export { default as PENCIL } from '../assets/images/Pencil.svg';
\ No newline at end of file
+export { default as PENCIL } from '../assets/images/Pencil.svg';
+export { default as GROUP_MINERS } from '../assets/images/groupofminers.svg';
```
mmumby: rename this SVG to follow naming conventions, I would suggest GroupOfMiners.svg

PR #191, opened 2018-10-25
MDS-475 feature/search by permit

mmumby:


frontend/src/tests/components/dashboard/MineSearch.spec.js
```diff
@@ -4,24 +4,27 @@ import {MineSearch} from '@/components/dashboard/MineSearch';
...
   dispatchProps.fetchMineNameList = jest.fn();
+  dispatchProps.handleMineSearch = jest.fn();
+  dispatchProps.handleCoordinateSearch = jest.fn()
 };
 
-const setupReducerProps = () => {
-  reducerProps.mineNameList = MOCK.MINE_NAME_LIST;
+const setupProps = () => {
+  props.mineNameList = MOCK.MINE_NAME_LIST;
+  props.isMapViuew = false;
```
JustinMacaulay: I think this is just a typo and not sure if it will affect the test but should be props.isMapView

mmumby: good catch!

PR #213, opened 2018-11-08
Mds 535

FreshGyan:
- Enable Automated Functional Testing in pipeline


functional-tests/src/test/groovy/pages/Dashboard.groovy
```diff
@@ -3,38 +3,38 @@ package pages
 import geb.Page
 import modules.*
 
-class DashboardPage extends Page { 
-    static at = { waitFor(10,1) {!loadingScreen.displayed}}
```
VivianCheungFW: Might want to keep the 10 sec wait, as the default only wait for 5 seconds, and for the Dashboard, sometimes it took longer than that, if not specify here, the test might fail. In the future we might improve performance, then we can change this.

functional-tests/src/test/groovy/pages/MineProfile.groovy
```diff
@@ -5,25 +5,24 @@ import modules.*
 import utils.Const
 
 class MineProfilePage extends Page {
-    static at = { waitFor(10,2) {activeTab == "Summary" }  }
```
VivianCheungFW: Same for this one;

PR #214, opened 2018-11-08
MDS-488 feature/updated create/edit mine

mmumby:


frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -107,3 +107,19 @@ export const fetchStatusOptions = () => (dispatch) => {
       dispatch(hideLoading('modal'));
     });
 };
+
+export const fetchRegionOptions = () => (dispatch) => {
+  dispatch(request(reducerTypes.GET_REGION_OPTIONS));
+  dispatch(showLoading('modal'));
+  return axios.get(ENVIRONMENT.apiUrl + API.MINE_REGION, createRequestHeader())
```
NWCalvank: Thoughts on string interpolation in this codebase? I don't think i've seen any instances of it being used.

This line is likely nicer visually this way, but I'm curious about the concatenation vs interpolation pattern across the codebase.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -107,3 +107,19 @@ export const fetchStatusOptions = () => (dispatch) => {
...
+    .then((response) => {
+      dispatch(success(reducerTypes.GET_REGION_OPTIONS));
+      dispatch(mineActions.storeRegionOptions(response.data));
+      dispatch(hideLoading('modal'));
+    })
+    .catch((err) => {
```
NWCalvank: Thoughts on destructuring out `response` here?

frontend/src/components/Forms/UpdatePermitteeForm.js
```diff
@@ -30,7 +28,7 @@ const permitteeOptions = (permit) => {
   permit.map((obj) => {
     const data = {
       value: obj.permittee[0].permittee_guid.concat(', ', obj.permit_guid),
-      option: obj.permittee[0].party.name.concat(", ", obj.permit_no)
+      label: obj.permittee[0].party.name.concat(", ", obj.permit_no)
```
NWCalvank: Destructure `obj` and set label to `${permittee.party.name},${permit_no}`

Thoughts?

Also, do we need to validate that `.permit_no` or `.name` will be present before accessing them, or are they always required?

frontend/src/components/common/RenderCheckbox.js
```diff
@@ -0,0 +1,36 @@
...
+
+const RenderCheckbox = ({
+  id,
+  input,
+  label,
+  meta: { touched, error},
+}) => (
+  <Form.Item 
+  validateStatus={(touched ? (error && 'error') : '')}
+>
```
NWCalvank: Looks like this should be vertically aligned

frontend/src/components/common/RenderCheckbox.js
```diff
@@ -0,0 +1,36 @@
...
+  <Checkbox 
+    id={id} 
+    checked={input.value || false}
```
NWCalvank: Is the `|| false` necessary here? My impression would be that `input.value` gets coerced?

But if it doesn't, how about `checked={Boolean(input.value)}` to be a bit more explicit about the issue?

frontend/src/components/dashboard/Dashboard.js
```diff
@@ -253,14 +254,15 @@ export class Dashboard extends Component {
   }
 
   render() {
+    console.log(this.props.mineRegionOptions);
```
NWCalvank: Remove console.log

frontend/src/components/mine/MineHeader.js
```diff
@@ -47,16 +48,18 @@ class MineHeader extends Component {
...
   
-  openModal(event, mineStatusOptions, onSubmit, title, mine) {
+  openModal(event, mineStatusOptions, mineRegionOptions, onSubmit, title, mine) {
     event.preventDefault();
     const initialValues = {
       "name": mine.mine_detail[0] ? mine.mine_detail[0].mine_name : null,
       "latitude": mine.mine_location[0] ? mine.mine_location[0].latitude : null,
       "longitude": mine.mine_location[0] ? mine.mine_location[0].longitude : null,
       "mine_status": mine.mine_status[0] ? mine.mine_status[0].status_values : null,
+      "major": mine.mine_detail[0] ? mine.mine_detail[0].major : false,
```
NWCalvank: Looks like we're inconsistent around single vs double-quotes. Do you have a preference?

frontend/src/components/mine/MineHeader.js
```diff
@@ -84,10 +87,12 @@ class MineHeader extends Component {
               isDropdown 
               overlay={menu} 
               string={<Icon type="ellipsis" theme="outlined" style={{fontSize: '30px'}}/>}
-            />
+              />
```
NWCalvank: This looks like an indentation error

frontend/src/tests/components/common/RenderCheckbox.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+  };
+}
+
+beforeEach(() => {
+  setupProps();
+});
+
+describe('RenderCheckbox', () => {
+
```
NWCalvank: The other spec file doesn't have newline padding around the `it` blocks. Which would you prefer? If we are using the Airbnb eslint config, they disallow newline padding.

https://github.com/airbnb/javascript#whitespace--padded-blocks

frontend/src/tests/mocks/dataMocks.js
```diff
@@ -277,6 +291,29 @@ export const STATUS_OPTIONS = {
   ]
 };
 
+export const REGION_OPTIONS = { 
+  'options': [{
```
NWCalvank: Any reason for using single quotes on these keys? They shouldn't be needed, and I don't think we're generally adding them in other files

frontend/src/components/common/RenderCheckbox.js
mmumby: Good catch, alignment was very off.

frontend/src/components/common/RenderCheckbox.js
mmumby: false is not needed.

frontend/src/components/dashboard/Dashboard.js
mmumby: went through and removed them all this am.

frontend/src/tests/components/common/RenderCheckbox.spec.js
mmumby: no newline padding, updated spec files that have it.

frontend/src/tests/mocks/dataMocks.js
mmumby: single quotes are removed, these are my mock objects, Usually copy and pasted from somewhere, will pay more attention to consistency!!

mmumby: also converted all the values to double quotes

frontend/src/components/mine/MineHeader.js
mmumby: fixed.

frontend/src/components/mine/MineHeader.js
mmumby: double quotes are preferred throughout.

Single quotes on all imports,

Will work towards removing these inconsistencies.

PR #220, opened 2018-11-13
MDS-481 bug/fixed forms from showing initialValues after submission

mmumby:


frontend/src/reducers/modalReducer.js
```diff
@@ -11,17 +11,19 @@ const initialState = {
   isModalOpen: false,
   props: {},
   content: null,
-}
+  clearOnSubmit: true,
```
NWCalvank: I think I'm changing my mind on this.

I'm noticing now that `initialState` is being set as the default for `state`. Since `clearOnSubmit` is being extracted from `action.payload` instead of from `state`, it doesn't really make sense for the default to be set in `initialState` after all.

My original thinking was that this would be more self-documenting, but I don't think it is actually, for the reason described above. Probably better off just assigning it to `true` instead of `initialState.clearOnSubmit` on line 20

Thoughts?

NWCalvank: Discussed IRL.

PR #245, opened 2018-11-22
mds-517 feature/tailings-and-expected-document-tests

Jsyro:


python-backend/tests/tailings/resources/test_mine_tailings_resource.py
```diff
@@ -0,0 +1,45 @@
...
+    assert get_resp.status_code == 200
+    org_mine_tsf_list_len = len(get_data['mine_storage_tailings_facilities'])
+
+    post_resp = test_client.post('/mines/tailings', data={'mine_guid':TEST_MINE_GUID, 'tsf_name':TEST_TAILINGS_STORAGE_FACILITY_NAME2}, headers=auth_headers['full_auth_header'])
+    post_data = json.loads(post_resp.data.decode())
+    assert post_resp.status_code == 200
+    assert len(post_data['mine_tailings_storage_facilities']) == (org_mine_tsf_list_len +  1)
+    assert all(tsf['mine_guid'] == TEST_MINE_GUID for tsf in post_data['mine_tailings_storage_facilities'])
+
+### #appears that these tests effect each other, insert from above persists when this test runs org_exp_document_list_len is 2, only 1 inserted on setup
```
NWCalvank: Is there a proposed solution here?

python-backend/app/api/documents/mines/resources/expected_documents.py
```diff
@@ -36,7 +36,7 @@ def post(self, mine_guid):
             mine_exp_doc = MineExpectedDocument(
                 req_document_guid=new_doc['req_document_guid'],
                 exp_document_name=new_doc['document_name'],
-                exp_document_description=new_doc['document_description'],
+                exp_document_description=new_doc.get('document_description'),
```
FreshGyan: I'm assuming this is here because Document description is optional and you want a default.
From my understanding, .get() if it doesnt find anything returns NULL. I dont know what the requirement is, but you may want an empty string or something else as default. I'll leave that decision to you.

Also if you do prefer the .get() convention over direct accessors like ['key']. Then maybe we can have this as something to change as you go. I think we should change all our accessors to .get() as we shouldnt break the app everytime the dictionary doesnt have the required value. So if you can update that as you go, that'd be 💯  Starting possibly with this method change all accessors from ['key'] to .get('key', 'default value')

python-backend/app/api/mines/tailings/resources/tailings.py
```diff
@@ -47,9 +49,11 @@ def post(self, mine_tailings_storage_facility_guid=None):
 
             if is_mine_first_tsf:  
                 try: 
-                    tsf_required_documents = requests.get(current_app.config['DOCUMENT_MS_URL'] + url_for('documents_required_document_resource') + '?category=MINE_TAILINGS', 
+                    get_resp = requests.get(current_app.config['DOCUMENT_MS_URL'] + current_app.config['BASE_PATH'] + '/documents/required' + '?category=MINE_TAILINGS',
```
FreshGyan: If we're going to end up doing this all over the place, maybe move this to a constant or env file somewhere?

python-backend/app/api/documents/mines/resources/expected_documents.py
Jsyro: I'm happy with null, also, the json str() casts the values when it sends it back from the api, so that would convert it to '' for the front end. 

exactly, .get is used because it's an optional field. are you thinking we should .get everywhere for consistency then validate fields that shouldn't be null with validators?

python-backend/app/api/mines/tailings/resources/tailings.py
Jsyro: This is the only place doing this kind of url construction, and it's still very WIP, when we find a pattern we can reliably repeat, definitely agree to pull it out to some kind of service/constants.

python-backend/app/api/documents/mines/resources/expected_documents.py
FreshGyan: I think we should change it everywhere so that if someone makes a call to the API using postman or whatever and doesn't put in one of the required fields. We can return an error we control, as opposed to the stack trace.

python-backend/app/api/mines/tailings/resources/tailings.py
FreshGyan: For the time being, can you move it to the Top of the class/file just for cleanliness?

python-backend/tests/tailings/resources/test_mine_tailings_resource.py
FreshGyan: Is this blocking other people from getting stuff done? If there are dependencies for frontend and such, we can merge this in as long as we address these comments in the next PR or so before the end of sprint.

python-backend/app/api/documents/mines/resources/expected_documents.py
Jsyro: Ok, let's make this a backlog item and try and get to it this sprint, Not in scope for this ticket? 

Does that work for you?

python-backend/tests/tailings/resources/test_mine_tailings_resource.py
Jsyro: no, this is a testing issue only. running the application normally doesn't cause any issues. this is mostly a werkzurg issue in how it handles/create api requests..

PR #260, opened 2018-11-23
mds-666 bug/added date validation to mine manager assignment

james-hollinger:
Date validation has been added to prevent adding a new mine manger who's start date is on or before the previous manager's start date.

Additionally, some refactoring to the party API endpoint code has been done. Reviewers, if you could look at lines 176 - 184 in particular. These lines seem to do nothing and so I have removed them. Perhaps they were a copy-paste error or an afterthought.


python-backend/app/api/parties/party/resources/party.py
```diff
@@ -141,28 +141,34 @@ def get(self, mgr_appointment_guid):
...
-            previous_mgr = MgrAppointment.find_by_mgr_appointment_guid(previous_mgr_info.mgr_appointment_guid)
+
+        # If there is a previous manager, set its expiry to the day before the new manager's start date
+        if mine.mgr_appointment:
+            previous_appointment_guid = mine.mgr_appointment[0].mgr_appointment_guid
+            previous_mgr = MgrAppointment.find_by_mgr_appointment_guid(previous_appointment_guid)
+
+            previous_mgr_expiry_date = (data['effective_date'] - timedelta(days=1)).date()
+            if previous_mgr_expiry_date < previous_mgr.effective_date:
+               self.raise_error(400, "Error: New manager's start date is on or before the previous manager's start date.")
```
NWCalvank: So, suppose that I add a manager to a mine, and set their start date to today. But they actually started yesterday.

If I try to correct that manager's start date, I'll get this error message. Is that what we want?

NWCalvank: But an implication of this fix for the history is that the current manager's start date can never be changed to an earlier date. So, if it is entered incorrectly once, it's permanently wrong.

This, to me, seems like trading one bug based on incorrect data input for another bug based on incorrect data input.

nayr974: Nathan, you're totally right. Those issues are tracked seperately by MDS-667 which would include more complex workflow changes. This item is a quick bug fix for the current workflow, but yes the current workflow has a poor user experience in that it doesn't handle erroneous input or adding historical data.

james-hollinger:
Added front-end validation

frontend/src/components/mine/ContactInfo/ViewMineManager.js
```diff
@@ -50,8 +50,21 @@ export class ViewMineManager extends Component {
...
   openModal(event, onSubmit, handleChange, handlePartySubmit, title) {
     event.preventDefault();
+
+    let mineManager = null;
+    let startDate = null;
+    let appointment = this.props.mine.mgr_appointment;
+    if (appointment != null && appointment.length > 0) {
+      mineManager = appointment[0].party_guid;
+      startDate = appointment[0].effective_date;
+    }
```
NWCalvank: It would be preferable and more consistent with our existing code to use `const`. You could do it something like this:
```
const hasAppointment = appointment != null && appointment.length > 0;
const mineManager = hasAppointment ? appointment[0].party_guid : null;
const startDate = hasAppointment ? appointment[0].effective_date : null;
```

frontend/src/components/Forms/UpdateMineManagerForm.js
```diff
@@ -14,13 +15,18 @@ const propTypes = {
...
+  initialValues: PropTypes.object,
 };
 
 const defaultProps = {
   parties: {},
   partyIds: [],
 };
 
+const validateStartDate = memoize(previousStartDate => value => 
+  value <= previousStartDate ? "New manager's start date cannot be on or before the previous manager's start date." : undefined
```
NWCalvank: @mmumby would `validateStartDate` be better off in a helper/util file?

Personal opinion here: if the `value` just needs to be an error string or a falsey value, I think `false` would make a bit more sense than `undefined`. The reason being that the error "check" would return false. But Meredith may be able to speak to an existing pattern here as well.

python-backend/tests/party/resources/test_mgr_appointment_resource.py
```diff
@@ -113,3 +113,28 @@ def test_post_manager_success(test_client, auth_headers):
     assert post_data['party_guid'] == test_manager_data['party_guid']
     assert post_data['mine_guid'] == test_manager_data['mine_guid']
     assert post_resp.status_code == 200
+
+
+def test_post_manager_invalid_date(test_client, auth_headers):
+    # TODO Currently, tests share the same database between test cases in modules. 
+    # This test assumes that it runs directly after test_post_manager_success, so the newest mine manager's start date is today.
```
NWCalvank: I haven't tried this yet with pytest, but could we POST twice in the same test to make it independent?

POST a valid date, check that it returns a valid response.. And then POST the invalid date based on the first POSTed date, check that it returns an invalid response?

Just spit-balling solutions here, since I'm not sure on the scope of getting these cleaning up the db properly between test cases.

frontend/src/components/Forms/UpdateMineManagerForm.js
mmumby: `validateStartDate` should be in kept in utils/Validate.js as It can be reused in the future.

NWCalvank: Thanks, Meredith. @james-hollinger having looked at the file Meredith mentioned, the `undefined` is in line with the existing validation functions. So, I revoke my suggestion to change it.

mmumby: Also, please update the tests in Validate.spec.js to reflect this new validate function. You should be able to follow the syntax of the other tests, If you need any help, let me know!

python-backend/tests/party/resources/test_mgr_appointment_resource.py
james-hollinger: Unfortunately, posting twice wouldn't make the tests independent. The first post is still dependent on the previous test and we wouldn't know beforehand if its valid. 

The real solution is that these tests should all run in isolation, ideally as unit tests with a mock database.

frontend/src/components/Forms/UpdateMineManagerForm.js
NWCalvank: That sounds like the best option to me, @james-hollinger

james-hollinger: Do you have any suggestions that are both clear and concise?

NWCalvank: We won't know where/when we'll want to reuse this validation, so I'm not sure choosing a general + helpful message is practical yet. The benefit now is keeping the validation functions all in the same spot.

So, I think the message is fine as-is for the time being. Moving the function over should be enough

frontend/src/components/mine/ContactInfo/ViewMineManager.js
james-hollinger: Sounds good, I have made this change.

frontend/src/components/Forms/UpdateMineManagerForm.js
james-hollinger: Added front-end unit tests

PR #263, opened 2018-11-26
MDS-786

itchison:
This PR supports MDS-786. This is the table structure for the document management feature.


migrations/sql/V1.7__Add_Documents.sql
```diff
@@ -0,0 +1,21 @@
...
+	create_user varchar(60) NOT NULL,
+	create_timestamp timestamptz NOT NULL DEFAULT now(),
+	update_user varchar(60) NOT NULL,
+	update_timestamp timestamptz NOT NULL DEFAULT now(),
+	CONSTRAINT document_manager_pk PRIMARY KEY (document_manager_id)
+);
+
+-- Permissions
+
+ALTER TABLE public.document_manager OWNER TO mds;
```
FreshGyan: We have an aftermigrate script that grants all permissions to the mds user, so feel free to remove this statement :)

PR #268, opened 2018-11-28
Fix/add permit info

mmumby:


frontend/src/tests/components/dashboard/Dashboard.spec.js
```diff
@@ -1,10 +1,12 @@
...
+import React from "react";
+import { shallow } from "enzyme";
+import { Dashboard } from "@/components/dashboard/Dashboard";
+import * as MOCK from "@/tests/mocks/dataMocks";
+import * as router from "@/constants/routes";
+import * as String from "@/constants/strings";
 
 const dispatchProps = {};
-const reducerProps = {};
+let reducerProps = {};
```
NWCalvank: Unless we're re-assigning `reducerProps` then we can still use `const`. Perhaps you're using this to signal to the reader that `reducerProps` will be mutated?

PR #271, opened 2018-11-29
MDS-855 feature/upload documents api

JustinMacaulay:
feature to allow document uploads to the system and the tests


python-backend/app/api/document_manager/models/document_manager.py
```diff
@@ -0,0 +1,30 @@
...
+    __tablename__ = 'document_manager'
+    document_manager_id = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
+    document_guid = db.Column(UUID(as_uuid=True), nullable=False)
+    full_storage_path = db.Column(db.String(150), nullable=True)
+    upload_date = db.Column(db.DateTime, nullable=False)
+    file_display_name = db.Column(db.String(40), nullable=False)
+    path_display_name = db.Column(db.String(150), nullable=True)
+
+    def __repr__(self):
+        return '<Document_manager %r>' % self.party_guid
```
NWCalvank: These are PascalCase elsewhere (DocumentManager)

python-backend/app/api/document_manager/models/document_manager.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def json(self):
+        return {
+            'document_manager_id' : str(self.document_manager_id),
```
NWCalvank: I don't believe we're putting a space before the colon in other dictionaries

Also, if this is just an integer then we shouldn't need to coerce it to a string to be valid JSON. Is there a reason why it should be a string?

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class DocumentManagerResource(Resource, UserMixin, ErrorMixin):
+    parser = reqparse.RequestParser()
+    
+    @jwt.requires_roles(["mds-mine-create"])
+    def post(self):
+
+        self.parser.add_argument('file', type=FileStorage, location='files', action='append')
+        self.parser.add_argument('folder', type=str, help='mine to the new document on')
```
NWCalvank: I don't understand this help message. Perhaps it makes sense in the context it would be shown in?

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -0,0 +1,85 @@
...
+                pretty_path = re.sub(r'\b'+str(file_guid)+r'\b', original_file_name, pretty_path)
+                
+
+                document_info = DocumentManager( 
+                    document_guid = file_guid,
+                    full_storage_path = real_path,
+                    upload_date = datetime.now(),
+                    file_display_name = (original_file_name + file_extension),
+                    path_display_name = pretty_path,
+                    **self.get_create_update_dict(),
```
NWCalvank: Has there been any discussion around trailing commas in the python-backend section of the codebase? I'm personally in-favour of them, but I notice this is pretty inconsistent, so I haven't been using them

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -0,0 +1,85 @@
...
+                original_file_name, file_extension = os.path.splitext(upload.filename)
+
+                upload_response = documents.save(upload, folder, (str(file_guid) + file_extension))
+                real_path = documents.path(upload_response)
+
+                #create the readable path by removing the guids and replacing them with the original file name and mine no
+                pretty_path = re.sub(r'\b'+folder+r'\b', pretty_folder, real_path)
+                pretty_path = re.sub(r'\b'+str(file_guid)+r'\b', original_file_name, pretty_path)
+                
+
```
NWCalvank: Probably keep the double newlines for separating methods? Instead of using them inside of a method. Personal opinion

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        'message': 'An unexpected error has occured: ' + str(e),
+                    }
+                )
+                continue
+
+        return {
+            'document_manager_guids' : document_guid_list,
+            'errors': errors,
+        }
```
NWCalvank: This is a really long method. I personally feel like many of our existing python methods are unnecessarily long.

Thoughts on breaking some of these pieces out? I'd be concerned with this method growing over time, as it is already around 60 lines long

python-backend/app/config.py
```diff
@@ -34,7 +34,12 @@ class Config(object):
     SQLALCHEMY_MAX_OVERFLOW = 20
     SQLALCHEMY_POOL_TIMEOUT = 300
 
-
+    #Flask-uploads configs
+    UPLOADED_DOCUMENT_DEST = '/app/document_uploads'
+    DOCUMENT_FILE_SET = set(['pdf', 'doc', 'docx'])
+    
+    #20MB file limit
+    FILE_BYTE_LIMIT = 20 * 1024 * 1024
```
NWCalvank: Is there a reason to save it as bytes? It looks like you save it as bytes here, just to convert it to MB later anyway

PR #272, opened 2018-11-29
MDS-13 Feature/ Add Mine Type + Mine Tenure Types

NWCalvank:
Adds:
- Ability to POST/PUT and view the Mine Tenure Type for a mine
- Small README update

Current limitations:
- Only supports 1 or 0 Mine Types per mine


python-backend/app/api/mines/mine/models/mine_detail.py
```diff
@@ -5,6 +5,7 @@
 from sqlalchemy.dialects.postgresql import UUID
 from ....utils.models_mixins import AuditMixin, Base
 from ...region.models.region import MineRegionCode
+from .mine_type import MineType
```
FreshGyan: Is this an unused import?

migrations/sql/V1.9__Add_Mine_Type_and_Mine_Tenure_Type.sql
```diff
@@ -0,0 +1,30 @@
+CREATE TABLE mine_tenure_type (
+  mine_tenure_type_id   smallint PRIMARY KEY NOT NULL,
```
FreshGyan: Can you add indexes to the foreign/join columns. AND/OR to the columns you know there will be a lot of querying/sorting on

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -132,6 +136,8 @@ def post(self, mine_no_or_guid=None):
             self.raise_error(400, 'Error: {}'.format(e))
         mine_identity.save()
         mine_detail.save()
+        if mine_tenure_type_id:
+            self._save_mine_type(mine_identity, mine_tenure_type_id)
```
FreshGyan: Is this the same behaviour as MineType(model).create_mine_type? If so, can we remove this internal method?
If not, can we move the method to the model? Just so the resource doesn't have to worry about how the data actually is created?

PR #275, opened 2018-11-30
MDS-855 feature/upload documents api

JustinMacaulay:
added a resource and model to allow for file uploads and the management of those files.


openshift/sonar.pod.json
```diff
@@ -36,7 +36,12 @@
             "name": "GIT_BRANCH",
             "description": "Current git branch name",
             "required": true
-        }
+        },
+        {
+			"name": "UPLOADED_DOCUMENT_DEST",
```
NWCalvank: Indentation appears to have gone a bit wonky here

python-backend/app/__init__.py
```diff
@@ -25,10 +26,14 @@ def create_app(test_config=None):
     else:
         # load the test config if passed in
         app.config.from_object(test_config)
+    
+    configure_uploads(app, documents)
```
NWCalvank: Maybe it's a mixed tabs/spaces issue? Things in this block aren't lined up with the existing code

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        'message': f'The type of the file: {original_file_name + file_extension} is not allowed.',
+                    }
+                )
+                continue
+
+            except Exception as e:
+                errors.append(
+                    {
+                        'message': 'An unexpected error has occured: ' + str(e),
```
NWCalvank: I think this will just dump the raw error into the message. Just checking that you've confirmed this doesn't spit a gross error to the client in the toast message or anything

python-backend/app/config.py
```diff
@@ -34,7 +34,10 @@ class Config(object):
     SQLALCHEMY_MAX_OVERFLOW = 20
     SQLALCHEMY_POOL_TIMEOUT = 300
 
-
+    #Flask-uploads configs
+    UPLOADED_DOCUMENT_DEST = os.environ.get('UPLOADED_DOCUMENT_DEST', '/app/document_uploads')
+    #100MB file limit
+    FILE_BYTE_LIMIT = 100 * 1024 * 1024
```
NWCalvank: Same question as before. Is there a reason to convert this to bytes here, and then convert it back to MB in the code?

python-backend/app/config.py
```diff
@@ -46,6 +49,12 @@ class TestConfig(Config):
     JWT_OIDC_TEST_CLIENT_SECRET = "test_secret"
     JWT_OIDC_TEST_ISSUER = "test_issuer"
     # Dummy Private Keys for testing purposes, can replace these keys with any other generated key.
+
+    #Flask-uploads configs
+    UPLOADED_DOCUMENT_DEST = os.environ.get('UPLOADED_DOCUMENT_DEST', '/app/document_uploads')
+    #100MB file limit
+    FILE_BYTE_LIMIT = 100 * 1024 * 1024
```
NWCalvank: Duplicate code?

python-backend/app/config.py
JustinMacaulay: It has to be in byte form for the configuration.

python-backend/app/config.py
JustinMacaulay: No one is for the test config and one is the normal config, unless I'm missing something and the test app will pull the normal config.

python-backend/app/api/document_manager/resources/document_manager.py
JustinMacaulay: At the moment this is just an API service that has no connection to the front end app and will only ever be talking to the backend. So this is more meant to be logged when we have logging available.

openshift/sonar.pod.json
JustinMacaulay: I will re push if I can but that is not how it shows upon my local version.

python-backend/app/config.py
JustinMacaulay: I swapped it around so that there is only one calculation now instead of two to go from MB to bytes and back again. Now it stays as MB and then for the configuration does one calculation to get the bytes.

FreshGyan: @JustinMacaulay @NWCalvank The TestConfig class subclasses from the Config class. So anything declared in Config is inherited by TestConfig as well.
You do not need to re-declare the variables

migrations/sql/V1.10__Add_Documents.sql
```diff
@@ -1,10 +1,10 @@
 CREATE TABLE public.document_manager (
 	document_manager_id serial NOT NULL,
-	document_guid uuid NOT NULL DEFAULT gen_random_uuid(),
```
FreshGyan: You can't change an exisiting migration. If you do need to remove/update these existing constraints. You should create a new migration.
Unless there was team concensus of deleting the test and prod environments and recreating them that i'm not aware of

python-backend/app/__init__.py
```diff
@@ -26,9 +27,14 @@ def create_app(test_config=None):
         # load the test config if passed in
         app.config.from_object(test_config)
 
+    configure_uploads(app, documents)
+    #convert to bytes from MB
+    patch_request_class(app, (Config.FILE_BYTE_LIMIT*1024*1024))
```
FreshGyan: You can also achieve the same behavior by setting the `MAX_CONTENT_LENGTH` app config variable for this setting. patch_request_class which is just there for backwards compatibility if we were using an older version of flask.
See the Configuration Tab
https://flask-uploads.readthedocs.io/en/latest/

python-backend/app/config.py
```diff
@@ -34,7 +34,10 @@ class Config(object):
     SQLALCHEMY_MAX_OVERFLOW = 20
     SQLALCHEMY_POOL_TIMEOUT = 300
 
-
+    #Flask-uploads configs
+    UPLOADED_DOCUMENT_DEST = os.environ.get('UPLOADED_DOCUMENT_DEST', '/app/document_uploads')
+    #100MB file limit
+    FILE_BYTE_LIMIT = 100
```
FreshGyan: As mentione above, this can probably be replaced by `MAX_CONTENT_LENGTH `

python-backend/app/config.py
```diff
@@ -46,6 +49,12 @@ class TestConfig(Config):
     JWT_OIDC_TEST_CLIENT_SECRET = "test_secret"
     JWT_OIDC_TEST_ISSUER = "test_issuer"
     # Dummy Private Keys for testing purposes, can replace these keys with any other generated key.
+
+    #Flask-uploads configs
+    UPLOADED_DOCUMENT_DEST = os.environ.get('UPLOADED_DOCUMENT_DEST', '/app/document_uploads')
+    #100MB file limit
+    FILE_BYTE_LIMIT = 100
```
FreshGyan: The TestConfig subclasses Config, so you dont need to duplicate this code.

migrations/sql/V1.10__Add_Documents.sql
JustinMacaulay: Right, I thought since this was the first time the migration was used that I could change it. I forgot it was a separate PR and already lived in the Sprint Branch I changed it to be done in a different script.

python-backend/app/__init__.py
JustinMacaulay: Changed this to use flasks built in MAX_CONTENT_LENGTH

PR #279, opened 2018-12-03
MDS-711 Bug/sonarqube coverage difference

mmumby:


frontend/src/components/mine/ContactInfo/ViewMineManager.js
```diff
@@ -25,127 +25,180 @@ const propTypes = {
...
+                    ModalContent.ADD_MINE_MANAGER
+                  )
+                }
+                string={ModalContent.ADD_MINE_MANAGER}
+                type="primary"
+              />
+            </div>
+          </Card>
+        )}
+        {mine.mgr_appointment[0] && (
```
NWCalvank: Did you decide that this was safe? Or do we need a `mine.mgr_appointment` check?

frontend/src/tests/routes/Routes.spec.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { shallow } from "enzyme";
+import Routes from "@/routes/Routes";
+// import * as routes from "@/constants/routes";
+// import * as String from "@/constants/strings";
```
NWCalvank: These got re-added somehow

PR #281, opened 2018-12-03
Feature/mds 701 document retrieval service

JustinMacaulay:
Adding a GET method to the document manager to retrieve the specified document by its document guid.


python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -83,3 +83,14 @@ def post(self):
...
         }
+
+    @api.doc(params={'document_guid': 'Required: Document guid. Returns the file associated to this guid.'})
+    @jwt.requires_roles(["mds-mine-create"])
+    def get(self, document_guid):
+
+        if document_guid == None:
+            return self.create_error_payload(401, 'Must provide a document guid.')
+
+        document_manager_doc = DocumentManager.find_by_document_manager_guid(document_guid)
```
NWCalvank: What happens if this is `None`? Will `send_file()` throw an exception while accessing `file_display_name`?

JustinMacaulay: I noticed that as well and added it while I was looking at the tests or how to test the byte stream being returned.

PR #284, opened 2018-12-03
Feature/mds 701 document retrieval service

JustinMacaulay:
Adding a GET method to the document manager to retrieve the specified document by its document guid.

recreated PR due to images not clearing.


python-backend/tests/document_manager/resources/test_document_manager_resource.py
```diff
@@ -29,14 +36,14 @@ def test_happy_path_file_upload(test_client, auth_headers, setup_info):
     }
     
     post_resp = test_client.post('/document-manager', data=data, headers=auth_headers['full_auth_header'])
+
+    assert post_resp.status_code == 200
+
     post_data = json.loads(post_resp.data.decode())
     filename = post_data['document_manager_guids'][0]+'.pdf'
     saved_file = open(f'{test_dir_path}/mines/{setup_info.get("mine_guid")}/tailings/{filename}')
```
NWCalvank: Does this get cleaned up after the test completes?

python-backend/app/api/document_manager/resources/document_manager.py
```diff
@@ -83,3 +83,18 @@ def post(self):
...
+    @api.doc(params={'document_guid': 'Required: Document guid. Returns the file associated to this guid.'})
+    @jwt.requires_roles(["mds-mine-create"])
+    def get(self, document_guid=None):
+
+        if document_guid == None:
+            return self.create_error_payload(401, 'Must provide a document guid.')
+
+        document_manager_doc = DocumentManager.find_by_document_manager_guid(document_guid)
+
+        if document_manager_doc:
```
NWCalvank: Small thing.. any reason for the difference in style for the `None` check on this line vs on like 91?

python-backend/tests/document_manager/resources/test_document_manager_resource.py
```diff
@@ -74,12 +80,57 @@ def test_bad_file_type(test_client, auth_headers, setup_info):
...
+    saved_file.seek(0)
+
+    assert get_resp.data == saved_file.read()
+
+def test_download_file_no_guid(test_client, auth_headers, setup_info):
+    
+    get_resp = test_client.get(f'/document-manager', headers=auth_headers['full_auth_header'])
+    get_data = json.loads(get_resp.data.decode())
+
+    assert get_resp.status_code ==200
```
NWCalvank: If you have to push a change for the cleanup, just add a space to this assertion and the same one in the test below.

python-backend/tests/document_manager/resources/test_document_manager_resource.py
JustinMacaulay: Yeah, there is a fixture at the top that runs for every function and yields in the middle. It checks for the folders before and after the test and removes them.

python-backend/app/api/document_manager/resources/document_manager.py
JustinMacaulay: No should be the same Ill change that now.

python-backend/tests/document_manager/resources/test_document_manager_resource.py
JustinMacaulay: fixed

PR #285, opened 2018-12-04
MDS-697

mmumby:
Fixed an incorrect test under mineSelectors, 
Fixed z-index of notifications to override ant design, 
Updated mineSelector getMineRegionHash to include reusable functions


frontend/src/selectors/mineSelectors.js
```diff
@@ -10,8 +10,9 @@ export const getMineGuid = (state) => mineReducer.getMineGuid(state);
 export const getMineStatusOptions = (state) => mineReducer.getMineStatusOptions(state);
 export const getMineRegionOptions = (state) => mineReducer.getMineRegionOptions(state);
 export const getMineTenureTypes = (state) => mineReducer.getMineTenureTypes(state);
-export const getExpectedDocumentStatusOptions = (state) =>  mineReducer.getExpectedDocumentStatusOptions(state);
-export const getMineTSFRequiredReports = (state) =>  mineReducer.getMineTSFRequiredReports(state);
+export const getExpectedDocumentStatusOptions = (state) =>
+  mineReducer.getExpectedDocumentStatusOptions(state);
+export const getMineTSFRequiredReports = (state) => mineReducer.getMineTSFRequiredReports(state);
```
NWCalvank: No need to change it now, but these can just be:
`export const getMineTSFRequiredReports = mineReducer.getMineTSFRequiredReports`

The extra function wrapping shouldn't be needed

PR #290, opened 2018-12-05
Apply ESLint Auto-fix

NWCalvank:


frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -37,10 +37,10 @@ export const updateMineRecord = (id, payload, mineName) => (dispatch) => {
   dispatch(request(reducerTypes.UPDATE_MINE_RECORD));
   dispatch(showLoading("modal"));
   return axios
-    .put(ENVIRONMENT.apiUrl + API.MINE + "/" + id, payload, createRequestHeader())
+    .put(`${ENVIRONMENT.apiUrl + API.MINE  }/${  id}`, payload, createRequestHeader())
```
Jsyro: Out of curiosity, are these backticks or single quotes?

frontend/package.json
```diff
@@ -95,9 +95,9 @@
     "ci-test": "CI=true jest --env=jsdom --testResultsProcessor=jest-sonar-reporter --updateSnapshot --runInBand --coverage",
     "lint": "eslint",
     "format:check": "prettier --config .prettierrc.json",
-    "format:check-all": "prettier --config .prettierrc.json src/**/*.js",
+    "format:check-all": "prettier --config .prettierrc.json src/*.js && prettier --config .prettierrc.json src/**/*.js && prettier --config .prettierrc.json src/**/**/*.js",
     "format:write": "prettier --config .prettierrc.json --write",
-    "format:write-all": "prettier --config .prettierrc.json src/**/*.js --write",
+    "format:write-all": "prettier --config .prettierrc.json src/*.js --write && prettier --config .prettierrc.json src/**/*.js --write && prettier --config .prettierrc.json src/**/**/*.js --write",
```
NWCalvank: @FreshGyan do you have any idea why this would be necessary to hit the different directory levels? I don't think I've ever needed to do this before. I added it like this to get us unblocked for now

PR #292, opened 2018-12-05
Develop Branch for CI/CD

ghost:


migrations/initdb-postgis.sh
```diff
@@ -0,0 +1,23 @@
...
+UPDATE pg_database SET datistemplate = TRUE WHERE datname = 'template_postgis';
+EOSQL
+
+# Load PostGIS into both template_database and $POSTGRES_DB
+for DB in template_postgis "$POSTGRES_DB"; do
+	echo "Loading PostGIS extensions into $DB"
+	"${psql[@]}" --dbname="$DB" <<-'EOSQL'
+		CREATE EXTENSION IF NOT EXISTS postgis;
+		CREATE EXTENSION IF NOT EXISTS postgis_topology;
+		CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
```
garywong-bc: FYI our PostGIS [image](https://github.com/bcgov/openshift-postgresql-oracle_fdw/blob/bbf74c4f571c8d064c3c300f5ded026ade04eff2/rhel7.rh-postgresql96/Dockerfile#L121) doesn't have all the PostGIS-related extensions .. if we need them we can add them too, but Clecio will have to re-generate a `new v2_stable`

PR #294, opened 2018-12-06
Adding persistant storage to openshift for document uploads

james-hollinger:
Added a PVC to the backend to hold our uploaded documents.
Currently a 1Gi provision.


openshift/_python36.dc.json
```diff
@@ -120,6 +120,27 @@
...
+                    "volume.beta.kubernetes.io/storage-provisioner": "kubernetes.io/glusterfs"
+                }
+            },
+            "spec": {
+                "accessModes": [
+                    "ReadWriteMany"
+                ],
+                "resources": {
+                    "requests": {
+                        "storage": "1Gi"
```
FreshGyan: Can you make this configurable as well?
Just so on dev we can always start off with I dunno, half a gig or something  and then on test we have 5ish? and on Prod 10?
Something similar to the DB stuff?

Thought behind:
Although the file size data will probably be larger in size than db entries so if we can get an estimate on how big Prod volume needs to be to support atleast up until the July release if not till end of next year, we should do that. Currently in our Quota we get upto 200Gigs of storage in PVCs. So maybe chat with business (Dan/Aaron?) and see what is our expected storage space?

Ideally once we have production envt, storage would remain permanent so it should be able to handle a good chunk of data

james-hollinger: Good points. I have set initial claims to be .5/5/10 Gi.

james-hollinger: Dev is now set to 1Gi, it seems our openshift has a minimum of 1Gi.

PR #295, opened 2018-12-06
MDS-950 Task/Add mine disturbance types to database

NWCalvank:


NWCalvank:
@FreshGyan The mine_disturbance_code table has 4 rows of static data. There are indexes on the primary keys by default, as well as the constraint. I took a serious look at adding additional index(es) for `mine_type_detail_xref.mine_type_guid` and `mine_type_detail_xref. mine_disturbance_code`, but came to the conclusion yesterday that I didn't think they'd be needed.

Certainly open to input on that, though.

migrations/sql/V1.16__Add_Disturbance_Type.sql
```diff
@@ -0,0 +1,34 @@
...
+
+    -- Foreign Key Definitions
+    FOREIGN KEY (mine_disturbance_code) REFERENCES mine_disturbance_code(mine_disturbance_code) DEFERRABLE INITIALLY DEFERRED,
+    FOREIGN KEY (mine_type_guid) REFERENCES mine_type(mine_type_guid) DEFERRABLE INITIALLY DEFERRED,
+
+    -- Constraints
+    CONSTRAINT mine_type_guid_mine_disturbance_code_uniqeness UNIQUE (mine_type_guid, mine_disturbance_code)
+);
+
+COMMENT ON TABLE mine_type_detail_xref IS 'A lookup of all valid disturbance types.';
```
klefler: minor suggestion: This table comment is ok for now, but once we add commodity, it'll need to be updated to include that in the description. It can be done later. Just wanted to mark it now :)

itchison: Is there a requirement to track mine commodity disturbances over time? It appears that this table has no effective changes.
Also, are you planning on adding commodity type later? (as per MDS-454)

NWCalvank: Perhaps @klefler can speak to this better than I can, but I believe `active_ind` is going to handle the expiration. If there are further requirements there, I'm not aware of them.

And yes, commodity is a separate ticket (MDS-455)

klefler: @NWCalvank @itchison ... right, this was the conversation we had on Slack last week. Not sure we came up with a solution? Tracking history of the "mine type" really depends on what we want to keep history of...

If we just want to know whether or not the "mine type" (which includes the record in the mine_type table PLUS the records in the mine_type_detail_xref table) is effective or not, then we could add an active_ind on the record in the mine_type table... if that is active, then it's assumed that the record(s) tied to it in the xref are also valid. Note; this doesn't include keeping a history of the xref changing over time.

If a history of changes to the commodity/disturbance changing needs to be tracked, then that's a whole other design. We'd likely need another "history" table to track that.

Either way, we might need a face to face chat to hash this out. I don't know that adding an active_ind on the xref is enough and we may need to consider the mine_type and xref tables as a package deal.

PR #298, opened 2018-12-06
MDS-454 Feature/ Add /mines/disturbance_codes endpoint

NWCalvank:
This endpoint returns the valid options for disturbance type.


python-backend/tests/constants.py
```diff
@@ -62,6 +62,9 @@
 TEST_MINE_TENURE_TYPE_IDS = [1,2,3,4]
 TEST_MINE_TENURE_TYPE_NAMES = ['Coal', 'Mineral', 'Placer', 'BC Land']
 
+TEST_MINE_DISTURBANCE_CODES = ['SUR', 'UND', 'CWA', 'MIL']
```
nayr974: I would really like to see the use of this constants file go away. Why aren't we returning these from the database?

nayr974: NM. Looks like these are just used for tests, and you are returning from the DB in the resource. Thanks!

PR #301, opened 2018-12-08
MDS-923 - Contacts / Engineer of Record

nayr974:
Added generic contacts (PartyRelationships) with dynamic form options, and conditionally rendered components for contact types. Add/Edit/Remove, and some EoR validation and logic.


frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
```diff
@@ -0,0 +1,99 @@
...
+
+const validate = (values) => {
+  const errors = {};
+  if (values.start_date && values.end_date) {
+    if (Date.parse(values.start_date) >= Date.parse(values.end_date)) {
+      errors.end_date = "Must be after start date.";
+    }
+  }
+  return errors;
+};
```
mmumby: Can this be kept in `utils/validate.js` to be reused in other places and for consistency

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
```diff
@@ -0,0 +1,99 @@
...
+
+export const EditPartyRelationshipForm = (props) => {
```
mmumby: Can `AddPartyRelationshipFrom` and `EditPartyRelationshipForm' be refactored to use the same component? 
If not, that's okay, just curious!

and Am I correct in assuming that `EngineerOfRecordOptions` was extracted out to be used in both places? I am not sure that the `Forms` file is the best place for it, maybe move it to `common` and keep `Forms` strictly for components that are Forms

frontend/src/components/common/NullScreen.js
```diff
@@ -81,6 +82,13 @@ const NullScreen = (props) => (
         <p>{String.NO_PREV_MINE_MANAGER}</p>
       </div>
     )}
+    {props.type === "contacts" && (
+      <div>
+        <img src={MINER_TWO} />
+        <h3>No contacts found</h3>
+        <p>Create a contact using the menu above</p>
```
mmumby: Create constants for these, reuse existing strings when necessary to keep `Null screens` consistent.

frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
```diff
@@ -0,0 +1,94 @@
...
+  removePartyRelationship: PropTypes.func.isRequired,
+};
+
+export class DefaultContact extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    const {
```
mmumby: This is one of the eslint rules we mentioned to look out for, As per the rule `react/destructuring-assignment` we have decided to go with 'never' for destructuring props. 

Reason being that this rule does not play nice with Redux props, so our options were to make things more complicated by choosing `always`, go with 'never', or disable the rule, 

We opted against disabling due to the gray area it will cause, and we wanted to make the frontend as consistent as possible. 

This, of course, can be discussed further if you disagree or need clarification.

nayr974: I thought that you and Nathan had decided on de-structuring as a recommendation. I'm happy to go the other way, I find de-structuring awkward. :)

frontend/src/components/mine/Tailings/MineTailingsInfo.js
```diff
@@ -88,22 +103,36 @@ export class MineTailingsInfo extends Component {
...
 
   openEditReportModal(event, onSubmit, title, statusOptions, doc) {
     this.setState({
-      selectedDocument: doc
-    })
+      selectedDocument: doc,
+    });
     event.preventDefault();
     const initialValues = {
-      "tsf_report_name": doc ? ((doc.exp_document_name === 'None') ? null : doc.exp_document_name) : null,
```
mmumby: Should avoid nested ternary, 
Since they are all checking the existence of `doc`, pull that out and create the initialValues if doc exists

frontend/src/components/mine/Tailings/MineTailingsInfo.js
```diff
@@ -200,34 +240,41 @@ export class MineTailingsInfo extends Component {
...
                       event,
                       this.handleAddReportSubmit,
                       ModalContent.ADD_TAILINGS_REPORT,
                       this.props.mineTSFRequiredReports
-                    )}
+                    )
+                  }
                 >
-+ Add TSF Report
+                  + Add TSF Report
```
mmumby: Use the title as the button content, I believe its `ModalContent.ADD_TAILINGS_REPORT`, that way if the name changes it can be updated in one spot.

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
nayr974: It is a different type of validator that can't be applied to single form items and a validator for the entire form (validation logic that crosses multiple fields). It looks re-usuable in this case because I'm only checking start and end dates, but I imagine every form will need it's own set of rules and for the form validate function you cannot pass it an array of functions in the way you can for an individual field.

frontend/src/types/index.js
```diff
@@ -0,0 +1,31 @@
+// types/index.js
```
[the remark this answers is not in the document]
nayr974: Renamed to customPropTypes.

frontend/src/components/mine/Tailings/MineTailingsInfo.js
nayr974: Changed.

frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
nayr974: Changed.

frontend/src/components/mine/Tailings/MineTailingsInfo.js
nayr974: Agree this is really awkward. Not sure why it was done this way, but changing.

nayr974: Changed.

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
mmumby: hmmm, I still think it belongs under the common directory, Most of the files under that directory as all children of forms as well.

frontend/src/components/common/NullScreen.js
mmumby: Yeah, you are correct in that.

I guess my point was to reuse as much as possible in terms of the strings so the null screens are consistent as possible with their verbiage. 

I'm happy to remove the constants within this file and show the actual strings.

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
mmumby: ah, okay.

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
nayr974: The components under common are common to the app, and where they are common to "forms" they are foundational components for all forms. 

These components will only be common to these specific two forms, and no others. Given that we may see 10+ of these child options, one for each contact type, this is going to clutter common with components that nobody else will care about.

mmumby: what about creating and adding them to `forms/common`? 
Assuming there will be 10+ this will clutter forms.

nayr974: To avoid clutter, I grouped them under a child folder called "PartyRelationships" along with the Add/Edit forms that relate to them. I can move them to forms/common, but I still have the same concern that they will apply only to these two forms and aren't really common re-usable components at the form level.

mmumby: ah no, that all works. 

Didn't notice the file grouping you had already created.

nayr974: Cool 👍

PR #304, opened 2018-12-10
MDS-163: Enable spacial capabilities (PostGIS)

ActionAnalytics:


migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -149,8 +149,7 @@ BEGIN
     INSERT INTO mine_location(
         mine_location_guid  ,
         mine_guid           ,
-        latitude            ,
-        longitude           ,
+        geom                ,
```
garywong-bc: I was assuming an 'expand and then contract'.. so we'd keep mine_location.latitude/longitude until we meed to contract.  This needs to be timed with the other MDS methods that currently work with textual lat/long (e.g. the map rendering of points); if that map still needs textual fields, we can get rid of the two columns but use the PostGIS functions (e.g. `ST_X(geom) as longitude`, `ST_Y(geom) as latitude`)
```

Comment (ActionAnalytics) [reply migrations/etl/1_Mine_profile_detail_location.sql]: Good catch.  Done

File: npm-debug.log [hunk 2]
```diff
@@ -0,0 +1,23 @@
+0 info it worked if it ends with ok
```
FreshGyan: Can you add a .gitignore entry for this?

migrations/Dockerfile.dev
```diff
@@ -1,4 +1,70 @@
 FROM postgres:9.6-alpine
+MAINTAINER Régis Belson <me@regisbelson.fr>
```
FreshGyan: Can we update the Maintainer to some email from MDS team?

npm-debug.log
ActionAnalytics: Done

migrations/Dockerfile.dev
ActionAnalytics: Thanks, as per our discussion, removed all Maintainer tags. 

FYI, Maintainer tags are deprecated now, replaced by Author tags.

migrations/sql/V1.21__Alter_Mine_Location_Add_Geoms.sql
```diff
@@ -0,0 +1,5 @@
+-- SRID=3005 is ‘BC Albers’ projection, chosen as the standard projection for this application.
+ALTER TABLE mine_location ADD COLUMN geom GEOMETRY(POINT, 3005);
+UPDATE mine_location SET geom = ST_SetSRID(ST_MakePoint(longitude, latitude),3005) WHERE longitude IS NOT NULL AND latitude IS NOT NULL;
```
FreshGyan: Where is the number 3005 coming from?
If i'm not mistaken all our maps use `4326`. Although im not an expert in these 😅

migrations/sql/V1.21__Alter_Mine_Location_Add_Geoms.sql
```diff
@@ -0,0 +1,5 @@
...
+CREATE INDEX mine_location_geom_idx ON mine_location USING GIST(geom);
+--VACUUM ANALYZE mine_location;
```
FreshGyan: Unused comment?

ActionAnalytics: Removed the VACUUM comment.  The 3005 one ads value so leaving it.

migrations/sql/V1.21__Alter_Mine_Location_Add_Geoms.sql
ActionAnalytics: We will move forward and create another ticket to discuss the projection we use.

PR #305, opened 2018-12-10
Feature/mds 938 create document service upload api

JustinMacaulay:
New api service for uploading documents, forwards the request to the document manager and creates new mine documents. Also, added new tests for this resource and mocking to mock the post to the document manager.

Several files have been updated due only to the fact that we added a new python formatter.


python-backend/app/api/documents/expected/resources/documents.py
```diff
@@ -13,31 +13,47 @@
...
+            return self.create_error_payload(
+                404, 'Must provide a expected document guid.'), 404
 
         exp_doc = ExpectedDocument.find_by_exp_document_guid(exp_doc_guid)
         if exp_doc is not None:
             data = self.parser.parse_args()
             updated_doc = data['document']
-            if str(exp_doc.exp_document_guid) != updated_doc['exp_document_guid']:
-                return self.create_error_payload(500, 'exp_document does not match guid provided'), 500
+            if str(exp_doc.
```
NWCalvank: What's determining the location of these line breaks?

JustinMacaulay: That would be the Python formatter YAPF and I don't understand why it would choose that as its break point.

Jsyro: pep8 standard is max line length of 80, YAPF and BLACK both enforce 85-90 as max line length as it breaks significantly fewer lines.

Jsyro: YAPF by default it 80, which is fairly strict. I'd look at changing that a bit, but if that's the case we should version the settings file so everyone has the same. Discussion for another time.

python-backend/app/config.py
```diff
@@ -26,6 +26,7 @@ class Config(object):
 
     # Microservice URLs
     DOCUMENT_MS_URL = os.environ.get('DOCUMENT_MS_URL', 'http://localhost:5000')
+    DOCUMENT_MANAGER_URL = os.environ.get('DOCUMENT_MS_URL', 'http://localhost:5000')
```
Jsyro: environment variable looks like it should be Document_Manager_URL, to be different. 
MS was supposed to mean microservice, maybe we should change that standard of follow it? 
maybe, Document_Manager_MS_URL

python-backend/app/api/documents/expected/resources/documents.py
garywong-bc: we had already had "PEP8 80 character limit is stoopid" discussion, with @samfresh  and @mmumby  @ConradGustafson  at the start.   Our other projects have more reasonable line limits set ... GWELLS is at least 110... if the MDS dev team agrees, we can shift back to "more than 80"

python-backend/app/config.py
JustinMacaulay: done

PR #309, opened 2018-12-12
Refactor mine List table to ant design table

mmumby:
Swapped out ant design Row/Columns on the mine list view for ant design Table. 

This is a very basic implementation of the table, filters/sorts can be added in the future.


frontend/src/components/dashboard/MineList.js
```diff
@@ -1,83 +1,132 @@
...
+      </div>
+    ),
+  },
+  {
+    title: "Permit #",
+    dataIndex: "permit",
+    width: 150,
+    render: (text, record) => (
+      <div>
+        {text && text.map(({ permit_no, permit_guid }) => <div key={permit_guid}>{permit_no}</div>)}
```
NWCalvank: Does this add a linting camelCase error? We'll need to disable that rule if we're going to write our code this way anyway

frontend/src/components/dashboard/MineList.js
```diff
@@ -1,83 +1,132 @@
...
-  mines: {},
-  mineIds: [],
-  pageData: {},
-};
-
 class MineList extends Component {
   render() {
-    const { mines, mineIds, mineRegionHash } = this.props;
+    const data = [];
+    this.props.mineIds.forEach((id) => {
```
NWCalvank: How about something like... ?
```
const data = this.props.mineIds.map((id) => ({ key: id, emptyField: String.EMPTY_FIELD, ... })
```

mmumby: yess, that is much better

frontend/src/components/dashboard/MineList.js
mmumby: it does indeed add the linting error. 

I was thinking the same thing.

frontend/src/components/dashboard/MineList.js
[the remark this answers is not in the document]
mmumby: Good catch, I agree. 

That has been updated, I aslo changed the component from stateful to stateless, as it does not currently have state and our linting rules suggested changing.

PR #311, opened 2018-12-12
MDS-455 Feature/Add commodity types and fully support Mine Type management

NWCalvank:
Commits `490e333` and beyond are relevant to this PR.

Feature summary:
* Seeded valid commodity values
* Add/remove commodities via API
* Delete mine types via API
* Validation around active commodity and mine types uniqueness


python-backend/app/api/mines/mine/models/mine_commodity_code.py
```diff
@@ -0,0 +1,28 @@
...
+from ....utils.models_mixins import AuditMixin, Base
+from app.api.constants import COMMODITY_CODES_CONFIG, METALS_AND_MINERALS
+from app.extensions import db
+
+
+class MineCommodityCode(AuditMixin, Base):
+    __tablename__ = 'mine_commodity_code'
+    mine_commodity_code = db.Column(db.String, nullable=False, primary_key=True)
+    description = db.Column(db.String, nullable=False)
+    active_ind = db.Column(db.Boolean, nullable=False, default=True)
```
Jsyro: default values should be set by the db. you can do that by adding this import: 

then going. 
active_ind = db.Column(db.Boolean, server_default=FetchedValue())

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -97,6 +161,31 @@ def test_post_mine_disturbance_duplicate(test_client, auth_headers):
...
+        'mine_commodity_code': TEST_MINE_COMMODITY_CODES[1]
+    }
+    post_resp1 = test_client.post('/mines/mine_types/details', data=test_data, headers=auth_headers['full_auth_header'])
+    assert post_resp1.status_code == 200
+
+    post_data1 = json.loads(post_resp1.data.decode())
+    assert post_data1['mine_type_guid'] == test_data['mine_type_guid']
+    assert post_data1['mine_commodity_code'] == test_data['mine_commodity_code']
+    assert post_data1['active_ind'] == True
+
```
Jsyro: When database defines the defaults active_ind == true will work as long as this is integration test, which we want to move away from in the future. Consider removing this assert as it is redundant and will fail when we mock the db?

python-backend/app/api/mines/mine/models/mine_commodity_code.py
NWCalvank: Sure. The only defaults I've come across in the codebase so far are timestamps, which have not been using fetchedValue. I can switch this default to use that

python-backend/app/api/mines/namespace/mines.py
```diff
@@ -18,6 +19,7 @@
 api.add_resource(MineTypeDetailResource, '/mine_types/details', '/mine_types/details/<string:mine_type_detail_xref_guid>')
 api.add_resource(MineTenureTypeCodeResource, '/mine_tenure_type_codes')
 api.add_resource(MineDisturbanceCodeResource, '/disturbance_codes')
+api.add_resource(MineCommodityCodeResource, '/commodity_codes')
 api.add_resource(MineLocationResource, '/location', '/location/<string:mine_location_guid>')
 api.add_resource(MineStatusResource, '/status', '/status/<string:mine_status_guid>')
 api.add_resource(MineRegionResource,'/region','/region/<string:mine_region_guid>')
```
Jsyro: We should be using kebab case for url paths as that is the http standard.

/commodity-codes
/disturbance-codes
/mine-tenure-type-codes

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
NWCalvank: The `active_ind = true` is definitely required behaviour upon record creation. I'd feel uncomfortable removing test coverage on this at this point

python-backend/app/api/mines/namespace/mines.py
NWCalvank: Sure

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -50,7 +80,23 @@ def test_post_mine_type_detail_missing_mine_disturbance_code(test_client, auth_h
     assert post_data == {
         'error': {
             'status': 400,
-            'message': 'Error: Missing mine_disturbance_code.'
+            'message': 'Error: Missing mine_disturbance_code or mine_commodity_code.'
```
Jsyro: We should not be asserting the exact error message unless it's defined as a constant or in some central object. Which we haven't does as a pattern for mds yet.  Should only be asserting status == 400.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -123,6 +212,56 @@ def test_post_mine_type_detail_invalid_mine_disturbance_code_for_tenure_type(tes
...
+        'mine_commodity_code': 'AE' # AE is not a valid COL commodity
+    }
+    post_resp = test_client.post('/mines/mine_types/details', data=test_data, headers=auth_headers['full_auth_header'])
+    assert post_resp.status_code == 400
+
+    post_data = json.loads(post_resp.data.decode())
+    assert post_data == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Invalid mine_commodity_code.'
```
Jsyro: Same as above, should not be testing error strings.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -123,6 +212,56 @@ def test_post_mine_type_detail_invalid_mine_disturbance_code_for_tenure_type(tes
...
+        'mine_commodity_code': TEST_MINE_COMMODITY_CODES[0]
+    }
+    post_resp = test_client.post('/mines/mine_types/details', data=test_data, headers=auth_headers['full_auth_header'])
+    assert post_resp.status_code == 400
+
+    post_data = json.loads(post_resp.data.decode())
+    assert post_data == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Unable to create mine_type_detail with disturbance and commodity.'
```
Jsyro: Same as above, should not be testing error strings.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -97,6 +161,31 @@ def test_post_mine_disturbance_duplicate(test_client, auth_headers):
...
+    assert post_data1['active_ind'] == True
+
+    post_resp2 = test_client.post('/mines/mine_types/details', data=test_data, headers=auth_headers['full_auth_header'])
+    assert post_resp2.status_code == 400
+
+    post_data2 = json.loads(post_resp2.data.decode())
+    assert post_data2 == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Unable to create mine_type_detail.'
```
Jsyro: Same as above, should not be testing error strings.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -72,6 +118,24 @@ def test_post_mine_type_detail_invalid_mine_disturbance_code(test_client, auth_h
...
+        'mine_commodity_code': 'ABC'
+    }
+    post_resp = test_client.post('/mines/mine_types/details', data=test_mine_type_data, headers=auth_headers['full_auth_header'])
+    assert post_resp.status_code == 400
+
+    post_data = json.loads(post_resp.data.decode())
+    assert post_data == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Invalid mine_commodity_code.'
```
Jsyro: See below, should not be testing error strings.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
```diff
@@ -50,7 +80,23 @@ def test_post_mine_type_detail_missing_mine_disturbance_code(test_client, auth_h
...
+        'mine_type_guid': TEST_MINE_TYPE_GUID,
+    }
+    post_resp = test_client.post('/mines/mine_types/details', data=test_mine_type_data, headers=auth_headers['full_auth_header'])
+    assert post_resp.status_code == 400
+
+    post_data = json.loads(post_resp.data.decode())
+    assert post_data == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Missing mine_disturbance_code or mine_commodity_code.'
```
Jsyro: Same as above, should not be testing error strings.

python-backend/tests/mines/resources/test_mine_type_resource.py
```diff
@@ -169,3 +169,64 @@ def test_put_mine_type_identical_record(test_client, auth_headers):
...
+    assert post_data1['mine_guid'] == test_data['mine_guid']
+
+    post_resp2 = test_client.post('/mines/mine_types', data=test_data, headers=auth_headers['full_auth_header'])
+    assert post_resp2.status_code == 400
+
+    post_data2 = json.loads(post_resp2.data.decode())
+    assert post_data2 == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Unable to create mine_type.'
```
Jsyro: Same as above, should not be asserting error strings.

python-backend/tests/mines/resources/test_mine_type_resource.py
```diff
@@ -169,3 +169,64 @@ def test_put_mine_type_identical_record(test_client, auth_headers):
...
+        '/mines/mine_types/' + TEST_MINE_GUID,
+        headers=auth_headers['full_auth_header']
+    )
+    assert delete_resp.status_code == 400
+
+    delete_data = json.loads(delete_resp.data.decode())
+    assert delete_data == {
+        'error': {
+            'status': 400,
+            'message': 'Error: Invalid mine_type_guid.'
```
Jsyro: Same as above, should not be asserting error strings.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
NWCalvank: I don't necessarily disagree, but this is the pattern other tests were following when I joined the project. Removing the error message check could also considerably reduce our test coverage

python-backend/app/api/mines/mine/models/mine_commodity_code.py
Jsyro: We haven't done a full refactor of the default values, but I have added as a backlog item and think we should avoid defining defaults in two places. active_ind is a safe one to always be 'true' but default date and guid policies might change in the future so using FetchedValue makes that much easier. Sounds good.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
Jsyro: Would it reduce our test coverage? the src that adds the message is still being executed even if the test isn't asserting it's value. We can have a larger discussion, but should changing the error message really cause a test to fail. and if the exact strings error messages should be is a high priority item, then we need to centrally define them so that we can test them (assert message = ERROR_MESSAGES['mine_type'][101]) or something like that.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
Jsyro: Ok, i'm also unsure if should actually be returning active_ind in our json functions, i started a discussion in slack on that.

NWCalvank: Saw the slack message. Will remove the `active_ind`. I misunderstood their purpose. I thought they were going to be used for showing historical data in the UI

Jsyro: Yea, it's the soft deleted flag. Another item that needs a look when the new db auditing/history is determined.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
Jsyro: This might need a larger discussion, but I believe we should not be checking that error messages are exacting a certain string. Maybe just check if it isn't null? 

I think it adds alot of fragility and maintenance effort when the test should be 'did it fail' not 'did it fail with this exact error message'.

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
NWCalvank: Just a heads-up, it's being returned in a few of the mine documents-related JSON responses as well

python-backend/tests/mines/resources/test_mine_type_detail_resource.py
NWCalvank: As I said, I agree. And I will gladly switch them all over to use constants when we've made that switch. I'll even write the constants. I just don't see that as in-scope for this PR. But this test coverage is in-scope since the specific error messages are part of the expected behaviour

Jsyro: I would tend to disagree, but that refactor would be outside of the scope of the PR. I'll approve it as you are following the pattern and I'll make a note to bring this up at the end of sprint.

migrations/sql/V1.25__Add_Commodity_Code.sql
```diff
@@ -0,0 +1,38 @@
...
+    active_ind            boolean NOT NULL DEFAULT TRUE,
+
+    --Audit Columns
+    create_user      character varying(60) NOT NULL,
+    create_timestamp timestamp with time zone NOT NULL DEFAULT current_timestamp,
+    update_user      character varying(60) NOT NULL,
+    update_timestamp timestamp with time zone NOT NULL DEFAULT current_timestamp
+);
+
+COMMENT ON TABLE mine_disturbance_code IS 'The valid options for a mine commodity type.';
```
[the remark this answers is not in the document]
NWCalvank: Good catch. Thanks

PR #314, opened 2018-12-13
feature/MDS-920 Tailings Report Due Date Bug

nayr974:
Fixes a bug where TSF reports were not getting created with the correct due date. Additional refactoring to remove awkward/unneeded code that had a separate "hash" selector.


frontend/src/components/mine/Tailings/MineTailingsInfo.js
```diff
@@ -77,11 +78,16 @@ export class MineTailingsInfo extends Component {
...
-    }).then(() => {
+    const requiredReport = this.props.mineTSFRequiredReports.find(
+      (x) => x.req_document_guid === value.req_document_guid
+    );
+    const newRequiredReport = {
+      document_name: requiredReport.req_document_name,
+      req_document_guid: requiredReport.req_document_guid,
+      document_due_date_type: requiredReport.req_document_due_date_type,
+      document_due_date_period_months: requiredReport.req_document_due_date_period_months,
+    };
```
Jsyro: I feel like the backend should get the req_document_guid and go look up the due date fields itself, and not let the front end set them (we don't want the front end to change them do we)?

frontend/src/components/mine/Tailings/MineTailingsInfo.js
```diff
@@ -209,8 +219,9 @@ export class MineTailingsInfo extends Component {
           <hr style={{ borderTop: "2px solid #c4cdd5" }} />
```
mmumby: Can this <hr> be switched to the ant designs <Divider> ? 

We have a few throughout the app already, most take on the default styling, others are custom.

nayr974: Yes addressed.

PR #316, opened 2018-12-13
feature/mds-937 Add file upload component to mine expected document modal

james-hollinger:
Adds a component that allows file upload. The component uses react-filepond.
 - Supports multiple simultaneous uploads
 - Can cancel uploads in progress
 - Filesize limit configured to 100MB
 - Currently no file type restrictions, though we can add those easily through a plugin

Could be made more generic for reuse in other parts of the app, but is currently specific to tailings expected document uploads. Accepts an expected document GUID through props to pass to the API so the upload can be associated with an (existing) expected document.


frontend/src/components/Forms/EditTailingsReportForm.js
```diff
@@ -59,6 +61,15 @@ export const EditTailingsReportForm = (props) => (
             validate={[]}
           />
         </Form.Item>
+        <Form.Item label="Upload Documents">
+          <Field
+            id="tsf_document_upload"
+            name="tsf_document_upload"
+            expDocumentGuid={props.expDocumentGuid}
+            component={FileUpload}
+            validate={[]}
```
NWCalvank: I think the validate attribute can be removed if no validations are applied

frontend/src/components/mine/Tailings/FileUpload.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { FilePond, File, registerPlugin } from "react-filepond";
+import "filepond/dist/filepond.min.css";
```
NWCalvank: I don't think we should be referencing anything in our `/dist` in our `/src` files. This would either go into a `vendors` directory or wouldn't need to be imported at all (since the module `react-filepond` should include all of its dependence source files.

frontend/src/components/mine/Tailings/FileUpload.js
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <FilePond
+        name={"file"}
+        allowRevert={false}
+        allowMultiple={true}
+        maxFileSize={"100MB"}
+        server={this.server}
+        onupdatefiles={(fileItems) => {
+          this.setState({
+            files: fileItems.map((fileItem) => fileItem.file),
```
NWCalvank: Stylistic personal preference: `files: fileItems.map(({ file }) => file)`

But I think we're using `lodash` in this project already. So, we could just:
```
import { map } from "lodash";

...

files: map(fileItems, 'file')
```

Relevant docs: https://lodash.com/docs/4.17.11#map

frontend/src/components/mine/Tailings/FileUpload.js
james-hollinger: The example code for filepond includes importing the .css styles separately like this. I'm not sure what you are suggesting is the alternative?

frontend/src/components/Forms/EditTailingsReportForm.js
james-hollinger: This is a fair point, but the existing items on the page all have empty validators. I could change all of them?

frontend/src/components/mine/Tailings/FileUpload.js
james-hollinger: Good point, made the change to use lodash.

frontend/src/components/Forms/EditTailingsReportForm.js
mmumby: Yeah remove them all.

frontend/src/components/mine/Tailings/FileUpload.js
```diff
@@ -0,0 +1,52 @@
...
+import FilePondPluginFileValidateSize from "filepond-plugin-file-validate-size";
+registerPlugin(FilePondPluginFileValidateSize);
+
+import { ENVIRONMENT } from "@/constants/environment";
+import { createRequestHeader } from "@/utils/RequestHeaders";
+
+class FileUpload extends React.Component {
+  constructor(props) {
+    super(props);
+
```
mmumby:
```
constructor(props) {
    super(props);
```

This can be removed. It is not serving any purpose. 

In older versions of react the constructor was needed and the state was initiated inside, like so: 
```
constructor(props) {
 super(props);
 this.state = { files: [] };
}
```

But state can be initiated on its own without the constructor, which is what you are doing on line 15, (also consistent with the other files)

so line 12 & 13 are not actually doing anything.

frontend/src/components/Forms/EditTailingsReportForm.js
```diff
@@ -59,6 +61,15 @@ export const EditTailingsReportForm = (props) => (
             validate={[]}
           />
         </Form.Item>
+        <Form.Item label="Upload Documents">
+          <Field
+            id="tsf_document_upload"
+            name="tsf_document_upload"
+            expDocumentGuid={props.expDocumentGuid}
+            component={FileUpload}
```
mmumby: Is FileUpload.js specific for tailings? Or can it be reused?

if so, can this be changed to follow the syntax of the other Field Components?

frontend/src/components/mine/Tailings/FileUpload.js
NWCalvank: Apologies. I misread and thought this was coming from our own `/dist` dir (an output from our build system). I reread it after your response and see what's actually happening.

Importing a CSS file like that still seems odd to me, but I checked the docs and didn't immediately see a reasonable alternative. Will mark as resolved

frontend/src/components/Forms/EditTailingsReportForm.js
james-hollinger: Empty validators have been removed.

frontend/src/components/mine/Tailings/FileUpload.js
james-hollinger: After discussion, this was a miss-understanding that the server parameter was set inside the constructor not outside of it. Marking as resolved.

frontend/src/components/Forms/EditTailingsReportForm.js
james-hollinger: Right now it is specific to tailings, as that is the only place where we are currently uploading docs.

The document upload API needs the GUID of the expected document to associate the file to. We could pass in a URL to this component rather than the GUID, but then a higher-level component would have to know and construct the URL. I felt it was better to put it inside this component to make the change footprint smaller.

When you say follow the syntax of other field components, do you mean using another name than expDocumentGuid?

mmumby: I was more so referring to the ```component={FileUpload}```

Can FileUpload be moved into src/components/common and be renamed to match the syntax other similar components? 

Any prop that is passed into to the <Field /> is accessible to FileUpload, it doesn't really matter where this component lives,

I could not have the full picture here, but FileUpload looks as though it can be reused in the future and thus should be kept in common and renamed to `renderFileUpload` for consistency

nayr974: I think moving it to common makes sense as we do intend to use it elsewhere. What should be renamed to renderFileUpload? The component?

nayr974: James re: offline convo with mmumby, please leave the comonent name as FileUpload as it is not specifically a form child component but can be used elsewhere.

james-hollinger: Took the time to make the component generic. FileUpload now accepts an upload URL rather than a GUID. The form generates the URL from a constant in constants/API.

As well, the tailings page now sends the entire doc down to the children rather than just the GUID, since the expected doc record has all of its related uploaded files which we will eventually be displaying in the modal.

mmumby: Thank you for making it more generic!

PR #322, opened 2018-12-14
MDS-672 feature/display data from nris

JustinMacaulay:
Takes the data provided from NRIS and displays it under the compliance tab.

- added new action creator, selector, reducer and actions for compliance.
- added more exception checking on the backend.
- added new deployment config variables for all environments.


openshift/_python36.dc.json
```diff
@@ -58,6 +58,11 @@
       "required": false,
       "value": "/opt/app-root/src/document_uploads"
     },
+    {
+      "name": "MDS_NRIS",
```
FreshGyan: Can you make the Name more descriptive? Like `MDS_NRIS_SECRET_NAME` or something?

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -8,65 +9,74 @@
...
-
-        return overview
+        try:
+            data = NRIS_service.get_EMPR_data_from_NRIS(mine_no)
+        except requests.exceptions.HTTPError as errhttp:
+            return self.raise_error(
+                errhttp.response.status_code,
+                "There has been an unexpected error with NRIS.")
+        except requests.exceptions.Timeout as errt:
+            return self.raise_error(408, "NRIS has timed out.")
```
nayr974: I like this. Explicit messaging that it's our communication with NRIS.

frontend/src/components/mine/Compliance/MineComplianceInfo.js
```diff
@@ -0,0 +1,138 @@
+import moment from "moment";
```
nayr974: Moment is cool. Please broadcast you're addition and use of moment to the group, so that others are aware of it and also don't add other libraries to deal with time.

openshift/_python36.dc.json
garywong-bc: I suggest MDS_NRIS_API_USERNAME and MDS_NRIS_API_PASSWORD to follow the WSO2/OAuth2 terms.

garywong-bc: I suggest MDS_NRIS_API_USERNAME and MDS_NRIS_API_PASSWORD to follow the WSO2/OAuth2 terms.

JustinMacaulay: Changed to MDS_NRIS_API_SECRET

frontend/src/components/mine/Compliance/MineComplianceInfo.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  render() {
+    // TODO need to work out how and what is being returned.
+    return (
+      <div>
+        {this.state.isLoading && (
+          <div>
+            <Loading />
+          </div>
```
mmumby: I don't think the added div is needed. 

```
{this.state.isLoading && <Loading />}
```
This should work!

frontend/src/components/mine/Compliance/MineComplianceInfo.js
```diff
@@ -0,0 +1,135 @@
...
+  state = { isLoading: true };
+
+  componentDidMount() {
+    this.props
+      .fetchMineComplianceInfo(this.props.mine.mine_detail[0].mine_no)
+      .then(() => this.setState({ isLoading: !this.state.isLoading }));
+  }
+
+  render() {
+    // TODO need to work out how and what is being returned.
```
NWCalvank: It looks like you've done that? Also, a small note, a `TODO` should have a colon (`TODO: message`). Most syntax highlighters pick it up as a keyword that way

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -8,65 +9,74 @@
 
 
 class MineComplianceResource(Resource, UserMixin, ErrorMixin):
-    @api.doc(params={'mine_region_guid': 'Mine region guid.'})
+    @api.doc(params={'Mine_no': 'Mine ID.'})
```
NWCalvank: Is `Mine_no` intentionally capitalized?

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -8,65 +9,74 @@
...
+                    stops_dict = stops[0]
+
+                    stop_orders = stops_dict.get('stopOrders')
+                    stop_advisories = stops_dict.get('stopAdvisories')
+                    stop_warnings = stops_dict.get('stopWarnings')
+
+                    if stop_orders:
+                        open_orders += sum(
+                            k.get('orderStatus') == 'Open'
+                            for k in stop_orders)
```
NWCalvank: A `lambda` seems more appropriate here. It would also avoid the type coercion. Something like...
```python
open_orders = sum(map(
  lambda k: 1 if k.get('orderStatus') == 'Open' else 0,
  stop_orders
))
```

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -8,65 +9,74 @@
...
+                        overdue_orders += sum(
+                            k.get('orderCompletionDate') is not None and k.get(
+                                'orderStatus') == 'Open' and datetime.strptime(
+                                    k.get('orderCompletionDate'),
+                                    '%Y-%m-%d %H:%M') < datetime.now()
```
NWCalvank: It might just be the janky line wrapping, but I'm finding this incredibly difficult to read. My same suggestion for using a `lambda` (or method if more appropriate) applies here and to the following for loops.

python-backend/app/api/nris_services/NRIS_service.py
```diff
@@ -48,13 +57,18 @@ def get_EMPR_data_from_NRIS(mine_no):
...
-        url=current_app.config['NRIS_EMPR_API_URL'],
-        params=params,
-        headers=headers)
-
-    if empr_nris_resp.status_code != 200:
-        return None
+    try:
+        empr_nris_resp.raise_for_status()
+    except requests.exceptions.HTTPError:
+        #TODO add logging for this error.
```
NWCalvank: Was this just missed or is it out of scope / blocked by something?

frontend/src/components/mine/Compliance/MineComplianceInfo.js
JustinMacaulay: changed and works.

frontend/src/components/mine/Compliance/MineComplianceInfo.js
JustinMacaulay: I noticed and removed it.

python-backend/app/api/mines/compliance/resources/compliance.py
JustinMacaulay: I don't see how the lambda would be better. List comprehension is usually the faster way when using a non defined function or logical expression to evaluate each element. Im also not sure about the type coercion as the list this is being run on is part of a JSON return and the field itself in the NRIS definitions is a string.

python-backend/app/api/nris_services/NRIS_service.py
JustinMacaulay: We currently do not have a way of logging and this was left in to make sure that when we do logging is added here.

python-backend/app/api/mines/compliance/resources/compliance.py
JustinMacaulay: changed just missed the capital.

python-backend/app/api/mines/compliance/resources/compliance.py
NWCalvank: My point about the type coercion was just that you're implicitly coercing booleans to zeros and ones. It's not a bit deal, but it does obfuscate the purpose of this code block a bit in my opinion

I'm not familiar with how python structures list comprehensions. This doesn't looks like list comprehensions that I'm familiar with. This looks very procedural, particularly with the `+=` operator.

I'm not going to push back super hard on this point, as I'm not familiar with what python considers to be most idiomatic.

PR #324, opened 2018-12-15
Switch production python to use WSGI server

FreshGyan:
This is the just the first attempt at replacing our production web server for backend with wsgi instead of flask.
Lots and lots of configuration options available.

Happy to chat and discuss.


python-backend/uwsgi.ini
```diff
@@ -0,0 +1,31 @@
...
+http-socket = 0.0.0.0:5000
+http-auto-gzip = true
+post-buffering = 8192
+
+# UWSGI settings
+plugins = router_cache
+master = true
+harakiri = 90
+processes = 4
+threads = 2
```
itchison: Are these tuned for how our Openshift is configured or were these defaults?

python-backend/uwsgi.ini
```diff
@@ -0,0 +1,31 @@
...
+enable-threads = true
+vacuum = true
+disable-logging = True
+
+# Cache
+cache2 = name=mines,items=100,keysize=10240
+route-if = contains:${HTTP_ACCEPT_ENCODING};gzip cache:key=gzipped_${REQUEST_URI},name=mines,content_encoding=gzip,expires=3600
+route = ^/$ cache:key=${REQUEST_URI},name=mines,expires=3600
+route = ^/$ cachestore:key=${REQUEST_URI},gzip=gzipped_mines_homepage,name=mines,expires=3600
```
itchison: Do you understand the cachestore? Does this mean a newly added mine would not appear on the endpoint for 3600 seconds?

python-backend/uwsgi.ini
FreshGyan: They're a combination.
The docs said there is no magic number and just to fine tune and play around.
So i started off with this and as we go on we may have to fine tune and play around.

python-backend/uwsgi.ini
FreshGyan: Not really. I don't intend to ship it off with this PR. Just wanted to play around and see how it behaves.

I dont think cache should be automagically controlled by a library coz we can run into magic issues. I just wanted to test and see how it behaves.

I'll play around with that a little tonight but if it's not working great, i'll remove it and we .can revisit that in future.

python-backend/app/__init__.py
```diff
@@ -4,17 +4,18 @@
 from flask import Flask
 from flask_cors import CORS
 from flask_restplus import Resource
-from flask_uploads import UploadSet, configure_uploads, patch_request_class
+from flask_uploads import configure_uploads
```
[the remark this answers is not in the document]
NWCalvank: I don't think there's any backend linting?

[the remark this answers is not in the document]
FreshGyan: No, I just noticed those imports were unused, so i removed em

PR #328, opened 2018-12-17
MDS-941: Feature/ Advanced search by mine status (frontend)

NWCalvank:


frontend/src/components/dashboard/MineSearch.js
```diff
@@ -19,11 +21,14 @@ const propTypes = {
   searchValue: PropTypes.string,
 };
 
-const defaultProps = {
-  mineNameList: [],
-};
+const checkAdvancedSearch = ({ status, region, tenure, commodity, TSF, major }) =>
+  TSF || major || some([status, region, tenure, commodity], negate(isEmpty));
```
mmumby: clever

NWCalvank: Bad clever? This should be fairly standard for a project using Lodash (or a similar library)

mmumby: no, good clever.

PR #329, opened 2018-12-17
Feature/mds 991

thegentlemanphysicist:
This PR allows the backend endpoint to filter by status, region_code, tsf, major_mine, and tenure.
The PR also refactors the search and filter features to remove nested if statements.
Note if a mine's details were edited the filter functionality will filter on both the old and new versions of a mine's details.
When a filter is applied to the query the results are not ordered alphabetically by mine_name


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -10,11 +10,14 @@
 from ...status.models.mine_status_xref import MineStatusXref
 from ..models.mine_identity import MineIdentity
 from ..models.mine_detail import MineDetail
+from ..models.mine_tenure_type_code import MineTenureTypeCode
 from ..models.mine_type import MineType
 from ..models.mineral_tenure_xref import MineralTenureXref
 from ....permits.permit.models.permit import Permit
+from ...region.models.region import MineRegionCode
```
NWCalvank: @james-hollinger @Jsyro is this an example of an incorrect way of using a method directly? I've only overheard some discussions about sending internal API requests, but I haven't been in those conversations. Is that required here?

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -50,48 +56,83 @@ def get(self, mine_no_or_guid=None):
...
-                    mine_query_with_status = MineIdentity.query\
-                        .join(MineDetail)\
-                        .join(MineStatus)\
-                        .join(MineStatusXref)\
-                        .filter(all_status_filter)
-                    sorted_mine_query = apply_sort(mine_query_with_status, sort_criteria)
-                else:
-                    sorted_mine_query = apply_sort(MineIdentity.query.join(MineDetail), sort_criteria)
+            # Filter by Major Mine, if provided
+            if major_mine_filter_term:
```
NWCalvank: Let's remove the duplication here. The whole block can be something like:
```python
if major_mine_filter_term:
    major_mine_filter = MineDetail.major_mine_ind.is_(major_mine_filter_term == "true")
    major_mine_query = MineIdentity.query.join(MineDetail).filter(major_mine_filter)
    mines_permit_join_query = mines_permit_join_query.intersect(major_mine_query)
```

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -50,48 +56,83 @@ def get(self, mine_no_or_guid=None):
...
+                    major_mine_filter = MineDetail.major_mine_ind.is_(True)
+                    major_mine_query = MineIdentity.query.join(MineDetail).filter(major_mine_filter)
+                    mines_permit_join_query = mines_permit_join_query.intersect(major_mine_query)
+                elif major_mine_filter_term == "false":
+                    major_mine_filter = MineDetail.major_mine_ind.is_(False)
+                    major_mine_query = MineIdentity.query.join(MineDetail).filter(major_mine_filter)
+                    mines_permit_join_query = mines_permit_join_query.intersect(major_mine_query)
+
+            # Filter by TSF, if provided
+            if tsf_filter_term:
```
NWCalvank: Same idea here. Something like...
```python
if tsf_filter_term:
    tsf_filter_value = MineIdentity.mine_tailings_storage_facilities != None if tsf_filter_term == "true" else MineIdentity.mine_tailings_storage_facilities == None
    tsf_query = MineIdentity.query.filter(tsf_filter_value)
    mines_permit_join_query = mines_permit_join_query.intersect(tsf_query)
```

python-backend/app/api/mines/mine/resources/mine.py
Jsyro: @NWCalvank Good eyes, this is fine as they are under the same module '/mines/mine' and '/mines/region' (aka, they will always be in the same microservice). if this was fetching a document or a party, we want to use REST calls as those modules may be separated into separate microservices in the future.  This is good as is.

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: done and pushed

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -50,48 +56,77 @@ def get(self, mine_no_or_guid=None):
...
-                status_search_term_array = status_search_term.split(',')
-                status_filter = MineStatusXref.mine_operation_status_code.in_(status_search_term_array)
-                status_reason_filter = MineStatusXref.mine_operation_status_reason_code.in_(status_search_term_array)
-                status_subreason_filter = MineStatusXref.mine_operation_status_sub_reason_code.in_(
-                    status_search_term_array)
-                all_status_filter = status_filter | status_reason_filter | status_subreason_filter
+            sort_criteria = [{'model': 'MineDetail', 'field': 'mine_name', 'direction': 'asc'}]
+            mines_permit_join_query = MineIdentity.query.join(MineDetail)
+            mines_permit_join_query = apply_sort(mines_permit_join_query, sort_criteria)
+            #TODO: Add 'isActive' column to mine status, details, type, location, and permmits tables to allow filtering
```
NWCalvank: Could you please update this to be an `active_ind` column. And change the `TODO` to be only about what happens here. Example: `TODO: Leverage active_ind column to filter out historical values, created in MDS-###`

And create a ticket for it, which you can reference in the comment. You may also want to check with @klefler to get more information about which tables will be receiving an `active_ind` column at all. Some may not be. She'll have the best perspective on the plans for the database as a whole.

The other option is to just create the ticket and to not leave a comment at all.

PR #332, opened 2018-12-19
MDS-1040 Fix/ Remove redundant PUT method

NWCalvank:
This is just a bit of cleanup for MineType since this functionality was determined to be unneeded.

**Explanation**
PUT (in this context) is the combination of a DELETE and a POST. This method does not update in place, as would likely be expected. In order to make MineType consistent with Disturbance and Commodity, the redundant PUT method is being removed.


NWCalvank:
@FreshGyan never used. @mmumby is adding the update/delete functionality currently

PR #334, opened 2018-12-20
Added file type checking to file upload component

james-hollinger:
Added configurable file type checking to the front-end file upload component.

A new prop acceptedFileTypesMap has been added. The keys of this object are the allowed MIME types, while the value is the human readable name for the extension. These names are shown if an invalid file is uploaded.

The new constants file _constants/fileTypes_ enumerates the currently allowed file types, but the types can be added add-hoc to the component if desired. 

Also, max file size is now a prop with a default of 100MB


frontend/src/components/common/FileUpload.js
```diff
@@ -1,13 +1,21 @@
...
-registerPlugin(FilePondPluginFileValidateSize);
+import FilePondPluginFileValidateType from "filepond-plugin-file-validate-type";
+registerPlugin(FilePondPluginFileValidateSize, FilePondPluginFileValidateType);
 
 import { ENVIRONMENT } from "@/constants/environment";
 import { createRequestHeader } from "@/utils/RequestHeaders";
 
+const propTypes = {
+  uploadUrl: PropTypes.string.isRequired,
+  maxFileSize: PropTypes.string,
```
NWCalvank: An optional prop requires a default value. Our linting rules will help spot things like this. If you haven't enabled the linter in your editor, I'd encourage you to do so, since I think everyone else working on the frontend has. This file, for example, has 9 linting errors and 1 warning.

If you have any questions about the linting, you can ask Justin and/or Ryan and/or read our contributions guidelines: https://github.com/bcgov/mds/tree/develop/frontend#code-contribution-standards

Thanks!

frontend/src/components/common/FileUpload.js
```diff
@@ -28,13 +36,16 @@ class FileUpload extends React.Component {
   }
 
   render() {
+    const { maxFileSize, acceptedFileTypesMap } = this.props;
```
NWCalvank: We've decided to never use destructuring in props assignment. Not all existing files have been updated, but they're being refactored as we touch them. The goal is to avoid introducing new linting errors/warnings while gradually removing existing ones.

frontend/src/components/common/FileUpload.js
```diff
@@ -28,13 +36,16 @@ class FileUpload extends React.Component {
...
     return (
       <FilePond
+        server={this.server}
         name={"file"}
         allowRevert={false}
         allowMultiple={true}
-        maxFileSize={"100MB"}
-        server={this.server}
+        maxFileSize={maxFileSize || "100MB"}
```
NWCalvank: This shouldn't be needed once you have a default props value

frontend/src/components/common/FileUpload.js
james-hollinger: Has been removed.

frontend/src/components/common/FileUpload.js
james-hollinger: Default props have been added, and some minor refactoring has been done to address linting issues.

frontend/src/components/common/FileUpload.js
```diff
@@ -1,13 +1,26 @@
...
+import FilePondPluginFileValidateType from "filepond-plugin-file-validate-type";
 import { ENVIRONMENT } from "@/constants/environment";
 import { createRequestHeader } from "@/utils/RequestHeaders";
 
+registerPlugin(FilePondPluginFileValidateSize, FilePondPluginFileValidateType);
+
+const propTypes = {
+  uploadUrl: PropTypes.string.isRequired,
+  maxFileSize: PropTypes.string,
+  acceptedFileTypesMap: PropTypes.object,
```
NWCalvank: The way to avoid a linting error on an object with dynamic/inconsistent keys is to type-check the object values: `PropTypes.objectOf(PropTypes.string)`

james-hollinger: That works, thanks.

PR #338, opened 2018-12-20
added widgets and removed widgets from summary view

itchison:
This resolves MDS-863, MDS-876, and MDS-921.


frontend/src/components/maps/MineMap.js
```diff
@@ -75,6 +79,19 @@ class MineMap extends Component {
       });
       widgetPositionArray["top-right"] = mapGallery;
 
+      const scaleBar = new ScaleBar({
+        view,
+        container: document.createElement("scale_bar"),
+      });
+      scaleBar.unit = "metric";
+      widgetPositionArray["bottom-right"] = scaleBar;
```
NWCalvank: Does the constructor not accept the `unit` property? Seems reasonable to just do something like:
```javascript
widgetPositionArray["bottom-right"] = new ScaleBar({
  view,
  container: document.createElement("scale_bar"),
  scaleBar.unit: "metric"
});
```

Same thing could be done for legend. It would make smaller and simpler

itchison: I was going to add this prop to the constructor but the arcgis documentation was a bit unclear. Looks like it works. I also like the pattern to just add the widget to the array. I refactored the other widgets into this pattern as well.

PR #341, opened 2018-12-20
MDS-1057 back end for list of open orders

JustinMacaulay:
Added a list of open orders to the return of the compliance resource get.
Modified the tests to look for this and fixed the returns on a few error cases.


python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -33,33 +33,59 @@ def get(self, mine_no=None):
...
-                            k.get('orderAuthoritySection') == 'Section 35' for k in stop_orders)
+                    for order in stop_orders:
+                        if order.get('orderStatus') == 'Open':
+
+                            legislation = order.get('orderLegislations')
+                            permit = order.get('orderPermits')
+                            section = None
+
+                            if legislation:
+                                section = legislation[0].get('section')
```
NWCalvank: This can still throw `AttributeError: 'NoneType' object has no attribute 'get'. To be safe here, you'll want to check that legislation has an item at index 0

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -33,33 +33,59 @@ def get(self, mine_no=None):
...
+                            elif permit:
+                                section = permit[0].get('permitSectionNumber')
```
NWCalvank: Same as legislation

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -33,33 +33,59 @@ def get(self, mine_no=None):
...
+
+                            order_to_add = {
+                                'order_no': f'{report.get("assessmentId")}-{order_count}',
```
NWCalvank: Have we confirmed `report["assessmentId"]` exists anywhere? If not, we could wind up with `None-order_count`

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -33,33 +33,59 @@ def get(self, mine_no=None):
...
+                                'code_viloation': section,
+                                'report_no': report.get('assessmentId'),
+                                'inspector': report.get('assessor'),
+                                'due_date': order.get('orderCompletionDate'),
+                                'overdue': False,
+                            }
+
+                            num_open_orders += 1
+
+                            if order.get('orderCompletionDate'
```
NWCalvank: This is the sort of line break that the formatter is adding that I don't think is appropriate. It's so difficult to read, and the break seems really arbitrary

Earlier in this file we have:
```python
return self.raise_error(errhttp.response.status_code,
                                     'NRIS error occurred.' + str(errhttp))
```

Which has the closing `)` on the same line. This break has the closing `)` on a new line. Then the check after the `and` has the opening `(` on its own line.

python-backend/app/api/mines/compliance/resources/compliance.py
JustinMacaulay: I might be wrong but I was under the impression that an empty list evaluates to false in python.

NWCalvank: Ahh fair enough. I can't say I like using that level of type coercion, as these things get confusing. Especially since our two main languages in this codebase do things differently:
Python
```python
>>> legislation = []
>>> bool(legislation)
False
```

JavaScript
```javascript
> Boolean([])
> true
```

python-backend/app/api/mines/compliance/resources/compliance.py
JustinMacaulay: This is an assumption of what NRIS will always return. If there is a report then it must have an ID.

python-backend/app/api/mines/compliance/resources/compliance.py
JustinMacaulay: I manually tried to break this in a more readable better way. The formatter doesn't always make the most readable choices.

python-backend/app/api/mines/compliance/resources/compliance.py
NWCalvank: I wouldn't consider this resolved. If you're not concerned about this approach to value checking then that's up to you.

But I think a `legislation && len(legislation) > 0` would be more explicit and less reliant on the dynamic type system. Perhaps I'm the only one with this opinion, in which case I will surrender to the majority

PR #350, opened 2018-12-24
MDS-1022 Feature/ Update colours

NWCalvank:


frontend/src/styles/elements/elements.scss
```diff
@@ -12,7 +12,7 @@ p {
 h1 {
   font-size: $default-h1;
   font-weight: $bold;
-  color: $font-black;
+  color: $h1-color;
```
mmumby: When the typography gets changed, they will likely be custom classes instead of h tags, so the naming `h1-color` may no longer work at that point, but that can be addressed when that refactor occurs.

frontend/src/styles/settings/themeOverride.scss
```diff
@@ -54,28 +54,28 @@
...
   margin: 10px 0;
-  color: $default-dark-green;
-  border-color: $default-dark-green;
+  color: $tertiary-btn-color;
+  border-color: $tertiary-btn-border-color;
   border-style: dashed;
   &:hover {
-    color: $default-light-green;
-    border-color: $default-light-green;
+    color: $tertiary-btn-color-hover;
```
mmumby: are the border-color and color the same? 
could `tertiary-btn-color` & `tertiary-btn-color-hover` work for both properties?

same for primary and secondary.

mmumby: disregard, I see the difference now.

PR #354, opened 2019-01-02
Tune Nginx and other cleanup

FreshGyan:
- Remove canvas dependancy
- Move gzipping of requests out of Express and onto Nginx
- Update README's to remove redundant parts
- Add comments to Nginx configuration and switch logs from stdout to files


frontend/src/setupTests.js
```diff
@@ -8,3 +8,15 @@ Enzyme.configure({ adapter: new Adapter() });
 global.requestAnimationFrame = (callback) => {
   setTimeout(callback, 0);
 };
+
+const location = JSON.stringify(window.location);
+delete window.location;
+
+Object.defineProperty(window, "location", {
+  value: JSON.parse(location),
```
NWCalvank: What's the purpose of stringifying the location and then parsing it again?

frontend/package.json
```diff
@@ -80,13 +78,16 @@
...
     "webpack-cli": "^3.1.1",
-    "webpack-dev-server": "^3.1.10",
+    "webpack-dev-server": "^3.1.14",
     "webpack-manifest-plugin": "^2.0.3",
     "webpack-merge": "^4.1.2"
   },
   "optionalDependencies": {
     "enzyme-adapter-react-16": "^1.1.1",
     "jest": "^22.4.3",
+    "jest-canvas-mock": "^1.1.0",
```
NWCalvank: It looks like you've removed canvas entirely? Is this unrelated to canvas?

FreshGyan: Yes and no. Canvas was needed by some lower level jsdom library that throws errors if tis not there. We don't actually use canvas anywhere. So this package just patches/mocks the canvas calls in jsdom so it doenst throw the error

frontend/src/setupTests.js
FreshGyan: I didn't spend too much time cleaning that up, but basically i believe that forced JS to create a new object with to override the old window.location with. and If you don't override it properly the following global.location override fails as well

PR #356, opened 2019-01-03
MDS-1040 Feature/delete mine types

mmumby:


frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -3,31 +3,64 @@ import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import { compose } from "redux";
 import { Field, reduxForm, FieldArray, getFormValues } from "redux-form";
-import { Form, Button, Col, Row, Popconfirm, Icon, Collapse } from "antd";
+import { Form, Button, Col, Row, Popconfirm, Icon, Collapse, notification, Tag } from "antd";
+import _ from "lodash";
```
NWCalvank: We're destructuring lodash imports elsewhere. May as well do that here for consistency

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -3,31 +3,64 @@ import PropTypes from "prop-types";
...
 };
 
 export class MineRecordForm extends Component {
   state = {
     activeKey: [],
+    usedTenureTypes: [],
   };
 
+  componentDidMount() {
+    const existingMineTypes = this.props.currentMineTypes
```
NWCalvank: You have a default prop for `currentMineTypes`, so it should never be `undefined`. Mapping over an empty array will return an empty array. So, this should work just fine:
```javascript
const existingMineTypes = this.props.currentMineTypes.map(({ mine_tenure_type_code }) => mine_tenure_type_code)
this.setState({ usedTenureTypes: existingMineTypes });
```

It's also possible to use the lodash `map()` function to reduce the boilerplate, if you would like to:
```javascript
// map() is overloaded to behave like .pluck() from other libraries
const existingMineTypes = map(this.props.currentMineTypes, "mine_tenure_type_code")
this.setState({ usedTenureTypes: existingMineTypes });
```

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -43,6 +76,21 @@ export class MineRecordForm extends Component {
...
+      const diff = _.difference(prevTenure, nextTenure);
+      this.setState((prevState) => ({
+        usedTenureTypes: prevState.usedTenureTypes.filter(
+          (tenureTypes) => !diff.includes(tenureTypes)
+        ),
+      }));
+    }
     // Do nothing if no mine_types, or no change to mine_types
     if (!nextProps.mine_types || nextProps.mine_types === this.props.mine_types) {
       return;
```
NWCalvank: This looks like an early exit. It should likely be above your other logic

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -65,39 +117,134 @@ export class MineRecordForm extends Component {
   removeField = (event, fields, index) => {
     event.preventDefault();
     fields.remove(index);
+    const removedMineType = this.props.mine_types[index].mine_tenure_type_code;
+    this.setState((prevState) => ({
```
NWCalvank: There's a lodash function `without` that might be handy here: https://lodash.com/docs/4.17.11#without
Should look something like...
```javascript
this.setState((prevState) => ({
  usedTenureTypes: without(prevState.usedTenureTypes, removedMineType)
}));
```

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -65,39 +117,134 @@ export class MineRecordForm extends Component {
...
+      usedTenureTypes: prevState.usedTenureTypes.filter(
+        (tenureTypes) => tenureTypes !== removedMineType
+      ),
+    }));
+  };
+
+  // addField allows users to create a max of 4 mine Types.
+  addField = (event, fields) => {
+    const totalTypes = this.props.currentMineTypes
```
NWCalvank: Probably not necessary, but I think this might be a bit clearer:
```javascript
const totalTypes = (this.props.currentMineTypes.length || 0) + fields.length;
```

It took me a few seconds to parse the ternary, but I think the above code is a bit more direct. Again, if you have a default value of `currentMineTypes` then it shouldn't be `undefined`

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -65,39 +117,134 @@ export class MineRecordForm extends Component {
...
+      </div>
     </div>
   );
 
   render() {
     const renderTypeSelect = ({ fields }) => (
       <div>
         <Collapse activeKey={this.state.activeKey} onChange={this.setActiveKey}>
+          {this.props.currentMineTypes &&
+            this.props.currentMineTypes.map((type) => (
```
NWCalvank: At a quick scan, it doesn't look like you reference `type` on its own after this. Could likely just destructure out `mine_tenure_type_code`

frontend/src/components/Forms/MineRecordForm.js
mmumby: much cleaner

frontend/src/components/Forms/MineRecordForm.js
mmumby: removed the entire block.

frontend/src/components/Forms/MineRecordForm.js
mmumby: done

frontend/src/components/Forms/MineRecordForm.js
mmumby: Checking the lengths without the ternary or `|| 0` works great now that there is a default for `currentMineTypes`
``` const totalTypes = fields.length + this.props.currentMineTypes.length;```

File: frontend/src/components/Forms/MineRecordForm.js (hunk 6 above)
Comment (mmumby) [reply frontend/src/components/Forms/MineRecordForm.js]: `mine_tenure_type_code, mine_disturbance_code, and mine_commodity_code` are all used,

is it preferred to destructure them all out?

Comment (NWCalvank) [reply frontend/src/components/Forms/MineRecordForm.js]: Your call. If it makes the code harder to read then possibly not

File: frontend/src/customPropTypes/mines.js [hunk 7]
```diff
@@ -7,3 +7,17 @@ export const mine = shape({
   guid: PropTypes.string.isRequired,
   mine_permit: arrayOf(minePermit),
 });
+
+export const mineTypes = shape({
+  mine_tenure_type_code: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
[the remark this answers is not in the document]
mmumby: It was initiated as `[]`... but now that I think about it, it should be initiated as a string instead.

PR #357, opened 2019-01-03
VS Code ESLint path alias fix

NWCalvank:
Using an absolute path instead of a relative path.

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -1,28 +1,30 @@
-{
-  "parser": "babel-eslint",
-  "env": {
-    "browser": true,
-    "node": true,
+const path = require("path");
+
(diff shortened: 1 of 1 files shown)


frontend/.eslintrc.js
```diff
@@ -1,35 +1,37 @@
...
     "no-case-declarations": 0,
     "react/display-name": 0,
     "react/no-deprecated": 0,
     "import/no-named-as-default": 0,
     "jsx-a11y/anchor-is-valid": 0,
     "react/destructuring-assignment": ["error", "never"],
-    "camelcase": 0
-  }
-}
+    camelcase: 0,
```
mmumby: why does everything need to be in quotations except `camelcase`?

PR #361, opened 2019-01-04
feature/mds-1110 Adding new access decorators to improve testing

james-hollinger:
The way we test the access modifiers on API endpoints can be improved. See [MDS-1110](https://bcmines.atlassian.net/browse/MDS-1110) for more details.

This PR is for the new decorators. The replacing of the existing jwt decorators will be in another PR, as well as a refactor of the current auth tests.

The names of the access roles have been pulled out into constants for use in the unit tests.


james-hollinger:
The way we currently test that an endpoint has the expected access modifiers is to actually call the end point and see if it returns a 200 or 401 response. With this change, we can now test the "required_roles" parameter on the the method objects directly, without having to do a full http request.

PR #367, opened 2019-01-04
MDS-699 & MDS-1113 Feature/Mine Manager & Permittee form validation + UX

NWCalvank:
Features and Fixes:

- Fixes validation for an invalid permittee
- Fixes validation for an invalid mine manager
- Improved UX for updating a mine manager or permittee
- Makes the "large select" component more generic
- New util function for validating form inputs against the form props
- Code cleanup (PropTypes, ESLint errors)


frontend/src/components/Forms/UpdateMineManagerForm.js
```diff
@@ -31,10 +35,11 @@ export const UpdateMineManagerForm = (props) => (
             id="mineManager"
             name="mineManager"
             label="Mine Manager *"
+            placeholder="Search for a Mine Manager"
             component={renderConfig.LARGE_SELECT}
             data={props.partyIds}
             options={props.parties}
-            validate={[required]}
+            validate={[required, validMineManager]}
```
mmumby: what is the benefit of creating a new variable and destructuring the params instead of passing them directly into `validSearchSelection` ?

ie, validate would look like so: 
```
validate={[required, validSearchSelection("parties", "Invalid Mine Manager")]}
```

NWCalvank: Due to the way that redux forms handle rerendering, passing a function in-line to the validate attribute causes a recursive rendering loop. You have to pull the function out of the render method to prevent this, hence the extra variable being created outside of the exported component.

As far as the destructuring, I think it is beneficial to pass only a single argument (an object in this case) when dealing with a manually-curried function. If we need to add arguments in the future (say, an options array to use instead of the props object), it is easier to add that as an object property than to add a new position argument.

Open to changing this, but that's why I opted for this approach here.

PR #369, opened 2019-01-04
updated user experience bug when changing tenure options for mine_types

mmumby:


frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -83,6 +83,16 @@ export class MineRecordForm extends Component {
         ),
       }));
     }
+    // when mine_type changes, update this.state.usedTenureTypes
+    if (this.props.mine_types !== nextProps.mine_types) {
+      const nextNewTenure = nextProps.mine_types
+        .filter(({ mine_tenure_type_code }) => mine_tenure_type_code)
```
NWCalvank: I think this use of type coercion is fairly common in JS, so it's probably not a big deal, but if we want to be a bit more explicit (which is my personal preference), we could do something like this:
```javascript
import { isEmpty } from "lodash";

const nextNewTenure = nextProps.mine_types
  .filter((x) => !isEmpty(x))
```
or
```javascript
import { isEmpty, negate } from "lodash";

const nextNewTenure = nextProps.mine_types
  .filter(negate(isEmpty))
```

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -83,6 +83,16 @@ export class MineRecordForm extends Component {
...
+        .map(({ mine_tenure_type_code }) => mine_tenure_type_code);
+      const nextExistingTenure = map(nextProps.currentMineTypes, "mine_tenure_type_code");
+      const combined = nextExistingTenure.concat(nextNewTenure);
```
NWCalvank: Could us ES6 syntax:
```javascript
const combined = [... nextExistingTenure, ...nextNewTenure];
```

frontend/src/components/Forms/MineRecordForm.js
mmumby: yeah, I like this better.

PR #372, opened 2019-01-04
MDS-1029 refactor mine identity and mine detail

JustinMacaulay:
Removing the mine_detail table and adding its relevant columns to the mine_identity table. Then renaming the mine_identity table to mine.


NWCalvank:
@JustinMacaulay This will require ETL updates as well. It looks like all of the existing ETL files reference the `mine_detail` table

JustinMacaulay:
Thanks, and I have all of that on a checklist for this branch. I just PR'd this so that I can have other people test it in dev to ensure that the changes I have made don't break everything, since it is such a fundamental change.

JustinMacaulay:
There is only one aspect of this branch that I'm still working on but I believe it is ready for review.

python-backend/app/api/mines/mine/models/mine_identity.py
```diff
@@ -62,24 +61,41 @@ def json(self):
...
+            'mine_name':
+            self.mine_name,
+            'mine_no':
+            self.mine_no,
+            'mine_note':
+            self.mine_note,
+            'major_mine_ind':
+            self.major_mine_ind,
+            'region_code':
+            self.mine_region,
```
NWCalvank: It doesn't look like these short lines need to be wrapped. The lines below this comment are longer and aren't wrapped.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -198,45 +210,28 @@ def put(self, mine_no_or_guid):
...
-                not tenure and
-                not (lat and lon) and
-                not mine_name and
-                not mine_note and
-                not status and
-                not region and
-                major_mine_ind is None
-           ):
+        if (not tenure and not (lat and lon) and not mine_name and not mine_note and not status
+                and not region and major_mine_ind is None):
```
NWCalvank: This format seems considerably worse to me than having them on separate lines.

JustinMacaulay:
That should be happening now I just was leaving it till last as long as I had the time. It's the last part I'm working on but the functional changes are in place. Im just name changing at the moment.

python-backend/tests/mines/models/test_mine_detail_model.py
```diff
@@ -1,57 +1,52 @@
...
-def test_mine_detail_model_validate_mine_no(test_client, auth_headers):
-    with pytest.raises(AssertionError) as e:
-        MineDetail(
-            mine_detail_guid=uuid.uuid4(),
-            mine_guid=uuid.uuid4(),
-            mine_no=''.join(['{}'.format(x) for x in range(11)]),
-            mine_name='test_name'
-        )
-    assert 'Mine number must not exceed 10 characters.' in str(e.value)
+# import uuid
```
NWCalvank: We should just remove this file, no?

JustinMacaulay: yeah 100%, deleted it .

python-backend/app/api/mines/mine/models/mine_identity.py
JustinMacaulay: Ive tried to manually make them stay on one line but the formatter forces them to another line.

migrations/sql/V1.31__Drop_Mine_Detail_Table.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE mine_detail DROP CONSTRAINT mine_detail_mine_region_fkey;
```
nayr974: You may need a COMMIT; here between altering and dropping, otherwise it may fail validation for the drop because the alter will not have executed.

JustinMacaulay: Added a commit just to be safe.

PR #374, opened 2019-01-07
MDS-1114: rename geom field

ActionAnalytics:


ActionAnalytics:
There is some discussion happening regarding keeping the "geom" convention or going to forward with this change.

ActionAnalytics:
Team decided to keep the original naming convention.

PR #378, opened 2019-01-07
updated logout button to be consistent with other dropdowns

mmumby:


frontend/src/components/navigation/NavBar.js
```diff
@@ -17,25 +17,38 @@ const propTypes = {
...
+        <Link
+          to={router.MINE_DASHBOARD.dynamicRoute({
+            page: String.DEFAULT_PAGE,
+            per_page: String.DEFAULT_PER_PAGE,
+          })}
+        >
+          <img alt="Home" className="menu__img" src={HOME} />
+        </Link>
+        <Dropdown overlay={this.menu} placement="bottomLeft">
+          <button type="button" style={{ color: "#ffffff" }}>
```
NWCalvank: Do we hardcode this into the other dropdowns? I'm curious about making a shared CSS class (like the `__dropdown-link` modifier removed in this PR) or starting a JS constants file for colours. Thoughts?

mmumby: yeah good question. 

The other dropdowns are all real buttons, ie secondary or primary buttons. 
This one has no button type styling.

so no, the other dropdowns do not have the inline-styling, and this could be a case for a specific class, but as it stands now it is the only style of this type. 

and yes, a color constant file is a very good idea.

mmumby: I'll switch this to be a class though.

PR #380, opened 2019-01-07
Display the basemap and warning on Arcgis api failure

FreshGyan:
- Added a warning message (Need a check on the language of it)
- Add a onFail handler to load the basic Map in case the Arcgis portal item fails
- Swapped the map on summary page to use Map instead of WebMap as well, since it's not using any of the WebMap functionalities

DEV URL TO VIEW - https://mds-empr-mds-dev.pathfinder.gov.bc.ca/380/dashboard/?page=1&per_page=25&map=true


frontend/src/components/maps/MineMap.js
```diff
@@ -23,7 +24,7 @@ const defaultProps = {
 };
 
 class MineMap extends Component {
-  state = { map: null, view: null, center: null, zoom: null };
+  state = { map: null, view: null, center: null, zoom: null, mapFailedToLoad: null };
```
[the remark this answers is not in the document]
NWCalvank: It's probably more self-documenting (saying what will eventually be set in the state), but if they're all initialized to `null`/`undefined` then it's the same result as not having initialized them at all.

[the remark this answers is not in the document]
NWCalvank: I like this a lot 👍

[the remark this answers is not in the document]
NWCalvank: So, here's a thought...

With default props appropriately set, is there any reason not to just do something like:
```javascript
state = { map: this.props.map, view: this.props.view, center: this.props.center, zoom: 5, mapFailedToLoad: false };
```

Required props will set the state to the appropriate value, and optional props will fallback to the default props (such as those you included in your comment above).

Also note: the default prop for `mine` shouldn't be `null` if the type is an object. And `object` should be replaced with our custom prop for `mine` @FreshGyan

frontend/src/components/maps/MineMap.js
```diff
@@ -14,16 +15,28 @@ import * as String from "@/constants/strings";
...
+  map: {},
+  view: {},
+  lat: null,
+  long: null,
 };
 
 class MineMap extends Component {
-  state = { map: null, view: null, center: null, zoom: null };
+  state = {
+    map: this.props.map,
```
NWCalvank: Looking at `handleLoadMap`, it looks like `map` and `view` are only ever used for state? Are they ever actually passed in as props to this component?

If they aren't then we don't need them in the propTypes.

If they are, you'll have to identify the properties/methods that exist on map and view respectively which this component leverages and update the propTypes to include those properties/methods using `PropTypes.shape()`. `PropTypes.object` is forbidden by our linting rules.

frontend/src/components/maps/MineMap.js
```diff
@@ -14,16 +15,28 @@ import * as String from "@/constants/strings";
...
   lat: PropTypes.number,
   long: PropTypes.number,
 };
 
 const defaultProps = {
   mine: null,
+  map: {},
+  view: {},
+  lat: null,
+  long: null,
```
NWCalvank: I trust that the component gracefully handles these `null` values?

frontend/src/components/maps/MineMap.js
FreshGyan: So if I leave the center as an empty array, a LocationPin appears on the top left of the map.
My Guess is, an empty list is somehow processed as [0,0] by the mapping library. But if you keep it null, it gets ignored

frontend/src/components/maps/MineMap.js
```diff
@@ -120,7 +138,25 @@ class MineMap extends Component {
...
+      );
+    }
+    if (this.state.mapFailedToLoad) {
+      return (
+        // Fallback to default map if any of the layers fail to load
+        <Map
+          style={{ width: "100vw", height: "100vh" }}
+          mapProperties={{ basemap: "topo" }}
+          viewProperties={{
+            center: [this.props.long, this.props.lat],
```
NWCalvank: If lat/long aren't required props then they need default props values. If this component breaks when they're `null` then they should be required.

Commenting here because I need code changes in order to comment on a line.

PR #387, opened 2019-01-09
MDS-865 Feature/ Pull Tenure Types from MMS

NWCalvank:


JustinMacaulay:
Looks good to me.

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -147,6 +150,43 @@ BEGIN
         (new.lat_dec <> 0 AND new.lon_dec <> 0);
     SELECT count(*) FROM mine into new_row;
     SELECT count(*) FROM mine_location into location_row;
+
+    -- Upsert data from new_record into mine_type
+    WITH new_record AS (
+        SELECT *
```
itchison: It looks like sometimes we * here and sometimes we list the columns explicitly. I prefer the latter especially when we start mapping them below.

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -147,6 +150,43 @@ BEGIN
...
+        mine_tenure_type_code,
+        create_user          ,
+        create_timestamp     ,
+        update_user          ,
+        update_timestamp     )
+    SELECT
+        gen_random_uuid()   ,
+        new.mine_guid       ,
+        CASE
+          when new.mine_typ = ANY('{CX,CS,CU}'::text[]) THEN 'COL'
```
itchison: Looks like we're collapsing types here. Where did we get this mapping?

NWCalvank: From business (Aaron and Brian)

migrations/etl/1_Mine_profile_detail_location.sql
NWCalvank: I can select the columns explicitly if that's the pattern we want to follow

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -147,6 +150,45 @@ BEGIN
         (new.lat_dec <> 0 AND new.lon_dec <> 0);
     SELECT count(*) FROM mine into new_row;
     SELECT count(*) FROM mine_location into location_row;
+
+    -- Upsert data from new_record into mine_type
+    WITH new_record AS (
+        SELECT
+            mine_guid,
+            mine_typ
+        FROM ETL_PROFILE
```
nayr974: It looks like ETL_PROFILE is only populated when a mine in MMS has not already been created in MDS. I think this will be fine for the upcoming release as we are planning on wiping and repopulating production data, but this pattern will not work in the future if we need to pull in new data for mines that already exist.

NWCalvank: Yeah, there was a discussion this morning about the plan for migrating to MDS as the source of truth. Apparently this discussion has been happening since March. When a business decision has been made on it, we can update ETL to exclude certain fields / pull all mines / etc.

nayr974: Personally, I find that unlikely as there will be a lot of functionality not implemented in MDS and users will likely continue to use and update data in MMS for years. That discussion is way outside of the scope of this ticket :) Thanks!

PR #388, opened 2019-01-09
Feature/mds 1159 unlink files from expected documents

JustinMacaulay:
This is just the backend to add functionality to remove the link between an uploaded document and the expectation it is fulfilling.


python-backend/app/api/documents/mines/models/mine_document.py
```diff
@@ -31,3 +31,7 @@ def json(self):
     @classmethod
     def find_by_mine_guid(cls, mine_guid):
         return cls.query.filter_by(mine_guid=mine_guid).filter_by(active_ind=True).all()
+
+    @classmethod
+    def find_by_mine_document_guid(cls, mine_document_guid):
+        return cls.query.filter_by(mine_document_guid=mine_document_guid).first()
```
Jsyro: i wrote basically the same method for my PR, this should filter by active_ind. Even though active_ind will be changed. better to have it here and change rather than remembering to add it later.

PR #389, opened 2019-01-09
Feature/mds 852 link exp doc to existing file

Jsyro:
post now accepts existing mine_document_guid OR file. and a test for it.


python-backend/app/api/documents/expected/resources/expected_document_uploads.py
```diff
@@ -46,50 +45,65 @@ def post(self, expected_document_guid):
             )
 
         expected_document = MineExpectedDocument.find_by_exp_document_guid(expected_document_guid)
+        if not expected_document:
+            return self.create_error_payload(500, f'expected document not found'), 500
```
NWCalvank: A 500 seems like an incorrect response code for this. A 4-something not-found seems more appropriate to me

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
```diff
@@ -46,50 +45,65 @@ def post(self, expected_document_guid):
...
-                    **self.get_create_update_dict())
-
-                expected_document.mine_documents.append(doc)
-                db.session.add(expected_document)
-                filenames.append(value)
-
+        if data.get('mine_document_guid'):
+            existing_mine_doc = MineDocument.find_by_guid(data.get('mine_document_guid'))
+            if not existing_mine_doc:
+                return self.create_error_payload(500, 'mine_document not found'), 500
```
NWCalvank: Same question about response code here

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
Jsyro: yea, this should be 400, not 500.

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
Jsyro: changed to 400

PR #397, opened 2019-01-11
Feature/mds-852 File picker component

james-hollinger:
Adding a new front-end component that complements fileUpload to allow a user to select from already uploaded documents on the mine. 

filePicker contains the fileUpload component as well as a dropdown to select existing files. These are toggled with a radio button selector. To keep the component generic, the component accept a list of dropdown items and a callback function to handle selecting files. To allow this, renderSelect was modified to expose the onSelect prop.

The next PR will include an adapter around this component to handle the redux store calls.


frontend/src/components/common/FilePicker.js
```diff
@@ -0,0 +1,76 @@
...
+      value: PropTypes.string,
+      label: PropTypes.string,
+    })
+  ).isRequired,
+  onSelectExisting: PropTypes.func.isRequired,
+};
+
+const defaultProps = {
+  maxFileSize: undefined,
+  acceptedFileTypesMap: undefined,
```
NWCalvank: I'm fine with this, but the pattern I've seen elsewhere in the codebase has been to use `null` instead

james-hollinger: I want to defer to the default props in FileUpload. Setting these to null won't use the defaults.

PR #399, opened 2019-01-11
MDS-944: Add geo to python

ActionAnalytics:
Did complete build, generated test data, logged in via front end and viewed entries in list and points on maps.  There is nothing hooked up to the geom field at the moment, just stubs in place until we retire the old latitude and longitude fields.


python-backend/app/api/mines/location/models/mine_location.py
```diff
@@ -2,6 +2,8 @@
 import uuid
 
 from sqlalchemy.dialects.postgresql import UUID
+from geoalchemy2 import Geometry, functions, types
+import json
```
NWCalvank: I see `Geometry` being used, but it looks like `functions`, `types`, and `json` unused imports?

PR #401, opened 2019-01-11
MDS-1159 unlink files from expected documents

JustinMacaulay:
Adds the front end capabilities for unlinking a document from an expected document.


frontend/src/components/common/UploadedFilesList.js
```diff
@@ -0,0 +1,113 @@
...
+      .then(this.props.fetchMineRecordById(this.props.mineGuid));
+  };
+
+  canUnlink = (selectedDoc) => {
+    const docStatus = this.props.expectedDocumentStatusOptions.find(
+      (status) => status.value === selectedDoc.exp_document_status_guid
+    );
+
+    if (moment(selectedDoc.due_date, "YYYY-MM-DD") > moment()) {
+      if (docStatus.label === "Not Received" || docStatus.label === "Received / Pending Review") {
```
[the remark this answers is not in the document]
JustinMacaulay: Ticket is made and Aaron and Dan are tagged.

frontend/src/customPropTypes/documents.js
```diff
@@ -0,0 +1,29 @@
...
+  mine_guid: PropTypes.string,
+  exp_document_name: PropTypes.string,
+  exp_document_description: PropTypes.string,
+  related_documents: arrayOf(mineDocument),
+  received_date: PropTypes.string,
+  exp_document_status_guid: PropTypes.string,
+  due_date: PropTypes.string,
+});
+
+export const documentStatus = shape({
```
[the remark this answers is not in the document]
JustinMacaulay: Removed.

frontend/src/components/common/UploadedFilesList.js
```diff
@@ -0,0 +1,113 @@
...
+import {
+  removeMineDocumentFromExpectedDocument,
+  fetchMineRecordById,
+} from "@/actionCreators/mineActionCreator";
+import { getMines } from "@/selectors/mineSelectors";
+import { getExpectedDocumentStatusOptions } from "@/selectors/staticContentSelectors";
+
+const propTypes = {
+  selectedDocGuid: PropTypes.string.isRequired,
+  mineGuid: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
JustinMacaulay: I replaced these with Id instead of guid

frontend/src/components/common/UploadedFilesList.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  canUnlink = (selectedDoc) => {
+    const docStatus = this.props.expectedDocumentStatusOptions.find(
+      (status) => status.value === selectedDoc.exp_document_status_guid
+    );
+
+    if (moment(selectedDoc.due_date, "YYYY-MM-DD") > moment()) {
+      if (docStatus.label === "Not Received" || docStatus.label === "Received / Pending Review") {
+        return true;
```
NWCalvank: Returning a boolean inside of a conditional is redundant. You can just return the condition instead 👍 

In this case, it'll be the two conditions, joined. For readability, it may make sense to give them helpful names and then just `return check1 && check2` (with the helpful names instead of check#)

This also means you don't need the second `return` statement

frontend/src/components/common/UploadedFilesList.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  render() {
+    const selectedDoc = this.props.mines[this.props.mineId].mine_expected_documents.find(
+      (expDoc) => expDoc.exp_document_guid === this.props.selectedDocId
+    );
+    const unlink = this.canUnlink(selectedDoc);
+    return (
+      <div>
+        {selectedDoc.related_documents.map((file, key) => (
```
NWCalvank: I don't have a super strong opinion on this, and I can see why you called it `key`, but I think it may be best to use the documented name here (`i`, `index`, or `idx`) since that's what this param actually is in the context of a `map`

frontend/src/customPropTypes/documents.js
```diff
@@ -0,0 +1,24 @@
+import { PropTypes, shape, arrayOf } from "prop-types";
+
+// This file is anticipated to have multiple exports
+// eslint-disable-next-line import/prefer-default-export
```
NWCalvank: This isn't needed now that there are two exports. Please remove

frontend/src/customPropTypes/mines.js
```diff
@@ -1,15 +1,18 @@
 import { PropTypes, shape, arrayOf } from "prop-types";
 import { minePermit } from "@/customPropTypes/permits";
+import { mineExpectedDocument } from "@/customPropTypes/documents";
 
 // This file is anticipated to have multiple exports
 // eslint-disable-next-line import/prefer-default-export
```
NWCalvank: It doesn't look like this comment is needed anymore either now that there are multiple exports. Could you remove this one two? Thanks

frontend/src/styles/settings/variables.scss
```diff
@@ -59,6 +59,9 @@ $pagination-item-hover: darken($violet, 20%);
 $pagination-item-border-hover: $pagination-item-hover;
 $card-border-color: $violet;
 
+//Backgrounds
+$secondary-background-colour: $lightest-grey;
```
NWCalvank: I'm not sure that we need two separate `// Backgrounds` sections. And I'm open to input here (cc: @mmumby), but I don't think we need to alias our background colours

If we are going to do that, could you do a quick find-and-replace of the current uses of `$lightest-grey` with your alias? I used `$lightest-grey` in a couple of places already, and having two references to the same background colour defeats the purpose of the variables

frontend/src/tests/components/common/UploadedFilesList.spec.js
```diff
@@ -0,0 +1,32 @@
...
+const props = {};
+const dispatchProps = {};
+
+const setupProps = () => {
+  props.selectedDocId =
+    MOCK.MINES.mines[MOCK.MINES.mineIds[0]].mine_expected_documents[0].exp_document_guid;
+  props.mineId = MOCK.MINES.mineIds[0];
+  props.expectedDocumentStatusOptions = MOCK.EXPECTED_DOCUMENT_STATUS_OPTIONS.options;
+  props.mines = MOCK.MINES.mines;
+};
```
NWCalvank: If these props are required in the component, should they not be required by the tests?
removeMineDocumentFromExpectedDocument
fetchMineRecordById
expectedDocumentStatusOptions

python-backend/app/api/documents/expected/resources/mine_documents.py
```diff
@@ -47,6 +48,8 @@ def post(self, mine_guid):
         data = self.parser.parse_args()
         doc_list = data['documents']
         mine_new_docs = []
+        doc_status_list = ExpectedDocumentStatus.find_all_document_status()
+        not_received = next(filter(lambda x: x.description == 'Not Received', doc_status_list))
```
NWCalvank: If "Not Received" is a special kind of status, perhaps it deserves its own method on the model. I'm not sure this is appropriate logic to have in the resource

migrations/sql/V1.35__alter_mine_expected_document.sql
```diff
@@ -0,0 +1,12 @@
...
+    SELECT exp_document_status_guid 
+    FROM mine_expected_document_status 
+    WHERE description = 'Not Received'
+    limit 1
+) 
+    WHERE exp_document_status_guid IS NULL;
+
+COMMIT;
+
+ALTER TABLE mine_expected_document ALTER COLUMN exp_document_status_guid SET NOT NULL;
```
NWCalvank: The model should be updated to reflect the non-nullable column

frontend/src/components/common/UploadedFilesList.js
```diff
@@ -0,0 +1,113 @@
...
+          <div className="padding-small margin-small lightest-grey-bg" key={key}>
```
mmumby: The index is being names and used as a key in this situation, there should be a warning in the console as react doesn't like the index being used as the key. 

instead, `file` should have a `guid` (perhaps `file.mine_document_guid`?), use that instead of the index.

frontend/src/styles/settings/variables.scss
JustinMacaulay: I only did this as it was suggested that I should by @mmumby, I can remove it if necessary.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -304,3 +304,30 @@ export const fetchMineBasicInfoList = (mine_guids) => (dispatch) => {
...
+      notification.success({
+        message: "Successfully removed the document from the report.",
+        duration: 10,
+      });
+      dispatch(success(reducerTypes.REMOVE_MINE_EXPECTED_DOCUMENT));
+      dispatch(hideLoading());
+      return response;
+    })
+    .catch(() => {
+      notification.error({ message: String.ERROR, duration: 10 });
```
mmumby: Does this endpoint return a custom error message? 
`String.ERROR` should be used as a backup.

frontend/src/customPropTypes/documents.js
mmumby: This can be deleted now as the file contains multiple exports and ESLint is no longer mad.

frontend/src/customPropTypes/mines.js
mmumby: same here, comment can be deleted.

python-backend/app/api/documents/expected/resources/mine_documents.py
JustinMacaulay: There is currently a ticket in the backlog to refactor all of this as the ExpectedDocumtneStatus is still using a guid instead of a 3 character code. That refactor was decided to be out of scope for this ticket, and that should handle this issue.

migrations/sql/V1.35__alter_mine_expected_document.sql
JustinMacaulay: I missed that, good chatch!

frontend/src/styles/settings/variables.scss
mmumby: Thanks for putting this in variable!!

JustinMacaulay: see @NWCalvank comment above.

frontend/src/components/common/UploadedFilesList.js
JustinMacaulay: Changed.

frontend/src/actionCreators/mineActionCreator.js
JustinMacaulay: It does return a custom message I just missed this when I grabbed another action creator to base mine off of.

frontend/src/components/common/UploadedFilesList.js
JustinMacaulay: changed to a single return.

frontend/src/styles/settings/variables.scss
mmumby: Yeah, I did suggest this to him as he is using it as the background for a smaller div, and if the background color changes I didn't want that div to get forgotten. 

I like Nathans suggestion to replace the other background colors with this variable.

frontend/src/tests/components/common/UploadedFilesList.spec.js
JustinMacaulay: All of these should be mocked appropriately in the test and I don't believe you can require props in tests.

python-backend/app/api/documents/expected/resources/mine_documents.py
NWCalvank: I don't see how this is related to using a guid vs a 3 character code? This check is being done against the description, and the logic is happening in the resource. My request was just that it be moved to the model

frontend/src/styles/settings/variables.scss
JustinMacaulay: All have been replaced.

frontend/src/tests/components/common/UploadedFilesList.spec.js
NWCalvank: I had just meant that the test should fail if not all required props are provided. I didn't notice that a couple of them were set up in `setupDispatchProps`

python-backend/app/api/documents/expected/resources/mine_documents.py
JustinMacaulay: I originally misread the question, I'm sorry for the confusion on my part. I understand what your asking and I have added a new function on the model to find_by_description as doing that at the SQL level will be far more efficient than filtering the entire returned list.

PR #402, opened 2019-01-11
Feature/admin view

mmumby:
Basic functionality for admin view to unblock copper. 

not fully admin protected, will continue building out the permissions in another PR.

### DIFF
--- a/frontend/src/tests/components/admin/AdminDashboard.spec.js
+++ b/frontend/src/tests/components/admin/AdminDashboard.spec.js
@@ -1,0 +1,10 @@
+import React from "react";
+import { shallow } from "enzyme";
+import { AdminDashboard } from "@/components/admin/AdminDashboard";
+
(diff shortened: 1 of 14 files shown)


frontend/src/tests/components/admin/AdminDashboard.spec.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { shallow } from "enzyme";
+import { AdminDashboard } from "@/components/admin/AdminDashboard";
+
+describe("AdminDashboard", () => {
+  it("renders properly", () => {
```
NWCalvank: I'd really like to see us use our snapshot tests to validate that things are rendering "properly." Not in scope of this PR.

mmumby: let's chat about this sometime this sprint!

PR #403, opened 2019-01-11
Feature/mds 1091 admin user management

Jsyro:
Sql and backend for minespace user management. these tables could be moved to a different application in the future.

If a row exists in the access table then they have access.


python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @api.doc(params={'user_id': 'User id.'})
+    @requires_role_mine_admin
+    def get(self, user_id=None):
+        if user_id or request.args.get('email'):
+            user = MinespaceUser.find_by_id(user_id)
+            if not user:
+                user = MinespaceUser.find_by_email(request.args.get('email'))
+                if not user:
+                    return self.create_error_payload(500, "user not found"), 500
```
JustinMacaulay: Should this be a 500? I think it would be better as a 404.

Jsyro: Done

python-backend/app/api/users/minespace/models/minespace_user.py
```diff
@@ -0,0 +1,41 @@
...
+    mines = db.relationship('MinespaceUserMine', backref='user', lazy='joined')
+
+    def json(self):
+        result = {
+            'id': str(self.id),
+            'keycloak_guid': str(self.keycloak_guid or ''),
+            'email': self.email,
+            'mines': [str(x.mine_guid) for x in self.mines]
+        }
+        return result
```
NWCalvank: Could just return the dict directly

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -0,0 +1,53 @@
...
+    parser = reqparse.RequestParser()
+    parser.add_argument('email', type=str, location='json', required=True)
+    parser.add_argument('mine_guids', type=list, location='json', required=True)
+
+    @api.doc(params={'user_id': 'User id.'})
+    @requires_role_mine_admin
+    def get(self, user_id=None):
+        if user_id or request.args.get('email'):
+            user = MinespaceUser.find_by_id(user_id)
+            if not user:
```
NWCalvank: I think this would be easier to read if the email check was not nested inside of the id check.

If you just check for and `user_id` first, you can check `if not user and email` after without needing to nest them. I think this should give you the same behaviour.

And then after that, the `if not user` -> not found works the same, but all of the nesting is removed.

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -0,0 +1,53 @@
...
+                user = MinespaceUser.find_by_email(request.args.get('email'))
+                if not user:
+                    return self.create_error_payload(404, "user not found"), 404
+            return user.json()
+        users = MinespaceUser.query.filter_by(deleted_ind=False).all()
```
NWCalvank: It seems like `deleted_ind` should be known/managed by the model. A `users` class method might be good here instead

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -0,0 +1,53 @@
...
+        return {'users': [x.json() for x in users]}
+
+    @api.doc(params={'user_id': 'User id.'})
+    @requires_role_mine_admin
+    def post(self, user_id=None):
+        if user_id:
+            return self.create_error_payload(400, "unexpected user guid"), 400
```
NWCalvank: Is there a benefit to this check compared to the default invalid route error that I thiiink would occur here?

migrations/sql/V1.35__add_user_table.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE minespace_users (
+    id serial PRIMARY KEY,
+    keycloak_guid uuid,
+    email character varying(100) NOT NULL,
```
nayr974: Can you please add comments to these tables describing their use? ie. Email is intended to be the unique BCEID email being used as our primary identifier, and that mds_mine_access is intended to represent any ability to access data related to these mines.

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -0,0 +1,53 @@
...
+        data = self.parser.parse_args()
+        new_user = MinespaceUser(email=data.get('email'))
```
NWCalvank: I'm seeing a mix of using class constructors like this as well as a `create_classname` class method throughout the codebase. Last I spoke to @FreshGyan (while building mine types), I believe the consensus was to use the class method approach.

As a team, I think we should determine which approach we're going to use.

Perhaps this was discussed in the "fatter" models part of the db standards presentation and I have since forgotten, but the following lines of code (db interfacing + record management) seem better suited for a model-level method to me.

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @api.doc(params={'user_id': 'User id.', 'mine_guid': 'MDS Mine Guid'})
+    @requires_role_mine_admin
+    def post(self, user_id=None, mine_guid=None):
+        if not user_id:
+            return self.create_error_payload(400, "user_id not found"), 400
+        if mine_guid:
+            return self.create_error_payload(400, "unexpected mine_guid"), 400
+        data = self.parser.parse_args()
+        guid = uuid.UUID(data.get('mine_guid'))  #ensure good formatting
```
NWCalvank: I definitely need this explained to me. Explicitly checking for the lack of a `mine_guid` and then pulling a `mine_guid` from the arguments?

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
```diff
@@ -0,0 +1,49 @@
...
+
+        user = MinespaceUser.find_by_id(user_id)
+        user.mines.append(MinespaceUserMine(user_id=user.id, mine_guid=guid))
+        user.save()
```
NWCalvank: I would expect we should wrap these save calls in a `try` to gracefully handle a failure

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
```diff
@@ -0,0 +1,49 @@
...
+        found = False
+        for mum in user.mines:
+            if str(mum.mine_guid) == mine_guid:
+                db.session.delete(mum)
+                found = True
+                break
+        if found:
+            user.save()
+            return ('', 204)
+        return self.create_error_payload(500, 'not user mine relation found')
```
NWCalvank: Neither of the 500 errors in this file are valid 500 errors, and there is a typo in this error message

migrations/sql/V1.35__add_user_table.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE minespace_users (
```
klefler: This may be picky, but we typically don't pluralize table names. Suggest we call this table minespace_user. However, if it is a ton of work to refactor, we can leave it.

migrations/sql/V1.35__add_user_table.sql
```diff
@@ -0,0 +1,17 @@
...
+    id serial PRIMARY KEY,
```
klefler: Please rename this column to "user_id" so that it matches the foreign key name we plan to use.

migrations/sql/V1.35__add_user_table.sql
garywong-bc: Is the `minespace_` prefix meant to denote a higher-order 'type of table'?  @kristimeredith  how do you feel about `mds_users` to denote that it's "application wide"?

klefler: @garywong-bc did you mean to tag me in this?
If so, yes, mds_user would be a good name... not plural ;)

Jsyro: Users in the table will come from the minespace (public facing) website, (Currently a different front end for the existing backend and db). The long term vision is for minespace to be freestanding application where this table would be transplanted into that application's db. mds_users would be inspectors and other EMPR staff. That's why the prefix minespace, is being used.

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: Agreed.

migrations/sql/V1.35__add_user_table.sql
Jsyro: changed,

migrations/sql/V1.35__add_user_table.sql
Jsyro: changed

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
Jsyro: the mine_guid url parameter is not used on post, all incoming data is pulled from form-data, not url parameters. There should not be a guid at the end of this resource `users/minespace/<user_id>/mines/<string:mine_guid>` but there needs to be a kvp of mine_guid for a list of strings..... it's definitely awkward.... Recommendations?

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
Jsyro: Given that these are values provided in the url, agreed.

migrations/sql/V1.35__add_user_table.sql
```diff
@@ -0,0 +1,20 @@
...
+    deleted_ind boolean NOT NULL DEFAULT FALSE,
+
+    UNIQUE (email),
+    UNIQUE (keycloak_guid)
+);
+
+
+COMMENT ON TABLE minespace_user IS 'A list of public mine proponents that can log into minespace and update mds information on mines listed in minespace_user_mds_mine_access';
+
+CREATE TABLE minespace_user_mds_mine_access(
```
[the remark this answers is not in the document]
Jsyro: good catch.

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
Jsyro: done.

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: Yea, old habit, having create methods is the pattern we are going to go with going forward. Refactored

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: `api.add_resource(MinespaceUserResource, '/minespace', '/minespace/<user_id>')` both of these urls send to this resource, but we need make sure for post they only use the first one, sending a POST with a guid is invalid. editing an existing record should be a PUT.

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: Old habit, refactored to use create methods.

python-backend/app/api/users/minespace/models/minespace_user.py
Jsyro: changed

python-backend/app/__init__.py
```diff
@@ -59,7 +60,7 @@ def register_routes(app):
     api.add_namespace(permits_api)
     api.add_namespace(document_api)
     api.add_namespace(document_manager_api)
-
+    api.add_namespace(users_api)
```
itchison: This doesn't support splitting the service out into minespace API in the future.

Jsyro: Users is registered to the app as one namespace and minespace is a submodule. I see your point. maybe it should be reversed. /minespace/users/ instead of users/minespace/....

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: Explored this suggestion but I prefer the current version of code for readability. Let me know if you want to revisit this though.

python-backend/app/api/users/minespace/resources/minespace_user.py
NWCalvank: Weird. But it looks like this is consistent with our existing patterns, so I'll mark these comments as resolved

PR #408, opened 2019-01-13
Restructure Database files and versioning

FreshGyan:
- Removed existing undo and create database scripts
- Added one big SQL script to encompass every database migration ran to this day.

Highly suggest for everyone to brush up on Flyway migrations while we're at it
https://flywaydb.org/documentation/migrations

NOTE to self: Inform team to not put commit statements in the middle of transactions.


migrations/sql/V2019.01.12.06.48__base_schema.sql
```diff
@@ -0,0 +1,1515 @@
+--
+-- Name: pgcrypto; Type: EXTENSION; Schema: -; Owner:
+--
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto WITH SCHEMA public;
```
garywong-bc: technically, this is un-necessary as https://github.com/bcgov/openshift-postgresql-oracle_fdw/blob/bbf74c4f571c8d064c3c300f5ded026ade04eff2/root/usr/share/container-scripts/postgresql/common.sh#L224 handles it (the oracle_fdw deployment actually has a parameter to choose Y/N for pgcrypto and postgis extensions)

FreshGyan: Agreed! I added that in there because our local database dockerfile doesnt come with pgcrypto. and it seemed more ugly to update the dockerfile 😬

PR #410, opened 2019-01-14
Feature/authentication upgrade

mmumby:
- renamed HOC to AuthenticationGuard and AuthorizationGuard. 

 - AuthorizationGuard  handles the rendering of Route component
Created AuthorizationWrapper to handle the rendering of non-route components, such as buttons, dropdowns, etc, 
- deleted ConditionalButton as AuthorizationWrapper is more versatile.


frontend-public/src/components/Home.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
-import { AuthGuard } from "../HOC/AuthGuard";
+import { AuthenticationGuard } from "../HOC/AuthenticationGuard";
 
 export const Home = () => <div>hello proponents</div>;
 
-export default AuthGuard(Home);
+export default AuthenticationGuard(Home);
```
nayr974: Nice. I like being more verbose with names.

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -13,4 +14,4 @@ export const AdminDashboard = () => (
   </div>
 );
 
-export default AdminDashboard;
+export default AuthorizationGuard("role_admin")(AdminDashboard);
```
NWCalvank: Do we want to put these auth keys into constants?

frontend/src/components/common/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+
+const propTypes = {
+  permission: PropTypes.string.isRequired,
+  children: PropTypes.element.isRequired,
+};
+
+export const AuthorizationWrapper = (props) =>
+  props.userRoles.includes(USER_ROLES[props.permission]) && <div>{...props.children}</div>;
```
nayr974: As we have constants defined for these, would it be better to use the constants in each component as well? Pass USER_ROLES.role_create as the permission prop instead of finding it here? 

Alternatively, we could use the role name directly in each view "mds-mine-admin". This approach seems like we still using magic strings in our components, but adding another set of magic strings we need to know. I don't really have strong feelings here, but thoughts on this?

frontend/src/components/common/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,48 @@
...
+import { connect } from "react-redux";
+import { getUserAccessData } from "@/selectors/authenticationSelectors";
+import { USER_ROLES } from "@/constants/environment";
+
+/**
+ * @constant AuthorizationWrapper conditionally renders react children depending
+ * if the role passed in matches the user permissions.
+ * 
+ * NOTE: If the childComponent is an ant design child component**
+ * IE. Menu.Item,
```
NWCalvank: IE made me think Internet Explorer. But I'm guessing that's not what this means

frontend/src/components/common/wrappers/AuthorizationWrapper.js
NWCalvank: So, the difference here is that the `AuthorizationGuard` wraps components that represent a "page" view? And this wraps smaller components?

frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
```diff
@@ -118,18 +118,21 @@ export class DefaultContact extends Component {
...
-                    this.props.handleChange,
-                    this.props.mine
-                  )
-                }
-              />,
+              <AuthorizationWrapper permission="role_create">
+                <Button
+                  type="primary"
+                  onClick={() =>
+                    this.props.openEditPartyRelationshipModal(
```
NWCalvank: This is quite a few positional arguments. If this function gets any more arguments (or if any become/are already optional), we may want to refactor this to accept an object, for the benefit of having named arguments.

Not necessarily related to this PR, but I'd like to flag it

frontend/src/styles/settings/themeOverride.scss
```diff
@@ -164,3 +164,22 @@
   padding: 0;
   margin: 0;
 }
+// MENU.ITEM
+// Same styling as the default for .ant-dropdown-menu-item,
+// see AuthorizationWrapper.js for the specific use case of this custom class
+.custom-menu-item {
```
NWCalvank: Are these the exact rules copied from the antd class? I think you can use `@extend` to inherit all of those rules directly instead

frontend/src/components/common/wrappers/AuthorizationWrapper.js
mmumby: Yeah I was thinking about passing the permissions around as constants instead of strings.

The reason I opted against it is because our permissions are so simple at this point, I was considering moving them into constant files when we make them more complex. 

I also don't have strong feelings, and will do which ever you think is better at this stage.

frontend/src/styles/settings/themeOverride.scss
mmumby: I also thought that, 

But it did not work, and I was assuming it was due to the class not being mentioned anywhere in our CSS files. Thoughts?

frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
mmumby: Good point, Should bring this to the attention of @nayr974

frontend/src/components/admin/AdminDashboard.js
mmumby: updated!

frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
nayr974: @NWCalvank My thought is that as this function is defined and consumed only within this component, and that all of our open modal functions in the app use only positional arguments that leaving it is better for consistency. I see a better argument for refactoring "Contact" proptypes into an object as those are passed through a couple layers. I'll keep both of these in mind if we are working on this screen again.

frontend/src/components/common/wrappers/AuthorizationWrapper.js
FreshGyan: I'm happy with roles as variables 💯

PR #415, opened 2019-01-15
Feature/mds 1020

mmumby:
Updated the font classes with the new typography, 
Updated the mine tombstone data to match figma.


frontend/src/selectors/mineSelectors.js
```diff
@@ -33,3 +33,27 @@ export const getCurrentMineTypes = createSelector(
...
+export const getTransformedMineTypes = createSelector(
+  [getMines, getMineGuid],
+  (mines, mineGuid) => {
+    const mine_types = {
+      mine_tenure_type_code: [],
+      mine_commodity_code: [],
+      mine_disturbance_code: [],
+    };
+    if (mineGuid) {
+      mines[mineGuid].mine_type.map((type) => {
```
NWCalvank: So, a couple thoughts on this..
If you're not returning anything then you'll be better off using `.forEach()` instead of `.map()`

But I also think we can clean this up quite a bit.. something like (haven't tested this):
```javascript
(mines, mineGuid) => {
  const extract = (key) => type.mine_type_detail.filter((detail) => detail[key]).map((detail) => detail[key]);
  return {
    mine_tenure_type_code: mineGuid ? mines[mineGuid].mine_type.(({ mine_tenure_type code }) => mine_tenure_type_code),
    mine_commodity_code: extract("mine_commodity_code"),
    mine_disturbance_code: extract("mine_disturbance_code"),
  };
}
```

frontend/src/styles/elements/elements.scss
```diff
@@ -156,3 +160,48 @@ hr {
     font-size: 16px;
   }
 }
+
+// special text styles/classes, all ant design overrides should @extend these.
```
NWCalvank: Good comment 👍

frontend/src/styles/settings/theme.scss
```diff
@@ -12,3 +12,10 @@ $menu-item-color: $white;
 
 $error-color: $alert-red;
 $success-color: $ok-green;
+
+$font-family: "Open Sans", sans-serif;
+
+$font-size-base: 16px;
+$font-size-lg: $font-size-base + 2px;
+$font-size-sm: 14px;
+// $line-height-base: 18px;
```
NWCalvank: Remove this?

frontend/src/components/mine/MineHeader.js
```diff
@@ -184,69 +149,86 @@ class MineHeader extends Component {
...
-                      {this.props.mineTenureHash[tenure.mine_tenure_type_code]}
-                    </span>
-                  ))}
-                  <Icon type="down" style={{ fontSize: "14px" }} />
-                </Button>
-              </Popover>
+          <div className="inline-flex padding-small">
+            <p className="field-title">Tenure</p>
+            <p>
+              {this.props.transformedMineTypes.mine_tenure_type_code.length > 0 ? (
```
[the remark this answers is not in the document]
mmumby: It's only there to display an "N/A" if the array is empty

frontend/src/selectors/mineSelectors.js
mmumby:
```
export const getTransformedMineTypes = createSelector(
  [getMines, getMineGuid],
  (mines, mineGuid) => {
    const mine_types = {
      mine_tenure_type_code: [],
      mine_commodity_code: [],
      mine_disturbance_code: [],
    };
    if (mineGuid) {
      mines[mineGuid].mine_type.forEach((type) => {
        mine_types.mine_tenure_type_code.push(type.mine_tenure_type_code);
        const getDetails = (key) =>
          type.mine_type_detail.forEach((detail) => {
            if (detail[key]) mine_types[key].push(detail[key]);
          });
        getDetails("mine_commodity_code");
        getDetails("mine_disturbance_code");
      });
      return mine_types;
    }
  }
);
```

@NWCalvank I updated it to this, but I think this is less clear. thoughts?

mmumby: also, how do you get your code blocks to be in colour??

NWCalvank: Ahh.. I see why you're doing that way. I didn't spot that it was using `type` lexically...

1) If you don't think my solution is any clearer then you don't need to use it.
2) But if you do, you can just pass `type` as an argument to the `getDetails` function.

Something like:
```javascript
(mines, mineGuid) => {
  const getDetails = (key, type) => type.mine_type_detail.filter((detail) => detail[key]).map((detail) => detail[key]);
  const types = mineGuid ? mines[mineGuid].mine_type : [];
  return {
    mine_tenure_type_code: types.map(({ mine_tenure_type code }) => mine_tenure_type_code),
    mine_commodity_code: types.map((type) => getDetails("mine_commodity_code", type)),
    mine_disturbance_code: types.map((type) => getDetails("mine_disturbance_code", type)),
  };
}
```
3) Putting the language name after the backticks gives you highlighting :)

PR #416, opened 2019-01-15
Update mine_status on mine put vs create new record.

nayr974:


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -132,7 +132,14 @@ def mine_status_processor(self, mine_status, mine_guid):
         except AssertionError as e:
             self.raise_error(400, 'Error: {}'.format(e))
         mine_status_xref.save()
+
         try:
+            existing_status = MineStatus.find_by_mine_guid(mine_guid)
+            if existing_status:
+                existing_status.mine_status_xref_guid = mine_status_xref.mine_status_xref_guid
+                existing_status.save()
+                return existing_status
```
NWCalvank: So, we're doing this to avoid creating new `mine_status` records on a `PUT`? I'm guessing for the benefit of update-in-place giving the freedom to not have to expire records?

My only concern is if we're leaving orphaned `MineStatusXref` records. I haven't worked with mine status before, so I could be missing something, but it looks like that's what's happening here.

nayr974: So far as I can tell MineStatusXref is actually the lookup table, which has permutations of code, reason_code, and sub_reason_code. It doesn't relate to mine_status and isn't actually a xref, so nothing will get orphaned.

NWCalvank: Alrighty. Sounds like you've looked at this further than I have, so I'll defer to you. Marking as resolved

python-backend/app/api/mines/status/models/mine_operation_status_code.py
```diff
@@ -22,23 +22,25 @@ def json(self):
         return {
             'mine_operation_status_code': self.mine_operation_status_code,
             'description': self.description,
-            'display_order': str(self.display_order),
-            'effective_date': self.effective_date.isoformat(),
-            'expiry_date': self.expiry_date.isoformat()
+            'display_order': str(self.display_order)
```
NWCalvank: Not in scope of these changes, but I'm not convinced that the backend should know about something like "display order" at all.

python-backend/app/api/mines/status/resources/status.py
```diff
@@ -8,7 +8,7 @@
 
 class MineStatusResource(Resource, UserMixin, ErrorMixin):
     @api.doc(params={'mine_status_guid': 'Mine status guid.'})
-    @requires_role_mine_view
+    #@requires_role_mine_view
```
NWCalvank: Is this an intentional change?

python-backend/app/api/mines/status/models/mine_operation_status_code.py
nayr974: Wouldn't this be for the front-end to consume and sort there? Alternative could be to have the backend return the collection sorted by display_order and then not include this property in the returned json. This is maybe better. Right now, I don't think we consume these at all as the get is still returned from the constants file.

python-backend/app/api/mines/status/resources/status.py
nayr974: No sir :)

python-backend/app/api/mines/status/models/mine_operation_status_code.py
NWCalvank: I think I agree.. My impression of the idea behind this attribute is to let the frontend know how to order the items. This seems silly. If the backend has an opinion on how things should be ordered then, like you said, it can just order them ahead of time and not expose this attribute.

Otherwise, (and this is my current stance on it) the frontend knows that it wants to display these items in an order based on a certain field and takes care of that on its own.

With the current approach, it seems like both frontend and backend are contributing to the "ordering," which I think is used in dropdowns? Again, out of scope, but I noticed this due to the diff.

migrations/sql/V2019.01.16.03.32__modify_status_code_xref.sql
```diff
@@ -0,0 +1,47 @@
...
+    ('CLD', 'ORP', 'LWT', 'system-mds', 'system-mds'),
+    ('CLD', 'ORP', 'RNS', 'system-mds', 'system-mds'),
+    ('CLD', 'ORP', 'SVR', 'system-mds', 'system-mds'),
+    ('CLD', 'UN', null, 'system-mds', 'system-mds'),
+    ('NS', null, null, 'system-mds', 'system-mds'),
+    ('OP', 'YR', null, 'system-mds', 'system-mds'),
+    ('OP', 'SEA', null, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
+drop table if exists etl_status;
```
FreshGyan: Probably should move this to ETL scripts if that's where the cleanup needs to happen.

nayr974: The ETL scripts execute daily, what I want this to do is to trigger a one time re-ETL because I'm intentionally truncating the data and I want it to be reimported.

migrations/sql/V2019.01.17.12.10__create_minespace_user_tables.sql
```diff
@@ -1,7 +1,9 @@
-CREATE TABLE minespace_user (
+drop table if exists minespace_user;
+CREATE TABLE minespace_user
+(
     user_id serial PRIMARY KEY,
     keycloak_guid uuid,
-    email character varying(100) NOT NULL, 
+    email character varying(100) NOT NULL,
```
klefler: Just thought of this now... is 100 characters going to be enough for email addresses? I think the max length allowable is 254 characters... so we may want to consider upping the number on this, eventually.

klefler: Also, I think it was mentioned in an earlier PR, but should we add the create_user, create_timestamp, update_user, and update_timestamp columns?

migrations/sql/V2019.01.17.12.10__create_minespace_user_tables.sql
```diff
@@ -11,12 +13,15 @@ CREATE TABLE minespace_user (
 
 COMMENT ON TABLE minespace_user IS 'A list of public mine proponents email addresses that can log into minespace and update mds information on mines listed in minespace_user_mds_mine_access';
 
-CREATE TABLE minespace_user_mds_mine_access(
+drop table if exists minespace_user_mds_mine_access;
+CREATE TABLE minespace_user_mds_mine_access
+(
     user_id integer NOT NULL,
-    mine_guid uuid NOT NULL, 
+    mine_guid uuid NOT NULL,
```
klefler: Should the 4 audit columns be added to this table too? Also, is there a need to keep active and inactive records? What happens if the user is no longer associated with a mine, does it get deleted? If so, should it?

migrations/sql/V2019.01.17.12.10__create_minespace_user_tables.sql
nayr974: Hi Kristen! For this PR I'm just renaming this script which was already commited to match our new versioning scheme. Jason is still working on user changes in another branch. I'll pass on your comments to him.

klefler: Gotcha! Ok thanks. Apologies for commenting in the wrong PR :)

nayr974: No, it's good to raise things when you see them. I did pass your comments to Jason.

Jsyro: @klefler 
Didn't know about the 254 email for bceid, we should change it to match the bceid/idir standard. 

These tables are built to support (and eventually be moved to) a separate application, so I didn't automatically apply that same pattern, not sure what we want to do here.

migrations/sql/V2019.01.17.12.10__create_minespace_user_tables.sql
Jsyro: @klefler, 
Same thing, didn't apply the pattern to this. i guess it's a question for the business if that information is required. 
Yes it currently hard deletes.

PR #417, opened 2019-01-15
Feature/mds 991

thegentlemanphysicist:
This tickent adds tsf, major mine, tenure, region, and commodity filtering.


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -42,7 +46,7 @@ class MineResource(Resource, UserMixin, ErrorMixin):
             'mine_no_or_guid':
             'Mine number or guid. If not provided a paginated list of mines will be returned.'
         })
-    @requires_role_mine_view
+    # @requires_role_mine_view
```
nayr974: Should this comment be removed or is this truly public?

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -61,48 +65,85 @@ def get(self, mine_no_or_guid=None):
...
+            if status_filter_term:
+                status_filter_term_array = status_filter_term.split(',')
+
+                status_filter = MineStatusXref.mine_operation_status_code.in_(status_filter_term_array)
+                status_reason_filter = MineStatusXref.mine_operation_status_reason_code.in_(status_filter_term_array)
+                status_subreason_filter = MineStatusXref.mine_operation_status_sub_reason_code.in_(
+                    status_filter_term_array)
+                all_status_filter = status_filter | status_reason_filter | status_subreason_filter
+                status_query = Mine.query \
+                    .join(MineStatus) \
```
nayr974: Mine to mine_status should be 1:1 after my change so this should work. Still want @JustinMacaulay to take a look

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: good catch

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -57,61 +61,91 @@ def get(self, mine_no_or_guid=None):
...
-                        .join(MineStatus)\
-                        .join(MineStatusXref)\
-                        .filter(all_status_filter)
-                    result_query = apply_sort(mine_query_with_status, sort_criteria)
-                else:
-                    result_query = apply_sort(Mine.query, sort_criteria)
-
-            paginated_mine_query, pagination_details = apply_pagination(
-                result_query, page, items_per_page)
+    def mine_filter_and_search_method(self, args):
```
NWCalvank: Thanks for breaking this out! A couple things:
1) `_method` is redundant here
2) The `get` method should still return the JSON payload
3) Personal opinion: either use a verb method name (like `apply_search_filters`) or noun method name (like `filtered_results`)

You should be able to do something like this to separate these concerns:
```python
# inside of the get method:
mines, pagination_details = mine_filter_and_search(request.args)
return {
  'mines': list(map(lambda x: x.json_for_list(), mines)),
  ...
}
```

PR #425, opened 2019-01-17
MDS-1153 style and content for public home page

JustinMacaulay:
Added some base styling for the new public site as well as the landing page, header and footer.


frontend-public/src/components/LandingPage.js
```diff
@@ -0,0 +1,136 @@
...
+        <br />
+        <p>
+          MineSpace is a pilot project as a part of the MDS digital product. MineSpace will be the
+          portal through which mines can interact with the Ministry. In time, Mines will be able to
+          submit documentation to the Ministry, as well as check the status of reporting
+          requirements, and documentation that they have submitted to the ministry for review. The
+          pilot project will run in the spring of 2019 with full integration in early 2020.
+        </p>
+      </Col>
+      <Col xs={0} md={2} lg={6} />
```
mmumby: What is the purpose of these empty columns?

JustinMacaulay: I need them to be there in lg and md to add white space around the content but not on the xs screen size.

frontend-public/src/styles/settings/theme.scss
```diff
@@ -0,0 +1,8 @@
+// theme.scss holds overrides for ant-design
+// refer to (https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less) for ant design variable names;
+$blue: #003366;
+$white: #f1f1f2;
+$violet: #5e46a1;
+
```
mmumby: Violet is not needed as its an MDS color

frontend-public/src/styles/settings/theme.scss
```diff
@@ -0,0 +1,8 @@
+// theme.scss holds overrides for ant-design
+// refer to (https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less) for ant design variable names;
+$blue: #003366;
+$white: #f1f1f2;
+$violet: #5e46a1;
```
mmumby: I think all of these can be removed actually, they aren't being used anywhere. This file can be left blank unless you're overriding anything with those variables.

frontend-public/src/components/LandingPage.js
mmumby: Okay, 

There is a better way to accomplish this than using empty elements, especially in this quantity. 
I can assist in this area if you'd like.

frontend-public/src/styles/settings/theme.scss
JustinMacaulay: made the file empty, I needed it for the web pack loader but forgot about the variables in it. Thanks.

PR #426, opened 2019-01-17
Refactor createItemMap for performance

NWCalvank:
The original implementation was causing a timeout on the map view.


frontend-public/src/utils/helpers.js
```diff
@@ -22,8 +22,15 @@ export const createReducer = (reducer, name) => (state, action) => {
   return reducer(state, action);
 };
 // Function to create state object using the id as the key (used in src/reducers/<customReducer>)
-export const createItemMap = (array, idField) =>
-  array.reduce((result, item) => ({ [item[idField]]: item, ...result }), {});
+export const createItemMap = (array, idField) => {
```
nayr974: Not sure that it's "better", but does lodash's keyby solve this problem already?

nayr974: Yeah, getting this change in as is lower risk and we can look for an opportunity to refactor later if it makes sense to do.

PR #432, opened 2019-01-18
MDS-1111 public user dashboard

JustinMacaulay:
Added a new view for users to have a list of the mines they may interact with displayed. If the user is not associated with any mines then they are shown a nullscreen.


frontend-public/src/components/LandingPage.js
```diff
@@ -1,6 +1,8 @@
 import React from "react";
 import { Divider, Row, Col, Button } from "antd";
 
+import QuestionSidebar from "./common/QuestionsSidebar";
```
nayr974: "Home" might be a better place to put this rather than common. I doubt other screens will use this in the future.

frontend-public/src/components/dashboard/UserDashboard.js
```diff
@@ -0,0 +1,91 @@
...
+import { getUserMineInfo } from "@/selectors/userMineInfoSelector";
+import { fetchUserMineInfo } from "@/actionCreators/userDashboardActionCreator";
+import NullScreen from "@/components/common/NullScreen";
+import CustomPropTypes from "@/customPropTypes";
+import QuestionSidebar from "@/components/common/QuestionsSidebar";
+import Loading from "@/components/common/Loading";
+import * as routes from "@/constants/routes";
+
+const propTypes = {
+  userInfo: PropTypes.objectOf(PropTypes.string).isRequired,
```
nayr974: is objectOf needed here or can you just put PropTypes.string.isRequired?

frontend-public/src/components/LandingPage.js
JustinMacaulay: I can move it but its also used on both of the userDashboard pages.

frontend-public/src/components/dashboard/UserMineInfo.js
```diff
@@ -9,11 +10,11 @@ const propTypes = {
 
 // This file is anticipated to use state
 // eslint-disable-next-line react/prefer-stateless-function
-export class ProponentDashboard extends Component {
+export class UserMineInfo extends Component {
```
nayr974: I find this naming a bit confusing. Should this just be MineInfo maybe?

JustinMacaulay: I think that was meant to be UserDashboard

JustinMacaulay: Sorry I was just confused by the proponentDashboard above it and MineInfo would be good

frontend-public/src/components/dashboard/UserDashboard.js
JustinMacaulay: I think it is because it is an object that has many strings in it. It's not just one string.

frontend-public/src/components/dashboard/UserMineInfo.js
JustinMacaulay: All renamed.

frontend-public/src/components/LandingPage.js
nayr974: Ok. We can move it later if we start adding a lot of other functionality.

frontend-public/src/components/common/QuestionsSidebar.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+/**
+ * @constant RenderField - Ant Design `Input` component for redux-form.
+ */
+
```
mmumby: Update this comment to reflect the correct component.

PR #434, opened 2019-01-21
Feature/mds 1091 frontend for user managment

Jsyro:
Add and delete minespace users. 

hard deletes and no edit for simplicity. Aaron's eyes only when deployed to prod.


frontend/src/components/admin/MinespaceUserList.js
```diff
@@ -0,0 +1,90 @@
...
+const propTypes = {
+  minespaceUsers: PropTypes.arrayOf(CustomPropTypes.minespaceUser),
+  mines: PropTypes.arrayOf(CustomPropTypes.mineName),
+  handleDelete: PropTypes.func,
+};
+
+const defaultProps = {
+  minespaceUsers: [],
+  mines: [],
+  handleDelete: null,
```
mmumby: null is not a valid default prop for a function, 
use: 

```
handleDelete:  ()=> {};
```

frontend/src/components/admin/MinespaceUserManagement.js
```diff
@@ -0,0 +1,80 @@
...
+
+  handleDelete = (userId) => {
+    this.props.deleteMinespaceUser(userId).then(() => {
+      this.props.fetchMinespaceUsers();
+    });
+  };
+
+  render() {
+    // console.log(this.props.minespaceUsers);
+    // console.log(this.props.mines);
```
mmumby: remove

frontend/src/components/admin/MinespaceUserList.js
Jsyro: thanks, fixed.

frontend/src/components/admin/MinespaceUserManagement.js
Jsyro: good catch, thanks.

frontend/src/customPropTypes/mines.js
```diff
@@ -26,3 +26,11 @@ export const minePageData = shape({
   total: PropTypes.numnber,
   total_pages: PropTypes.numnber,
 });
+
+export const mineName = shape({
```
mmumby: `mineName` is very specific for an object that contains more than a name. This should be renamed.

frontend/src/components/admin/NewMinespaceUser.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+  };
+
+  render() {
+    return (
+      <div>
+        <h3>Add BCEID User</h3>
+        {this.props.mines.mines && (
+          <AddMinespaceUser
+            mines={this.props.mines.mines.map((mine) => ({
```
mmumby: The data should be transformed before passing it in as a prop.

frontend/src/components/admin/MinespaceUserManagement.js
```diff
@@ -0,0 +1,78 @@
...
+  mines: PropTypes.arrayOf(CustomPropTypes.mineName),
+  minespaceUsers: PropTypes.arrayOf(CustomPropTypes.minespaceUser),
+  fetchMineNameList: PropTypes.func.isRequired,
+  fetchMinespaceUsers: PropTypes.func.isRequired,
+  deleteMinespaceUser: PropTypes.func.isRequired,
+};
+
+const defaultProps = {
+  mines: [],
+  minespaceUsers: {},
```
mmumby: should this be? `minespaceUsers: []`

frontend/src/components/admin/NewMinespaceUser.js
Jsyro: removed second .mines. i think that's what you were looking for.

frontend/src/components/admin/MinespaceUserManagement.js
Jsyro: yes it should.

frontend/src/customPropTypes/mines.js
Jsyro: `mineNameList: getMineNames(state).mines,` existing places that used that object structure use a selector called getMineNames, so i figure the object type should be MineName. I don't disagree and I'd be happy to rename it. But then the selector name and the customproptype don't match in name, which may be awkward.

mmumby: ah, okay, I'm following. 

This is fine then.

PR #436, opened 2019-01-21
Feature/mds 1148

thegentlemanphysicist:
Front end of 991, this relies on the backend change to work, so we should avoid merging until the backend is merged.


frontend/src/components/Forms/AdvancedSearchForm.js
```diff
@@ -25,77 +33,96 @@ export class AdvancedSearchForm extends Component {
...
+                  name="commodity"
+                  placeholder="Select Mine Commodity"
+                  component={renderConfig.MULTI_SELECT}
+                  data={this.props.mineCommodityOptions}
+                />
+              </Col>
+            </Row>
+            <Row gutter={6}>
+              <Col md={12} xs={24}>
+                {/* TODO: Remove label, get placeholder text displaying properly */}
```
mmumby: good comment!

frontend/src/components/dashboard/Dashboard.js
```diff
@@ -68,12 +72,13 @@ export class Dashboard extends Component {
       params: {
         page: String.DEFAULT_PAGE,
         per_page: String.DEFAULT_PER_PAGE,
-        major: false,
-        tsf: false,
+        major: ["blah"],
```
mmumby: remove.

frontend/src/components/dashboard/MineSearch.js
```diff
@@ -61,8 +60,8 @@ export class MineSearch extends Component {
    * filter mineList with new search input;
    */
   handleSearch = (value = {}) => {
-    const search = value.target && value.target.value;
-    const { commodity, region, status, tenure, tsf, major } = search ? {} : value;
+    const { commodity, region, status, tenure, tsf, major, search } = value;
+
```
mmumby: now that I think about it, we should be able to pass value in without destructuring it first

```
this.props.handleMinesearch(value)
```

and actually, may be able to delete `handleSearch` all together and pass `handleMinesearch` into the form, since we are no longer formatting the data we should no longer need this middle function.

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -70,3 +70,8 @@ export const getCommodityOptionHash = createSelector(
       {}
     )
 );
+
+export const getDropdownCommodityOptions = createSelector(
```
mmumby: ah, we will have to create a test for this. I forgot about this selector, I can help with that test.

frontend/src/components/dashboard/Dashboard.js
thegentlemanphysicist: done

PR #437, opened 2019-01-21
MDS-1092 user document upload page

JustinMacaulay:
Added functionality to the public site which allows users to see all of the required reports for a mine as well as upload and delete documents from them.


frontend-public/src/components/Forms/EditReportForm.js
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+  </Form>
+);
+
+EditReportForm.propTypes = propTypes;
+
+export default reduxForm({
+  form: FORM.EDIT_TAILINGS,
+  touchOnBlur: false,
+  onSubmitSuccess: resetForm(FORM.EDIT_TAILINGS),
```
mmumby: I dont think you copied this funtion over, 

```
import { resetForm } from "@/utils/helpers";
```

JustinMacaulay: It was already copied over from the initial creation.

frontend-public/src/customPropTypes/mines.js
```diff
@@ -0,0 +1,28 @@
...
+export const mine = shape({
+  guid: PropTypes.string.isRequired,
+  mine_no: PropTypes.string,
+  mine_name: PropTypes.string,
+  mine_note: PropTypes.string,
+  region_code: PropTypes.string,
+  major_mine_ind: PropTypes.bool,
+  mine_permit: arrayOf(minePermit),
+  mine_expected_documents: arrayOf(mineExpectedDocument),
+});
```
mmumby: I don't think the user is exposed to the majority of this, delete what is not needed. 

```
export const mine = shape({
  guid: PropTypes.string.isRequired,
  mine_no: PropTypes.string,
  mine_name: PropTypes.string,
  mine_expected_documents: arrayOf(mineExpectedDocument),
});
```

frontend-public/src/customPropTypes/permits.js
```diff
@@ -0,0 +1,18 @@
+import { PropTypes, shape } from "prop-types";
+
```
mmumby: Do Proponents see permits?

frontend-public/src/styles/components/MineInfo.scss
```diff
@@ -0,0 +1,38 @@
...
+  box-shadow: $default-box-shadow;
+  margin: 24px 0px;
+  h4 {
+    padding-top: $default-padding-lg;
+    padding-bottom: $default-padding-md;
+  }
+}
+
+.user-upload-document-btn {
+  color: $font-colour !important;
```
mmumby: Concerned about the number of `!importants` used here, 
should set defaults for ant designs buttons instead and use those instead of these custom classes.

frontend-public/src/styles/settings/themeOverride.scss
```diff
@@ -28,3 +24,34 @@
...
+  }
+}
+
+.ant-modal-title {
+  font-size: $default-h2;
+  color: $p-color;
+  font-weight: $semi-bold;
+}
+
+.ant-radio-button-wrapper {
```
mmumby: This is an ant class correct?? 
if so, the `!importants` used in this situation are not alarming compared to the other situation.
but is the CSS not being overwritten without them?

JustinMacaulay: Without the important it leaves the right side of the border the wrong colour

frontend-public/src/customPropTypes/permits.js
JustinMacaulay: No but i needed this to make the mine custom prop type work.

frontend-public/src/customPropTypes/mines.js
JustinMacaulay: Removed all of the unused props information as well as removed the permittee.

frontend-public/src/customPropTypes/permits.js
JustinMacaulay: Removed.

frontend-public/src/styles/components/MineInfo.scss
JustinMacaulay: There are a lot of different button styles on this site so a lot of them I need styling that isn't overriding the ant styles. This one though had to be important as it was ignoring it otherwise.

NWCalvank: Is the goal to override this class to use this font colour everywhere, in all instances? If not, we should probably try to get more specific with the selector instead of using `!important`. If so, we should have this in the `theme.scss` or `themeOverrides.scss`. I'm not sure which file is more appropriate

JustinMacaulay: I have refactored as much as I can into the theme overrides and am using primary, secondary, tertiary, and ghost buttons.

PR #438, opened 2019-01-22
MDS-404 Feature/ Expose Mine Manager history

NWCalvank:
Expose an admin-only endpoint for fetching a CSV of mine manager history changes
Ingest and download that CSV on the frontend via the Admin Dashboard


frontend/src/components/admin/AdminDashboard.js
```diff
@@ -1,18 +1,49 @@
...
+  };
+
+  render() {
+    return (
+      <div className="landing-page">
+        <div className="landing-page__header" />
+        <div className="landing-page__content">
+          <h1>Admin View</h1>
+          <h2>Pull Mine Manager History</h2>
+          <Input
```
mmumby: Would this be better as a redux form and then we wouldn't have to worry about setting the state at all with mineNo?

NWCalvank: I knew you'd ask or have some comment about it! But no, I personally don't see any reason for this component to have application-wide state. The only state is the input field, which is relevant to this component only, does not need to persists between routes, etc. I think leaving out redux here makes sense (at this point, anyway).

We may choose to connect this to the store if/when we add more to the Admin dashboard.

migrations/sql/V2019.01.16.11.23__alter_mine_party_appt.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE mine_party_appt
+ADD COLUMN processed_by character varying(60) DEFAULT '' NOT NULL,
```
nayr974: I don't like defaulting here, personally I'd rather see null if we don't know what the value is, because that's what null represents, but it's definitely easier to deal with if we know it's always a string. Not requesting a change, but it stands out to me. I see we've done this on the mine table for a few columns.

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -40,6 +40,11 @@ class Mine(AuditMixin, Base):
         lazy='joined')
     mine_type = db.relationship(
         'MineType', backref='mine', order_by='desc(MineType.update_timestamp)', lazy='joined')
+    mine_party_appt = db.relationship(
```
nayr974: I really don't like seeing MinePartyAppointment on the mine model, which is already extremely heavy. In various API discussions we talked about treating each resource (/mines, /parties, /documents) as if it were a separate micro-service. Either loop back and call the /parties endpoint to fetch this data if it's really needed here (there is a pattern to copy in our API for this), or have the front-end fetch it from /parties when it's needed. It's already in the store and there are actions around it.

migrations/sql/V2019.01.16.11.23__alter_mine_party_appt.sql
klefler: I tend to agree with Ryan. It's far cleaner to have a NULL value that a default space. It could make things tricky for someone running a query on the data... if it looks like a NULL but isn't.

Could you also maybe add a comment for the new column? We're going to eventually clean up the table/column comments to make sure we have them all defined... so best to start now on anything new :)

nayr974: Kristen, maybe we can talk about this another time but I'd like to push back on column comments, unless there is some real exception in design. They are somewhat time consuming and don't tend to get maintained.

klefler: yep sure thing :) I didn't know they were time consuming. Good to know. Thanks for the heads up.

python-backend/app/api/mines/mine/models/mine.py
nayr974: Actually it doesn't look like you're returning party appointments with the mine. I'm curious why you needed the relationship defined here?

migrations/sql/V2019.01.16.11.23__alter_mine_party_appt.sql
nayr974: After some face to face discussion, "processed" probably isn't actually that intuitive. If a comment helps make it clear, I'm not opposed 👍

NWCalvank: Will happily change that to nullable if that's what we want to do. I wrote this early last week, so I don't remember exactly what I looked at, but this must have fallen in line with some sort of pattern that I saw.

python-backend/app/api/mines/mine/models/mine.py
NWCalvank: This was a requirement when I was leveraging the relationship in the `MinePartyAppointment` model to query for the related mine, prior to switching to the internal API request. I don't think this is needed anymore and should be removed. Good catch

python-backend/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -98,6 +103,36 @@ def find_parties_by_mine_party_appt_type_code(cls, code):
...
+        # send internal API call to fetch matching mine record
+        mines_url = current_app.config['MINES_URL'] + '/mines/' + str(mine_no)
+        auth_headers = request.headers.get('Authorization')
+        if not auth_headers:
+            return None, 401, 'Unauthorized'
+
+        headers = {'Authorization': auth_headers}
+        response = requests.get(url=mines_url, headers=headers)
+        if not response:
+            return None, 400, 'Unexpected error'
```
Jsyro: This would be an internal web call 400., to the one sent to the server was fine. I feel like this should return a 500.

NWCalvank: That's fair. The two places I put a 400 marked errors that I wasn't totally sure about how to describe. Open to input on either of those and will update the codes accordingly

PR #440, opened 2019-01-22
Feature/mds-1241 Integrate file upload components

james-hollinger:
In the file upload modal, improves user experience by updating the list of uploaded documents when either a new file is uploaded or an existing file is selected.

FilePicker now has an optional prop called onFileLoad that is a callback for when a file successfully loads in the FileUpload component.


frontend/src/components/common/FilePicker.js
```diff
@@ -13,11 +13,13 @@ const propTypes = {
   uploadUrl: PropTypes.string.isRequired,
   existingFilesDropdown: PropTypes.arrayOf(CustomPropTypes.dropdownListItem).isRequired,
   onSelectExisting: PropTypes.func.isRequired,
+  onFileLoad: PropTypes.func,
 };
 
 const defaultProps = {
   maxFileSize: undefined,
   acceptedFileTypesMap: undefined,
+  onFileLoad: undefined,
```
mmumby: defaultProps shouldn't be set to be undefined, set them according to their proptype, if the prop is an array, initial it with an empty array etc. 
```
maxFileSize: " " //set an actual default size here
acceptedFileTypesMap: {},
onFileLoad: () => {},
```

frontend/src/components/common/FileUpload.js
```diff
@@ -14,11 +14,13 @@ const propTypes = {
   uploadUrl: PropTypes.string.isRequired,
   maxFileSize: PropTypes.string,
   acceptedFileTypesMap: PropTypes.objectOf(PropTypes.string),
+  onFileLoad: PropTypes.func,
 };
 
 const defaultProps = {
   maxFileSize: "100MB",
   acceptedFileTypesMap: {},
+  onFileLoad: null
```
mmumby: ah, yes, this is exactly what should be used above, Except functions cannot be null, user `() => {}` instead.

frontend/src/components/common/FilePicker.js
james-hollinger: The reason they are set to undefined in FilePicker is because I don't want to overwrite or duplicate the defaults that are set in FileUpload.

james-hollinger: I would choose to not have a defaultProps in FilePicker but its a linting error.

frontend/src/components/common/FileUpload.js
james-hollinger: After discussion, added a comment explaining that FilePond wants this to be null.

PR #442, opened 2019-01-22
MDS-1016: Initial table create WIP

ActionAnalytics:


FreshGyan:
Talked with Julian and closing for now. But can be resurrected later.

PR #446, opened 2019-01-23
Add database backups

FreshGyan:
- Added a container to periodically create rolling database backups.
 - Updated migration scripts to create the test database without killing any active connections.

Backups:
- They only run in production and run every 24 hours

Test db creation:
Currently: When we run migrations, we kill every connection to mds and mds_test db and then copy over the structure of mds to mds_test. This can result in incomplete queries and when deploys happen, we get a tiny downtime between backend deploys to db connections.

New implementation: 
Only create the test database if it doesnt exist. The beforemigrate file is to install the extensions on a clean mds_test database. For mds database it should just skip and not cause any grief.
This way we dont kill any active connections and basically run flyway migrations on both test and non test to keep the seed data and structure exactly same.

Docs - https://apps.nrs.gov.bc.ca/int/confluence/display/MDS/Database+backup+and+restore+steps


migrations/sql/afterMigrate__99__create_test_db.sql
```diff
@@ -1,2 +1,13 @@
-DROP DATABASE IF EXISTS mds_test;
-CREATE DATABASE mds_test WITH TEMPLATE mds OWNER mds;
\ No newline at end of file
+DO
+$do$
+DECLARE
+BEGIN
+  CREATE EXTENSION IF NOT EXISTS dblink;
```
garywong-bc: FYI https://www.postgresql.org/docs/9.5/postgres-fdw.html is the new way and 'create extension postgres_fdw' is available out of the box.

garywong-bc: it's what oracle_fdw is based upon..  which is part of our postgres build

migrations/sql/beforeMigrate__0__add_extensions.sql
```diff
@@ -0,0 +1,3 @@
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
```
garywong-bc: @FreshGyan  knows this but for other MDS reviewers, this is simply a 'failsafe' as our image already has pgcrypto and postgis enabled.

migrations/sql/afterMigrate__99__create_test_db.sql
FreshGyan: yes! that works great. 
We don't do anything with the fdw in our test schema, so that should stay unaffected for now. :D

PR #449, opened 2019-01-23
Feature/mds 1164 secure api to specific mines

nayr974:
Applies a filter to restrict mine access on all queries when a user has the proponent role. 

Implementation based on the multi-tenancy approach outlined in this presentation, and reading way too much sqlalchemy documentation: https://www.infoq.com/presentations/saas-python?utm_source=infoq&utm_medium=slideshare&utm_campaign=slidesharelondon


python-backend/app/auth.py
```diff
@@ -0,0 +1,65 @@
...
+def get_mine_access():
+    user = get_current_user()
+    return list(x.mine_guid for x in user.mines)
+
+
+def get_current_user():
+    rv = getattr(g, 'current_user', None)
+    if rv == None:
+        email = get_user_email()
+        rv = MinespaceUser.query.unbound_unsafe().filter_by(email=email).filter_by(
```
NWCalvank: In Ruby on Rails, we'd do something like `@current_user ||= some_query.first()` to avoid running the query every time the method is called by storing the value in an instance variable. Is that being handled by the `if rv == None` check?

python-backend/app/auth.py
```diff
@@ -0,0 +1,65 @@
...
+            deleted_ind=False).first()
+        g.current_user = rv
+    return rv
+
+
+def get_user_is_proponent():
+    token_data = User().get_user_raw_info()
+    is_proponent = MINESPACE_PROPONENT in token_data["realm_access"]["roles"]
```
NWCalvank: If we're accessing the keys on `token_data` like this, we probably want a `try` and catch a key error.

Otherwise, a `.get()` call; possibly with a `None` check to ensure desired behaviour in returning `is_proponent`

nayr974: Yeah that's a good idea. I'm in the process of adding some handling around whether the token exists or not as it throws an error if there is no token. I'll add some protection here as well.

python-backend/app/auth.py
nayr974: Yes, g is the flask "current request" context. It's global to the current request, and dies when the current request is returned. So, this query will be called a maximum of once per API request, even if within the request you make several queries, and only if you are a proponent.

python-backend/requirements.txt
```diff
@@ -37,4 +37,5 @@ SQLAlchemy==1.2.10
 sqlalchemy-filters==0.8.0
 uwsgi==2.0.17.1
 uwsgitop==0.10
-Werkzeug==0.14.1
\ No newline at end of file
+Werkzeug==0.14.1
+cached_property==1.5.1
```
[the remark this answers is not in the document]
nayr974: Moved to the right spot. Sorry :)

PR #458, opened 2019-01-25
bug/mds 1259 return proper null instead of ''

Jsyro:
so the backend parser accepts its own json


python-backend/app/commands.py
```diff
@@ -24,6 +24,9 @@
 from .api.utils.random import generate_mine_no, generate_mine_name, random_geo, random_key_gen, random_date, random_region, random_mine_category
 from .api.parties.party_appt.models.mine_party_appt import MinePartyAppointment
 from .extensions import db
+from app import auth
+
+auth.apply_security = False
```
james-hollinger: Are we sure this doesn't disable security for the whole app?

Jsyro: @james-hollinger More security changes are coming (along with tests). This change allows create_data to work. it's an auth flag for @nayr974 row level security layer. IT can delayed it you prefer.

PR #462, opened 2019-01-28
Feature/auth

mmumby:
Work In progress,

Frontend changes for BCeID login/logout with 2 protected routes. 

All that needs to be done is testing, and setting the proper keycloak-url for local/dev/prod.


frontend-public/src/HOC/AuthenticationGuard.js
```diff
@@ -1,79 +1,43 @@
...
-      if (this.props.keycloak) {
-        if (this.props.isAuthenticated) {
-          return <WrappedComponent {...this.props} />;
-        }
-        // Add this block back in once there are user permissions to check against, ie user logs in with BCeID but permissions have not been updated on keycloak.
-        // if (!this.props.isAuthenticated) {
-        //   return <NullScreen type="unauthorized" />;
-        // }
-        return <Loading />;
+      if (this.props.isAuthenticated || isPublic) {
```
NWCalvank: What's the purpose of `isPublic`?

frontend-public/src/components/Header.js
```diff
@@ -7,23 +7,25 @@ import * as routes from "@/constants/routes";
...
-            <span className="header-title"> MineSpace</span>
-          </h1>
-        </div>
-      </Col>
-      <Col xs={24} sm={8} md={8} lg={8}>
-        <Authentication />
-      </Col>
-    </Row>
+  <div className="header inline-flex between tablet-block tablet-center">
+    {/* <Row> */}
```
NWCalvank: Comments in this file can likely be removed?

frontend-public/src/components/ReturnPage.js
```diff
@@ -0,0 +1,72 @@
...
+    authenticateUser: PropTypes.func.isRequired,
+    redirect: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]).isRequired,
+  };
+
+  componentDidMount() {
+    // grab the code and redirect type from the redirect url
+    const { type, code } = queryString.parse(this.props.location.search);
+    switch (type) {
+      case RETURN_PAGE_TYPE.LOGIN:
+        if (code) {
```
NWCalvank: Does this need to take any action if the code isn't present?

frontend-public/src/styles/elements/elements.scss
```diff
@@ -78,6 +78,12 @@ link {
     color: $a-color-hover;
     text-decoration: underline;
   }
+  & #header-link {
```
NWCalvank: Why are these ids?

mmumby: Forgot to switch these back to classes

frontend-public/src/HOC/AuthenticationGuard.js
mmumby: I needed this authGuard to serve two purposes, only return the `WrappedComponent` if the user is authenticated, and also check for authentication at the highest level, that being App.js. 

So instead of creating two guards that do essentially the same thing, I added an `isPublic` param that is able to authenticate the user (if a token exists) and also renders the `WrappedComponent` because it's a public route, (ie the home page).

NWCalvank: Okay. That makes sense. I was getting the impression that the entire app was being set to public, circumventing the need to be authenticated. The idea of having lower-level components requiring authentication makes sense. Thanks for clarifying

frontend-public/src/components/ReturnPage.js
mmumby: simplified.

frontend-public/src/styles/elements/elements.scss
mmumby: these are now classes

PR #464, opened 2019-01-28
MDS 540: Performance Improvements

nayr974:
This PR contains the following changes to improve user performance:

- Changes to the lazy loading options in the mine model
- Enabling of caching functionality we PR'd into the JWT_OIDC library
- Caching/invalidation of the mine map location data in Redis
- A faster method of transforming map point data in the MinePin component
- Allow browsers to cache content given a modified check
- Webpack configuration change to apply module optimization to all imports
- A property in the static content reducer that indicates if static content has been loaded or not, and changes to only fetch this content if needed.
- Remove unneeded ComponentDidMount data load on MineSearch component
- Modification of generate_mine_no to allow more than 10,000 test mines to be generated


frontend/src/components/dashboard/Dashboard.js
```diff
@@ -95,11 +101,15 @@ export class Dashboard extends Component {
...
+    if (!this.props.optionsLoaded) {
+      this.props.fetchStatusOptions();
+      this.props.fetchRegionOptions();
+      this.props.fetchMineTenureTypes();
+      this.props.fetchMineDisturbanceOptions();
+      this.props.fetchMineCommodityOptions();
+      this.props.fetchPartyRelationshipTypes();
+      this.props.setOptionsLoaded();
+    }
   }
```
mmumby: I really like this change. Good job!

What are your thoughts on these GET requests being moved to the highest level parent, ie Home.js, 
Then they can then be imported and called once, instead of in Dashboard.js and MineDashboard.js

frontend/server.js
```diff
@@ -11,10 +11,15 @@ if (dotenv.parsed) {
...
+// maxAge and mustRevalidate control how the client caches application files. The settings
+// below allows the client to cache content, but the client must check to see if the content
+// is stale. Our app serves content with eTags, so this results in a status 304 Not Modified
+// response, unless the content has been updated.
 const app = express();
 app.use(
   cacheControl({
-    noStore: true,
+    mustRevalidate: true,
+    maxAge: 0,
```
mmumby: is `maxAge` needed? It doesn't seem to be doing anything.

nayr974: It should override any default max-age behavior for a browser or proxy server that interprets the response. I would think that some proxy servers are pretty aggressive about what they're trying to cache so this makes it explicit. Max-age: 0 and must-revalidate are commonly used together and I think make the combo makes sense for our app.

mmumby: Thanks for the explanation!

frontend/src/components/dashboard/Dashboard.js
nayr974: That's probably best. If I don't hit any snags in doing so I'll move it.

python-backend/app/api/mines/location/models/mine_location.py
```diff
@@ -4,7 +4,8 @@
 from sqlalchemy.dialects.postgresql import UUID
 from geoalchemy2 import Geometry
 from ....utils.models_mixins import AuditMixin, Base
-from app.extensions import db
+from app.extensions import db, cache
+from ....constants import MINE_MAP_CACHE
```
FreshGyan: Unused import?

python-backend/app/api/mines/location/models/mine_map_view_location.py
```diff
@@ -38,4 +38,13 @@ def json_for_map(self):
                 'latitude': str(self.latitude),
                 'longitude': str(self.longitude)
             }
-        }
\ No newline at end of file
+        }
+
+    def compact_for_map(self):
```
FreshGyan: What's this doing?

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -5,6 +5,11 @@
...
 from app.extensions import db
 
+# NOTE: Be careful about relationships defined in the mine model. lazy='joined' will cause the relationship
+# to be joined and loaded immediately, so that data will load even when it may not be needed. Setting
+# lazy='select' will lazy load that data when the property is first accessed. There are other options as well
+# that may be best in different situations: https://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html
+
 
 class Mine(AuditMixin, Base):
     __tablename__ = 'mine'
```
FreshGyan: Hmm 🤔 My understanding was that `joined` only loaded the relationship but not the data until you access it.
I may be wrong though. Excellent job on exploring and adding more performing options

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -60,9 +62,41 @@ def get(self, mine_no_or_guid=None):
...
+                    records = MineMapViewLocation.query.filter(MineMapViewLocation.latitude != None)
+                    last_modified = datetime.now()
+
+                    # jsonify then store in cache
+                    map_result = json.dumps(
+                        {
+                            'mines': list((map(lambda x: x.json_for_map(), records)))
+                        },
+                        separators=(',', ':'))
+                    cache.set(MINE_MAP_CACHE, map_result, timeout=43140)
```
FreshGyan: What's the reasoning behind this timeout value?

python-backend/app/config.py
```diff
@@ -64,14 +62,15 @@ class TestConfig(Config):
...
-        DB_NAME_TEST)
+    DB_URL = "postgresql://{0}:{1}@{2}:{3}/{4}".format(Config.DB_USER, Config.DB_PASS,
+                                                       Config.DB_HOST, Config.DB_PORT, DB_NAME_TEST)
     SQLALCHEMY_DATABASE_URI = DB_URL
     JWT_OIDC_TEST_MODE = True
     JWT_OIDC_TEST_AUDIENCE = "test_audience"
     JWT_OIDC_TEST_CLIENT_SECRET = "test_secret"
     JWT_OIDC_TEST_ISSUER = "test_issuer"
+    # Below enables functionalty we PR'd into the JWT_OIDC library to add caching
+    JWT_OIDC_CACHING_ENABLED = True
```
FreshGyan: This should get inferred from the above class and shouldn't need reinitializing

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -60,9 +62,41 @@ def get(self, mine_no_or_guid=None):
...
-                return {'mines': result}
+
+                # Below caches the mine map response object in redis with a timeout.
+                # Generating and jsonifying the map data takes 4-7 seconds with 50,000 points,
+                # so caching seems justified.
+                #
+                # TODO: Use some custom representation of this data vs JSON. The
+                # json string is massive (with 50,000 points: 16mb uncompressed, 2.5mb compressed).
+                # A quick test using delimented data brings this down to ~1mb compressed.
+                map_result = cache.get(MINE_MAP_CACHE)
```
FreshGyan: Any reason why you're explicitly setting and getting from cache instead of using the header?
`@cache.cached(timeout=43140, key_prefix=MINE_MAP_CACHE)`

My understanding was if it's in cache, flask will just return the json response from the cache. but if not, it'll run the whole function and return data from there?

frontend/src/components/maps/MinePin.js
```diff
@@ -126,22 +106,22 @@ export class MinePin extends Component {
...
-        const fclData = mineIds.reduce((result, id) => {
-          const point = this.points(id);
-          if (!point) {
-            return result;
+        const mapPopupString = renderToString(<MapPopup id="{mineId}" />);
+
+        // The previous code processing point data was nicely seperated into functions, but slow with
+        // large datasets taking 5-10 seconds for 50000 points. The code below is ~50ms for 50000 points
+        // and uses significantly lower memory.
+        const fclData = [];
```
FreshGyan: 🙏 💯 🔥 
Makes me almost wonder if we should refactor all uses of `.reduce` in our frontend given all our datasets are relatively large

python-backend/app/api/mines/location/models/mine_location.py
nayr974: Yes, removed.

python-backend/app/api/mines/location/models/mine_map_view_location.py
nayr974: It was used for testing a more compact representation of the data. It's not used right now, but I didn't see harm in leaving it as it's something I'd still like to work on.

FreshGyan: If it helps, there's a FLASK flag set which removes any whitespaces and tabs and newlines from our JSON responses

python-backend/app/api/mines/mine/models/mine.py
nayr974: No, lazy='joined' is a form of eager loading that will include those tables as part of a join statement:
select from mine m join mine_locaion ml on m.id=ml.id

lazy='selectin' is a form of eager loading that runs a seperate query to load the data instead of joining

lazy='select' is actual lazy loading

In my experiments, lazy='joined' is faster for 1:1 relationships, and lazy='selectin' is faster for 1:* relationships if you want to eager load.

python-backend/app/api/mines/mine/resources/mine.py
nayr974: Because it is explicit and readable. As well, since this map=true code path isn't on a separate endpoint I thought this would have lower risk of unintended consequences for the search and other uses. I don't want the search results to get cached or the map data to be returned on a search.

python-backend/app/api/mines/mine/resources/mine.py
nayr974: I thought 12 hours would cover use during business hours, but ensure that when our ETLs run at night the cache will be expired the next day.

python-backend/app/config.py
nayr974: Removed

python-backend/app/api/mines/mine/resources/mine.py
FreshGyan: and later on, we can use the scheduler to add that as well 👌

python-backend/app/api/mines/mine/resources/mine.py
FreshGyan: Hmm 🤔 Okay that makes sense to be safe. But I dont think anything is using this method, except when you pass map=true.
I believe you can use cache.memoize to ensure it takes into account the parameters passed to the function.

python-backend/app/api/mines/location/models/mine_map_view_location.py
nayr974: The whitespace / tab behavior is overridden by flask_restplus's marshaling. We are already setting restplus to remove whitespace when it does so, but it does add newline's which can't be disabled.

Using delimiters would remove the need to define each property name eg "guid", "latitude", "longitude" over and over again, which is what I was testing. The difference is notable!

frontend/src/components/dashboard/Dashboard.js
nayr974: I gave it a shot and it's not happy. I think it's related to authenticationguard on that component conflicting with connecting to the store. I've rolled my WIP changes back for now and would like to keep this as is as it's still a solid improvement.

python-backend/app/api/mines/mine/resources/mine.py
nayr974: The main page mine list and the advanced search call this same method, there is a just a different code path for mine results. At one point I was going to make this a separate endpoint, but with the way the frontend handles the List / Map tab view it would have needed a whoooole bunch of frontend refactoring as well. Out of scope for now.

frontend/src/components/dashboard/Dashboard.js
mmumby: Understandable! Thank you for attempting.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: From my understanding you cant use memoize here since this method takes in a reference to self. When memoize caches the function it uses the function and variables as the key. Since the reference to self changes every time you will just run this and cache it every time instead of grabbing the value from the cache.

PR #466, opened 2019-01-28
bug/mds 1276 null mine locations causing issues

Jsyro:
backend nulls mine_location if lat is not populated. Makes two cases instead of three for the front end to handle. other cases where it was expecting list. modified view query to match normal querys.

### DIFF
--- a/python-backend/app/api/mines/mine/resources/mine.py
+++ b/python-backend/app/api/mines/mine/resources/mine.py
@@ -296,7 +296,6 @@ class MineResource(Resource, UserMixin, ErrorMixin):
             self.raise_error(400, 'latitude and longitude must both be empty, or both provided')
         if mine.mine_location:
             #update existing record
-            #raise Exception("update the mine_location")
             if "latitude" in data.keys():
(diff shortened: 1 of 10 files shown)


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -60,7 +60,7 @@ def get(self, mine_no_or_guid=None):
             # Handle MapView request
             _map = request.args.get('map', None, type=str)
             if _map and _map.lower() == 'true':
-                records = MineMapViewLocation.query.all()
+                records = MineMapViewLocation.query.filter(MineMapViewLocation.latitude != None)
```
nayr974: Can you modify the view's query to include a where lat and long are not null clause instead?

Jsyro: db view will need to be updated when mine_location and mine tables are merged. (and possibly obsolete?). This is the quickest fix for the current issue for end of sprint. I don't disagree though.

PR #467, opened 2019-01-29
MDS-1185 cache NRIS major mine data

JustinMacaulay:
Added a scheduler as well as 2 scheduled jobs to pull and cache NRIS data for all major mines every night.


python-backend/app/extensions.py
```diff
@@ -1,15 +1,22 @@
+import os
 from flask_caching import Cache
 from flask_jwt_oidc import JwtManager
 from flask_sqlalchemy import SQLAlchemy
 from flask_uploads import UploadSet, AllExcept, EXECUTABLES, SCRIPTS
 from flask_restplus import Api
+from pytz import timezone
```
nayr974: Did you end up using this? What's it for?

JustinMacaulay: No, I will remove it. I was trying to set the schedulers timezone but it wouldn't let me.

python-backend/app/__init__.py
```diff
@@ -39,13 +41,17 @@ def create_app(test_config=None):
...
     cache.init_app(app)
     db.init_app(app)
     jwt.init_app(app)
-
+    if app.config.get('ENVIRONMENT_NAME') == 'prod':
+        sched.init_app(app)
+        if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == 'true':
+            sched.start()
+            _schedule_jobs(app)
     # Following is a simple example to demonstrate redis connection working
```
FreshGyan: Can you remove the comment here?

python-backend/app/__init__.py
```diff
@@ -39,13 +41,17 @@ def create_app(test_config=None):
...
     api.app = app
     api.init_app(app)
-
     cache.init_app(app)
     db.init_app(app)
     jwt.init_app(app)
-
+    if app.config.get('ENVIRONMENT_NAME') == 'prod':
+        sched.init_app(app)
+        if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == 'true':
```
FreshGyan: Okay this takes away the concern I had partially. :)

Now for a single uwsgi context, this will only run on one main thread. But for our production instances, we have two separate pods serving the app.
Each pod has its own uwsgi context. So at the least, the scheduler will run twice for each thing. I dont have a good solution off the top of my head. But this will definitely be a concern we should look out for

openshift/_python36.dc.json
```diff
@@ -99,6 +99,13 @@
     {
       "name": "REPLICA_MAX",
       "value": "1"
+    },
+    {
+      "name": "ENVIRONMENT_NAME",
+      "displayName": "Environment Name (Environment Id)",
+      "description": "The name or Id of the environment.  This variable is used by the webhook integration to identify the environment in which the backup notifications originate.",
+      "required": false,
```
FreshGyan: Can you put in a default value here?
Usually if something is not required we need to give it a default value or vice versa.

python-backend/app/api/constants.py
```diff
@@ -404,7 +404,16 @@
...
+#Cache prefixes
+NRIS_MMLIST_JOB = "processed_list"
+NRIS_MAJOR_MINE_LIST = "major_mine_list"
+NRIS_JOB_PREFIX = "nris_sched_job_"
+NRIS_CACHE_PREFIX = "nris_"
 
+#Cache Timeouts
+TIMEOUT_60_MINUTES = 3600
+TIMEOUT_24_HOURS = 86340
+TIMEOUT_12_HOURS = 43140
```
FreshGyan: 🙏 💯

python-backend/app/api/nris_services/NRIS_service.py
```diff
@@ -4,45 +4,57 @@
...
+        }
+        url = current_app.config['NRIS_TOKEN_URL']
+        if url is None:
+            raise TypeError('Could not load the NRIS URL.')
+        else:
+            resp = requests.get(
+                url=url,
+                params=params,
+                auth=(current_app.config['NRIS_USER_NAME'], current_app.config['NRIS_PASS']))
+            try:
```
FreshGyan: Instead of raising the exception and checking the response like this. Can you not check the status code and then return appropriate response?
Is NRIS doing someting funky like returning 200 at HTTP level and then 400 in the body because of which this is required?

python-backend/app/api/nris_services/NRIS_service.py
```diff
@@ -4,45 +4,57 @@
...
+                resp.raise_for_status()
+            except:
+                raise
 
-        return resp.json().get('access_token')
+            if resp.status_code != 200:
```
FreshGyan: Is this check even relevant? Whatever value you set in here, gets overriden in the next line, it seems

python-backend/app/scheduled_jobs/NRIS_jobs.py
```diff
@@ -0,0 +1,46 @@
...
+        trigger='cron',
+        id='get_major_mine_NRIS_data',
+        hour=9,
+        minute=10)
+
+
+# caches a list of mine numbers for all major mines and each major mine indavidually
+# to indicate whether of not it has been processed.
+def _cache_major_mines_list():
+    if sched.app.config.get('ENVIRONMENT_NAME') == 'prod':
```
FreshGyan: Do you need this check here in addition to the one in init.py?

python-backend/app/config.py
```diff
@@ -57,6 +58,8 @@ class Config(object):
     # 100MB file limit
     MAX_CONTENT_LENGTH = 100 * 1024 * 1024
 
+    SCHEDULER_API_ENABLED = True
```
FreshGyan: If this flag is what controls for the scheduler to run, can we just set this to false in dev and test and True in Prod?
Would that save us from the hassle of checking the environment variable all over the place?

python-backend/app/scheduled_jobs/NRIS_jobs.py
```diff
@@ -0,0 +1,46 @@
...
+                cache.set(NRIS_JOB_PREFIX + mine.mine_no, 'False', timeout=TIMEOUT_60_MINUTES)
+            cache.set(
+                NRIS_JOB_PREFIX + NRIS_MAJOR_MINE_LIST, major_mine_list, timeout=TIMEOUT_60_MINUTES)
+
+
+# Using the cached list of major mines procees them if they are not already set to true.
+def _cache_all_NRIS_major_mines_data():
+    if sched.app.config.get('ENVIRONMENT_NAME') == 'prod':
+        with sched.app.app_context():
+            major_mine_list = cache.get(NRIS_JOB_PREFIX + NRIS_MAJOR_MINE_LIST)
```
FreshGyan: Are we using redis as a queue management system?  Basically mark each mine job that needs to be executed gets marked as false and then when it gets queued, it gets marked to true.

That may or may not cause some grief if lets say some kubernetes pod is restarted in the middle of night either for python or redis.
For now, this implementation makes sense, but lets make a note of that and think about implementing celery or something more permanent to manage state of the jobs 👍

python-backend/app/scheduled_jobs/NRIS_jobs.py
```diff
@@ -0,0 +1,46 @@
...
+            if major_mine_list is None:
+                return
+
+            for mine in major_mine_list:
+                if cache.get(NRIS_JOB_PREFIX + mine) == 'False':
+                    cache.set(NRIS_JOB_PREFIX + mine, 'True', timeout=TIMEOUT_60_MINUTES)
+                    data = NRIS_service._get_EMPR_data_from_NRIS(mine)
+                    NRIS_service._process_NRIS_data(data, mine)
```
FreshGyan: Don't need to do it as part of this PR, but It might be beneficial to use threads here. As each thread doesn't depend on one another you can spawn several for each mine guid and let them do their things at night. Given that we don't really expect any high volumes of traffic, the extra CPU hit should be okay.  An example is in create_data under commands.py

openshift/_python36.dc.json
JustinMacaulay: Will do, this is copied from your database backup though so you might need one there too.

python-backend/app/__init__.py
JustinMacaulay: The way the jobs are set up it shouldn't be a problem if they run on both of the pods.

python-backend/app/__init__.py
JustinMacaulay: Done.

python-backend/app/api/nris_services/NRIS_service.py
JustinMacaulay: This call is nested in another api call so I need to raise the exception up a level so that the top level request can return the correct response.

python-backend/app/api/nris_services/NRIS_service.py
JustinMacaulay: It is not, I missed that when I was refactoring things. Thanks.

python-backend/app/config.py
JustinMacaulay: When I was first adding the scheduler I brought this over thinking it was needed but this just adds endpoints to the app that allow for api scheduling which I don't like and have now set that to false.

python-backend/app/scheduled_jobs/NRIS_jobs.py
JustinMacaulay: Not any more. I added the check in the init because I was getting a strange issue in the test phase where they were all failing because the scheduler was being started for every test.

openshift/_python36.dc.json
FreshGyan: Good catch! Can you update that in your PR? If not, im happy to do in one of my follow ups :)

PR #470, opened 2019-01-29
APM Implementation, Adding Flask Depend, and Secret Keys.

itchison:


python-backend/app/__init__.py
```diff
@@ -16,6 +17,7 @@
 from app.commands import register_commands
 from app.config import Config
 from app.extensions import db, jwt, api, documents, cache
+from elasticapm.contrib.flask import ElasticAPM
```
FreshGyan: Is this the only place we'll be using this?
If this is going to be imported in multiple places can we move this to extensions.py?

Im assuming this is a one time setup and then it just monitors the application on its own?

itchison: Yes, this will be the only place it's used.

PR #478, opened 2019-01-31
MDS-1243 Feature/Add warning banners for both CORE and minespace

mmumby:
IE Warning for CORE/minespace
Test env banner for CORE


frontend-public/src/components/LandingPage.js
```diff
@@ -78,7 +78,6 @@ export const LandingPage = () => (
             Wait for an email confirming that your employees/delegates have been added to MineSpace.
           </li>
           <li>Log in to the system.</li>
-          <li>If you require a tutorial on how to use the system, watch this video.</li>
```
mmumby: In case anyone is wondering about this deletion, Aaron asked for it to be removed until a video is created.

itchison:
When I open https://mds-empr-mds-dev.pathfinder.gov.bc.ca/478/ in IE I only get a white screen. Url works fine in Chrome.

frontend-public/src/utils/environmentUtils.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * detect IE
+ * @returns version of IE or false, if browser is not Internet Explorer
+ */
+// This file is anticipated to have multiple exports
+// eslint-disable-next-line import/prefer-default-export
+export const detectIE = () => {
```
NWCalvank: Were you able to test this somehow? I would think with the current IE issue, that wouldn't be possible

frontend/src/components/Home.js
```diff
@@ -1,45 +1,61 @@
...
-};
+export class Home extends Component {
+  state = { isIE: false, isTest: false };
+
+  componentDidMount() {
+    this.setState({ isIE: detectIE(), isTest: detectTestEnvironment() });
+  }
+
+  handleIEClose = () => {
+    // this.setState({ isIE: false });
```
NWCalvank: This looks like it should be removed or enabled

frontend-public/src/utils/environmentUtils.js
mmumby: Haven't tested yet, need to deploy to dev to test the logic!

It doesn't show up for chrome or safari.

frontend/src/components/Home.js
mmumby: uncommented!

PR #480, opened 2019-01-31
[MDS-1213] Fixing Mines List Responsive Views

jeznorth:


frontend/src/components/dashboard/MineList.js
```diff
@@ -77,7 +99,7 @@ const columns = [
     dataIndex: "commodity",
     width: 150,
     render: (text, record) => (
-      <div>
+      <div className="td-inner" data-label="Commodity" title="Commodity">
```
mmumby: Is `title` doing anything here? (And on all the divs where its been added?)

jeznorth: We have some ambiguous naming conventions here (see TSF) which I've added a title tag for accessibility purposes. I've translated this across to each "td-inner" div for consistency.

frontend/src/styles/components/MineList.scss
```diff
@@ -0,0 +1,16 @@
+.mine-list table {
+  min-width: 1500px;
+}
```
mmumby: When inspecting the code through dev tools, this line doesn't seem to be taking effect. 

Also, `.ant-table-wrapper` appears to be the same outer div as your custom `.mine-list`, 
might be better to use that as the parent class instead. 

screen shot 2019-02-04 at 10 38 05 am
screen shot 2019-02-04 at 10 37 57 am

frontend/src/styles/components/Tables.scss
```diff
@@ -1,3 +1,53 @@
+.ant-table-body {
+  overflow-x: auto;
+}
+
+@include tablet {
```
mmumby: This is great! Looks very generic

frontend/src/components/dashboard/MineList.js
```diff
@@ -25,28 +25,44 @@ const columns = [
...
-          ))}
+      <div title="Permit Number">
+        <ul className="mine-list__permits">
+          {text &&
+            uniqBy(text, "permit_no").map(({ permit_no, permit_guid }) => (
+              <li key={permit_guid}>{permit_no}</li>
+            ))}
+        </ul>
         {!text && <div>{record.emptyField}</div>}
       </div>
```
mmumby: The alignment on the 'N/A' is off if there are no permits

frontend/src/styles/components/MineList.scss
jeznorth: This is specific to the mine-list - I don't think we'd want to assume that every '.ant-table-wrapper' in the application should default to a min-width of 1500px.

The div you are inspecting is sizing as expected in those screenshots. This selector is targeting the TABLE element specifically, not it's encapsulating DIV.

frontend/src/styles/components/MineList.scss
```diff
@@ -0,0 +1,16 @@
...
+
+// Permit Column List
+.mine-list__name {
+  font-weight: 500;
+}
```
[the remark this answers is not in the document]
jeznorth: Removed as requested.

frontend/src/components/dashboard/MineList.js
jeznorth: Updated

frontend/src/styles/components/MineList.scss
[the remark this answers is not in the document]
mmumby: Thank you! 

Yeah that's a known issue, we have adopted a "fix-going-forward" approach to some things, this being one of them, The typography was updated a few sprints ago, and not every component has been updated to reflect that yet, but every new implementation should follow the new standards, (which are found in elements.scss)

frontend/src/styles/components/MineList.scss
mmumby: two things, 
1) why is this specific to mine-list? We do want a standard across all ant design tables. 
2) When inspecting the css through the dev tools, I didn't see this styling being applied anywhere, that's what led me to believe it wasn't being targeted correctly.

frontend/src/components/dashboard/MineList.js
```diff
@@ -121,6 +148,7 @@ const transformRowData = (mines, mineIds, mineRegionHash, mineTenureHash, mineCo
 export const MineList = (props) => (
   <Table
     align="center"
+    className="test"
```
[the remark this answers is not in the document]
jeznorth: Test isn't - i've updated to use the correct selector

frontend/src/styles/components/MineList.scss
jeznorth: I'm not sure. The previous developer had left this table with a hard-coded width of 1500px - I'm just continuing on with that pattern - but leveraging CSS instead of hard-coding the issue (as it was done previously).

PR #483, opened 2019-01-31
added config setting for env name

JustinMacaulay:


JustinMacaulay:
All of the other required were made to the NRIS cache branch that was already merged in. This was the one piece that was over looked.

PR #484, opened 2019-01-31
MDS-1271-no-longer-show-deleted-mines-in-mds

JustinMacaulay:


migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -71,7 +71,8 @@ BEGIN
         longitude      = mms.mmsmin.lon_dec,
         major_mine_ind = transform_major_mine_ind(mms.mmsmin.min_lnk),
         mine_region    = transform_mine_region(mms.mmsmin.reg_cd)    ,
-        mine_type      = transform_mine_type_code(mms.mmsmin.mine_typ)
+        mine_type      = transform_mine_type_code(mms.mmsmin.mine_typ),
+        deleted_ind    = CASE WHEN lower(mms.mmsmin.mine_nm) LIKE '%delete%' OR lower(mms.mmsmin.mine_nm) LIKE '%deleted%' OR lower(mms.mmsmin.mine_nm) LIKE '%reuse%' THEN TRUE ELSE FALSE
```
NWCalvank: It looks like this is missing and `END`

Also, the `CASE` itself should be redundant here. This should give you the same thing:
```sql
deleted_ind = LOWER(mms.mmsmin.mine_nm) LIKE '%delete%' OR LOWER(mms.mmsmin.mine_nm) LIKE '%deleted%' OR LOWER(mms.mmsmin.mine_nm) LIKE '%reuse%'
```

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -108,7 +110,8 @@ BEGIN
         transform_mine_type_code(mms_new.mine_typ),
         mms_new.lat_dec    ,
         mms_new.lon_dec    ,
-        transform_major_mine_ind(mms_new.min_lnk)
+        transform_major_mine_ind(mms_new.min_lnk),
+        CASE WHEN lower(mms.mmsmin.mine_nm) LIKE '%delete%' OR lower(mms.mmsmin.mine_nm) LIKE '%deleted%' OR lower(mms.mmsmin.mine_nm) LIKE '%reuse%' THEN TRUE ELSE FALSE
```
NWCalvank: Your choice, but this seems like a good option for a function, like the `transform_` functions above it. I stuck repeated business logic into functions to help DRY things up

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -166,9 +171,10 @@ BEGIN
         new.mine_no         ,
         new.mine_name       ,
         new.mine_region     ,
-        major_mine_ind      ,
+        new.major_mine_ind  ,
+        new.deleted_ind     ,
         'mms_migration'     ,
-        now()               ,
+        now()               ,
```
NWCalvank: Trailing whitespace

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -18,7 +18,7 @@ class Mine(AuditMixin, Base):
     mine_name = db.Column(db.String(60), nullable=False)
     mine_note = db.Column(db.String(300), default='')
     major_mine_ind = db.Column(db.Boolean, nullable=False, default=False)
-    deleted_ind = db.Column(db.Boolean, nullable=False, default=True)
+    deleted_ind = db.Column(db.Boolean, nullable=False, default=False)
```
NWCalvank: I think this could use the fetch function to pull the default from the db, which is the pattern to use going forward, instead of hardcoding the defaults into the application level

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -123,6 +123,8 @@ def apply_filter_and_search(self, args):
         tsf_filter_term = args.get('tsf', None, type=str)
         # Base query:
         mines_query = Mine.query
+        # Adding a filter to remove deleted mines
+        mines_query = Mine.query
```
NWCalvank: It looks like you're immediately reassigning `mines_query` to the same value?

migrations/etl/1_Mine_profile_detail_location.sql
JustinMacaulay: since the case is in a select statement or the part of a statement you don't want the END.If I add the END it won't run and this currently works on my dev branch.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: Good catch I added this before I used apply_filters instead and forgot to remove it.

python-backend/app/api/mines/mine/models/mine.py
JustinMacaulay: changed.

migrations/etl/1_Mine_profile_detail_location.sql
JustinMacaulay: removed.

migrations/etl/1_Mine_profile_detail_location.sql
JustinMacaulay: I also tried swapping out the case for this logical statement and postgres wouldn't run it.

NWCalvank: Hmm... that doesn't seem right to me, since I can run this query just fine:
```
mds=> select LOWER(mine_nm) LIKE '%delete%' OR LOWER(mine_nm) LIKE '%deleted%' OR LOWER(mine_nm) LIKE '%reuse%' AS deleted_ind from mms.mmsmin LIMIT 10;
 deleted_ind
-------------
 f
 f
 f
 f
 f
 f
 f
 f
 f
 f
(10 rows)
```

JustinMacaulay: I added this back in to the first case but I think I had added it for both the first time. It worked this time around.

PR #493, opened 2019-02-04
mine manager ETL update

Jsyro:
Working with @klefler to pull Regional mine manager from the latest notice of work, not the most recently updated contact.


migrations/etl/2_Mine_manager.sql
```diff
@@ -26,47 +26,35 @@ BEGIN
...
-        INNER JOIN mms.mmsccn contact_info on
-            contact_info.cid=now_contact.cid_ccn
-        WHERE
-            SUBSTRING(now_contact.type_ind from 3 for 1) = 'Y'
+
+    --Step 1: Get regional mine_no's
+    WITH regional_mine AS (
+        SELECT mine_no
+        FROM mms.mmsmin
+        WHERE (min_lnk = 'N' OR min_lnk is null)
```
NWCalvank: Might make sense to say "not a major mine" `WHERE min_lnk != 'Y'` instead?

migrations/etl/2_Mine_manager.sql
```diff
@@ -187,51 +175,38 @@ BEGIN
...
-    --4. Select new manager record
+    -- Step 4, get contact from contact connection
     new_manager AS (
-        SELECT
-            mine_no                         ,
-            contact_cid AS person_combo_id  ,
-            mine_no||contact_cid AS mgr_combo_id
-        FROM latest_manager
-        where mine_no||contact_cid NOT IN (
+        SELECT latest_now_ccc.mine_no mine_no, cn.cid person_combo_id, latest_now_ccc.mine_no||cn.cid as mgr_combo_id
```
NWCalvank: Could you put the selected columns on separate lines to make it easier to read with the aliases?

migrations/etl/2_Mine_manager.sql
Jsyro: Postgres is weird, that way results being excluded when min_lnk is null.

(https://stackoverflow.com/questions/36508815/not-equal-and-null-in-postgres).

migrations/etl/2_Mine_manager.sql
Jsyro: done

PR #496, opened 2019-02-05
Bug/mds-1354 Fix file downloads

james-hollinger:
Currently, file downloads are broken.
This fix corrects the issue (no permissions to access Document Manager from the database). 
This also fixes the wonky workaround of opening the file download in a new window, which also allows us to restore the authorization on the doc manager GET endpoint.

Some linting issues are pending.


frontend/package.json
```diff
@@ -17,6 +17,7 @@
     "filepond-plugin-file-validate-type": "^1.2.1",
     "filepond-polyfill": "^1.0.2",
     "hoist-non-react-statics": "^3.2.1",
+    "js-file-download": "^0.4.4",
```
mmumby: Did you get a chance to check that this new dependency works with IE?

frontend-public/src/utils/helpers.js
```diff
@@ -44,3 +51,15 @@ export const createLabelHash = (obj) =>
...
 export const formatDate = (dateString) => moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");
+
+export const downloadFileFromDocumentManager = (docManagerGuid, filename) => {
+  if (!docManagerGuid || !filename) {
+    throw new Error("Must provide both docManagerGuid and filename");
+  }
+
+  // TODO: Update url when Document Manager moves to its own microservice.
+  const url = `${ENVIRONMENT.apiUrl + DOCUMENT_MANAGER_FILE_GET_URL}/${docManagerGuid}`;
+  axios.get(url, { responseType: "arraybuffer", ...createRequestHeader() }).then((response) => {
```
mmumby: this seems to be more of an actionCreator than a util function.

james-hollinger: I wasn't sure if it should be an action creator or not. This isn't touching the store or modifying any state, which is the normal definition of an action creator.

james-hollinger: After discussion: Move the method into a new utils file, as its neither a static helper function nor an action creator.

frontend/package.json
james-hollinger: Not yet, this is a good point. Will add this note to PR#490 as we merge them together; PR#490 also has a new js library to test with IE.

james-hollinger:
To better test both this PR and PR#490, I will be merging this PR into PR#490.

PR #498, opened 2019-02-05
Feature/MDS-1217 Chunked file uploads for public site

james-hollinger:
This should be a carbon copy of the changes made in PR#490 for the public site.

Merge after PR#490


frontend-public/src/tests/actionCreators/userDashboardActionCreator.spec.js
```diff
@@ -127,26 +127,26 @@ describe("`removeMineDocumentFromExpectedDocument` action creator", () => {
...
 
-describe("`addMineDocumentToExpectedDocument` action creator", () => {
+describe("`addDocumentToExpectedDocument` action creator", () => {
   const expDocGuid = "12345-6789";
   const url = ENVIRONMENT.apiUrl + API.UPLOAD_MINE_EXPECTED_DOCUMENT_FILE(expDocGuid);
 
-  it("Request with new file, dispatches `success` with correct response", () => {
-    const payload = { file: "a file" };
+  it("Request with new upload, dispatches `success` with correct response", () => {
+    const payload = { filename: "a file", document_manager_guid:"a GIUD" };
```
FreshGyan: GUID typo

frontend-public/src/components/common/FileUpload.js
```diff
@@ -15,11 +16,15 @@ const propTypes = {
...
   acceptedFileTypesMap: PropTypes.objectOf(PropTypes.string),
+  onFileLoad: PropTypes.func,
+  chunkSize: PropTypes.number,
 };
 
 const defaultProps = {
   maxFileSize: "100MB",
   acceptedFileTypesMap: {},
+  onFileLoad: () => {},
+  chunkSize: 1048576, //1MB
```
FreshGyan: Can you make this like 5MB or something? 1MB was cool to see in testing. But if we are getting like 50 MB files, we probably don't want 50 network calls and overwhelm uwsgi and/or nginx and such

nayr974: These aren't async so 1 connection will still be consumed persistently while uploading. Maybe we need metrics on this but I feel like even 5mb could be a lot to ask on rural / spotty networks...

frontend-public/src/tests/actionCreators/userDashboardActionCreator.spec.js
james-hollinger: Fixed

PR #503, opened 2019-02-06
added cli command to run the NRIS jobs manually.

JustinMacaulay:
Added a simple command to allow the NRIS jobs to be run from the terminal in the backend pods.


python-backend/app/commands.py
```diff
@@ -162,3 +163,22 @@ def delete_data():
             db.session.rollback()
             click.echo(f'Error, failed on commit.')
             raise
+
+    #Had to define this function twice since the annotation used is different based on the environment.
```
NWCalvank: Is there a reason not to put the scheduled tasks in their own file? The rest of these commands are meant for development only to manage mock seed data. This command serves an entirely different concern.

Also, why is it necessary to provide this method at all in a non-production environment? If someone tried to run this in a test or dev instance, I'm sure a 'method not defined' error would be acceptable.

JustinMacaulay: I don't really see a reason to put them in separate files if we had a ton of commands registered to the app I could see splitting them up but at the moment we only have a handful. As for the If statement I have removed it as it was needed and then became unnecessary so I have removed it.

NWCalvank: I would argue that even just two commands for entirely different environments/purposes would still warrant two separate files.

What's the harm in having a separate scheduled tasks or production tasks file?

JustinMacaulay: I'm unsure what you asking for here. There are separate files for the actual tasks this is just a flask command being registered so that they can be force started from the pods CLI. This command doesn't have anything to do with scheduling.

NWCalvank: My mistake. I thought this was your ETL automation PR.

Unless I'm totally missing something here, my point still stands about having a file for production-only commands, though, no?

python-backend/app/commands.py
```diff
@@ -22,7 +22,9 @@
 from .api.mines.status.models.mine_operation_status_sub_reason_code import MineOperationStatusSubReasonCode
 from .api.utils.random import generate_mine_no, generate_mine_name, random_geo, random_key_gen, random_date, random_region, random_mine_category
 from .api.parties.party_appt.models.mine_party_appt import MinePartyAppointment
-from .extensions import db
+from .extensions import db, sched
+from .scheduled_jobs import NRIS_jobs
```
NWCalvank: Unused import?

python-backend/app/commands.py
```diff
@@ -161,3 +163,15 @@ def delete_data():
             db.session.rollback()
             click.echo(f'Error, failed on commit.')
             raise
+
+    if app.config.get('ENVIRONMENT_NAME') == 'prod':
+
+        @sched.app.cli.command()
```
NWCalvank: Can you explain the use of `@sched` and `sched.` in these changes if this has nothing to do with scheduling? I'm honestly not clear on what these are doing if these changes aren't related to scheduling

JustinMacaulay: So, since the function I'm calling lives in the scheduled jobs files I need to use the same application context that it is expecting which is that of sched.app.app_context. So when I register these I need to use the app that is associated with the scheduler.

python-backend/app/commands.py
JustinMacaulay: The function being called by the CLI command lives in the NRIS_jobs file.

python-backend/app/commands.py
JustinMacaulay: I tried running this using the app.cli.command() instead but since the scheduled jobs or the functions the scheduled jobs are using rely on the context of the app registered to the scheduler it doesn't work. So it has to be registered using the app reference in the scheduler.

python-backend/app/commands.py
```diff
@@ -161,3 +163,15 @@ def delete_data():
             db.session.rollback()
             click.echo(f'Error, failed on commit.')
             raise
+
+    if app.config.get('ENVIRONMENT_NAME') == 'prod':
```
Jsyro: Do we want a placeholder for if someone is not in prod, so they get a 'This method is only available in prod' to make it obvious, or are we ok with a basic 'command not found'?

Jsyro: I see this has been discussed above , disregard.

python-backend/app/commands.py
Jsyro: I'd rather see all the commands in the commands.py files, the condition (or a comment) explains the intent. If these commands start growing any sort of interdependence then maybe it makes sense to break them down, i think it's unnecessary to have multiple 'commands' files right now.

JustinMacaulay: I don't see a reason to split up the commands at the moment, there are so few of them that I don't think adding another file will benefit us.

PR #509, opened 2019-02-07
MDS-1356 Fix/ Refine mine map sort by name

NWCalvank:
Summary of changes:
- Break out the overloaded `search` query param into `name` (name-only search) and `term` (name, mine_no, permit_no search)
- Rename `/names` route to `search`
- Fix some linting errors
- Update package-lock.json (`npm ci` found it was out of sync with package.json)


python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -4,6 +4,7 @@
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
 from ....utils.models_mixins import AuditMixin, Base
+from ....permits.permit.models.permit import Permit
```
nayr974: This does break our discussions on treating resources as separate services. I get the benefit here because filtering on the relationship is easier and probably faster than getting a set separately by another request. If we're doing this, can you please add a comment describing the exception? 

I've thought about search a few times and I think what we might want in the future is a separate search model that just kind of has access to everything and returns results in a generic way. Then, our other models are more self-concerned and even if everything is on separate micro-services it can deal with that centrally for searching.

NWCalvank: I was definitely aware of this as I did it. The reason I didn't worry about it is because this model is already being imported into the Mine resource. The reason why I needed to pull it in here is because I moved that logic to the model from the resource.

So, while this 'breaks' the pattern, it had already been broken here. My hope was to get this in with the naive implementation for now.

I can add a comment in the meantime. I'll add it to the resource as well.

nayr974: Sounds good, thanks!

PR #510, opened 2019-02-08
MDS-1362-cron-job-to-run-ETL

JustinMacaulay:


python-backend/app/__init__.py
```diff
@@ -45,15 +46,18 @@ def register_extensions(app):
...
     jwt.init_app(app)
     apm.init_app(app) if app.config['ELASTIC_ENABLED'] == '1' else None
+    sched.init_app(app)
 
     CORS(app)
     Compress(app)
 
-    if app.config.get('ENVIRONMENT_NAME') == 'prod':
-        sched.init_app(app)
+    if app.config.get('ENVIRONMENT_NAME') == 'test' or app.config.get('ENVIRONMENT_NAME') == 'prod':
```
FreshGyan: You can probably do something like `condition in ['test', 'prod']` for this.

python-backend/app/__init__.py
```diff
@@ -45,15 +46,18 @@ def register_extensions(app):
...
         if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == 'true':
             sched.start()
             _schedule_NRIS_jobs(app)
+            #This is her to prevent this from running in production until we are confident in the permit data.
```
FreshGyan: typo in comment and the line below should just not be there if we're not running it 🤷‍♂️ It's not being controlled by a flag or anything so im not certain why the always failing if is there

python-backend/app/commands.py
```diff
@@ -164,7 +165,7 @@ def delete_data():
             click.echo(f'Error, failed on commit.')
             raise
 
-    if app.config.get('ENVIRONMENT_NAME') == 'prod':
+    if app.config.get('ENVIRONMENT_NAME') == 'test' or app.config.get('ENVIRONMENT_NAME') == 'prod':
```
FreshGyan: Can you look at refactoring the scheduler initialization a little to ensure we dont keep this check all over the place?

python-backend/app/scheduled_jobs/ETL_jobs.py
```diff
@@ -0,0 +1,20 @@
...
+
+#the schedule of these jobs is set using server time (UTC)
+def _schedule_ETL_jobs(app):
+    app.apscheduler.add_job(func=_run_ETL, trigger='cron', id='ETL', hour=11, minute=0)
+
+
+@register_apm
+def _run_ETL():
+    with sched.app.app_context():
+        db.session.execute('select transfer_mine_information();')
```
FreshGyan: Just wondering how the naming conventions for these functions came to be

python-backend/app/scheduled_jobs/NRIS_jobs.py
```diff
@@ -23,14 +23,16 @@ def _schedule_NRIS_jobs(app):
...
     with sched.app.app_context():
-        cache.set(NRIS_JOB_PREFIX + NRIS_MMLIST_JOB, 'True', timeout=TIMEOUT_24_HOURS)
-        major_mines = Mine.query.unbound_unsafe().filter_by(major_mine_ind=True).all()
-        major_mine_list = []
-        for mine in major_mines:
-            major_mine_list.append(mine.mine_no)
-            cache.set(NRIS_JOB_PREFIX + mine.mine_no, 'False', timeout=TIMEOUT_60_MINUTES)
-        cache.set(
-            NRIS_JOB_PREFIX + NRIS_MAJOR_MINE_LIST, major_mine_list, timeout=TIMEOUT_60_MINUTES)
+        job_running = cache.get(NRIS_JOB_PREFIX + NRIS_MMLIST_JOB)
```
FreshGyan: Nice check and a good reminder for everyone that our redis isn't persistent 💯

python-backend/app/__init__.py
JustinMacaulay: Type aside, this comment is here to only prevent the ETL script from running nightly in production until we are confident in the permit changes and the information that is being brought in.

python-backend/app/scheduled_jobs/ETL_jobs.py
JustinMacaulay: If its the _ prefix your wondering about it's there because these are internal use functions.

python-backend/app/__init__.py
JustinMacaulay: Changed.

python-backend/app/commands.py
JustinMacaulay: Removed as with the refactor there shouldn't be a problem here.

python-backend/app/__init__.py
FreshGyan: Oh I understand the comment. What I was trying to ask is why is there a 
```
            if False:
                _schedule_ETL_jobs(app)
```

Where you can have no code in there if the code block never runs.
Or even just a 
```
# Comment describing line below
# _schedule_ETL_jobs(app)
```

Just bringing it up because its just a redundant code cycle check and may be picked up as sonarqube as code smell

python-backend/app/scheduled_jobs/ETL_jobs.py
FreshGyan: Oh i didn't mean the private _ prefix for python.
The function for postgres with the `transfer_` prefix. Just curious on that

JustinMacaulay: Oh, I just tried to keep the stored procedure names for the ETL as close to the original names of the files as I could.

python-backend/app/__init__.py
JustinMacaulay: Yeah, not an ideal situation but this code is what will be run when we are more comfortable with all of the permit and permitee information we are expecting and want to see. The if false was chosen as it was thought to look better than commented out code.

PR #511, opened 2019-02-09
Bug/MDS-1180 Refactor document status codes

james-hollinger:
The Document Status table used GUIDs as primary keys, causing duplication and other problems. The table has been refactored to be in line with our other code tables. This includes:

 - Table has been renamed to mine_expected_document_status_code
 - Primary key is now exp_document_status_code, a three character code
 - Expected document has been updated to reference the new table/column
 - Seed_data has been modifed to follow out other code tables

The backend model for expected documents now has a relationship to a status object rather than just a GUID/code. Of note this means the backend JSON for expected documents has an exp_document_status property rather than exp_document_status_code. The frontend model has been updated accordingly, and the workarounds for the old representation in the frontend have been removed. 

Expected document status has been almost entirely removed from the public site code, as it was copied over from the core code and is now unneeded. The workaround code has been removed and there are no document status drop-downs on the proponent site. This has resulted in quite a few empty reducer/selector/actionCreator files as the document status machinery was the only code there. I feel deleting this code is good as it means we don't have to maintain the code in two places, but I could easily be convinced to revert this change.


migrations/sql/V2019.02.07.02.27__alter_expected_document_status.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE mine_expected_document DROP CONSTRAINT mine_expected_document_exp_document_status_guid_fkey;
+ALTER TABLE mine_expected_document DROP COLUMN exp_document_status_guid;
+ALTER TABLE mine_expected_document ADD COLUMN exp_document_status_code character varying(3) NOT NULL DEFAULT 'MIA';
+
+DROP TABLE mine_expected_document_status;
+CREATE TABLE mine_expected_document_status_code (
+    exp_document_status_code character varying(3) PRIMARY KEY DEFAULT 'MIA',
+    description character varying(100) NOT NULL,
+    display_order smallint NOT NULL,
```
klefler: minor thought... would it make sense to have display_order as an optional field? In the case where a code is inactive, would there still be a display order value?

james-hollinger: Good point, thats probably a good change for maximum future compatability

frontend-public/src/actionCreators/staticContentActionCreator.js
```diff
@@ -1,32 +1,2 @@
...
-import { request, success, error } from "@/actions/genericActions";
-import * as reducerTypes from "@/constants/reducerTypes";
-import * as staticContentActions from "@/actions/staticContentActions";
-import * as String from "@/constants/strings";
-import * as API from "@/constants/API";
-import { ENVIRONMENT } from "@/constants/environment";
-import { createRequestHeader } from "@/utils/RequestHeaders";
-
 // This file is anticipated to have multiple exports
 // eslint-disable-next-line import/prefer-default-export
```
NWCalvank: It looks like this file can just be removed. This comment is just there to prevent the eslint rule from barking about having a single non-default export.

frontend-public/src/actions/staticContentActions.js
```diff
@@ -1,8 +1,2 @@
-import * as ActionTypes from "../constants/actionTypes";
-
 // This file is anticipated to have multiple exports
 // eslint-disable-next-line import/prefer-default-export
```
NWCalvank: Same comment here

frontend-public/src/components/common/UploadedFilesList.js
```diff
@@ -30,12 +28,8 @@ export class UploadedFilesList extends React.Component {
   };
 
   canUnlink = (selectedDoc) => {
-    const docStatus = this.props.expectedDocumentStatusOptions.find(
-      (status) => status.value === selectedDoc.exp_document_status_guid
-    );
     const dueDateCheck = moment(selectedDoc.due_date, "YYYY-MM-DD") > moment();
-    const statusCheck =
-      docStatus.label === "Not Received" || docStatus.label === "Received / Pending Review";
+    const statusCheck = selectedDoc.exp_document_status.exp_document_status_code in ['MIA', 'PRE']
```
NWCalvank: `in` operator is for objects, not arrays
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

My understanding is that this will always return `false`. You'll be looking for `Array.prototype.includes()`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

migrations/sql/V2019.02.07.02.27__alter_expected_document_status.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE mine_expected_document DROP CONSTRAINT mine_expected_document_exp_document_status_guid_fkey;
+ALTER TABLE mine_expected_document DROP COLUMN exp_document_status_guid;
+ALTER TABLE mine_expected_document ADD COLUMN exp_document_status_code character varying(3) NOT NULL DEFAULT 'MIA';
+
+DROP TABLE mine_expected_document_status;
```
Jsyro: DROP TABLE IF EXISTS, incase something has to run multiple times.

frontend-public/src/reducers/staticContentReducer.js
```diff
@@ -1,28 +1,4 @@
-import * as actionTypes from "@/constants/actionTypes";
```
NWCalvank: Is there any reason to keep this file? It looks like it should be deleted.

frontend/src/components/common/UploadedFilesList.js
```diff
@@ -31,12 +29,8 @@ export class UploadedFilesList extends React.Component {
   };
 
   canUnlink = (selectedDoc) => {
-    const docStatus = this.props.expectedDocumentStatusOptions.find(
-      (status) => status.value === selectedDoc.exp_document_status_guid
-    );
     const dueDateCheck = moment(selectedDoc.due_date, "YYYY-MM-DD") > moment();
-    const statusCheck =
-      docStatus.label === "Not Received" || docStatus.label === "Received / Pending Review";
+    const statusCheck = selectedDoc.exp_document_status.exp_document_status_code in ['MIA', 'PRE']
```
NWCalvank: Same comment here about `in` operator

frontend/src/components/mine/Tailings/MineTailingsInfo.js
```diff
@@ -212,10 +201,7 @@ export class MineTailingsInfo extends Component {
                     </Col>
                     <Col id={`status-${id}`} span={4}>
                       <h6 className={isOverdue ? "bold" : null}>
-                        <DocumentStatusText
-                          doc={doc}
-                          expectedDocumentStatusOptions={this.props.expectedDocumentStatusOptions}
-                        />
+                        {doc ? doc.exp_document_status.description : String.LOADING}
```
NWCalvank: Could we add this constant to `frontend-public` as well?

migrations/sql/V2019.02.07.02.27__alter_expected_document_status.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE mine_expected_document DROP CONSTRAINT mine_expected_document_exp_document_status_guid_fkey;
```
NWCalvank: Will this leave orphaned `mine_expected_document` records?

frontend-public/src/tests/reducers/staticContentReducer.spec.js
```diff
@@ -1,28 +1,5 @@
...
-  it("receives STORE_DOCUMENT_STATUS_OPTIONS", () => {
-    const expectedValue = getBaseExpectedValue();
-    expectedValue.expectedDocumentStatusOptions = MOCK.EXPECTED_DOCUMENT_STATUS_OPTIONS.options;
-    const result = staticContentReducer(
-      undefined,
-      storeDocumentStatusOptions(MOCK.EXPECTED_DOCUMENT_STATUS_OPTIONS)
-    );
-    expect(result).toEqual(expectedValue);
-  });
-});
```
mmumby: This file should be deleted, The code that is left behind is serving no purpose without a test suite.

frontend-public/src/tests/actionCreators/staticContentActionCreator.spec.js
```diff
@@ -19,25 +18,3 @@ beforeEach(() => {
...
-
-  it("Request failure, dispatches `error` with correct response", () => {
-    mockAxios.onGet(url, MOCK.createMockHeader()).reply(400, MOCK.ERROR);
-    return fetchExpectedDocumentStatusOptions()(dispatch).then(() => {
-      expect(requestSpy).toHaveBeenCalledTimes(1);
-      expect(errorSpy).toHaveBeenCalledTimes(1);
-      expect(dispatch).toHaveBeenCalledTimes(4);
-    });
-  });
-});
```
mmumby: This file can also be deleted.

frontend-public/src/selectors/staticContentSelectors.js
```diff
@@ -1,4 +1,2 @@
-import * as staticContentReducer from "@/reducers/staticContentReducer";
 // This file is anticipated to have multiple exports
 // eslint-disable-next-line import/prefer-default-export
-export const { getExpectedDocumentStatusOptions } = staticContentReducer;
```
mmumby: remove this file.

frontend-public/src/components/common/UploadedFilesList.js
james-hollinger: My mind was still in python mode 😅

james-hollinger: Good catch.

james-hollinger:
Files with no remaining logic have been removed.

frontend/src/components/mine/Tailings/MineTailingsInfo.js
james-hollinger: Fixed, thank you

migrations/sql/V2019.02.07.02.27__alter_expected_document_status.sql
james-hollinger: This is removing the foreign key relationship to the old mine document status table. I'm not sure how this would leave orphaned mine_expected_document records.

NWCalvank: I meant the file as a whole. I wasn't sure if removing the constraint without cleaning up the related records would lead to any issues.

nayr974: I don't see how it would orphan anything. If there are mine_expected_document records, they may lose any status details they had. the status table gets dropped completely, so nothing orphaned in the foreign table. I think we're good with this, as we should have no expected documents in production right now.

PR #512, opened 2019-02-09
Misc backend updates and fixes

FreshGyan:
- Fixed local unit tests database creation
- Changes most list accessors around the codebase to use .get() to avoid exceptions being thrown
- Updated APM invocation to be more clean and importable across codebase through extensions
- Removed certain unused imports
- Added some indexes and enabled parallelization on joins for ETL script #1. On dev the boost didn't seem significant. Only reduced the times by like 20secs but hoping in Test/Prod it'll be faster with more resources. Alternatively, doesn't seem to do any harm


itchison:
Looks good to me. APM is getting the details from this PR.

docker-compose.yml
```diff
@@ -42,7 +42,19 @@ services:
   flyway:
     container_name: mds_flyway
     image: boxfuse/flyway:5.2.3-alpine
-    command: -url=jdbc:postgresql://postgres/mds -user=mds -password=test -connectRetries=60 -outOfOrder=true migrate
```
FreshGyan: I changed the way we were creating test database a while back
https://github.com/bcgov/mds/commit/01bbb61ce16722e255c32e1eb1a64afb8e3fa2a1

Didn't update local to exhibit the same behavior. So this fixes that. Also cleaned up the dockerfile a little

load_testing/requirements.txt
```diff
@@ -1,3 +1,2 @@
-locustio
-dotenv
-os
\ No newline at end of file
+locustio==0.9.0
+python-dotenv==0.8.2
```
FreshGyan: Always a better practise to fix the package versioning and fixed the dotenv package name

migrations/etl/1_Mine_profile_detail_location.sql
```diff
@@ -1,5 +1,5 @@
--- 1. ETL mine data from MMS
--- Create the ETL_MINE table
+-- Query performance optimization
+SET max_parallel_workers_per_gather TO 8;
```
FreshGyan: https://www.postgresql.org/docs/9.6/parallel-query.html

python-backend/app/__init__.py
```diff
@@ -80,15 +80,3 @@ def get(self):
     def default_error_handler(error):
         _, value, traceback = sys.exc_info()
         return json.loads({"error": str(traceback)})
-
-
-def register_apm(app):
```
FreshGyan: Moved this to extensions so that the init is a single line. Also so that people can now import APM module from extensions and do stuff like `apm.log(AAHHHHH)` for custom exceptions

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -183,12 +183,14 @@ def validate_mine_name(self, key, mine_name):
 
     @validates('mine_note')
     def validate_mine_note(self, key, mine_note):
+        mine_note = mine_note if mine_note else ''
```
FreshGyan: So the validators return the data back to database. I changed it to be empty string on None. Because . the database . has a NOT NULL constraint on mine_note 🤷‍♂️

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -183,12 +183,14 @@ def validate_mine_name(self, key, mine_name):
...
         if len(mine_note) > 300:
             raise AssertionError('Mine note must not exceed 300 characters.')
         return mine_note
 
     @validates('mine_no')
     def validate_mine_no(self, key, mine_no):
-        if len(mine_no) > 10:
+        mine_no = mine_no if mine_no else ''
```
FreshGyan: Same as above

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -121,7 +119,7 @@ def put(self, party_guid):
             party_exists.first_name = first_name
             party_exists.party_name = party_name
         elif party_type_code == PARTY_STATUS_CODE['org']:
-            party_name = data['party_name'] if data['party_name'] else party_exists.party_name
+            party_name = data.get('party_name', party_exists.party_name)
```
FreshGyan: Another pattern i really like. If you're looking at a fallback key, .get() saves you time from not having to write inline if else

python-backend/app/api/utils/include/user_info.py
```diff
@@ -22,13 +22,13 @@ def get_user_raw_info(self):
...
     def get_user_given_name(self):
         raw_info = self.get_user_raw_info()
-        return raw_info['given_name']
+        return raw_info.get('given_name')
 
     def get_user_username(self):
         raw_info = self.get_user_raw_info()
         realms = list(set(VALID_REALM) & set(raw_info['realm_access']['roles']))
-        return realms[0] + '\\' + raw_info['preferred_username'] if realms else raw_info['preferred_username']
+        return realms[0] + '\\' + raw_info.get('preferred_username')
```
FreshGyan: Also this if statement seemed to be ineffective as far I could tell. Whoever . the original author was, please confirm I didn;t misunderstood some complex logic

python-backend/app/auth.py
```diff
@@ -66,7 +64,7 @@ def get_user_email():
 
 def get_current_user_security():
     rv = getattr(g, 'current_user_security', None)
-    if rv == None:
+    if rv is None:
```
FreshGyan: Usually a good practise . to not use string comparators with None

python-backend/app/api/utils/include/user_info.py
FreshGyan: my understanding was this gets processed as A + (B if condition else B) So i changed it to A + B.

If twas doing (A+B if condition else B). Please let me know so i can update that

python-backend/app/api/mines/mine/models/mine.py
Jsyro: Sounds like the NOT NULL constraint should be removed, but this will work for now.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -232,20 +232,20 @@ def post(self, mine_no_or_guid=None):
...
+        note = data.get('note')
         location = None
         mine_region = None
-        status = data['mine_status']
-        major_mine_ind = data['major_mine_ind']
-        mine_region = data['mine_region']
+        status = data.get('mine_status')
+        major_mine_ind = data.get('major_mine_ind')
+        mine_region = data.get('mine_region')
         mine = Mine(mine_guid=uuid.uuid4(), **self.get_create_update_dict())
```
Jsyro: Why do we create a mine here?

python-backend/app/auth.py
nayr974: I don't believe == is a string comparator?  == checks equivalence in terms of values, is checks equivalence in terms of identity (object reference). I suppose you can argue that None isn't a value, but we're not interested in object identity here.

python-backend/app/api/mines/mine/resources/mine.py
NWCalvank: It's a `POST` request to the mine endpoint. That's the expected behaviour, no?

python-backend/app/auth.py
nayr974: In a quick search "is None" seems to be preferred, so still a valid change.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -232,20 +232,20 @@ def post(self, mine_no_or_guid=None):
...
+        lon = data.get('longitude')
+        note = data.get('note')
         location = None
         mine_region = None
-        status = data['mine_status']
-        major_mine_ind = data['major_mine_ind']
-        mine_region = data['mine_region']
+        status = data.get('mine_status')
+        major_mine_ind = data.get('major_mine_ind')
+        mine_region = data.get('mine_region')
```
Jsyro: The string indexing was a form of safety (crash if this doesn't exist). If this code continues with values being null (that were never null before). Are there potential issues? Are there now validators that none of these are None?

nayr974: My concern as well. I remember there being discussion around this, although I didn't actively research or weight in on it.

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -85,15 +83,15 @@ def post(self, party_guid=None):
...
 
         try:
             party = Party(
                 party_guid=uuid.uuid4(),
-                phone_no=data['phone_no'],
-                email=data['email'],
-                phone_ext=data['phone_ext'] if data['phone_ext'] else None,
+                phone_no=data.get('phone_no'),
+                email=data.get('email'),
+                phone_ext=data.get('phone_ext') if data.get('phone_ext') else None,
```
Jsyro: Don't need in line ternary with .get()

python-backend/app/api/mines/mine/resources/mine.py
NWCalvank: So, perhaps revert the `.get()`s and wrap it in a `try` `except KeyError` so that we can handle it gracefully?

nayr974: The point of the change was to save boilerplate code though?

python-backend/app/api/mines/mine/resources/mine.py
FreshGyan: 🤷‍♂️ I didn't look into why the logic is happening 😅 but you can make a note or ticket and we can address that

python-backend/app/api/mines/mine/resources/mine.py
FreshGyan: Hmm 🤔 Your point makes sense. On one hand we risk crashing the app on other flushing the DB with null values.
I say we stick to crashing for now and we make a ticket to address the bigger issue here of handling exceptions properly.
The point of this PR was just a bit of housekeeping, I wasn't expecting to change any patterns or behavior

FreshGyan:
@Jsyro Changes reverted to mine resource

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -280,14 +280,14 @@ def post(self, mine_no_or_guid=None):
...
-        major_mine_ind = data['major_mine_ind']
-        region = data['mine_region']
+        tenure = data.get('tenure_number_id')
+        lat = data.get('latitude')
+        lon = data.get('longitude')
+        mine_name = data.get('name')
+        mine_note = data.get('note')
+        status = data.get('mine_status')
+        major_mine_ind = data.get('major_mine_ind')
+        region = data.get('mine_region')
```
Jsyro: I presume this will not explode because all of these values were bring provided, but same thing. these will have to None 'checked where appropriate' but this PUT method needs to some real TLC.

python-backend/app/api/mines/tailings/resources/tailings.py
```diff
@@ -45,48 +45,48 @@ def get(self, mine_tailings_storage_facility_guid=None):
...
-                        headers={'Authorization': request.headers['Authorization']})
+                        headers={'Authorization': request.headers.get('Authorization')})
 
                     if get_tsf_docs_resp.status_code != 200:
                         self.raise_error(
                             500,
                             'get_tsf_req_docs returned error' + str(get_tsf_docs_resp.status_code))
 
-                    tsf_required_documents = get_tsf_docs_resp.json()['required_documents']
+                    tsf_required_documents = get_tsf_docs_resp.json().get('required_documents')
```
Jsyro: for example, now that tsf_required_documents can be none, the loop below will silently be skipped, rather than crashing should something be changed on other endpoint.

python-backend/app/api/mines/tailings/resources/tailings.py
```diff
@@ -45,48 +45,48 @@ def get(self, mine_tailings_storage_facility_guid=None):
...
                     new_expected_documents = []
                     for tsf_req_doc in tsf_required_documents:
                         new_expected_documents.append({
                             'req_document_guid':
-                            tsf_req_doc['req_document_guid'],
+                            tsf_req_doc.get('req_document_guid'),
```
Jsyro: same thing with these, this was a really quick and dirty way to make sure the other endpoint didn't change, cause if it did, this would crash, not just use None.... really strong unit tests will be needed to make sure a change to the json will not cause unexpected behavior.

FreshGyan:
@Jsyro Good point. Not being the Original Author on most of these endpoints, I underestimated the ripple effect and assumptions that were made around data sanity and checks.
I've reverted most of the changes and propose making a ticket for someone to go deeply through them and update the unit tests. Thanks for looking in depth and making sure I didn't introduce any unknown changes 💯

PR #514, opened 2019-02-11
Feature/mds 924

thegentlemanphysicist:
This loads the pins without an http call.  This should fix the bug listed.  Testing is needed on test, the bug did not show up locally.


frontend/src/components/maps/MinePin.js
```diff
@@ -134,6 +134,7 @@ export class MinePin extends Component {
 
         const fcl = FlareClusterLayer.FlareClusterLayer(options);
         fcl.title = "Mine Pins";
+        fcl.refreshOnStationary = true;
```
FreshGyan: This seems to be true by default. Did you notice some funky behavior here?

thegentlemanphysicist: you are right I was just playing around with that and forgot to take it out

frontend/src/assets/images/small-pin.svg
```diff
@@ -1,6 +1,8 @@
-<svg width="40" height="40" viewBox="0 0 40 40" fill="none" xmlns="http://www.w3.org/2000/svg">
-<ellipse cx="20.0004" cy="6.74412" rx="1.53846" ry="1.62791" fill="#F1F1F2"/>
-<ellipse cx="20.8002" cy="5.9575" rx="1.6" ry="1.70213" fill="#F1F1F2"/>
-<path fill-rule="evenodd" clip-rule="evenodd" d="M25.7054 10.9374C26.5225 9.86298 27 8.48487 27 6.82927C27 3.05757 23.6421 0 19.5 0C15.3579 0 12 3.05757 12 6.82927C12 8.48705 12.4788 9.86662 13.2978 10.9417L19.4843 20L25.7054 10.9374Z" fill="#795F48"/>
-<circle cx="19.5" cy="7.5" r="2.5" fill="#F1F1F2"/>
-</svg>
+<svg width="70" height="70" viewBox="0 0 70 70" fill="none" xmlns="http://www.w3.org/2000/svg">
```
FreshGyan: What's the change to the SVG? Was this approved by Tatianna?

frontend/src/components/maps/MinePin.js
```diff
@@ -62,12 +61,16 @@ export class MinePin extends Component {
...
-          url: `${window.location.origin}${process.env.BASE_PATH}/public/small-pin.svg`,
+        // The svg for the map pin is encoded directly into the default symbol
+        const defaultSym = new SimpleMarkerSymbol({
+          size: 25,
           width: this.state.isFullMap ? "40" : "80",
           height: this.state.isFullMap ? "40" : "80",
-          type: "picture-marker",
-        };
+          path:
+            "M16,3.5c-4.142,0-7.5,3.358-7.5,7.5c0,4.143,7.5,18.121,7.5,18.121S23.5,15.143,23.5,11C23.5,6.858,20.143,3.5,16,3.5z M16,14.584c-1.979,0-3.584-1.604-3.584-3.584S14.021,7.416,16,7.416S19.584,9.021,19.584,11S17.979,14.584,16,14.584z",
```
FreshGyan: Good job on implementing this. 💯 🙌

frontend/src/assets/images/small-pin.svg
mmumby: He just removed the pin that was saved in `/assets`, but never used, and replaced it with the pin that was saved in `/public` which was the current pin in use

PR #519, opened 2019-02-12
Upgrade/404

mmumby:
- Updated 404 views for both MineSpace and CORE
- Added a possible fix to the mineSpace toast message bug


frontend-public/src/components/common/NullScreen.js
```diff
@@ -29,8 +29,14 @@ const NullScreen = (props) => (
     )}
     {props.type === "404" && (
       <div className="no-nav-bar">
-        <h3>404 - Page not found</h3>
-        <Link to={routes.HOME.route}>Go Home</Link>
+        <h1>Uh Oh!</h1>
+        <p>
+          We can't seem to find the page you're looking for. It may have moved, or no longer exists.
```
NWCalvank: Grammar feedback here. The first sentence ends with a preposition, and the second sentence is technically an incorrect use of a conjunction.
We could avoid these with a single sentence, like:
`The page you're looking for has moved or no longer exists`

Or a small adjustment to the existing two sentences:
`The page you're looking for can't be found. It may have moved, or it no longer exists.`

frontend/src/components/common/NullScreen.js
```diff
@@ -107,6 +103,19 @@ const NullScreen = (props) => (
         <p>Create a contact using the menu above</p>
       </div>
     )}
+    {props.type === "404" && (
```
NWCalvank: Do we have page 'types' elsewhere in the app? This feels like it should be a constant, based on our existing constants patterns.

frontend-public/src/components/authentication/Authentication.js
```diff
@@ -24,65 +24,70 @@ const defaultProps = {
...
-      <Divider style={{ margin: "0" }} />
-      <Menu.Item>
-        <Button type="tertiary" onClick={signOutFromSiteMinder}>
-          Log out
-        </Button>
-      </Menu.Item>
-    </Menu>
-  );
+export class Authentication extends Component {
+  handleLogout = () => {
```
NWCalvank: What are these changes for? Do they relate to the 404 somehow or to something else?

frontend/src/components/common/NullScreen.js
mmumby: the `type` is for the null screen, not a page `type`

I felt that 404 explained the type of null screen the best, happy to change the string to something else. 

If you expand this file, you'll see the other types being displayed.

frontend-public/src/components/authentication/Authentication.js
mmumby: This relates to the second part of my PR, " - Added a possible fix to the mineSpace toast message bug"

frontend-public/src/components/common/NullScreen.js
mmumby: Thanks for the feedback! will update with the second option.

frontend-public/src/components/authentication/Authentication.js
NWCalvank: Oh, glorious! Marking as resolved

frontend/src/components/common/NullScreen.js
NWCalvank: I agree that 404 is an accurate description. Anything that feels like a magic number or magic string, though, is probably good to question. A typo on this wouldn't get caught by a linter, for example.

But creating a constant called `NOT_FOUND`, for example, can be picked up by a linter and reused (if appropriate) elsewhere.

PR #528, opened 2019-02-14
add realtime search as admin types

Jsyro:
fix for admin page mine multi-select to search as user types.

### DIFF
--- a/frontend/src/components/Forms/AddMinespaceUser.js
+++ b/frontend/src/components/Forms/AddMinespaceUser.js
@@ -12,6 +12,13 @@ import CustomPropTypes from "@/customPropTypes";
 const propTypes = {
   handleSubmit: PropTypes.func.isRequired,
   mines: CustomPropTypes.options.isRequired,
+  handleChange: PropTypes.func,
+  handleSearch: PropTypes.func,
+};
+
+const defaultProps = {
+  handleChange: {},
+  handleSearch: {},
 };
 
 export const AddMinespaceUser = (props) => (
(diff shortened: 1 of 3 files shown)


frontend/src/components/admin/NewMinespaceUser.js
```diff
@@ -35,6 +35,17 @@ export class NewMinespaceUser extends Component {
     });
   };
 
+  handleSearch = (name) => {
+    if (name.length > 2) {
+      // Only used by the map view, which searches by name
```
NWCalvank: This comment is incorrect here. It's only relevant to the implementation of `handleSearch` in the frontend dashboard.

python-backend/app/api/mines/mine/models/mine.py
```diff
@@ -165,7 +165,7 @@ def find_by_mine_name(cls, term = None):
 
     @classmethod
     def find_by_name_no_permit(cls, term = None):
-        MINE_LIST_RESULT_LIMIT = 500
+        MINE_LIST_RESULT_LIMIT = 50
```
NWCalvank: This seems a lot more reasonable to me 👍

frontend/src/components/Forms/AddMinespaceUser.js
```diff
@@ -38,6 +45,8 @@ export const AddMinespaceUser = (props) => (
               placeholder="Please Select a Mine"
               component={renderConfig.MULTI_SELECT}
               data={props.mines}
+              onChange={props.handleChange}
```
NWCalvank: This one seems unnecessary, since events on the input handle both the default and the filtered search.

frontend/src/components/admin/NewMinespaceUser.js
Jsyro: Yea, that was a copy pasta error from the MineSearch (there i took this snippet from).

frontend/src/components/Forms/AddMinespaceUser.js
Jsyro: Multi-select keeps everything focused, so onChange is needed.

PR #530, opened 2019-02-14
Bug/mds 1139

thegentlemanphysicist:
This fixes the pulsing location pin in the map search.
Note this should be tested on mobile to make sure it works properly.


frontend/src/styles/components/Maps.scss
```diff
@@ -2,13 +2,7 @@
...
 */
 
-/* FIXME: Add pulsating animation to the LocationPin under the BaseMap Layer
-div.esri-view-root
-  > div.esri-view-surface.esri-view-surface--inset-outline
-  > div
-  > svg
-  > g
-  > circle {
+//This modifies the custum drawn path in the dom.  It is created in the 'LocationPin.js' file
```
NWCalvank: *custom and *DOM

Otherwise, this looks good. I can't think of a way around the path duplication at the moment, so this seems like a reasonable solution.

PR #531, opened 2019-02-14
input prop was being weird

Jsyro:
input prop was not updating the redux state when changes were made.

### DIFF
--- a/frontend/src/components/Forms/AddMinespaceUser.js
+++ b/frontend/src/components/Forms/AddMinespaceUser.js
@@ -46,7 +46,7 @@ export const AddMinespaceUser = (props) => (
               component={renderConfig.MULTI_SELECT}
               data={props.mines}
               onChange={props.handleChange}
-              input={{ onBlur: props.handleChange, onSearch: props.handleSearch }}
+              onSearch={props.handleSearch}
             />
           </Form.Item>
         </Col>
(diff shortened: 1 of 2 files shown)


frontend/src/components/Forms/AddMinespaceUser.js
```diff
@@ -46,7 +46,7 @@ export const AddMinespaceUser = (props) => (
               component={renderConfig.MULTI_SELECT}
               data={props.mines}
               onChange={props.handleChange}
-              input={{ onBlur: props.handleChange, onSearch: props.handleSearch }}
+              onSearch={props.handleSearch}
```
mmumby: When I first saw the above, I was curious as to how it was working, but didn't ask about it.... should have. 

This new implementation looks much better!

frontend/src/components/common/RenderMultiSelect.js
```diff
@@ -44,6 +46,7 @@ export const RenderMultiSelect = (props) => (
         getPopupContainer={() => document.getElementById(props.id)}
         placeholder={props.placeholder}
         id={props.id}
+        onSearch={props.onSearch}
         {...props.input}
```
mmumby: did you test to make sure it didn't break the previous multi selects that aren't passing in an onSearch function?

PR #532, opened 2019-02-15
Fix/mds 1388 existing minespace user mines

Jsyro:
had to serve separate reference list for front end to resolve mine_guids to mine names. and use that list to resolve the table.


Jsyro:
@mmumby The frontend was using the same list in the multi-select to also resolve the mine_guids into mine names, however that list for the multi-select is being changed on the fly to help users search (and is only 50 long). The other part of the page showing existing minespace users also have associated guids that need to be resolved into names, so it has to use a different list (and the back needs to get that list and serve it).  Does that make more sense?

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -26,7 +27,18 @@ def get(self, user_id=None):
...
             result = user.json()
         else:  #get all users (not deleted)
-            result = {'users': [x.json() for x in MinespaceUser.get_all()]}
+            ms_users = MinespaceUser.get_all()
+            all_mines = []
+            for user in ms_users:
+                for user_mine in user.mines:
+                    all_mines.append(str(user_mine.mine_guid))
+
+            mine_basic_info_resp = requests.post(
```
NWCalvank: Why not just use the response with all minespace users (the 'before' in this diff) to fetch the basic mine data with an API call to '/basicinfo' from the frontend?

I get the impression here that this is a concern for the frontend (displaying mine names for mines associated with users -- which is what I think this change is meant to do). The backend shouldn't care that the frontend wants to do that.

Correct me if I'm misunderstanding the goal of this PR.

Jsyro: That would be the other approach, both are valid, I went this way because the endpoint is not useful without this information also being returned, so I figured it would just make the hop from the backend.

NWCalvank: I think this falls in line with what @nayr974 and I were discussing the other day on Slack. This adds the logic for a UI concern to the backend. If we don't want business logic in our endpoints then I think we should be consistent with that decision.

A `mine_guid` is totally sufficient for a consumer of the API to use to then fetch whatever mine data they want. Fetching that data for the consumer on the backend is preemptive and should be avoided, in my opinion.

Jsyro: Working right now to change this to the frontend fetching this.

NWCalvank: Ryan said this was addressed IRL. Resolving the conversation

Jsyro: @NWCalvank Changed it so frontend makes the second call, same pattern as PartyRelationships. Let me know if it looks good.

python-backend/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -8,6 +8,7 @@
 from ..models.minespace_user import MinespaceUser
 from ..models.minespace_user_mine import MinespaceUserMine
 from app.extensions import db
+from app.api.utils.url import get_mines_svc_url
```
NWCalvank: This is no longer used

frontend/src/components/admin/MinespaceUserManagement.js
```diff
@@ -38,15 +45,22 @@ export class MinespaceUserManagement extends Component {
     });
   };
 
+  refreshUserData = () => {
+    this.props.fetchMinespaceUsers().then(() => {
+      const mine_guids = [...new Set(flatMap(this.props.minespaceUsers, (user) => user.mines))];
```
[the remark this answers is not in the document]
Jsyro: Thanks, changed.

python-backend/app/api/users/minespace/resources/minespace_user.py
Jsyro: good catch.

PR #534, opened 2019-02-15
Feature/mds 1141

mmumby:
Completed: 
- Updated PartyProfile.js to use ant design tables
- Updated RelationshipProfile.js to use ant design tables

- Updated the tables to be left aligned and have the same colour heading as figma,


frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -24,40 +24,66 @@ const columns = [
     title: "Permit No.",
     dataIndex: "permitNo",
     key: "permitNo",
+    render: (text) => <div title="Permit No.">{text}</div>,
```
mmumby: This allows CSS to access the title attribute and add this as a label for responsive table views.

frontend/src/components/parties/PartyProfile.js
```diff
@@ -36,112 +37,108 @@ const propTypes = {
...
 export class PartyProfile extends Component {
+  state = { isLoaded: false };
+
   componentDidMount() {
     const { id } = this.props.match.params;
     this.props.fetchPartyById(id);
     this.props.fetchPartyRelationships({ party_guid: id }).then(() => {
       const mine_guids = uniq(this.props.partyRelationships.map(({ mine_guid }) => mine_guid));
-      this.props.fetchMineBasicInfoList([...mine_guids]);
+      this.props.fetchMineBasicInfoList([...mine_guids]).then(() => {
```
NWCalvank: Looks like `mine_guids` is an array? Spreading an array inside of an array cancels each other out.
`(mine_guids)` should be equivalent

Also, it looks like `partyRelationships` and `basicMineInfoListHash` are being passed as optional props AND being fetched on mount. If that's true, we should only fetch if the props were not provided.

frontend/src/components/parties/PartyProfile.js
```diff
@@ -36,112 +37,108 @@ const propTypes = {
...
-      this.props.partyRelationshipTypes.length > 0 &&
-      (this.props.partyRelationships.length > 0 && parties);
+    const transformRowData = (partyRelationships) =>
+      partyRelationships.map((relationship) => ({
+        key: relationship.mine_party_appt_guid,
+        mineGuid: relationship.mine_guid,
+        mineName: this.props.mineBasicInfoListHash[relationship.mine_guid],
+        role: this.props.partyRelationshipTypeHash[relationship.mine_party_appt_type_code],
+        endDate:
+          relationship.end_date === "9999-12-31" ? "Present" : formatDate(relationship.end_date),
```
NWCalvank: James recently merged a change that sets Mine Manager `mine_party_appt`s to `null` in the ETL process instead of to this bunk value. We should probably handle both cases here, to be safe.

frontend/src/components/parties/RelationshipProfile.js
```diff
@@ -102,11 +104,55 @@ export class RelationshipProfile extends Component {
...
+
+    const transformRowData = (historyRelationships) =>
+      historyRelationships.map((relationship) => ({
+        key: relationship.mine_party_appt_guid,
+        contact: relationship.party.name,
+        partyGuid: relationship.party.party_guid,
+        role: this.state.partyRelationshipTitle,
+        permit: this.state.permitsMapping[relationship.related_guid],
+        endDate:
+          relationship.end_date === "9999-12-31" ? "Present" : formatDate(relationship.end_date),
```
NWCalvank: Same comment here about `null` values

frontend/src/components/parties/RelationshipProfile.js
```diff
@@ -137,50 +183,15 @@ export class RelationshipProfile extends Component {
...
-                      <Col span={width}>
-                        <h2>Permit No.</h2>
-                      </Col>
-                    )}
-                    <Col span={width}>
-                      <h2>Dates</h2>
-                    </Col>
-                  </Row>
-                  <Divider style={{ height: "2px", backgroundColor: "#013366", margin: "0" }} />
+                <div className="tab__content ">
```
NWCalvank: Extra space after `tab__content`?

frontend/src/components/parties/RelationshipProfile.js
mmumby: The end_date is `null` you mean? instead of `9999-12-31` ?

frontend/src/components/parties/RelationshipProfile.js
mmumby: removed.

frontend/src/components/parties/PartyProfile.js
mmumby: `mine_guids` was fixed. 

Chatted to Nathan about the props in person.

frontend/src/components/parties/RelationshipProfile.js
NWCalvank: That's right. James updated ETL to no longer set `mine_party_appt` `end_date`s to "9999-12-31". Mine manager `mine_party_appt`s `end_date`s are now set to `null` instead. We'll need to handle both for the time being

mmumby: updated!

PR #535, opened 2019-02-15
Improvement/MDS-1445 Increase Document Manager file name and path limit

james-hollinger:
Adds a migration to increase the file name and file path limits in the document manager. Linux has a 255 character limit for file names and a 4096 character limit for file paths. Windows has a 260 character limit for file paths in older systems.


migrations/sql/V2019.02.15.01.27__increase_document_manager_file_name_limits.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE document_manager ALTER COLUMN full_storage_path TYPE VARCHAR(4096);
+ALTER TABLE document_manager ALTER COLUMN file_display_name TYPE VARCHAR(260);
```
Jsyro: Looks like Linux likes 255 character file names. `Linux has a maximum filename length of 255 characters for most filesystems (including EXT4), and a maximum path of 4096 characters.`. So we might want to match that.

Jsyro: Also might need UI changes to support on-screen elements and links that are this long, but that can be addressed in a separate PR.

james-hollinger: Looks like the filename restriction on windows is actually 255 characters as well. Changing this.

PR #536, opened 2019-02-15
Bug/mds 1046

thegentlemanphysicist:
This prevents people from creating a mine with the same name as an existing one
This prevents people from editing  a mine name to be the same as an existing one
This does not cause a problem if there already exists a mine with a redundant name
Backend tests included.


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -245,6 +254,8 @@ def post(self, mine_no_or_guid=None):
         mine_region = data['mine_region']
         mine = Mine(mine_guid=uuid.uuid4(), **self.get_create_update_dict())
         try:
+            # query the mine tables and check if that mine name exists
+            self.throw_error_if_mine_exists( data['name'])
```
NWCalvank: Let's add `mine_name = data['name']` above this line and then use that variable here & on line 262

python-backend/tests/mines/resources/test_mine_resource.py
```diff
@@ -301,6 +316,17 @@ def test_put_mine_name(test_client, auth_headers):
     assert put_resp.status_code == 200
 
 
+def test_put_redundant_mine_name(test_client, auth_headers):
+    # this relies on  test test_post_mine_name_and_coord
+    test_tenure_data = {
+        "name": "test_create_mine",
+        "note": ""
```
NWCalvank: Is the `note` field required here? I don't think it should be. If it isn't, we should probably remove it to keep the test as specific as possible.

PR #537, opened 2019-02-15
Upgrade/mds 1018

mmumby:
Updated dates throughout the app to use the correct format.


frontend/src/components/mine/Permit/MinePermitCard.js
```diff
@@ -23,7 +23,7 @@ export const PermitCard = (props) => {
       <h4>{formatTitleString(props.permit.permit_no)}</h4>
       <br />
       <h6>Last Amended</h6>
-      <span>{props.permit.issue_date}</span>
+      <span>{formatDate(props.permit.issue_date)}</span>
```
NWCalvank: I see that this had no checks on it before, but now that we're passing it into a function.. I'd like to confirm that we know this value will exist. A `null` or `undefined` value here may cause issues

mmumby: will fix this!

PR #538, opened 2019-02-19
MDS 1364 Permit Amendment Refactor

nayr974:
Move to parent permits table with child amendment records vs multiple permit / permittee records representing amendments.


frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -60,44 +61,19 @@ const columns = [
 ];
 
 const childColumns = [
+  { title: "", dataIndex: "key", key: "Key" },
+  { title: "Received Date", dataIndex: "ReceivedDate", key: "ReceivedDate" },
+  { title: "Issue Date", dataIndex: "IssueDate", key: "IssueDate" },
```
mmumby: the render line is important for responsive views,

```
render: (text) => <div title="Permit No.">{text}</div>
```

CSS uses the title attribute to add a label for the card. 
Please add these back in.

mmumby: (note that the responsive nested table still needs some work, but it is being addressed in the next sprint)

nayr974: Good to know, thanks! I will add a render line. 👍

mmumby: Thank you!

nayr974: Added these back in. FYI I didn't intentionally remove them, it was merge side effect :)

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -61,43 +62,34 @@ const columns = [
 
 const childColumns = [
   {
-    title: "Permit No.",
-    dataIndex: "permitNo",
-    key: "childPermitNo",
-    render: (text) => <div title="Permit No.">{text}</div>,
+    title: "",
+    dataIndex: "key",
+    key: "Key",
```
NWCalvank: Any reason why this isn't something like `amendmentNo`?

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -107,49 +99,54 @@ const getPermittees = (partyRelationships, permits) =>
...
   const permitteeName =
     partyRelationships.length === 0 ? Strings.LOADING : getPermitteeName(permittees);
 
   return {
-    key: latest.permit_guid,
-    lastAmended: formatDate(latest.issue_date),
-    permitNo: latest.permit_no || Strings.EMPTY_FIELD,
-    firstIssued: formatDate(first.issue_date) || Strings.EMPTY_FIELD,
+    key: permit.permit_guid,
+    lastAmended: (latestAmendment && formatDate(latestAmendment.issue_date)) || Strings.EMPTY_FIELD,
```
NWCalvank: It looks like `.issue_date` can be `null`?
https://mds-schemaspy-538-empr-mds-dev.pathfinder.gov.bc.ca/tables/permit_amendment.html

We should probably check for that before passing it into the format function

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -107,49 +99,54 @@ const getPermittees = (partyRelationships, permits) =>
...
-const transformChildRowData = (permittee, record) => ({
-  key: record.key,
-  permitNo: record.permitNo,
-  Date: permittee.start_date,
-  permittee: permittee.party.name,
+const transformChildRowData = (amendment, record, amendmentNumber) => ({
+  key: amendmentNumber,
+  ReceivedDate: amendment.received_date,
+  IssueDate: amendment.issue_date,
+  AuthorizationEndDate: amendment.authorization_end_date,
```
NWCalvank: Why are these PascalCase?

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -107,49 +99,54 @@ const getPermittees = (partyRelationships, permits) =>
...
-    transformRowData(permits, props.partyRelationships)
-  );
+  const rowData = props.mine.mine_permit
+    .filter((permit) => permit.permit_no.toUpperCase().charAt(1) !== "X")
+    .map((permit) => transformRowData(permit, props.partyRelationships))
+    .concat(
+      props.mine.mine_permit
+        .filter((permit) => permit.permit_no.toUpperCase().charAt(1) === "X")
+        .map((permit) => transformRowData(permit, props.partyRelationships))
+    );
```
NWCalvank: I'm sure there's a good reason for this, but at a quick look, it seems like two mutually-exclusive parts of the `mine_permit` list are being transformed in the same way and then being joined.

What's the reason for splitting them apart before passing them to `transformRowData` and then joining them again? Any chance this reason could be made a bit more self-documenting in the code?

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -147,9 +125,16 @@ export const MineSummary = (props) => {
...
+              {props.mine.mine_permit
+                .filter((permit) => permit.permit_no.toUpperCase().charAt(1) !== "X")
+                .map((permit) =>
+                  renderSummaryPermit(permit, props.partyRelationships.filter(isActive))
+                )}
+              {props.mine.mine_permit
+                .filter((permit) => permit.permit_no.toUpperCase().charAt(1) === "X")
+                .map((permit) =>
+                  renderSummaryPermit(permit, props.partyRelationships.filter(isActive))
+                )}
```
NWCalvank: Same question here. Also, it seems like it might be useful to name the filter function(s). I know from previous discussions that this is separating them into exploratory permits, but that may not be clear to a future reader of this code (including me, if/when I forget)

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,88 @@
...
+    permit_amendment_status_code = db.Column(
+        db.String(2), db.ForeignKey('permit_amendment_status_code.permit_amendment_status_code'))
+    permit_amendment_type_code = db.Column(
+        db.String(2), db.ForeignKey('permit_amendment_type_code.permit_amendment_type_code'))
+
+    def json(self):
+        return {
+            'permit_amendment_id': str(self.permit_amendment_id),
+            'permit_amendment_guid': str(self.permit_amendment_guid),
+            'permit_id': self.permit_id,
```
NWCalvank: You touched on this in your tech demo, but I think I missed the reasoning. Why are we referencing the permit by id here instead of by guid? Is that the preferred pattern for relationships going forward? Or is it specific to permits/amendments?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,88 @@
...
+        if issue_date.isoformat() == '9999-12-31':
+            return issue_date
+        if issue_date > datetime.today():
+            raise AssertionError('Permit amendment issue date cannot be set to the future.')
+        return issue_date
+
+    @validates('authorization_end_date')
+    def validate_authorization_end_date(self, key, authorization_end_date):
+        if authorization_end_date.isoformat() == '9999-12-31':
+            return authorization_end_date
```
NWCalvank: Good thinking. I'm wondering if we should convert these all to `null` at some point so that the backend and frontend can stop worrying about these special values

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
JustinMacaulay: I thought the plan going forward was to use serial Id's where we could going forward along with the guid's. That way we can use the serial ID for the primary key and not propagate the guid's across the relationships to save space.

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: No, I'm not sure why I went with key. Let's change it to amendmentNumber which is what I pass into it.

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: Done. formatDate does handle nulls, but it renders as an empty string instead of 'N/A', so either way I've now updated it to check for null and render Strings.EMPTY_FIELD.

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: ¯\_(ツ)_/¯ . Changed 👍

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: I figured I would get called out on this, but it seemed like the quickest way to achieve the separation between exploration and other types of permits. I'm not confidant that we have all the requirements for "permit sorting", so I thought going this approach for now wouldn't require changing any of the other logic on this page.

Might also actually make some sense for the back-end to do this, if we want this to be the order universally (like on the mine search results). If I can confirm that with Aaron, I'll make that change and refactor this out in our current sprint.

frontend/src/components/mine/Summary/MineSummary.js
nayr974: Do agree. I'll rename and add comments for clarity.

frontend/src/components/mine/Permit/MinePermitInfo.js
NWCalvank: Perfect! Thanks 👍

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
NWCalvank: Yep. We came across a ticket for that in the backlog in Sprint Planning 2 today. I wasn't aware of that, but now I am. Thanks!

nayr974: That's my understanding. As part of the DB changes Julian proposed auto-incrementing integers should be primary keys going forward as they are faster for queries and indexes, so I took this as an opportunity to add this to the permit table.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
nayr974: I hate the whole idea of storing '9999-12-31' with a burning passion so I'm just going to make this change now for amendments.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
```diff
@@ -0,0 +1,161 @@
...
+
+CREATE TABLE permit_amendment
+(
+	permit_amendment_id serial PRIMARY KEY,
+	permit_amendment_guid uuid DEFAULT gen_random_uuid() NOT NULL,
+	permit_id integer NOT NULL,
+	received_date date,
+	issue_date date,
+	authorization_end_date date,
+	permit_amendment_type_code character varying(2) DEFAULT 'W' not null ,
```
JustinMacaulay: Does this Default 'W' mean something particular? Shouldn't this default to a valid value?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,97 @@
...
+        db.String(2), db.ForeignKey('permit_amendment_type_code.permit_amendment_type_code'))
+
+    def json(self):
+        return {
+            'permit_amendment_id':
+            str(self.permit_amendment_id),
+            'permit_amendment_guid':
+            str(self.permit_amendment_guid),
+            'permit_id':
+            self.permit_id,
```
JustinMacaulay: Shouldn't this be wrapped in str() also?

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
```diff
@@ -0,0 +1,161 @@
...
+INSERT INTO permit_amendment_status_code
+	(
+	permit_amendment_status_code,
+	description,
+	display_order,
+	create_user,
+	update_user
+	)
+VALUES
+	('A', 'Active', 10, 'system-mds', 'system-mds'),
```
klefler: This is a minor comment re: status values. My recommendation would be to use 3-character codes rather than 1. The reason is we want the codes to be able to properly reflect the value. If we use "A" for active, we can't use "A" for anything else (i.e. "Accepted")... this would mean needed to use another letter which may not make sense. If we used 3-character codes, we could have "ACT" for Active and "ACC" for Accepted (just as an example)... and these codes could bring more value to the business.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
NWCalvank: Is there a reason why we'd want the sequential IDs to be converted from integers to strings?

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
klefler: What code value is "W"?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
JustinMacaulay: Consistency since the first id is converted to a string.

NWCalvank: My question applies to both IDs in that case ;) I'd expect IDs that are ints in the backend to be returned as JavaScript Numbers in the JSON payload, personally. So, I'd like to make sure we're making an intentional choice to do this.

If it's a matter of "that's how we've done it elsewhere" then I would suggest we ask ourself if that's how we want to do it before it's been done in too many places.

In `frontend/src/customPropTypes/permits.js` `permit_id` is a number type. So, I get the impression that this is Ryan's intentional behaviour for this ID. But I'll leave it to him to weigh in now.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
```diff
@@ -0,0 +1,161 @@
...
+table temp_old_permit from permit;
+
+truncate table permit
+cascade;
+
+insert into permit (mine_guid, permit_no, permit_status_code, create_user, update_user)
+select *, 'system-mds' as create_user, 'system-mds' as update_user
+from temp_permits;
+
+insert into permit_amendment (permit_id, received_date, issue_date, authorization_end_date, create_user, update_user)
```
klefler: Question... does the permit table have values for permit_id at this point?

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
nayr974: W was originally a Work amendment, and there were Work and Non-Work types. Missed changing the default when that was identified as not required. Good catch, thanks.

nayr974: W was originally a Work amendment, and there were Work and Non-Work types. Missed changing the default when that was identified as not required. Good catch, thanks.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
nayr974: Yes, since the column was defined as serial it will populate auto-incrementing values for the dataset immediately when the column is added.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
nayr974: Sure, I'm always happy to be more verbose 👍

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
nayr974: I don't see why we would want to represent it as a string. str(self.permit_amendment_id) was copy/paste without thought and I will change it. IMO our API should properly return the few types JSON defines such as numbers, booleans, and null types, not numbers as strings, the word "True", or empty strings, etc.

frontend/src/components/mine/Summary/MineSummary.js
nayr974: Ended up moving this to the API and removing this logic from the front-end entirely.

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: Ended up moving this to the API and removing this logic from the front-end entirely.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
nayr974: Updated with three letter codes.

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
nayr974: Updated to default to AMD

migrations/sql/V2019.02.11.10.23__permit_amendments.sql
klefler: Looks great! Thanks :)

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,94 @@
...
+            raise AssertionError('Permit amendment received date cannot be set to the future.')
+        return received_date
+
+    @validates('issue_date')
+    def validate_issue_date(self, key, issue_date):
+        if issue_date.isoformat() == '9999-12-31':
+            raise AssertionError('Permit amendment issue date should be set to null if not known.')
+        if issue_date > datetime.today():
+            raise AssertionError('Permit amendment issue date cannot be set to the future.')
+        return issue_date
```
Jsyro: What is the intent of these validators? is it to transform/coalesce '9999-12-31' to None/null? Because this is currently going to throw an attribute error in it's current form.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class PermitAmendment(AuditMixin, Base):
+    __tablename__ = 'permit_amendment'
+    permit_amendment_id = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
+    permit_amendment_guid = db.Column(UUID(as_uuid=True))
+    permit_id = db.Column(db.Integer, db.ForeignKey('permit.permit_id'), nullable=False)
+    received_date = db.Column(db.DateTime, nullable=False, server_default=FetchedValue())
+    issue_date = db.Column(db.DateTime, nullable=False, server_default=FetchedValue())
+    authorization_end_date = db.Column(db.DateTime, nullable=False, server_default=FetchedValue())
```
Jsyro: We have FetchedValue here, but below we enforce that they are set. do we allow the db to set the default, or does the python need to?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @validates('permit_id')
+    def validate_permit_id(self, key, permit_id):
+        if not permit_id:
+            raise AssertionError('Permit id is not provided.')
+        return permit_id
+
+    @validates('received_date')
+    def validate_received_date(self, key, received_date):
+        if not authorization_end_date:
```
JustinMacaulay: I believe this is a typo but I think this should be received_date.

nayr974: Thanks. Changed :)

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
nayr974: Not to transform 9999 to null, but issue date does not expect to follow the 9999 pattern for any use. It wouldn't make sense to set 9999 unless this app actually is being used in the year 9999.

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -107,48 +99,55 @@ const getPermittees = (partyRelationships, permits) =>
...
   description: Strings.EMPTY_FIELD,
 });
 
 export const MinePermitInfo = (props) => {
-  const groupedPermits = Object.values(groupPermits(props.mine.mine_permit));
   const amendmentHistory = (record) => {
-    const childRowData = record.permittees.map((permittee) =>
-      transformChildRowData(permittee, record)
+    const childRowData = record.amendments.map((amendment, index) =>
+      transformChildRowData(amendment, record, record.amendments.length - index)
```
NWCalvank: So, it looks like we can change `transformChildRowData` slightly and remove a lot of the busy code here. Plus, it looks like record is an unused argument in `transformChildRowData`?
```javascript
const transformChildRowData = (amendment, index, amendments) => ({
  amendmentNumber: amendments.length - index,
 // the rest
})

const amendmentHistory = (record) => {
  const childRowData = record.amendments.map(transformChildRowData));
  // the rest
};
```

migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
```diff
@@ -576,77 +578,149 @@ CREATE OR REPLACE FUNCTION transfer_premit_permitee_information() RETURNS void A
...
             --Select only new entry in ETL_PERMIT table
             new_permit AS (
                 SELECT *
-                FROM ETL_PERMIT
+                FROM ETL_PERMIT etl
                 WHERE permit_guid NOT IN (
                     SELECT permit_guid
                     FROM permit
                 )
+				AND issue_date = (select max(issue_date) from ETL_PERMIT where etl.permit_no = ETL_PERMIT.permit_no)
```
NWCalvank: Weird indentation?

migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
```diff
@@ -576,77 +578,149 @@ CREATE OR REPLACE FUNCTION transfer_premit_permitee_information() RETURNS void A
...
+                    create_timestamp       			,
+                    update_user            			,
+                    update_timestamp       			
+                )
+                SELECT
+                    permit.permit_id				         	,
+                    new_permit_amendment.received_date       	,
+                    new_permit_amendment.issue_date          	,
+                    new_permit_amendment.authorization_end_date	,
+                    'AMD'									  	,
```
NWCalvank: I won't block a PR for crazy comma alignment, but I will point it out

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: Cool, tweaked!

migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
nayr974: It doesn't show up like this locally in either VSCode or Notepad++ 👎

migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
nayr974: Also not sure why it's showing up on github like this but locally in both VSCode and Notepad++ it appears as expected with all commas lined up :)

PR #542, opened 2019-02-21
MDS-1418 Feature/ Add sorting to Mine List View columns

NWCalvank:
This is 1/2 of the tables to be updated with this behaviour. The compliance tab table is not yet converted to using AntDesign. Once that has been done, and this behaviour becomes top priority again, that table will be updated.


NWCalvank:
Closing this PR because this feature is not a priority and must be scoped-out more fully before being implemented.

PR #543, opened 2019-02-21
MDS-1438 Feature/ Add address columns to party & province_code table

NWCalvank:
This copies existing patterns for code tables in the database.


migrations/sql/V2019.02.21.9.59__add_address_columns_to_party.sql
```diff
@@ -0,0 +1,54 @@
...
+ALTER TABLE party ADD COLUMN city           varchar(50);
+ALTER TABLE party ADD COLUMN province_code  varchar(2);
+ALTER TABLE party ADD COLUMN postal_code    varchar(6);
+
+COMMENT ON TABLE party IS 'Party references an individual or an organization that does business with EMPR, and any relevant contact information, such as address.';
+
+CREATE TABLE IF NOT EXISTS province_code (
+    province_code    character varying(2)                   NOT NULL,
+    description      character varying(50)                  NOT NULL,
+    display_order    integer                                NOT NULL,
```
klefler: We could make this nullable... that way if a code is marked as "not active" we don't have to assign it a display value?

migrations/sql/V2019.02.21.9.59__add_address_columns_to_party.sql
```diff
@@ -0,0 +1,54 @@
...
+
+INSERT INTO province_code
+    (
+    province_code,
+    description,
+    display_order,
+    create_user,
+    update_user
+    )
+VALUES
```
klefler: do we want to set these values in alphabetical order?

NWCalvank: Sure. I just did it based on the SGC code in this table, since that's how it's ordered in this table:
https://www12.statcan.gc.ca/census-recensement/2011/ref/dict/table-tableau/table-tableau-8-eng.cfm

Given that the sort order is currently just used for dropdowns, I guess alphabetical probably makes the most sense. I'll update

migrations/sql/V2019.02.21.9.59__add_address_columns_to_party.sql
NWCalvank: Sure, that makes sense.

migrations/sql/V2019.02.21.9.59__add_address_columns_to_party.sql
```diff
@@ -0,0 +1,54 @@
+ALTER TABLE party ADD COLUMN suite_no       varchar(5);
+ALTER TABLE party ADD COLUMN address_line_1 varchar(50);
```
nayr974: I think supporting multiple addresses for a party in the future was discussed. If we can **confirm** that, it could be an opportunity to put this in a related table and avoid some future refactoring, even if today we only ever treat it as a 1:1 entity. If we can't confirm this 100% then it does make sense not to pay that cost today. Not a blocking comment.

NWCalvank: This was discussed. I asked Kristin if we should do it in a separate address table to support multiple, and the consensus was to put the columns directly on the party table at this time.

nayr974: Sounds good. My intuition leans the other way, but I also don't think we'll have any complexities refactoring this if needed.

PR #546, opened 2019-02-22
MDS-1431 Feature/Add contact view at a system level

mmumby:
This PR includes new view for contacts, route updates, file structure updates, new navigation, 

Note: The new navigation links have been added for admin only, and the new route is protected for admin only, therefore users will not see this change. (This is temporary until we have a permission developer role for feature flagging - coming soon!)


frontend/src/actionCreators/partiesActionCreator.js
```diff
@@ -31,11 +31,11 @@ export const createParty = (payload) => (dispatch) => {
     });
 };
 
-export const fetchParties = (value = null) => (dispatch) => {
+export const fetchParties = (params = null) => (dispatch) => {
```
NWCalvank: Reason for a default value of `null`? Is `undefined` not acceptable here?

frontend/src/components/Home.js
```diff
@@ -28,13 +44,13 @@ export class Home extends Component {
     return (
       <Layout className="layout">
         <div className="header">
-          <NavBar />
+          <NavBar activeButton={this.state.activeNavButton} />
           <LoadingBar
             style={{
-              backgroundColor: "#6b6363",
+              backgroundColor: "#5e46a1",
```
NWCalvank: Bringing this up again. Constants for JSX colours?

frontend/src/components/dashboard/contactsHomePage/ContactHomePage.js
```diff
@@ -0,0 +1,146 @@
...
+    } else {
+      this.props.history.push(
+        router.CONTACT_HOME_PAGE.dynamicRoute({
+          page: Strings.DEFAULT_PAGE,
+          per_page: Strings.DEFAULT_PER_PAGE,
+        })
+      );
+    }
+    this.props.fetchParties(params).then(() => {
+      this.setState({ isContactList: true });
```
NWCalvank: Reason for `isContactList` instead of `isLoading`?

frontend/src/components/dashboard/contactsHomePage/ContactList.js
```diff
@@ -0,0 +1,69 @@
...
+    render: (text) => <div title="phone">{text}</div>,
+  },
+];
+
+const transformRowData = (parties, partyIds) =>
+  partyIds.map((id) => ({
+    key: id,
+    emptyField: Strings.EMPTY_FIELD,
+    name: parties[id].name ? parties[id].name : Strings.EMPTY_FIELD,
+    email: parties[id].email === "Unknown" ? Strings.EMPTY_FIELD : parties[id].email,
```
NWCalvank: Is the API returning values as the string "Unknown"?

There's a db constraint that `party.email` cannot be `null`. So, I'm just wondering where this value would be coming from

frontend/src/components/dashboard/contactsHomePage/ContactList.js
```diff
@@ -0,0 +1,69 @@
...
+    phone:
+      parties[id].phone_no && parties[id].phone_no !== "Unknown"
```
NWCalvank: Same question as with `party.email`

frontend/src/components/mine/ContactInfo/MineContactInfo.js
```diff
@@ -33,7 +33,8 @@ export class MineContactInfo extends Component {
   }
 
   handleChange = (value) => {
-    this.props.fetchParties(value);
+    const param = `?search=${value}`;
+    this.props.fetchParties(param);
```
NWCalvank: Why does `fetchParties` accept a query string?

It seems more idiomatic to me to pass an object to your `fetch...` methods, and have the route function handle the query string building with the `querystring` library. Like we're currently doing here:
https://github.com/bcgov/mds/blob/develop/frontend/src/constants/routes.js#L17

Thoughts?

frontend/src/constants/API.js
```diff
@@ -6,7 +6,8 @@ export const MINE_LIST = "/mines";
 export const MINE_LIST_QUERY = (params) => `/mines${params}`;
 export const MINE_BASIC_INFO_LIST = `/mines/basicinfo`;
 export const PARTY = "/parties";
-export const PARTIES = (search) => (search ? `/parties?search=${search}` : "/parties");
+// export const PARTIES = (search) => (search ? `/parties?search=${search}` : "/parties");
```
NWCalvank: Comment that can be removed

frontend/src/constants/routes.js
```diff
@@ -11,26 +12,33 @@ export const DASHBOARD = {
...
-    return `/dashboard?page=${page}&per_page=${perPage}${searchParam}&map=true`;
+    return `/dashboard/mines?page=${page}&per_page=${perPage}${searchParam}&map=true`;
   },
   component: Dashboard,
 };
 
+export const CONTACT_HOME_PAGE = {
+  route: "/dashboard/contacts",
+  dynamicRoute: ({ page, per_page }) =>
+    `/dashboard/contacts/?${queryString.stringify({ page, per_page })}`,
```
NWCalvank: This brings me joy

frontend/src/styles/components/Tables.scss
```diff
@@ -5,7 +5,7 @@
 @include tablet {
   .ant-table table {
     min-width: 100%;
-    font-size: 0.875rem; // 14px to allow some breathing room on smaller devices
+    // font-size: 0.875rem; // 14px to allow some breathing room on smaller devices
```
NWCalvank: Remove?

frontend/src/components/dashboard/contactsHomePage/ContactHomePage.js
```diff
@@ -0,0 +1,146 @@
...
+      this.renderDataFromURL(params);
+    }
+  }
+
+  componentWillUnmount() {
+    this.setState({ params: {} });
+  }
+
+  renderDataFromURL = (params) => {
+    const { page, per_page } = queryString.parse(params);
```
NWCalvank: It seems to me like `params` should be the object, not the query string. That way we can pass around the params object in the code, and then stringify it with `queryString.stringify` in the route function (as mentioned in another comment)

frontend/src/components/Home.js
mmumby: yeah I'll make that change!

frontend/src/components/dashboard/contactsHomePage/ContactList.js
mmumby: yeah it's coming back as 'Unknown', the phone number is also returned as "Unknown"

frontend/src/constants/API.js
mmumby: removed.

frontend/src/styles/components/Tables.scss
mmumby: deleted.

frontend/src/components/dashboard/contactsHomePage/ContactList.js
NWCalvank: Okay. I just took a look at the codebase, and that's what we're putting in during ETL when the email isn't available in MMS. I feel like that should just be a nullable field and should use `null` instead if we can't actually guarantee to always have a useful value there.

I'll start a chat on Slack.

frontend/src/components/mine/ContactInfo/MineContactInfo.js
mmumby: I followed the same Logic that `fetchMines` does, The difference here, is that the value in this particular situation doesn't come from the URL, it is the return of a `handleChange` function.

With that example, the params are all fetched from the URL and passed into `fetchMines`, The map search uses a different endpoint that accepts just the payload, not a query string and constructs the URL with the payload.
```
export const MINE_NAME_LIST = (params = {}) => `/mines/search?${queryString.stringify(params)}`;
```
In this situation, we are using the same endpoint for both so I had to create the query myself.
Does that explain my reasoning enough?

```
export const MINE_LIST_QUERY = (params) => `/mines${params}`;
export const PARTIES_LIST_QUERY = (params) => (params ? `/parties${params}` : "/parties");
```
Will likely need to change when this when the advanced filters are added anyways..

frontend/src/components/dashboard/contactsHomePage/ContactList.js
mmumby: good idea! The Unkown has caused some issues with our null checks.

frontend/src/components/mine/ContactInfo/MineContactInfo.js
NWCalvank: Yep, totally. As discussed IRL, we can use a params object / refactor when implementing the advanced search 👍 

Marking as resolved

frontend/src/components/dashboard/contactsHomePage/ContactHomePage.js
mmumby: fixed!

PR #548, opened 2019-02-23
MDS-1438 Feature/ Support GET/POST/PUT for address fields at /parties endpoint

NWCalvank:
Adds an `address` array to the JSON response payload for the /parties endpoint as well as support for setting these fields on create/update


python-backend/app/api/parties/party/models/party.py
```diff
@@ -15,15 +15,22 @@
...
     effective_date = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
     expiry_date = db.Column(db.DateTime, nullable=False, default=datetime.strptime('9999-12-31', '%Y-%m-%d'))
-    party_type_code = db.Column(db.String(3), db.ForeignKey('party_type_code.party_type_code'))
+    party_type_code = db.Column(db.String, db.ForeignKey('party_type_code.party_type_code'))
+
+    suite_no = db.Column(db.String, nullable=True)
+    address_line_1 = db.Column(db.String, nullable=True)
+    address_line_2 = db.Column(db.String, nullable=True)
+    city = db.Column(db.String, nullable=True)
+    province_code = db.Column(db.String, nullable=True)
```
nayr974: At once point support for international addresses was raised, as companies possibly have international contacts. May want to validate that that is not in scope. I think there's a standard set of fields for international addresses (4 address lines, locale, region, zip, country or along that line).

NWCalvank: Yep, we discussed this as well before I moved ahead on implementing this. Kristin said to stick with province_code and postal_code at this time.

It may be worth re-addressing this if you think it could lead to breaking API changes.

nayr974: Seems short sighted not to nail this down. If we intend to support it in the future we use those fields today. I don't think its extra complexity, and if we want to show "province" in our UI and have more generic fields in our DB we can do that. @klefler @itchison thoughts?

https://stackoverflow.com/questions/929684/is-there-common-street-addresses-database-design-for-all-addresses-of-the-world

    Address Lines 1-4
    Locality
    Region
    Postcode (or zipcode)
    Country


https://stackoverflow.com/questions/126207/best-practices-for-consistent-and-comprehensive-address-storage-in-a-database
    name,
    company-name,
    street-lines[], // up to 4 free-type street lines
    county/sublocality,
    city/town/district,
    state/province/region/territory,
    postal-code,
    country

klefler: Hi @nayr974 - yep, if we want to talk about a more robust address model, I've got some ideas. Let's chat!

python-backend/app/api/parties/party/models/party.py
```diff
@@ -81,24 +96,71 @@ def search_by_name(cls, search_term, party_type=None, query_limit=50):
...
+            if first_name:
+                name = first_name + ' ' + party_name
+            else:
+                name = party_name
+            raise AssertionError(
+                'Party with the name: {} already exists'.format(name))
+        return party_name
+
+    @validates('party_type_code')
+    def validate_email(self, key, party_type_code):
```
Jsyro: Function name here seems off.

NWCalvank: Yep! That'll need to be fixed. These validate functions were moved around quite a bit while refactoring. Thanks for catching that

python-backend/app/api/parties/party/models/party.py
NWCalvank: Let's address this tomorrow morning, so that I can implement the changes on this and close the branch. Thanks!

migrations/sql/V2019.02.26.9.21__alter_party_table_rename_address_columns.sql
```diff
@@ -0,0 +1,46 @@
...
+    display_order     integer                                        ,
+    active_ind        boolean                  DEFAULT true  NOT NULL,
+    create_user       character varying(60)                  NOT NULL,
+    create_timestamp  timestamp with time zone DEFAULT now() NOT NULL,
+    update_user       character varying(60)                  NOT NULL,
+    update_timestamp  timestamp with time zone DEFAULT now() NOT NULL
+);
+
+ALTER TABLE address_type_code OWNER TO mds;
+COMMENT ON TABLE address_type_code IS 'Lookup table for valid address types';
```
klefler: Just a picky thing, but I can see how this code table might be confusing to some... can we update the table description to "Lookup table for valid address types, used to decipher a range of addressing styles used in different countries."  (or something like that)

NWCalvank: Yep. I'll copy/paste that description directly

migrations/sql/V2019.02.26.9.21__alter_party_table_rename_address_columns.sql
```diff
@@ -0,0 +1,46 @@
...
+    create_user,
+    update_user
+    )
+VALUES
+    ('CAN', 'Canada', 10, 'system-mds', 'system-mds'),
+    ('USA', 'United States', 20, 'system-mds', 'system-mds'),
+    ('INT', 'International', 30, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
+ALTER TABLE province_code RENAME TO region_code;
```
klefler: We might want to consider using a different name for this table... "region" has a very specific meaning in EMPR (and the rest of Natural Resource Ministries). I had suggested "sub_division_code" which would in essence be a child of a table called "country" (at some point maybe)... but it's a generic term used to describe things like states, provinces, territories, etc.

NWCalvank: I can rename the db table, yeah. I'm really trying to just go with consensus on these, so can we make sure there's a proper consensus if I go through another rename? @nayr974 

I think region_code still makes sense at the API level, though. Since it'll be in an object, in an array that's labelled `addresses`

Thanks!

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -72,77 +75,60 @@ def get(self, party_guid=None):
...
-                party_guid=uuid.uuid4(),
-                phone_no=data['phone_no'],
-                email=data['email'],
-                phone_ext=data.get('phone_ext'),
-                **self.get_create_update_dict(),
-                **party_context)
+            party = Party.create(data['party_name'],
+                                 data['email'],
+                                 data['phone_no'],
+                                 data['type'],
```
Jsyro: Will this send back a stack trace if one of these is missing? I believe we want a controlled error message.

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -72,77 +75,60 @@ def get(self, party_guid=None):
...
+            existing_party.party_name      = data.get('party_name') or existing_party.party_name
+            existing_party.email           = data.get('email') or existing_party.email
+            existing_party.phone_no        = data.get('phone_no') or existing_party.phone_no
+            existing_party.party_type_code = data.get('type') or existing_party.party_type_code
+            existing_party.first_name      = data.get('first_name') or existing_party.first_name
+            existing_party.suite_no        = data.get('suite_no') or existing_party.suite_no
+            existing_party.address_line_1  = data.get('address_line_1') or existing_party.address_line_1
+            existing_party.address_line_2  = data.get('address_line_2') or existing_party.address_line_2
+            existing_party.city            = data.get('city') or existing_party.city
+            existing_party.region_code     = data.get('region_code') or existing_party.region_code
```
nayr974: Kristin was thinking subdivision instead of region, otherwise looks great 👍

python-backend/app/api/parties/party/resources/party_resource.py
NWCalvank: This should reflect all existing behaviour. I didn't feel comfortable refactoring these in this PR. This shows up in the diff, but the key access style of these properties has not been changed.

I would argue that we should convert these to use `.get()` for key access and check with an appropriate error, but I don't think it's appropriate in this PR. There's already scope creep happening.

All of the new functionality that I add has used `.get()` and proper null checks.

 Hope that seems reasonable

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -72,77 +75,60 @@ def get(self, party_guid=None):
...
+                                 first_name=data.get('first_name'),
+                                 phone_ext=data.get('phone_ext'),
+                                 suite_no=data.get('suite_no'),
+                                 address_line_1=data.get('address_line_1'),
+                                 address_line_2=data.get('address_line_2'),
+                                 city=data.get('city'),
+                                 region_code=data.get('region_code'),
+                                 post_code=data.get('post_code'))
         except AssertionError as e:
-            self.raise_error(400, 'Error: {}'.format(e))
```
Jsyro: I don't think this will run the '@validates' methods... as the create method adds it to the transaction but doesn't commit the transaction. Let me know if i'm mistaken.  I think this try/except needs to be on the save below, i may be mistaken though.

python-backend/app/api/parties/party/resources/party_resource.py
NWCalvank: I'm going to move the `AssertionError` `try` block wrap the `save()`, as I believe your other comment is correct. And I'll catch the key error here

NWCalvank:
Closing to destroy everything from migration changes and starting a new env.

NWCalvank:
Latest `develop` killed my build. Trashing the entire environment and rebuilding to see if that fixes it.

NWCalvank:
🤞

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -72,77 +75,62 @@ def get(self, party_guid=None):
...
+            existing_party.email             = data.get('email') or existing_party.email
+            existing_party.phone_no          = data.get('phone_no') or existing_party.phone_no
+            existing_party.party_type_code   = data.get('type') or existing_party.party_type_code
+            existing_party.first_name        = data.get('first_name') or existing_party.first_name
+            existing_party.suite_no          = data.get('suite_no') or existing_party.suite_no
+            existing_party.address_line_1    = data.get('address_line_1') or existing_party.address_line_1
+            existing_party.address_line_2    = data.get('address_line_2') or existing_party.address_line_2
+            existing_party.city              = data.get('city') or existing_party.city
+            existing_party.sub_division_code = data.get('sub_division_code') or existing_party.sub_division_code
+            existing_party.post_code         = data.get('post_code') or existing_party.post_code
```
Jsyro: Was going to wait until tech demo, but this is coming up more. 

We are unable to differentiate between a None being intentionally set, or if the value was meant to unchanged. The pattern I think we are going with is... if the key is in the dictionary (form data on put), then take it's value (assuming it passes validation), if it is not included in the form data then leave it unchanged. 

However by default, reqparser.parse_arg() makes a key for all arguments defined with reqparser.add_argument() (even if it isn't in the request body), this means we cannot detect if the kvp was actually received. The solution is to add store_missing=False to the add_argument() at the top of the resource class, then the condition here is, if 'key' in data.keys(), copy the value. the store_missing=False tells reqparser, only make a kvp if it was actually provided in the request body. 

Let me know if that makes sense. I have an example on my current branch.

PR #553, opened 2019-02-27
MDS-1431 Feature/ Conditional nesting of party.mine_party_appt & mine_party_appt.party

NWCalvank:
The requirement was to support returning `party.mine_party_appt`. We have chosen to do this conditionally, based on the `relationships` param

Because `mine_party_appt` blindly returned its `party` relationship, this meant that `party.mine_party_appt` would include a second, nested instead of `party`. For this reason, I have added the `relationships` condition to the `mine_party_appt` GET method, with respect to the `party` relationship.

To avoid any regressions on the frontend, all references to this endpoint have been updated to provide the `relationships=party` query param. Some of these may not be necessary, but it maintains existing behaviour and can be optimized at a later date. The priority here is avoiding regressions while gradually implementing the new pattern.


NWCalvank:
Latest develop appears to have killed my dev deployment with a migration file. I believe the fix is in now, so I will tear everything down here and rebuild fresh.

NWCalvank:
🤞

NWCalvank:
Burn it down... 🔥

NWCalvank:
...and hope it passes

python-backend/app/api/parties/party_appt/resources/mine_party_appt_resource.py
```diff
@@ -26,18 +26,20 @@ class MinePartyApptResource(Resource, UserMixin, ErrorMixin):
     @api.doc(params={'mine_party_appt_guid': 'mine party appointment serial id'})
     @requires_role_mine_view
     def get(self, mine_party_appt_guid=None):
+        relationships = request.args.get('relationships')
+        relationships = relationships.split(',') if relationships else []
```
Jsyro: I think in theory the HTTP standard calls for the query params to look like `/mines?relationship=parties&relationship=permits` and then we can use `request.args.getlist('relationship')` but this exists in other parts. Just an FYI, not a blocker.

python-backend/tests/parties/party/resources/test_party_resource.py
```diff
@@ -93,20 +93,6 @@ def test_post_person_no_email(test_client, auth_headers):
     assert post_resp.status_code == 400
 
 
-def test_post_person_name_exists(test_client, auth_headers):
```
Jsyro: Just wondering about removed tests, can your provide insight? where they duplicated? was the behavior no longer required?

NWCalvank: Some people have the same name. The unique name validation was removed (shh.. don't tell anyone). So, the tests had to go as well

python-backend/app/api/parties/party_appt/resources/mine_party_appt_resource.py
NWCalvank: I'm fine with either pattern. The comma-separated approach is used elsewhere for listed values, so I did that here. If it's not the pattern we want to use then I'll update this -- but I'll ask for team consensus around that first, since we already have one pattern in place

python-backend/tests/parties/party/resources/test_party_resource.py
Jsyro: There is a ticket for that behavior (https://bcmines.atlassian.net/browse/MDS-1393) and larger issues that need to be considered, like how do we display a list of parties when we need more than just name to uniquely identify. I agree the constraint needs to go, but i'd like to see that problem tackled as a whole. Up to Aaron ultimately if he wants the constraint removed without the additional changes.

NWCalvank: I'm fine with reverting this change. I don't think this is a validation that should be in the codebase, though. It seems more problematic than anything. The choice to remove it was made during a conversation with @nayr974 and @mmumby this morning.

Jsyro: I totally agree the validation rule shouldn't exist, but I don't support that validation being removed without a discussion with Aaron and explanation of the impact, as there is already a prioritized item in the backlog that contains this change.

NWCalvank: Reverted

PR #554, opened 2019-02-28
Mds 1364 Permit Amendments - Minor Tweaks

nayr974:
Only show one Permittee for a permit on the Mine Summary; Minor code tweaks.


frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -139,9 +139,7 @@ const transformChildRowData = (amendment, record, amendmentNumber) => ({
 
 export const MinePermitInfo = (props) => {
   const amendmentHistory = (record) => {
-    const childRowData = record.amendments.map((amendment, index) =>
-      transformChildRowData(amendment, record, record.amendments.length - index)
-    );
+    const childRowData = record.amendments.map(transformChildRowData(transformChildRowData));
```
NWCalvank: Why is `transformChildRowData` being passed to `transformChildRowData`? It doesn't look like this should work or be necessary.

It looks like this should be all that's needed:
```javascript
const childRowData = record.amendments.map(transformChildRowData)
```

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -94,14 +94,24 @@ export const MineSummary = (props) => {
...
                     props.partyRelationshipTypes
                   )
                 )}
+              {props.mine.mine_permit.map((permit) =>
+                renderPartyRelationship(
+                  props.mine,
+                  props.partyRelationships
+                    .filter(isActive)
+                    .filter((pr) => pr.mine_party_appt_type_code === "PMT")
+                    .filter((pr) => pr.related_guid === permit.permit_guid)[0],
```
NWCalvank: I doubt it makes much/any difference with the amount of data that we're dealing with in this part of the code, but it could be worth mentioning since it's part of the rendered JSX.. these chained filters are creating unnecessary iterations through the array.

These could be composed into a single conditional, and perhaps assigned to a variable for readability.
```javascript
const aGoodConditionName = (pr) => isActive(pr) &&
  pr.mine_party_appt_type_code === "PMT" && 
  pr.related_guid === permit.permit_guid

// stuff
props.partyRelationships
                    .filter(aGoodConditionName)[0]

```

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -94,14 +94,24 @@ export const MineSummary = (props) => {
             <Row gutter={16} type="flex">
               {props.partyRelationships
                 .filter(isActive)
-                .filter((pr) => ["MMG", "PMT"].includes(pr.mine_party_appt_type_code))
+                .filter((pr) => pr.mine_party_appt_type_code === "MMG")
```
NWCalvank: Same comment about chained filters here

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: Not sure why I did that. Changed 👍

frontend/src/components/mine/Summary/MineSummary.js
nayr974: That crossed my mind as well, but I think Permittee import logic may change next sprint (MDS-1516) as a different source for the "current" permittee should be considered and I'm not sure yet how that will be represented in the collection of Permittees.

nayr974: Actually, let's pull it out anyways even if the logic changes. Might as well and I have the file open.

nayr974: Updated so that each only has one filter 👍

PR #555, opened 2019-02-28
BUG: MDS-1465 and MDS-1210

mmumby:
- Added a few null checks to remove "Unkown" from being an email link, showing "N/A" if an email does not exist. 

- Added a null check to all dates being passed into the `formatDate` function.

### DIFF
--- a/frontend/src/components/parties/RelationshipProfile.js
+++ b/frontend/src/components/parties/RelationshipProfile.js
@@ -150,9 +150,9 @@ export class RelationshipProfile extends Component {
         role: this.state.partyRelationshipTitle,
         permit: this.state.permitsMapping[relationship.related_guid],
         endDate:
-          relationship.end_date === "9999-12-31" || null
-            ? "Present"
-            : formatDate(relationship.end_date),
+          relationship.end_date && relationship.end_date !== "9999-12-31"
+            ? formatDate(relationship.end_date)
+            : "Present",
         startDate: relationship.start_date ? formatDate(relationship.start_date) : "Unknown",
       }));
 
--- a/frontend/src/components/parties/RelationshipProfile.js
+++ b/frontend/src/components/parties/RelationshipProfile.js
@@ -153,11 +153,8 @@ export class RelationshipProfile extends Component {
         partyGuid: relationship.party.party_guid,
         role: this.state.partyRelationshipTitle,
         permit: this.state.permitsMapping[relationship.related_guid],
-        endDate:
-          relationship.end_date && relationship.end_date !== "9999-12-31"
-            ? formatDate(relationship.end_date)
-            : "Present",
-        startDate: relationship.start_date ? formatDate(relationship.start_date) : "Unknown",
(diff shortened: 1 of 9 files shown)


frontend/src/components/parties/RelationshipProfile.js
```diff
@@ -154,9 +154,9 @@ export class RelationshipProfile extends Component {
         role: this.state.partyRelationshipTitle,
         permit: this.state.permitsMapping[relationship.related_guid],
         endDate:
-          relationship.end_date === "9999-12-31" || null
-            ? "Present"
-            : formatDate(relationship.end_date),
+          relationship.end_date && relationship.end_date !== "9999-12-31"
+            ? formatDate(relationship.end_date)
+            : "Present",
```
NWCalvank: Not necessary, but we could move this logic into the helper function? Then we only have the magic string in one place
```javascript
export const formatDate = (dateString) => dateString && dateString !== "9999-12-31" && moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");

// Then we could just do something like:
formatDate(relationship.end_date) || "Present"
```

mmumby: I just started to update this, and then realized that this doesn't actually help in all situations. 

For example, it appears that the date is sometimes null, "None" or  
"9999-12-31"... and the outcome is "N/A", "Unknown", or "Present" 

So checks would still need to happen before we passed any date into this function to know what out outcome should be. This function would need to be turned into a ternary, and since the outcome is different every time that makes this difficult to do. 

If we always showed "N/A" if the data was null, then it would make sense to enhance this function to handle that.

thoughts?

NWCalvank:
```
it appears that the date is sometimes null, "None" or
"9999-12-31"... and the outcome is "N/A", "Unknown", or "Present"
```
^ This is a serious concern that needs to be addressed before third parties start consuming the API. This sort of inconsistency and use of magic strings is not acceptable for a third-party to deal with.

As far as the frontend goes, though, the formatData function could still just chain those potential values.
```javascript
dateString => 
  dateString &&
  dateString !== "9999-12-31" &&
  dateString !== "None" &&
```

That will return a falsey value or a valid date string. Which is why you could still do this at the callsite:
```javascript
formatDate(relationship.end_date) || "Present"
// or
formatDate(relationship.end_date) || "Unknown"
// or
formatDate(relationship.end_date) || "N/A"
```

So, the `formatDate` function doesn't know anything about the default/fallback value. It just returns a valid string or `null`/`undefined`/`false`

Does that clarify what I meant? Or maybe I'm misunderstanding your concern. If so, let's chat IRL

mmumby: Yes, this clarifies. 

Logically, I was overlooking the `falsey` return from `dateFormat` and assuming that checks would still need to happen at the callsite to some extent. This implementation makes sense. 


and I agree 100% with the API concerns.

PR #557, opened 2019-02-28
Feature/mds 1452 1457 permit amendment crud and uploads 

Jsyro:
this also temporarliity mitigates some of our sqlalchemy code first testing model issues. including skipping the test


python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -24,14 +24,16 @@ class PermitAmendment(AuditMixin, Base):
...
         return {
             'permit_amendment_id':
-            self.permit_amendment_id,
+            str(self.permit_amendment_id),
             'permit_amendment_guid':
             str(self.permit_amendment_guid),
-            'permit_id':
-            self.permit_id,
+            'permit_guid':
+            str(self.permit.permit_guid),
```
NWCalvank: What happens if you don't convert a guid to a string here? This seems redundant to me

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -41,16 +43,44 @@ def json(self):
...
+        return cls.query.filter_by(permit_amendment_id=_id).filter_by(deleted_ind=False).first()
+
+    @classmethod
+    def find_by_permit_amendment_guid(cls, _guid):
+        return cls.query.filter_by(permit_amendment_guid=_guid).filter_by(deleted_ind=False).first()
 
     @classmethod
     def find_by_permit_id(cls, _id):
-        return cls.query.filter_by(permit_id=_id)
+        return cls.query.filter_by(permit_id=_id).filter_by(deleted_ind=False).all()
```
NWCalvank: It seems like it could be helpful to add `.filter_by(deleted_ind=False)` to a base query for this class. And maybe do something similar for `active_ind` on some other classes?

Any idea if we could do something like this:
```python
def not_deleted(cls):
    return cls.query.filter_by(deleted_ind=False)

@classmethod
def find_by_permit_id(cls, _id):
    return cls.not_deleted().filter_by(permit_id=_id).all()
```
or
```python
not_deleted = PermitAmendment.query.filter_by(deleted_ind=False)

@classmethod
def find_by_permit_id(cls, _id):
    return not_deleted.filter_by(permit_id=_id).all()
```

None of this is tested -- just spit-balling ideas. Thoughts?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -72,29 +102,26 @@ def validate_permit_id(self, key, permit_id):
...
     def validate_received_date(self, key, received_date):
-        if not received_date:
-            return received_date
-        if received_date.isoformat() == '9999-12-31':
-            raise AssertionError(
-                'Permit amendment received date should be set to null if not known.')
-        if received_date > datetime.today():
-            raise AssertionError('Permit amendment received date cannot be set to the future.')
+        if received_date:
+            if received_date.isoformat() == '9999-12-31':
```
NWCalvank: What's the reason for changing this from the early return style?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -72,29 +102,26 @@ def validate_permit_id(self, key, permit_id):
...
     @validates('issue_date')
     def validate_issue_date(self, key, issue_date):
-        if not issue_date:
-            return issue_date
-        if issue_date.isoformat() == '9999-12-31':
-            raise AssertionError('Permit amendment issue date should be set to null if not known.')
-        if issue_date > datetime.today():
-            raise AssertionError('Permit amendment issue date cannot be set to the future.')
+        if issue_date:
+            if issue_date.isoformat() == '9999-12-31':
```
NWCalvank: Also here. Why refactor away the early return? This was creates deeper conditional nesting while appearing to have the same behaviour

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -41,16 +43,44 @@ def json(self):
...
+    @classmethod
+    def create(cls,
+               permit,
+               received_date,
+               issue_date,
+               authorization_end_date,
+               user_kwargs,
+               permit_amendment_type_code='AMD',
+               permit_amendment_status_code='ACT',
+               save=False):
```
NWCalvank: I think the other `save` values are defaulted to `True`.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
```diff
@@ -21,7 +44,89 @@ def get(self, permit_guid=None, permit_amendment_id=None):
...
+        if not permit_amendment_guid:
+            return self.create_error_payload(400, 'permit_amendment_id must be provided'), 400
+        pa = PermitAmendment.find_by_permit_amendment_guid(permit_amendment_guid)
+        if not pa:
+            return self.create_error_payload(404, 'permit amendment not found'), 404
+
+        data = self.parser.parse_args()
+        current_app.logger.info(f'updating {pa} with >> {data.keys()}')
+
+        if 'received_date' in data.keys():
```
NWCalvank: Quick test in my terminal shows that the `keys()` method is redundant?
```
>>> 'received_date' in { 'received_date': 1 }
True
>>> 'received_date' in { 'issue_date': 1 }
False
```

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
```diff
@@ -21,7 +44,89 @@ def get(self, permit_guid=None, permit_amendment_id=None):
...
+            pa.received_date = data.get('received_date')
+        if 'issue_date' in data.keys():
+            pa.issue_date = data.get('issue_date')
+        if 'authorization_end_date' in data.keys():
+            pa.authorization_end_date = data.get('authorization_end_date')
+        if 'permit_amendment_status_code' in data.keys():
+            pa.permit_amendment_status_code = data.get('permit_amendment_status_code')
+        if 'permit_amendment_type_code' in data.keys():
+            pa.permit_amendment_type_code = data.get('permit_amendment_type_code')
```
NWCalvank: Could you test for throwing an `AssertionError` here? I have a feeling that the `validates` methods are executed when values are pushed to the object. I could be wrong since the constructor isn't being called here.

I'm just not sure if the `AssertionError` check around the `save()` will cover all our bases. Not certain, so I'd just like this confirmed. Thanks!

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
```diff
@@ -0,0 +1,125 @@
...
+from ....utils.access_decorators import requires_role_mine_create
+from ....utils.resources_mixins import UserMixin, ErrorMixin
+from ....utils.url import get_document_manager_svc_url
+
+
+class PermitAmendmentDocumentResource(Resource, UserMixin, ErrorMixin):
+    parser = reqparse.RequestParser()
+    parser.add_argument('document_manager_guid', type=str)
+    parser.add_argument('filename', type=str)
+    #permit_guid, it could be in the request and we don't want to disallow it, but it is not used.
```
NWCalvank: Is this kept around for a reason or just missed?

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
Jsyro: objects of type uuid aren't serializable by default sadly.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
Jsyro: I'm been looking to add it to the default behaviour of the db.query object. like how we have the tenant level control for mines. https://bcmines.atlassian.net/browse/MDS-1460, all queries automatically have the deleted filter unless you call a 'include_deleted()' method. The design principle is that deleted objects would never be returned by the api (sql forensics only).

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
Jsyro: Just how we have talked about the pattern, early positive returns can be harder to manage as function size grows, this instance was probably fine, but I had already done it before the catch was implemented this way, so I kept my version.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
Jsyro: Correct, good catch. Thanks.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
Jsyro: Learned something new, thanks.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
Jsyro: It explains what appears to be an unnecessary parameter in all the methods in the file. I'm not attached if you think the comment is out of place.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
Jsyro: I feel like I tested and found them invoked on save... I was mistaken, you are correct, it is checked on assignment.. changed.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -13,8 +13,8 @@
 
 class PermitAmendment(AuditMixin, Base):
     __tablename__ = 'permit_amendment'
-    permit_amendment_id = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
-    permit_amendment_guid = db.Column(UUID(as_uuid=True))
+    permit_amendment_id = db.Column(db.Integer, primary_key=True)
```
Jsyro: I want to add a comment about how the id should be FetchedValue, but with the current test infrastructure that isn't possible.

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
NWCalvank: Okay! Weird

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
NWCalvank: That would be my preference as well. I just wasn't sure if that sort of global change could be made in the microservice approach that we're supposed to be implementing

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
NWCalvank: Where/when was this pattern discussed? I favour an early return when exiting due to an error.

It says "we don't have a received date, so there's nothing else to do here. Exiting."

This, to me, seems clearer.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
NWCalvank: I'm fine with it. Just wanted to confirm that it wasn't left accidentally

python-backend/app/api/permits/permit_amendment/models/permit_amendment.py
Jsyro: To clarify, i think we call agree that returning early on an error makes sense and it totally fine, we do it lots in the resources. But I prefer to have happy path always make it to the end of the function.

NWCalvank: Ahh.. perfect. Thanks for clarifying

PR #564, opened 2019-03-04
Feature/MDS-1428 Load NRIS in backgroud

james-hollinger:
NRIS now loads in background when the mine dashboard is pulled up.
 - The compliance endpoint will now return a default compliance info object rather than null if there are no inspections in NRIS.
 - the complianceActionCreator now does not have a cached-only mode (this was originally to support loading the compliance data in the background, but only if it was in the cache). Instead, it now has a "silent" mode that does not show notifications when errors occur during the loading of NRIS.
 - the complianceActionCreator now puts an empty object into the store when called. This is to prevent the previously loaded mine's compliance data from showing while loading the new mine.
 - MineComplianceInfo no longer fetches the data, it only reads from the store.
 - While the compliance data is loading, the dashboard has a complianceInfoLoading flag that is passed to the summary tab and compliance tab.
 - Compliance cards on the summary tab now render while loading, and display a "Loading..." message.
 - Compliance cards are now shown on the summary tab if NRIS returns no data. The "Last inspection date" is shown as "No inspections in NRIS". This is now a new string constant.
 - If NRIS returns an error, the compliance cards are not shown.
 - The summary screen only show the NullScreen ("No data") if there is both no mine info and NRIS returns an error.
 - The compliance info tab shows the NullScreen only if NRIS returns an error. It shows zeros in all fields if there are no inspections in NRIS.

In addition, the compliance endpoint was returning an "inspector" field that was unused by the app. To support future development, this field has been renamed to "last_inspector", and the logic has been updated accordingly.

Lastly, the string keys used for Redis have been altered to follow Redis' recommended standards.


frontend/src/components/mine/Compliance/MineComplianceInfo.js
```diff
@@ -140,7 +133,9 @@ export class MineComplianceInfo extends Component {
                   <Col span={4}>
                     <div className="center">
                       <p className="info-display">
-                        {formatDate(this.props.mineComplianceInfo.last_inspection) || "-"}
+                        {this.props.mineComplianceInfo.last_inspection === null
+                          ? String.NO_NRIS_INSPECTIONS
+                          : formatDate(this.props.mineComplianceInfo.last_inspection)}
```
NWCalvank: Meredith and I recently updated the `formatDate` function to handle falsey values so that it cane be used like this:
```javascript
formatDate(this.props.mineComplianceInfo.last_inspection) || String.NO_NRIS_INSPECTIONS
```

The idea was to move all the repeated null checks into the function because they were being written over and over

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -73,6 +75,9 @@ const isActive = (pr) =>
 const activePermitteesByPermit = (pr, permit) =>
   isActive(pr) && pr.mine_party_appt_type_code === "PMT" && pr.related_guid === permit.permit_guid;
 
+const displayLastInspectionDate = (dateString) =>
+  dateString === null ? String.NO_NRIS_INSPECTIONS : formatDate(dateString);
```
NWCalvank: Same comment about `formatDate` here

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -190,7 +196,9 @@ export const MineSummary = (props) => {
                       <img alt="Open Orders" src={DOC} style={{ height: 40, paddingRight: 5 }} />
                       &nbsp;
                       <span className="info-display">
-                        {props.mineComplianceInfo.num_open_orders}
+                        {props.complianceInfoLoading
+                          ? String.LOADING
+                          : props.mineComplianceInfo.num_open_orders}
```
NWCalvank: Do we need any safety checks on `props.mineComplianceInfo` to avoid `Cannot read property num_open_orders of undefined`?

python-backend/app/api/nris_services/NRIS_service.py
```diff
@@ -80,7 +83,7 @@ def _get_EMPR_data_from_NRIS(mine_no):
         try:
             empr_nris_resp.raise_for_status()
         except requests.exceptions.HTTPError:
-            #TODO add logging for this error.
+            # TODO add logging for this error.
```
NWCalvank: Was this missed or is this blocked by something?

python-backend/app/api/nris_services/NRIS_service.py
```diff
@@ -156,15 +155,17 @@ def _process_NRIS_data(data, mine_no):
                 advisories += len(stop_advisories)
                 warnings += len(stop_warnings)
 
+    latest_report = data[0] if data else None
     overview = {
-        'last_inspection': most_recent.get('assessmentDate'),
-        'inspector': inspector,
+        'last_inspection': latest_report.get('assessmentDate') if latest_report else None,
```
NWCalvank: I could just be derping, but I think `latest_report.get('assessmentDate')` returns `None` if it doesn't exist, so the ternary should be redundant

python-backend/app/api/nris_services/NRIS_service.py
james-hollinger: Not sure honestly.

python-backend/app/api/nris_services/NRIS_service.py
james-hollinger: If latest_report is none, calling get() on it will throw an error.

frontend/src/components/mine/Summary/MineSummary.js
james-hollinger: This entire section is wrapped in props.mineComplianceInfo && so only runs if mineComplianceInfo exists.

frontend/src/components/mine/Compliance/MineComplianceInfo.js
james-hollinger: Thanks for this, has been updated 👍

frontend/src/components/mine/Summary/MineSummary.js
james-hollinger: Updated.

PR #565, opened 2019-03-04
MDS-1441 Feature/ Contact Advanced Search on Frontend

NWCalvank:
This hits the `/parties` endpoint unless a `type=` query parameter is provided, at which point it hits the `/parties/search` endpoint (@thegentlemanphysicist is implementing this on the backend).

Behaviour is copied from the Mine Advanced Search form. Implementation is somewhat simplified by comparison.


frontend/src/components/dashboard/contactsHomePage/ContactList.js
```diff
@@ -51,19 +50,16 @@ const uniqueRolesString = (mine_party_appt, relationshipTypeHash) =>
     )
   ).join(", ");
 
-const transformRowData = (parties, partyIds, relationshipTypeHash) =>
-  partyIds.map((id) => ({
-    key: id,
+const transformRowData = (parties, relationshipTypeHash) =>
+  toArray(parties).map((party) => ({
```
mmumby: I like this lodash function!

PR #566, opened 2019-03-05
Feature/MDS-1432 Add and Edit Permits and Amendments

nayr974:
Enables adding and editing permits, adding and editing permit amendments, and uploading amalgamated permits.


frontend/src/components/Forms/AddPermitAmendmentForm.js
```diff
@@ -0,0 +1,77 @@
...
+  submitting: PropTypes.bool.isRequired,
+};
+
+const { TabPane } = Tabs;
+
+export const AddPermitAmendmentForm = (props) => (
+  <Form layout="vertical" onSubmit={props.handleSubmit}>
+    <Tabs defaultActiveKey="1" style={{ marginTop: "-30px" }}>
+      <TabPane tab="Info" key="1">
+        <br />
```
NWCalvank: Are we using hard line breaks on our other forms? Styling with our HTML seems like a tough pattern to manage since it blurs the line between HTML/CSS

frontend/src/components/Forms/AddPermitForm.js
```diff
@@ -0,0 +1,196 @@
...
+  },
+  {
+    label: "Sand & Gravel",
+    value: "G",
+  },
+  {
+    label: "Quarry",
+    value: "Q",
+  },
+];
```
NWCalvank: It looks like this should live on the server at an options endpoint

frontend/src/components/Forms/AddPermitForm.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+  return errors;
+};
+
+export const AddPermitForm = (props) => (
+  <Form layout="vertical" onSubmit={props.handleSubmit}>
+    <Tabs
+      defaultActiveKey="1"
+      onChange={(activeKey) => {
+        // validate
```
NWCalvank: Is this a TODO?

frontend/src/components/Forms/AddPermitForm.js
```diff
@@ -0,0 +1,196 @@
...
+const selector = formValueSelector(FORM.ADD_PERMIT);
+const { TabPane } = Tabs;
+
+const validate = (values) => {
+  const errors = {};
+  if (
+    values.permit_activity_type === "X" &&
+    !(values.permit_type === "C" || values.permit_type === "M")
+  ) {
+    errors.permit_activity_type = "Exploration activity is only valid for Coal and Placer permits";
```
NWCalvank: This looks like business logic in the frontend

frontend/src/components/mine/Permit/MinePermitInfo.js
```diff
@@ -1,166 +1,126 @@
...
-      expandedRowRender={amendmentHistory}
-      locale={{ emptyText: <NullScreen type="permit" /> }}
-    />
-  );
-};
+  render() {
+    return [
+      <div>
+        <div className="inline-flex between">
+          <div />
```
NWCalvank: Empty div?

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,467 @@
...
+        </Menu>
+      );
+      return (
+        <AuthorizationWrapper inTesting>
+          <AuthorizationWrapper permission={Permission.CREATE} isMajorMine={text.major_mine_ind}>
+            <Dropdown className="full-height full-mobile" overlay={menu} placement="bottomLeft">
+              <Button type="secondary" className="permit-table-button">
+                <div className="padding-small">
+                  <img className="padding-small--right icon-svg-filter" src={EDIT} alt="Add/Edit" />
+                  Add/Edit&nbsp;&nbsp;
```
NWCalvank: There are a number of non-breaking spaces in this file. I could be wrong, but I don't think we've needed them much elsewhere. What's the use-case here?

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,467 @@
...
+  const firstAmendment = permit.amendments[permit.amendments.length - 1];
+
+  const permittees = getPermittees(partyRelationships, permit);
+  const permitteeName = partyRelationships.length === 0 ? "" : getPermitteeName(permittees);
+  const hasAmalgamated = permit.amendments.find((pa) => pa.permit_amendment_type_code === "ALG");
+
+  return {
+    key: permit.permit_guid,
+    lastAmended:
+      (latestAmendment && latestAmendment.issue_date && formatDate(latestAmendment.issue_date)) ||
```
NWCalvank: I believe Meredith updated `formatDate` to handle `undefined`/`null`/magic string(s). So, the `latestAmendment.issue_date` check (and other checks like it) should be unnecessary

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,467 @@
...
+    description: amendment.description,
+  },
+  openEditAmendmentModal,
+});
+
+export const MinePermitTable = (props) => {
+  // Edit permit Amendment
+  const handleEditPermitAmendment = (data) => {
+    props.updatePermitAmendment(data.permit_amendment_guid, data).then(() => {
+      props.closeModal();
```
NWCalvank: Not a big deal, but the outer function is unnecessary here:
```javascript
.then(props.closeModal);
```

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,467 @@
...
+    transformRowData(
+      permit,
+      props.partyRelationships,
+      props.major_mine_ind,
+      openEditPermitModal,
+      openAddPermitAmendmentModal,
+      openAddAmalgamatedPermitModal,
+      props.permitStatusOptions
+    )
+  );
```
NWCalvank: There are so many functions inside of this component. It looks like a lot of them are or could be pure functions that could be moved out of the exported function. This would improve readability, in my opinion.

Many of these actions (closing modals, sending API requests) look like things that should be passed in as props

frontend/src/constants/API.js
```diff
@@ -38,3 +38,12 @@ export const REMOVE_MINE_EXPECTED_DOCUMENT = (expectedDocumentGuid, mineDocument
 export const MINE_MANAGER_HISTORY = (mineNo) =>
   `/parties/mines/manager-history/csv?mine_no=${mineNo}`;
 export const MINESPACE_USER = "/users/minespace";
+
+// permits
+export const PERMIT = (params) =>
+  params ? `/permits?${queryString.stringify(params)}` : "/permits";
```
NWCalvank: Setting `params` to a default value of an empty object should mean that you can just pass it in directly without the ternary

frontend/src/styles/generic/layout.scss
```diff
@@ -46,6 +46,14 @@ body {
...
 }
 
+.nowrap {
+  white-space: nowrap;
+}
+
+.flex {
+  display: flex;
+}
+
```
mmumby: This is the exact same as `.inline-flex`

the only difference being `.inline-flex` has additional browser support with the flex mixin. 

```
@mixin flex() {
  display: -webkit-box;
  display: -ms-flexbox;
  display: flex;
}
```

frontend/src/components/Forms/AddPermitForm.js
```diff
@@ -0,0 +1,196 @@
...
+  {
+    label: "Exploration",
+    value: "X",
+  },
+];
+
+const selector = formValueSelector(FORM.ADD_PERMIT);
+const { TabPane } = Tabs;
+
+const validate = (values) => {
```
james-hollinger: Maybe rename this? This is a little ambiguous as we already use validate for our forma validation (see line 99)

frontend/src/components/Forms/AddPermitForm.js
Jsyro: This dropdown generates a prefix for the permit number, which is then passed to the backend. It only exists to enforce from formatting on the input. It 'could' be on the backend, but wouldn't have any FK or data integrity/quality improvements.  Same for the permit activity type below.

frontend/src/actionCreators/permitActionCreator.js
```diff
@@ -0,0 +1,128 @@
...
+    })
+    .catch((err) => {
+      notification.error({
+        message: err.response ? err.response.data.error.message : String.ERROR,
+        duration: 10,
+      });
+      dispatch(error(reducerTypes.UPDATE_PERMIT_AMENDMENT));
+      dispatch(hideLoading());
+    });
+};
```
mmumby: I love that you broke permits out into their own Redux file structure, great job.

frontend/src/components/Forms/AddPermitForm.js
mmumby: Minus, the `Quarry` these are all the same as tenure_types that we already fetch from the backend, and have various selectors for (including dropdown format and hash format)

Two questions, are permits based on the tenures? If so, should we update that table to include `Quarry` and use the tenur_type options in this situation?

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -110,6 +111,7 @@ export class Dashboard extends Component {
       this.props.fetchMineDisturbanceOptions();
       this.props.fetchMineCommodityOptions();
       this.props.fetchPartyRelationshipTypes();
+      this.props.fetchPermitStatusOptions();
       this.props.setOptionsLoaded();
     }
```
mmumby: Did you have any annoyances when adding another staticContentActionCreator?
I'm just wondering because I did when adding `ProvinceCodes`, and I'm thinking of creating a ticket to move all of these to the highest level parent, ie `App.js`

frontend/src/constants/assets.js
```diff
@@ -17,6 +17,7 @@ export { default as BRAND_PENCIL } from "../assets/images/BrandPencil.svg";
 export { default as BRAND_DOCUMENT } from "../assets/images/BrandDocument.svg";
 export { default as RED_CLOCK } from "../assets/images/late-clock-red.svg";
 export { default as EDIT } from "../assets/images/Edit.svg";
+export { default as EDITOUTLINE } from "../assets/images/EditOutline.svg";
```
mmumby: For consistency, should be in SCREAMING_SNAKE_CASE

frontend/src/components/Forms/AddPermitAmendmentForm.js
nayr974: I actually removed this before seeing your comment, as the tabs were placeholder/wip for upload, which won't end up using tabs anyways.

frontend/src/components/Forms/AddPermitForm.js
nayr974: I actually removed this before seeing your comment, as the tabs were placeholder/wip for upload, which won't end up using tabs anyways.

frontend/src/components/Forms/AddPermitForm.js
nayr974: Correct. Unless we're moving to server side validation for form validation rules, this is the only way for a good user experience here.

frontend/src/components/Forms/AddPermitForm.js
NWCalvank: I guess my question is where did these options come from? Did they require business knowledge that isn't available in our API? Or was it a case of the frontend wanting to provide some options to the end user that has no business value?

If it's the latter then I think it's fine to leave it in the frontend. It just **looks** like something that I'd expect to see in a backend code table

frontend/src/styles/components/MinePermitTable.scss
```diff
@@ -0,0 +1,50 @@
...
+    align-items: center;
+    padding-right: 10px;
+  }
+}
+
+.add-permit-dropdown-button-icon {
+  font-size: 18px;
+  padding-left: 8px;
+  padding-right: 20px;
+}
```
mmumby: This file concerns me a little.
All of our buttons have pre-defined CSS or can use a combination of existing classes the meet the desired outcome. 
These existing classes should be used over creating custom CSS for permit-specific buttons. 
Same goes for tables.

frontend/src/components/mine/Permit/MinePermitInfo.js
nayr974: It expands to fill empty space in the flex-box.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: I didn't see an announcement on that. I believe this was already in place, but makes sense to update if that behavior has changed.

frontend/src/components/Forms/AddPermitForm.js
NWCalvank: That's fair. Do we have comparable validation for this on the backend as well?

frontend/src/components/dashboard/minesHomePage/Dashboard.js
nayr974: Justin missed that he had to make changes in two files the first time, which confused him for 20 minutes or so. We talked about this before and it makes sense to do so I totally support creating a ticket for that!

frontend/src/components/Forms/AddPermitForm.js
nayr974: As far as the back-end is concerned, the permit is a free-form string. Although we could, it feels like a smell to do string parsing and validate specific characters. This is data entry nice-ity mainly.

Storing things as separate type codes at the data level is actually an item in the backlog and a longer term goal. If or when we do that, applying API validation would make a lot of sense.

frontend/src/components/Forms/AddPermitForm.js
nayr974: These are specifically for data-entry nicety and the result is a string permit_no "CX-1234". The list was given to us, and although we could have had a table and API for this we thought the value wasn't there. As this explicit type does get discarded, I don't see a case where this would be used elsewhere.

I didn't know they were almost exactly the same list as tenure types. If permits are based on tenures, it would totally make sense to use that as the source and I would have pushed for that at the start. Let's see if we can confirm that.

If we have to start adding flags or other meta to identify which should and shouldn't appear on this drop down or that dropdown, I think this will be contained enough that the hard-coded list here makes more sense.

frontend/src/components/Forms/AddPermitForm.js
nayr974: Renamed to validateBusinessRules.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
JustinMacaulay: Yeah, as Ryan said I was confused by it for 15-20 until Ryan pointed our the two separate entry points issue. I would be all for removing the need to have this in two places.

frontend/src/constants/assets.js
nayr974: Yes, updated.

frontend/src/components/Forms/AddPermitForm.js
NWCalvank: > "This is data entry nice-ity mainly."
Fair enough. I'll mark this as resolved

frontend/src/components/mine/Permit/MinePermitTable.js
NWCalvank: My apologies. I may have only had the discussion with her on the PR. I had thought it was a Slack thread with more involvement.

frontend/src/constants/API.js
nayr974: True. Modified.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: Let me make a refactoring pass as a bunch of piecemeal changes were made by different people. MinePermitInfo is the container component here so passing some of these in as props is the right approach.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: True. Modified.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: Updated

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: Looks only to be padding. Updated to a style.

frontend/src/styles/generic/layout.scss
nayr974: Cool, removed and used inline-flex.

frontend/src/styles/components/MinePermitTable.scss
nayr974: The drop-down list component being wrapped around buttons has an effect on the rendering of the child button and it's own child parts (image, text). I had to do some specific tweaking to get it to render as one would expect.

Additionally, there are are mixed icon/images in the dropdown that have mixed spacing. They required different padding to align. They are independent components rendered by the dropdown list, so no way to have columns for alignment.

I'm happy moving the permit-table-row class to a common place and naming it generically "center-table-row" or something. There's currently no other class I can see that does this.

mmumby: Okay, Thank you for explaining! 

aaaaand Table.scss would be a good spot for all the table related css!

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: I committed a refactor that I think improves this. The open modal actions are passed in as props, and the handlers all exist on the PermitInfo level.

frontend/src/components/Forms/AddPermitAmendmentForm.js
```diff
@@ -0,0 +1,70 @@
...
+        <Form.Item>
+          <Field
+            id="issue_date"
+            name="issue_date"
+            label="Issue date"
+            component={RenderDate}
+            validate={[required, dateNotInFuture]}
+          />
+        </Form.Item>
+        {props.initialValues.permit_amendment_type_code !== "OGP" && (
```
NWCalvank: Not a blocking change, but something I did in the Contact advanced search was put a similar condition like this into a function so that the logic was a bit more explicit (via a variable name -- `isPerson`, in that case).

Reason being, it's not clear to me as a reader what "OGP" means

frontend/src/components/Forms/AddPermitForm.js
```diff
@@ -0,0 +1,183 @@
...
+          Cancel
+        </Button>
+      </Popconfirm>
+      <Button
+        className="full-mobile"
+        type="primary"
+        htmlType="submit"
+        disabled={props.submitting}
+        onClick={() => {
+          // Change tabs
```
NWCalvank: Is this WIP?

frontend/src/components/mine/MineDashboard.js
```diff
@@ -148,7 +150,7 @@ export class MineDashboard extends Component {
                 </TabPane>
                 <TabPane tab="Permit" key="permit">
                   <div className="tab__content">
-                    <MinePermitInfo mine={mine} />
+                    <MinePermitInfo mine={mine} {...this.props} />
```
NWCalvank: I know we do this all over the codebase, so it's not a blocking change, but I'd like to propose that we start passing props in explicitly as a pattern going forward. It's a team decision that requires a bit of discussion, but the existing pattern of passing in all props like this is generally considered an anti-pattern in React.

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,349 @@
...
+
+  return {
+    key: permit.permit_guid,
+    lastAmended: (latestAmendment && formatDate(latestAmendment.issue_date)) || Strings.EMPTY_FIELD,
+    permitNo: permit.permit_no || Strings.EMPTY_FIELD,
+    firstIssued: (firstAmendment && formatDate(firstAmendment.issue_date)) || Strings.EMPTY_FIELD,
+    permittee: permitteeName,
+    authorizationEndDate:
+      latestAmendment && latestAmendment.authorization_end_date
+        ? formatDate(latestAmendment.authorization_end_date)
```
NWCalvank: One remaining redundant null check

frontend/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -0,0 +1,349 @@
...
+  major_mine_ind,
+  openEditAmendmentModal
+) => ({
+  amendmentNumber,
+  amendmentType: amendment.permit_amendment_type_code,
+  receivedDate:
+    (amendment.received_date && formatDate(amendment.received_date)) || Strings.EMPTY_FIELD,
+  issueDate: (amendment.issue_date && formatDate(amendment.issue_date)) || Strings.EMPTY_FIELD,
+  authorizationEndDate:
+    (amendment.authorization_end_date && formatDate(amendment.authorization_end_date)) ||
```
NWCalvank: Looks like these null checks are redundant as well

frontend/src/constants/API.js
NWCalvank: Did that change not get pushed?

frontend/src/components/Forms/AddPermitAmendmentForm.js
nayr974: Good practice. Now reads "originalPermit".

frontend/src/components/Forms/AddPermitForm.js
nayr974: Missed removing this. Removed.

frontend/src/components/mine/MineDashboard.js
nayr974: I'm all for it and I don't like passing in this.props, but I'll address this in a future PR we are immediately following this up with more changes to these components.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: Removed. Thought the conditional was still needed for the empty string but it's not.

frontend/src/components/mine/Permit/MinePermitTable.js
nayr974: Removed. Thought the conditional was still needed for the empty string but it's not.

frontend/src/constants/API.js
nayr974: I reverted it. I think the API is behaving differently if the question mark is in the call. I can pursue this but would prefer to do it in our uploading permit documents branch where we are making other API changes.

NWCalvank: Ahh gotcha. Not a big deal. Marking this as resolved

frontend/src/components/mine/MineDashboard.js
NWCalvank: Definitely not in scope for this PR. Happy to hear we're on the same page with it. Marking as resolved and we can revisit this as we touch components going forward

PR #567, opened 2019-03-05
Bug/ Make email optional/nullable

NWCalvank:
Business has determined that email should be optional.

Changes:
- Make `party.email` nullable in db
- Remove presence validation for `party.email` in backend application layer
- Remove backend test for required email
- Make `email` an optional field on the frontend


james-hollinger:
I'm getting a sqlalchemy.exc.DataError on your branch when running the contact search.

PR #568, opened 2019-03-05
Add storage for logs to nginx

FreshGyan:
- Added a new PVC for logging called `mds-logs-pr-xxx` for dev and `mds-logs-test/prod` just so we can use same PVC for logs across different envts.
- Storage sizes are dev,test 1 gig and prod 5gigs
- nginx logs are stored at /var/logs/nginx

Sample log
```
- 172.51.20.1 -  [12/Mar/2019:00:13:58 +0000] "GET /568/api/mines/commodity-codes HTTP/1.1" 200 1548 "https://minesdigitalservices-dev.pathfinder.gov.bc.ca/568/dashboard/mines/?page=1&per_page=25" "Bearer TOKEN" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36" "24.68.227.0" "BODY_CONTENT"
```


openshift/_nginx.dc.json
```diff
@@ -181,12 +204,26 @@
                     "memory": "${MEMORY_REQUEST}"
                   }
                 },
+                "volumeMounts": [
+                  {
+                    "mountPath": "/var/log/nginx",
+                    "name": "${NAME}${SUFFIX}-log"
```
Jsyro: Will there be one log PVC for all logs (nginx, flask, and postgres)? or one for each? Just wondering about folder/PVC naming.

FreshGyan: When @itchison and I chatted, we figured twas probably better to keep a single PVC to share between environments, that only holds logs, than assigning logs to every single environment.
Mostly to keep our PVC #'s down and if we lose the PVC there is no real harm, given that filebeat will immediately send it to elastic

PR #577, opened 2019-03-07
Bug/MDS-1069 Frontend Hangs when JWT token is expired

james-hollinger:
1. Have a page of our application open long enough that your JWT token expires
2. Click on a link that posts to the backend.

The app will return a 401 error code, and the frontend will sit spinning forever as it doesn't know what to do.

This PR adds Axios interceptors to refresh the page on a 401 status code to force SiteMinder re-login. This interceptor is added globally across the app, and runs before any .then() or .catch() methods attached to the promise.

This PR also addresses a concern from MDS-1407, refreshing the page on a 503 response code (maintenance mode)


james-hollinger:
Slight update to address magic numbers

PR #578, opened 2019-03-07
MDS-1512 Feature/add contacts at a system level

mmumby:
Still a work in progress
- Add contacts at a system level on 'Contact View' (this will NOT be visible to users after merging)
- Update the add contact flow from the mine profile on Contact Tab. (This WILL BE VISIBLE to users after merging)


frontend/src/actionCreators/staticContentActionCreator.js
```diff
@@ -11,144 +10,140 @@ import { createRequestHeader } from "@/utils/RequestHeaders";
...
+  dispatch(request(reducerTypes.GET_PROVINCE_CODES));
+  return axios
+    .get(ENVIRONMENT.apiUrl + API.PROVINCE_CODES, createRequestHeader())
+    .then((response) => {
+      dispatch(success(reducerTypes.GET_PROVINCE_CODES));
+      dispatch(staticContentActions.storeProvinceCodes(response.data));
+    })
+    .catch((err) => {
+      notification.error({
+        message: "this is causing an error",
```
NWCalvank: Is this our desired error message?

frontend/src/actionCreators/staticContentActionCreator.js
```diff
@@ -11,144 +10,140 @@ import { createRequestHeader } from "@/utils/RequestHeaders";
 
 export const fetchMineDisturbanceOptions = () => (dispatch) => {
   dispatch(request(reducerTypes.GET_DISTURBANCE_OPTIONS));
-  dispatch(showLoading("modal"));
```
mmumby: Removed the loading state on all StaticContent actionCreators as they are being fetched when a parent loads, and not in a modal, therefore this loading state is never seen.

frontend/src/components/modalContent/AddPartyModal.js
```diff
@@ -0,0 +1,182 @@
...
+      });
+  };
+
+  next() {
+    if (!this.props.addPartyForm.syncErrors) {
+      const current = this.state.current + 1;
+      this.setState({ current });
+    } else {
+      // submit form to trigger validation errors.... alternate suggestions??
+      this.props.submit(FORM.ADD_FULL_PARTY);
```
NWCalvank: If the validation errors are client-side then you should be able to update the validation errors to show up when the field is dirty (maybe with a 1-3 second delay?)

frontend/src/components/modalContent/AddPartyModal.js
```diff
@@ -0,0 +1,182 @@
...
+      />
+    );
+  }
+
+  // WIP
+  renderStepTwo() {
+    return (
+      <div className="center">
+        <p>
+          If you would like to add another contact, click on the add another contact below. Your
```
NWCalvank: `click on the add another contact ____ below`. There should probably be something in that blank. Looks like `button` might make sense

frontend/src/actionCreators/staticContentActionCreator.js
mmumby: no, absolutely not. Forgot to change it, Good catch! 🤦🏼‍♀️

frontend/src/components/modalContent/AddPartyModal.js
mmumby: Usually, yes. 
The way that redux-forms works, if it the form submission fails it shows all validations errors, but this form is bypassing the built-in features and submitting it from an outside component, instead of having the submit function connected to the form itself. (where all the built-in magic happens)

In this situation, I was trying to force the submission to fail to get all the form benefit
Redux-forms has a whole host of functions I could use, but none of them seemed to force the display of the errors, except trying to submit the invalid form and letting redux-forms do its thing. 
(https://redux-form.com/6.4.3/docs/api/reduxform.md/)

We can take this offline If I didn't explain it well enough!

frontend/src/components/modalContent/AddPartyModal.js
mmumby: fixed!

PR #579, opened 2019-03-08
MDS-1528 Bug/ Changes to Received Date and Status Behaviours

NWCalvank:
Original AC:
- When users upload a file, the Status should be changed to Received and the Received column should show the date the file was received.
- When a file is deleted, and there is only 1 file for the requirement, the Status will reset to Not Received and there should be no date in the Received column.
- When multiple files are deleted for one requirement but there is still at least one linked file, the Status should not change and the Received date should also not change.
- When multiple files are deleted for one requirement and there a no linked files remaining, Status resets to Not Received and Received date should be cleared.
- The reported date is the one for the first submission.


python-backend/app/api/documents/expected/resources/expected_document_uploads.py
```diff
@@ -98,8 +99,17 @@ def put(self, expected_document_guid):
             db.session.commit()
         else:
             return self.create_error_payload(400, 'Must specify either Mine Document GIUD or Docuemnt Manager GUID'), 400
- 
-        return expected_document.json()           
+
+        try:
+            expected_document.exp_document_status_code = 'PRE'
```
james-hollinger: This endpoint is used by both minespace and CORE. This will set the status back to PRE whenever an inspector or proponent uploads a file, regardless of what the status was before.

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
```diff
@@ -98,8 +99,17 @@ def put(self, expected_document_guid):
...
+            # set received_date to first upload date
+            if len(expected_document.mine_documents) == 1:
+                expected_document.received_date = datetime.now()
+            expected_document.save()
+        except AssertionError as e:
+            return self.create_error_payload(500, 'Unable to update document status'), 500
+
+        return expected_document.json()
```
james-hollinger: It was desired for only minespace to automatically set received date and status.

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
james-hollinger: It was desired for only minespace to automatically set received date and status.

NWCalvank: So, you're saying that Minespace and CORE need to contain their own version of the business logic?

My impression was that any upload would need to be under review. If this isn't correct, I'll need a bit more clarity on how the described behaviour fits with the existing behaviour in CORE.

frontend-public/src/components/dashboard/mineInfo/MineInfo.js
```diff
@@ -41,8 +42,30 @@ export class MineInfo extends Component {
...
+    // Set status to received/pending review
+    updatedDocument.exp_document_status.exp_document_status_code = "PRE";
+
+    // Only set received_date when going from 0 to 1 uploaded documents
+    if (
+      updatedDocument.related_documents.length === 0 &&
+      updatedMineDocument.related_documents.length === 1
+    ) {
+      updatedDocument.received_date = moment();
+    }
```
james-hollinger: This fails to set the date if you upload two files and hit submit.

NWCalvank: Ah... I missed that use-case. I'll rethink this solution

frontend-public/src/components/Forms/EditReportForm.js
```diff
@@ -49,7 +48,10 @@ export const EditReportForm = (props) => (
       <Popconfirm
         placement="topRight"
         title="Are you sure?"
-        onConfirm={props.closeModal}
+        /* Cancelling here is too late because all serverside actions have been
+         * taken. Calling handleSubmit will update the props with latest data,
+         * handle document status/received_date, and close the modal */
+        onConfirm={props.handleSubmit}
```
james-hollinger: I think it might be time to remove the cancel button, since its not really cancelling anything at this point.
Also, you get a toast message saying that you "Successfully updated expected document" when you hit cancel, which is kinda not what you would expect to see when you hit a cancel button.

NWCalvank: I agree. It's pointless. If there's no requirement for it to exist then I'm happy to remove it

frontend-public/src/components/dashboard/mineInfo/MineInfo.js
NWCalvank: Updated the logic to handle multiple documents uploaded before a submission

PR #581, opened 2019-03-11
Bug/responsive menu

mmumby:
Add responsive navigation to CORE.


frontend/src/components/navigation/NavBar.js
```diff
@@ -1,125 +1,283 @@
...
+            >
+              <img
+                alt="Mine"
+                className="padding-small--right"
+                style={{ verticalAlign: "-0.125em" }}
+                src={MINE}
+              />
+              Mines
+            </Button>
+          </Link>
```
NWCalvank: I don't know if this would be better, but I'll suggest it since you mentioned the duplication issue. You could create a pure function that returns this JSX. Something like:
```javasctipt
const menuButton = (activeButton, { link, button, image}) => (
<Link
            to={link.to}
          >
            <Button
              id={includes(activeButton, button.route) ? "active-mine-btn" : ""}
              className={button.className || ""}
            >
              <img
                alt={image.alt}
                className={image.className || ""}
                style={image.style}
                src={image.src}
              />
              {image.label}
            </Button>
          </Link>
)
```
and call it like:
```javascript
<AuthorizationWrapper permission={Permission.ADMIN}>
{menuButton(this.props.activeButton, {
  link: {to: router.ADMIN_DASHBOARD.route},
  button: {route: router.CONTACT_HOME_PAGE.route},
  image: {alt: "team", className: "padding-small--right icon-sm", src: TEAM, style: { verticalAlign: "-0.125em" }, label: "Admin"}
})}
// other stuff
</AuthorizationWrapper>
```

mmumby: Great input. Thank you Nathan! 

Will keep this abstraction in mind when the navigation grows larger than it currently is.

PR #582, opened 2019-03-12
MDS-1496 capture initial applications

JustinMacaulay:
- PUT, POST, and GET for applications.
 - Migration to add application tables.
 - Integration tests for the application endpoints.


python-backend/app/api/applications/resources/apllication.py
```diff
@@ -0,0 +1,91 @@
...
+    parser = reqparse.RequestParser()
+    parser.add_argument('application_no', type=str, help='Number of the application being added.')
+    parser.add_argument('mine_guid', type=str, help='guid of the mine.')
+    parser.add_argument(
+        'application_status_code',
+        type=str,
+        help='Status of the application being added.',
+        store_missing=False)
+    parser.add_argument(
+        'received_date', type=lambda x: datetime.strptime(x, '%Y-%m-%d') if x else None)
```
Jsyro: should this have a store_missing flag

python-backend/app/api/applications/resources/apllication.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @api.doc(params={'application_guid': 'Application guid.'})
+    @requires_role_mine_view
+    def get(self, application_guid=None):
+
+        mine_guid = request.args.get('mine_guid', None, type=str)
+        if mine_guid:
+            applications = Application.find_by_mine_guid(mine_guid)
+            result = [app.json() for app in applications]
+        elif application_guid:
```
Jsyro: I've done this reversed, if someone provides a guid in the url, that should supersede any query string parameters.

python-backend/app/api/applications/resources/apllication.py
```diff
@@ -0,0 +1,91 @@
...
+        mine = Mine.find_by_mine_guid(data.get('mine_guid'))
+
+        if not mine:
+            return self.create_error_payload(
+                404, 'There was no mine found with the provided mine_guid.'), 404
+
+        try:
+            application = Application.create(mine.mine_guid, data.get('application_no'),
+                                             data.get('application_status_code'),
+                                             data.get('received_date'), data.get('description'),
```
Jsyro: if these are optional values for creating an application, I think the create method should have them as optional parameters as well.

python-backend/tests/applications/resources/test_application_resource.py
```diff
@@ -0,0 +1,133 @@
...
+        '/applications/' + setup_info.get('mine_1_application_guid'),
+        headers=auth_headers['full_auth_header'],
+        data=data)
+    assert post_resp.status_code == 400
+
+
+#Put
+def test_put_application(test_client, setup_info, auth_headers):
+    application_guid = setup_info.get('mine_1_application_guid')
+    application = Application.find_by_application_guid(application_guid)
```
Jsyro: Is there a reason not to yield the application object (instead of just it's guid. I have been doing that in other tests.

python-backend/app/api/applications/resources/apllication.py
JustinMacaulay: No, The only time that is entered and used is on the creation of the application.

python-backend/app/api/applications/resources/apllication.py
JustinMacaulay: True, I'll swap that around.

python-backend/tests/applications/resources/test_application_resource.py
JustinMacaulay: I did it this way because the object can change from when it was created. I know its not very likely the guid would change but just in case id rather grab a instance of the object.

python-backend/app/api/applications/resources/apllication.py
JustinMacaulay: Changed all of these to the dictionary access without get as they are all required.

python-backend/app/api/permits/permit/resources/permit.py
```diff
@@ -6,8 +6,8 @@
 from ...permit_amendment.models.permit_amendment import PermitAmendment
 from ....mines.mine.models.mine import Mine
 from app.extensions import api
-from ....utils.access_decorators import requires_role_mine_view, requires_role_mine_create
-from ....utils.resources_mixins import UserMixin, ErrorMixin
+from app.api.utils.access_decorators import requires_role_mine_view, requires_role_mine_create
```
nayr974: Oh good. Let's move away from relative paths. 👍

migrations/sql/V2019.03.07.11.16__add_application_tables.sql
```diff
@@ -0,0 +1,35 @@
...
+
+CREATE TABLE application
+(
+    application_id serial PRIMARY KEY,
+    application_guid uuid DEFAULT gen_random_uuid() NOT NULL,
+    mine_guid uuid NOT NULL,
+    application_no character varying(16) NOT NULL,
+    description character varying(280) NOT NULL,
+    application_status_code character varying(3),
+    recieved_date date DEFAULT '9999-12-31'::date NOT NULL,
```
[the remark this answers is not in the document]
nayr974: Or if it's truly not allowed to be null, no default required.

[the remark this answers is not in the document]
JustinMacaulay: Yes, I grabbed this from another migration I will remove it now.

python-backend/app/api/applications/models/application_status_code.py
```diff
@@ -0,0 +1,37 @@
...
+            'application_status_code': self.application_status_code,
+            'description': self.description,
+        }
+
+    @classmethod
+    def find_by_application_status_code(cls, code):
+        return cls.query.filter_by(application_status_code=code).first()
+
+    @classmethod
+    def find_all_application_status_code(cls):
```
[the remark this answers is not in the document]
JustinMacaulay: I have changed the name as I think its better and there is probably a situation in the future where you want all not just active.

PR #585, opened 2019-03-12
Task/mds 1533 require document category refactor

Jsyro:
Preparation for more powerful behaviour regarding required documents, now we use a type code for the required document category. And updated some endpoints to use established patterns.


python-backend/app/api/mines/tailings/resources/tailings.py
```diff
@@ -56,7 +56,8 @@ def post(self, mine_tailings_storage_facility_guid=None):
             db.session.add(mine_tsf)
             if is_mine_first_tsf:
                 try:
-                    req_documents_url = get_documents_svc_url('/required?category=MINE_TAILINGS')
+                    req_documents_url = get_documents_svc_url('/required?category=TSF')
+                    current_app.logger.debug(req_documents_url)
```
NWCalvank: Is this output for monitoring? Or can this be removed?

Jsyro: Removed, good catch.

PR #588, opened 2019-03-13
Feature/mds 1021

thegentlemanphysicist:
I've fixed the order and grouping of layers in the map widget drop down.  This modifies the order in which layers are displayed on the map as well as well as how they are grouped in the layers widget.


frontend/src/components/maps/MineMap.js
```diff
@@ -75,6 +75,7 @@ class MineMap extends Component {
 
   /**
    * Adds widgets to a given MapView instance
+   * MapImageLayer is required for the collection operation add to work.
```
NWCalvank: It looks like this is outdated

frontend/src/components/maps/MineMap.js
```diff
@@ -83,7 +84,75 @@ class MineMap extends Component {
...
+        if (newTenureLayer) {
+          tenureLayer.add(newTenureLayer);
+        }
+      });
+      roadLayerArray.forEach((layerTitle) => {
+        const newTenureLayer = view.map.allLayers.find(({ title }) => title === layerTitle);
+        if (newTenureLayer) {
+          roadsLayer.add(newTenureLayer);
+        }
+      });
```
NWCalvank: This is looking a lot better than the WIP code we had yesterday!

Since these all do the exact same thing, how about creating a function to handle it? Something like:
```javascript
const addLayers = (layerArray) =>
    layerArray.forEach((layerTitle) => {
        const newTenureLayer = view.map.allLayers.find(({ title }) => title === layerTitle);
        if (newTenureLayer) {
          roadsLayer.add(newTenureLayer);
        }
      });

addLayers(adminLayerArray);
addLayers(tenureLayerArray);
addLayers(roadLayerArray);
```

frontend/src/components/maps/MinePin.js
```diff
@@ -58,7 +58,7 @@ export class MinePin extends Component {
           this.setState({ isFullMap: false });
         } else {
           this.setState({ isFullMap: true });
-          mineIds = this.props.mineIds;
+          mineIds = [...this.props.mineIds];
```
NWCalvank: Does something break if this change is reverted? It looks like this should be logically equivalent to `mineIds = this.props.mineIds`

thegentlemanphysicist: This was a linting error :P

frontend/src/components/maps/MineMap.js
thegentlemanphysicist: done

frontend/src/components/maps/MineMap.js
thegentlemanphysicist: done

PR #595, opened 2019-03-14
Feature/mds 1462 change tsf reports list

Jsyro:
Added infrastructure for sub-typing required documents. using it to have a default list of tsf reports and allowing CORE to add more for an expanded list.


migrations/sql/V2019.03.12.15.29__add_required_document_sub_categorization.sql
```diff
@@ -0,0 +1,37 @@
+COMMENT ON TABLE mine_required_document_category IS 'Business context that the required document is used in, type code is also used by document manager for folder name';
+
+CREATE TABLE required_document_sub_category(
+    req_document_sub_category_code character varying(3) PRIMARY KEY, 
+    description character varying(200)
+);
+ALTER TABLE required_document_sub_category OWNER TO mds;
+COMMENT ON TABLE required_document_sub_category IS 'This code can be used when additional behaviour is needed for required reports';
```
klefler: Ya, so for this table comment, maybe just throw an example that states something like "INI would be used as a flag to determine default documents... etc."

Jsyro: Added example.

PR #596, opened 2019-03-14
Feature/mds 1496 front end to display applications

JustinMacaulay:
- Adds a new tab to the mine view.
- Allows for the creation and updating of applications.
- Tests added and the appropriate snapshots were updated.


frontend/src/actionCreators/applicationActionCreator.js
```diff
@@ -0,0 +1,85 @@
...
+import { request, success, error } from "@/actions/genericActions";
+import * as reducerTypes from "@/constants/reducerTypes";
+import * as applicationActions from "@/actions/applicationActions";
+import * as String from "@/constants/strings";
+import * as API from "@/constants/API";
+import { ENVIRONMENT } from "@/constants/environment";
+import { createRequestHeader } from "@/utils/RequestHeaders";
+
+// This file is anticipated to have multiple exports
+// eslint-disable-next-line import/prefer-default-export
```
mmumby: This comment can be removed.

frontend/src/components/mine/Applications/MineApplicationInfo.js
```diff
@@ -0,0 +1,138 @@
...
+                </Button>
+              </AuthorizationWrapper>
+            </AuthorizationWrapper>
+          </div>
+        </div>
+      </div>,
+      <br />,
+      <MineApplicationTable
+        applications={this.props.applications}
+        major_mine_ind={this.props.mine.major_mine_ind}
```
mmumby: should be in camelCase. 
`isMajorMine` is used in other places throughout the app for this boolean.

frontend/src/components/mine/Applications/MineApplicationTable.js
```diff
@@ -0,0 +1,137 @@
...
+import * as Permission from "@/constants/permissions";
+import { getDropdownApplicationStatusOptions } from "@/selectors/staticContentSelectors";
+import CustomPropTypes from "@/customPropTypes";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import { EDIT_OUTLINE } from "@/constants/assets";
+
+/**
+ * @class  MinePermitTable - displays a table of permits and permit amendments
+ */
```
mmumby: update comment to correct Class

frontend/src/components/mine/Applications/MineApplicationTable.js
```diff
@@ -0,0 +1,137 @@
...
+      props.openEditApplicationModal
+    )
+  );
+
+  return (
+    <Table
+      className="nested-table"
+      rowClassName={() => "table-row-align-middle"}
+      align="left"
+      pagination={false}
```
mmumby: className `table-role-align-middle` and the next line is `align="left"`

Are these targeting the same alignment?

frontend/src/components/common/wrappers/AuthorizationWrapper.js
```diff
@@ -43,6 +43,7 @@ const propTypes = {
   isMajorMine: PropTypes.bool,
   inDevelopment: PropTypes.bool,
   inTesting: PropTypes.bool,
+  tab: PropTypes.bool,
   children: PropTypes.element.isRequired,
```
mmumby: Are you doing anything with this tab prop? or forgot to remove?

frontend/src/components/mine/Applications/MineApplicationInfo.js
```diff
@@ -0,0 +1,138 @@
...
+import { modalConfig } from "@/components/modalContent/config";
+import {
+  fetchApplications,
+  updateApplication,
+  createApplication,
+} from "@/actionCreators/applicationActionCreator";
+
+/**
+ * @class  MinePermitInfo - contains all permit information
+ */
```
mmumby: Update this comment too

frontend/src/actionCreators/applicationActionCreator.js
JustinMacaulay: Removed.

frontend/src/components/mine/Applications/MineApplicationTable.js
JustinMacaulay: Changed.

frontend/src/components/mine/Applications/MineApplicationTable.js
JustinMacaulay: Isn't one of those applying to the table and one is applying to the row in the table?

frontend/src/components/common/wrappers/AuthorizationWrapper.js
JustinMacaulay: Yeah, good catch. I have removed it.

frontend/src/components/mine/Applications/MineApplicationInfo.js
JustinMacaulay: Changed

frontend/src/components/mine/Applications/MineApplicationInfo.js
JustinMacaulay: Changed.

frontend/src/components/mine/Applications/MineApplicationTable.js
mmumby: I'm not sure on that, never used the `rowClassName` before.
Is this table supposed to look different than the others? If it's supposed to be the same, Id removes the `rowClassName`, and use the `align="left"`

If not, disregard this comment.

NWCalvank: I can't speak to how it's being used in this context, but I introduced `rowClassName` to the Permit Table when dealing with nested tables. Because each 'row' was a table, it needed custom classes to deal with on-hover behaviour.

If there's nothing special about this table then `rowClassName` is _likely_ not needed

JustinMacaulay: I brought this over from the permit table I was just trying to keep them the same, Im not super sure what this is doing. I feel like its applying the class to the row content but I could be wrong.

JustinMacaulay: ahhhhh, that makes sense. I'll remove it.

python-backend/app/api/applications/resources/application_status_code.py
```diff
@@ -0,0 +1,14 @@
...
+from app.extensions import api
+from ..models.application_status_code import ApplicationStatusCode
+from app.api.utils.resources_mixins import UserMixin, ErrorMixin
+from app.api.utils.access_decorators import requires_role_mine_view
+
+
+class ApplicationStatusCodeResource(Resource, UserMixin, ErrorMixin):
+    @requires_role_mine_view
+    def get(self):
+        return [x.json() for x in ApplicationStatusCode.find_all_active_application_status_code()]
```
NWCalvank: `ApplicationStatusCode.find_all_active_application_status_code()` can return `None`

From what I can tell, `None.json()` will throw an exception. You should probably update the method to return an empty list instead of `None`, or handled it here with a null check

frontend/src/components/mine/Applications/MineApplicationTable.js
mmumby: Thanks!

frontend/src/components/mine/MineDashboard.js
```diff
@@ -148,6 +155,13 @@ export class MineDashboard extends Component {
                     />
                   </div>
                 </TabPane>
+                {mine.major_mine_ind && (
+                  <TabPane tab="Applications" key="applications">
+                    <div className="tab__content">
+                      <MineApplicationInfo mine={mine} {...this.props} />
```
NWCalvank: As per a brief thread with @nayr974 , we'd like to be more explicit with the props that are being passed in and move away from the blind passing of all `props`

frontend/src/components/mine/Applications/MineApplicationTable.js
```diff
@@ -0,0 +1,137 @@
...
+    key: "applicationNo",
+    render: (text) => <div title="Application No.">{text}</div>,
+  },
+  {
+    title: "Status",
+    dataIndex: "status",
+    key: "status",
+    render: (text, record) => (
+      <div title="Status">
+        {record.applicationStatusOptions.find((item) => item.value === text).label}
```
NWCalvank: If this doesn't find a match, it will return `undefined`. `undefined.label` will throw a runtime error. We should add some safety around that

frontend/src/components/mine/Applications/MineApplicationTable.js
nayr974: It's likely not needed here. I used this as it was a clean way to align buttons consisting of mixed parts (Icon, button, svg image) that were also taller than the default row height causing weird spacing on the permit table. The alternative looked like a bunch of padding/margin overrides. We'll require conditional classes on the permit table to highlight amalgamated permits, so I thought using the rowClassName hook made sense because of that too.

frontend/src/components/mine/MineDashboard.js
nayr974: Agree here. Let's avoid new uses of {...this.props} and start refactoring existing uses.

python-backend/app/api/applications/resources/application_status_code.py
JustinMacaulay: Yes, I didn't think about the value error section. I have modified this to check and return an empty list if there is a value error.

frontend/src/components/mine/MineDashboard.js
JustinMacaulay: Done.

frontend/src/components/mine/Applications/MineApplicationTable.js
JustinMacaulay: I agree, at the moment however it is not a pattern we use and the cases in which this could happen are: 

1. For some reason the application status options have not been loaded. 
2. Somehow an application was added to our system with an invalid foreign key.

I can add a check in if you'd like, but we should probably take a look at all of our find() calls on the front end to see if this is a needed change.

JustinMacaulay: Changed.

PR #598, opened 2019-03-15
MDS-1437 Feature/Edit party information

mmumby:
Users have the ability to edit contact information of parties.

- Added a reusable Address.js component to handle address formatting - as users can enter 1 or all fields, Address.js will always display a cleanly formatted address regardless of the amount of data.
- Removed backend validations surrounding unique names.


frontend/src/components/common/Address.js
```diff
@@ -0,0 +1,81 @@
...
+export class Address extends Component {
+  addressContainsTruthyValues = (object) =>
+    // address_type_code is always "CAD" && if "suit_no" exists but nothing else, this field isn't helpful
+    Object.keys(object)
+      .filter((key) => key !== "address_type_code" && key !== "suite_no")
+      .some((key) => object[key]);
+
+  formatRowContent = (itemArr) =>
+    itemArr
+      .filter((item) => item)
```
NWCalvank: What datatype is each `item` here? Are they all strings?

If we know they'll be strings, I would like to see a type-specific check here (`item !== ""`). This is more self-documenting for this fairly abstract function.

frontend/src/components/common/Address.js
```diff
@@ -0,0 +1,81 @@
...
+      .map((item, i) => (
+        <span key={item}>{i !== itemArr.length - 1 ? `${item}${" "}` : item}</span>
```
NWCalvank: Did Prettier automatically format the whitespace like that? It seems like the `${" "}` should be able to be replaced with normal whitespace

frontend/src/components/common/Address.js
```diff
@@ -0,0 +1,81 @@
...
+
+  renderNextLine = (itemArr) => (
+    <div className="padding-large--left">
+      <p>{this.formatRowContent(itemArr)}</p>
+    </div>
+  );
+
+  // since nothing in the address object is required, users can fill out 1 or more fields.
+  // the following check tests all the edge cases, and formats the address into a 1-3 line section, with an Icon on line 1.
+  renderAddress = (address) => (
```
NWCalvank: If you have time, this would be a method to add regression tests for if possible. It's a pure function, so you can easily cover all of the edge-cases based on just a mock `address` object and then compare the actual output with the expected output.

It could also help the next person who touches this component considerably

frontend/src/components/modalContent/EditPartyModal.js
```diff
@@ -0,0 +1,22 @@
...
+  title: PropTypes.string,
+};
+
+const defaultProps = {
+  title: "",
+};
+
+export const EditPartyModal = (props) => (
+  <div>
+    <EditFullPartyForm {...props} />
```
NWCalvank: Pass props explicitly here

frontend/src/components/parties/PartyProfile.js
```diff
@@ -63,6 +79,46 @@ export class PartyProfile extends Component {
...
 
+  openEditPartyModal = (event, party, onSubmit, title, isPerson, provinceOptions) => {
+    const initialValues = {
+      email: party.email && party.email !== "Unknown" ? party.email : null,
+      first_name: party.first_name ? party.first_name : null,
+      party_name: party.party_name ? party.party_name : null,
+      phone_ext: party.phone_ext ? party.phone_ext : null,
+      phone_no: party.phone_no ? party.phone_no : null,
+      address_line_1:
+        party.address[0] && party.address[0].address_line_1
```
NWCalvank: You could pull `party.address[0]` out into a variable above `initialValues` to avoid repeating that every time.

Or... it looks like every address property includes a check for `party.address[0]`. A lot of this duplication could be removed with something like:
```javascript
// If the explicit nulls aren't needed:
const addressValues = party.address[0] || {}

// OR If the explicit nulls are needed:
const partyAddress = party.address[0];
const addressValues = !partyAddress ? {} : {
  address_line_1: partyAddress.address_line_1 || null,
  // the rest
}

// Regardless...
const initialValues = {
  email: party.email && party.email !== "Unknown" ? party.email : null,
  // other stuff
  ...addressValues
};
```

frontend/src/utils/helpers.js
```diff
@@ -49,6 +49,8 @@ export const formatDate = (dateString) =>
   dateString !== "None" &&
   moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");
 
+export const formatPostalCode = (code) => code && code.replace(/.{3}$/, " $&");
```
NWCalvank: Why is this formatting needed?

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -1,4 +1,5 @@
 import uuid
+import logging
```
NWCalvank: Logging that can be removed :)

frontend/src/components/common/Address.js
```diff
@@ -0,0 +1,81 @@
...
+  address: PropTypes.objectOf(PropTypes.string).isRequired,
+};
+
+/**
+ * @constant Address- reusable component that formats any given address object
+ */
+
+export class Address extends Component {
+  addressContainsTruthyValues = (object) =>
+    // address_type_code is always "CAD" && if "suit_no" exists but nothing else, this field isn't helpful
```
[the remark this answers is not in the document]
mmumby: changed.

frontend/src/components/common/Address.js
mmumby: The data will either be a string or null, never an empty string.

frontend/src/utils/helpers.js
mmumby: The backend expects only 6 ch - `V6V4H4`
The UI would like to display a space between `V6V 4H4`

python-backend/app/api/parties/party/resources/party_resource.py
mmumby: removed

frontend/src/components/common/Address.js
mmumby: Great suggestion! will work on that this afternoon.

frontend/src/components/modalContent/EditPartyModal.js
mmumby: fixed.

frontend/src/components/common/Address.js
mmumby: yes it did, and fixed!

frontend/src/components/parties/PartyProfile.js
mmumby: Updated. Looks like party.address[0] is always returned. Removed that check from all the address fields.

frontend/src/components/common/Address.js
NWCalvank: Okay. So, should we strictly check against `null`? I'm just leery of implicit type coercion because it can be a bit of a footgun. When we've done this filter elsewhere, it's (mostly) been in a very specific context, where it's easy to tell what the value is/would be as it passes through the filter.

frontend/src/components/parties/PartyProfile.js
```diff
@@ -63,6 +79,38 @@ export class PartyProfile extends Component {
...
+      first_name: party.first_name ? party.first_name : null,
+      party_name: party.party_name ? party.party_name : null,
+      phone_ext: party.phone_ext ? party.phone_ext : null,
+      phone_no: party.phone_no ? party.phone_no : null,
+      address_line_1: address.address_line_1 ? address.address_line_1 : null,
+      address_line_2: address.address_line_2 ? address.address_line_2 : null,
+      city: address.city ? address.city : null,
+      post_code: address.post_code ? address.post_code : null,
+      sub_division_code: address.sub_division_code ? address.sub_division_code : null,
+      suite_no: address.suite_no ? address.suite_no : null,
```
NWCalvank: These values are coming from the API, so the JSON payload should have them set to `null` if they're not present (ie. already failing the truthy ternary checks here). Is there a reason why we need to use the ternary at all?

I get the impression that (for example) `suite_no` would be `null`. It would fail the ternary check, and the property would be set to `null`. The same thing would result from `suite_no: address.suite_no`.

frontend/src/components/common/Address.js
NWCalvank: Marking this as resolved, since we've discussed that you may do this separately

mmumby:
Closing this PR to re-start the db and all that.

NWCalvank:
Closing to attempt a full clean up. Openshift is confused and is convinced that the PVC is being deleted, but it can't seem to delete it -- causing a failed deployment.

NWCalvank:
Hopefully this will trigger a successful rebuild/deploy

PR #600, opened 2019-03-18
ETL nightly job issue.

JustinMacaulay:
- fixed an issue where the ETL would be run on all available pods.
- fixed some of the ETL where some index's were being created multiple times.


python-backend/app/scheduled_jobs/ETL_jobs.py
```diff
@@ -10,11 +11,14 @@ def _schedule_ETL_jobs(app):
...
-        db.session.execute('commit;')
-        db.session.execute('select transfer_mine_manager_information();')
-        db.session.execute('commit;')
-        db.session.execute('select transfer_permit_permitee_information();')
-        db.session.execute('commit;')
-        db.session.execute('select transfer_mine_status_information();')
-        db.session.execute('commit;')
+        job_running = cache.get(ETL())
+        if not job_running:
+            cache.set(ETL, 'True', timeout=TIMEOUT_24_HOURS)
```
Jsyro: Is there any concern about a race condition where they both read false before one has a chance to write true? What would the impact be if this ran side by side?

Jsyro: I don't think there is any graceful way of handling this race condition with one boolean. This race condition causes an error to be raised, but does not affect data, as per discussion with @JustinMacaulay. I think it's fine to go in as is.

Jsyro: Took a look at possible solutions but can only come up with a partial solution. This will be addressed if/when we create a pod that handles all scheduled jobs. I didn't want to put in a partial solution that looks good but doesn't work.

PR #602, opened 2019-03-18
Bug/MDS-1353 Remove double toast messages on Minespace login

james-hollinger:
In the authorization wrapper for Minspace, the component checks to see if you have a JWT token and makes a getUserInfo call if you do. However if the token is invalid (such as if it's expired if you return to the site after 24hrs), an error is returned [toast message #1], and the catch handler for getUserInfo logs the user out [toast message #2].

This PR changes the auth flow slightly to handle this case, and expands error handling slightly for other failure cases.

 - getUserInfo now accepts an error message, or re-throws errors if none are supplied. 
- unAuthenticateUser accepts a toast message, and only shows the toast if supplied
- authenticateUser now unAuthenticates if it catches an error.


frontend-public/src/tests/actionCreators/authenticationActionCreator.spec.js
```diff
@@ -39,7 +39,7 @@ describe("`getUserInfoFromToken` action creator", () => {
 
   it("Request failure, dispatches `error` with correct response", () => {
     mockAxios.onGet(url, MOCK.createMockHeader()).reply(400, MOCK.ERROR);
-    return getUserInfoFromToken()(dispatch).then(() => {
+    return getUserInfoFromToken("A Token", "An Error Message")(dispatch).then(() => {
```
nayr974: "A Verbose String"

frontend-public/src/actionCreators/authenticationActionCreator.js
```diff
@@ -44,19 +50,20 @@ export const authenticateUser = (code) => (dispatch) => {
     redirect_uri: ENV.BCEID_LOGIN_REDIRECT_URI,
     client_id: ENV.KEYCLOAK.clientId,
   };
-  dispatch(request(reducerTypes.AUTHENTICATE_USERR));
+  dispatch(request(reducerTypes.AUTHENTICATE_USER));
```
nayr974: THIS_IS_GOODERR

frontend-public/src/HOC/AuthenticationGuard.js
```diff
@@ -15,7 +15,9 @@ export const AuthenticationGuard = (isPublic) => (WrappedComponent) => {
     componentDidMount() {
       const token = localStorage.getItem("jwt");
       if (token && !this.props.isAuthenticated) {
-        this.props.getUserInfoFromToken(token);
+        this.props.getUserInfoFromToken(token).catch(() => {
+          // Silently fail
```
nayr974: I like to create a const doNothing = () => {}; as it makes itself describing without comments.

frontend-public/src/actionCreators/authenticationActionCreator.js
james-hollinger: US  IS  YOU
ER

frontend-public/src/HOC/AuthenticationGuard.js
james-hollinger: Fair, but seems a waste if its only used once. Maybe we should add it to constants.js?

NWCalvank: Lodash provides a `noop` function 🤷‍♂️

Also, we use `() => {}` all the time as a default in `defaultProps`, so we'll definitely be able to use this constant or lodash function many times

PR #603, opened 2019-03-18
MDS-1438 Feature/ Add 0, 1, or n roles to a new party

NWCalvank:
Features:
- Support adding >= 0 roles when creating a new party
- Add validation to user inputs to help ensure a positive user experience
- Properly clean-up Redux store after submission or cancellation

I recommend some manual testing during the review process to ensure that I haven't overlooked some typical use case in the validation/user flow.


frontend/src/components/Forms/parties/AddRolesForm.js
```diff
@@ -0,0 +1,126 @@
...
+              </Col>
+              <Col span={12}>
+                <Field
+                  label="End Date *"
+                  id={`end_date-${roleNumber}`}
+                  name={`end_date-${roleNumber}`}
+                  placeholder="yyyy-mm-dd"
+                  component={renderConfig.DATE}
+                  validate={[required]}
+                />
```
mmumby: Are `start_date` and `end_date` required?

frontend/src/components/Forms/parties/AddRolesForm.js
```diff
@@ -0,0 +1,126 @@
...
+                <Form.Item label="Mine *">
+                  <Field
+                    id={`mine_guid-${roleNumber}`}
+                    name={`mine_guid-${roleNumber}`}
+                    component={renderConfig.AUTOCOMPLETE}
+                    placeholder="Please add Mine"
+                    data={transformMineNames(props.mineNameList)}
+                    handleChange={props.handleChange(roleNumber)}
+                    handleSelect={props.handleSelect(roleNumber)}
+                    iconColor="rgba(0, 0, 0, 0.25)"
```
mmumby: Is there a reason why this is rgba and not hex? 
Also can this be stored in a variable with the rest of the colors?

frontend/src/components/Forms/parties/AddRolesForm.js
NWCalvank: I don't know if they're supposed to be. I had made them required on Friday for whatever reason. If that's inconsistent then I will gladly remove them from the validation

frontend/src/components/Forms/parties/AddRolesForm.js
NWCalvank: 1) I just pulled it directly from Ant Design (it's used to match their icon colour in an input field). I can likely get it to be a HEX -- not an expert on how that works with transparency, though.
2) Sure!

frontend/src/constants/styles.js
```diff
@@ -3,5 +3,6 @@
 export const COLOR = {
   violet: "#5e46a1",
   errorRed: "#D40D0D",
-  mediumGray: "#6b6363",
+  mediumGrey: "#6b6363",
+  layoutGrey: "e8e8e8",
```
mmumby: does this work without #?

NWCalvank: That's definitely a typo

thegentlemanphysicist:
Manual testing notes:
Not sure if this is part of this ticket's scope, but if I click the 'Next' button on the first 'add new contact ' screen without  filling out any fields,  the app goes blank with console errors: 
image

thegentlemanphysicist:
The disableing of the next button means that companies can no longer be submitted as contacts.

frontend/src/components/Forms/parties/AddRolesForm.js
mmumby: As per response in the dev channel, Looks like it was a business decision to have neither dates be required.
Can dig into this further if you want, otherwise remove the validations and the '*' from the label.

PR #606, opened 2019-03-18
Improve Node build times

FreshGyan:
Reduces base builds from an avg of ~20mins for core and minespace each to ~7mins each.

Reduces application builds from an avg of ~12mins for core and ~10mins for minespace to ~5mins for core and ~3.5mins for minespace.

- Upgraded Resources for builds to allow multi threading
- Added thread loader for webpack caching of assets between builds
- Refactored some code to follow the syntax we have for plugins
- Updated our base node image just to sneak in any updates red hat made

### DIFF
--- a/openshift/_nodejs.bc.json
+++ b/openshift/_nodejs.bc.json
@@ -168,11 +168,11 @@
         },
         "resources": {
           "limits": {
-            "cpu": "2",
+            "cpu": "1",
             "memory": "1Gi"
           },
           "requests": {
-            "cpu": "1",
+            "cpu": "500m",
             "memory": "512Mi"
           }
         },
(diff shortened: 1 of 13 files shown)


openshift/_nodejs.bc.json
```diff
@@ -168,12 +168,12 @@
         },
         "resources": {
           "limits": {
-            "cpu": "250m",
-            "memory": "2Gi"
+            "cpu": "2",
+            "memory": "1Gi"
```
james-hollinger: This is increasing our limit by nearly 10x; do you have any concerns with our resource allocation if we have a lot of PRs building at the same time?

FreshGyan: Yes! We have a maximum number of 16 cores available under our tools project for builds.
As long as we don't have 8 PR's all running the builds at the same time we should be good. 

Which sometimes seems to be the case, but i'd also like for people to stop updating their branches with develop as they do and see if we can come up with some way of merging PR's one after another

FreshGyan: I vote we try this out for a while and if start running out of resources we can dial it down a tad bit

PR #613, opened 2019-03-20
Fix/update timestamp update user

Jsyro:
SQLAlchemy can define `onupdate` for given columns, meaning the audit columns can update themselves, the method to do so involved reading the auth token, so I a made a flag to use a fake one when creating objects without a request context. We shouldn't have to code any assignments to these 4 columns anywhere in the application. it is backwards compatible (if it does assign the audit columns, it doesn't explode).


python-backend/app/api/utils/models_mixins.py
```diff
@@ -56,7 +57,12 @@ def save(self, commit=True):
...
     create_timestamp = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    update_user = db.Column(db.String(60), nullable=False)
-    update_timestamp = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    update_user = db.Column(
+        db.String(60),
+        nullable=False,
+        default=User().get_user_username,
+        onupdate=User().get_user_username)
+    update_timestamp = db.Column(
+        db.DateTime, nullable=False, default=datetime.utcnow, onupdate=datetime.now)
```
JustinMacaulay: is there a reason one of these is now() and the other is utcnow()?

Jsyro: good catch, i went through and changed all .now to ,utc now (except one nris comparison)

PR #617, opened 2019-03-22
Hackathon: Added Metabase Dashboard to Admin Page

MaxWardle:


frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
```diff
@@ -0,0 +1,36 @@
...
+const METABASE_SITE_URL = "https://mds-metabase-empr-mds-prod.pathfinder.gov.bc.ca";
+const METABASE_SECRET_KEY = "623e8cbb9f381e40651494194032c005ba8e92559ba291a8587d1b8e021cb454";
+
+const payload = {
+  resource: { dashboard: 1 },
+  params: {}
+};
+const token = jwt.sign(payload, METABASE_SECRET_KEY);
+
+export class DocumentUploadsMetabaseDashboard extends Component{
```
NWCalvank: It looks like this doesn't leverage state or context at all. So, it should be able to be converted to a pure function / stateless component

frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
```diff
@@ -0,0 +1,36 @@
...
+const jwt = require("jsonwebtoken");
+
+const METABASE_SITE_URL = "https://mds-metabase-empr-mds-prod.pathfinder.gov.bc.ca";
+const METABASE_SECRET_KEY = "623e8cbb9f381e40651494194032c005ba8e92559ba291a8587d1b8e021cb454";
+
+const payload = {
+  resource: { dashboard: 1 },
+  params: {}
+};
+const token = jwt.sign(payload, METABASE_SECRET_KEY);
```
NWCalvank: Looking at how we're handling token here, it would probably be better for this to be in the component state. It could be fetched on one of the component lifecycle hooks and added to the state when the token is returned. Then referenced as `this.state.token` inside of the component

frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from "react";
+import { AuthorizationGuard } from "@/HOC/AuthorizationGuard";
+import * as Permission from "@/constants/permissions";
+
+/**
+ * @class DocumentUploadsMetabaseDashboard displays an iframe housing Document Upload information.
+ */
+
+const jwt = require("jsonwebtoken");
```
NWCalvank: Use import instead of common js modules

frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
```diff
@@ -0,0 +1,36 @@
...
+
+const METABASE_SITE_URL = "https://mds-metabase-empr-mds-prod.pathfinder.gov.bc.ca";
+const METABASE_SECRET_KEY = "623e8cbb9f381e40651494194032c005ba8e92559ba291a8587d1b8e021cb454";
```
MaxWardle: Add this key to OpenShift secrets

frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
```diff
@@ -0,0 +1,36 @@
...
+    super(props);
+    this.state = {
+
+    };
+  }
+
+  render() {
+    const iframeUrl = `${METABASE_SITE_URL}/embed/dashboard/${token}#bordered=true&titled=true`;
+    return (
+      <iframe title="metabaseDashboard" src={iframeUrl} frameBorder="0" width="100%" height="2700px" allowTransparency/>
```
MaxWardle: Probably should change the height to dynamically change based on the size of the page in the iframe

frontend/src/components/admin/DocumentUploadsMetabaseDashboard.js
mmumby: @NWCalvank If you want this moved into the state, then the above comment to convert `DocumentUploadsMetabaseDashboard.js` to a pure function is no longer needed?

NWCalvank: Yes, they were two, independent suggestions that depended on which solution was taken.

NWCalvank:
Closing this PR as a POC. We have tracked the future work to be done here.

PR #618, opened 2019-03-22
Hackathon 2019/checkmark

Jsyro:


itchison:
Could use some refactoring...

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -27,30 +56,104 @@ export class AdminDashboard extends Component {
...
+  );
+
+AdminDashboard.propTypes = propTypes;
+
+export default compose(
+  connect(
+    null,
+    mapDispatchToProps
+  ),
+  AuthorizationGuard(Permission.ADMIN) // isPublic === true
```
mmumby: The `isPublic` comment isn't needed here, assuming this was copied over?

frontend/src/components/admin/AdminVerifiedMinesList.js
```diff
@@ -0,0 +1,63 @@
...
+import { Table } from "antd";
+import PropTypes from "prop-types";
+import CustomPropTypes from "@/customPropTypes";
+import * as router from "@/constants/routes";
+import { Link } from "react-router-dom";
+import NullScreen from "@/components/common/NullScreen";
+import { formatDate } from "@/utils/helpers";
+
+/**
+ * @class AdminDashboard houses everything related to admin tasks, this is a permission-based route.
```
mmumby: update comment to correct class

frontend/src/components/mine/MineDashboard.js
```diff
@@ -111,10 +111,19 @@ export class MineDashboard extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { activeTab } = nextProps.match.params;
+    const { id, activeTab } = nextProps.match.params;
     if (activeTab !== this.props.activeTab) {
       this.setState({ activeTab });
     }
+    if (this.props.match.params.id !== nextProps.match.params.id) {
```
mmumby: suggestion to DRY up the code, move the logic inside this block into its own function and call it in both `componentDidMount` and `componentWillReceiveProps`

Very happy with this update, by the way, good find.

frontend/src/components/mine/MineHeader.js
```diff
@@ -138,13 +151,91 @@ class MineHeader extends Component {
...
+          )}
+        </AuthorizationWrapper>
+        <AuthorizationWrapper inTesting>
+          {(!this.props.mine.verified_status || this.props.mine.verified_status.healthy) && (
+            <Menu.Item key="3">
+              <Popconfirm
+                placement="left"
+                title="Are you sure?"
+                onConfirm={() =>
+                  this.props.setMineVerifiedStatus(this.props.mine.guid, false).then(() => {
```
mmumby: To simplfy the jsx, move this logic into its own function and pass it into the `onConfirm`
you can also pass the event in here, and handle the propagation inside the new function. Then the `onClick` on the button below will no longer be needed.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -379,3 +379,49 @@ export const fetchMineDocuments = (mineGuid) => (dispatch) => {
       dispatch(hideLoading());
     });
 };
+
+// MineVerifcationStatus
+export const fetchMineVerifiedStatuses = (user_id) => (dispatch) => {
+  const params = user_id ? { user_id } : null;
```
mmumby: it looks like you're doing the null check inside `API.MINE_VERIFIED_STATUSES` as well, 
```
export const MINE_VERIFIED_STATUSES = (params) =>
  params ? `/mines/verified-status?${queryString.stringify(params)}` : "/mines/verified-status";
```

is `user_id` the only expected param? If so, this could be simplified and user_id could be passed directly in without the above const.

```
export const MINE_VERIFIED_STATUSES = (userID) =>
  userID ? `/mines/verified-status?user_id=${userID}` : "/mines/verified-status";
```

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -27,30 +56,104 @@ export class AdminDashboard extends Component {
     downloadMineManagerHistory(this.state.mineNo, { window, document });
   };
```
mmumby: Is this being sorted alphabetically? Looks like there is a JS function that does this, `localeCompare()`

If that's not the purpose of this function, can you rename to make it more clear whats happening?

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -27,30 +56,104 @@ export class AdminDashboard extends Component {
...
+  compareMineName = (a, b) => {
+    if (a.mine_name < b.mine_name) {
+      return -1;
+    }
+    if (a.mine_name > b.mine_name) {
+      return 1;
+    }
+    return 0;
+  };
+
```
mmumby: Is this being sorted alphabetically? Looks like there is a JS function that does this, localeCompare()

If that's not the purpose of this function, can you rename to make it more clear whats happening?

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -27,30 +56,104 @@ export class AdminDashboard extends Component {
...
+            animated={{ inkBar: true, tabPane: false }}
+          >
+            <TabPane tab="Verified Mines" key="verifiedMines">
+              <div className="tab__content">
+                <div>
+                  <h4>{this.state.verifiedMines.length}&nbsp;Verified Mines</h4>
+                  <div>
+                    {this.state.verifiedMines.length > 0 && (
+                      <AdminVerifiedMinesList
+                        minesVerifiedStatusList={this.state.verifiedMines.sort(
```
NWCalvank: Small thing, but if we sort the list before putting it into the state, we won't need to sort it again on every re-render. This helps keep the render method nice and light

frontend/src/components/admin/AdminDashboard.js
```diff
@@ -27,30 +56,104 @@ export class AdminDashboard extends Component {
...
+              </div>
+            </TabPane>
+            <TabPane tab="Mines to be Checked" key="unverifiedMines">
+              <div className="tab__content">
+                <div>
+                  <h4>{this.state.unverifiedMines.length}&nbsp;Mines Needing Re-Verification</h4>
+                  <div>
+                    {this.state.unverifiedMines.length > 0 && (
+                      <AdminVerifiedMinesList
+                        minesVerifiedStatusList={this.state.unverifiedMines.sort(
```
NWCalvank: Same comment about sorting in the render method here

frontend/src/components/admin/AdminVerifiedMinesList.js
```diff
@@ -0,0 +1,63 @@
...
+    width: 150,
+    dataIndex: "formatted_timestamp",
+  },
+];
+
+const transformRowData = (verifiedMinesList) =>
+  verifiedMinesList.map((status) => ({
+    key: status.mine_guid,
+    formatted_timestamp: formatDate(status.verifying_timestamp),
+    ...status,
```
NWCalvank: If we don't need to pass the other `status` properties onto the transformedRow object, we can pull out the properties that we want to modify and then just pass in the rest:
```javascript
verifiedMinesList.map(({mine_guid: key, verifying_timestamp, ...rest}) => ({
  key,
  formatted_timestamp: formatDate(status.verifying_timestamp),
  ...rest,
})
```
Renaming `mine_guid` during destructuring is entirely optional. I just included it since that's how I would do it.

frontend/src/constants/API.js
```diff
@@ -40,6 +40,9 @@ export const MINE_MANAGER_HISTORY = (mineNo) =>
 export const MINESPACE_USER = "/users/minespace";
 export const PROVINCE_CODES = "/parties/sub-division-codes";
 
+export const MINE_VERIFIED_STATUSES = (params) =>
+  params ? `/mines/verified-status?${queryString.stringify(params)}` : "/mines/verified-status";
```
NWCalvank: I'm in favour of setting the default `params` value to an empty object (as I've done on line 16). Then there's no need for the ternary. It also handles the `null` check, so the function will never blow up so long as it either receives `null`, `undefined` or an object.

python-backend/app/api/mines/mine/models/mine_verified_status.py
```diff
@@ -0,0 +1,56 @@
...
+    verifying_timestamp = db.Column(db.DateTime, nullable=False, default=datetime.datetime.utcnow)
+    update_user = db.Column(
+        db.String(60),
+        nullable=False,
+        default=User().get_user_username,
+        onupdate=User().get_user_username)
+    update_timestamp = db.Column(
+        db.DateTime,
+        nullable=False,
+        default=datetime.datetime.utcnow,
```
NWCalvank: Does `FetchedValue()` not work for dates?

frontend/src/components/admin/AdminDashboard.js
Jsyro: fixed

frontend/src/components/admin/AdminVerifiedMinesList.js
Jsyro: fixed

frontend/src/components/mine/MineDashboard.js
Jsyro: fixed

frontend/src/components/mine/MineHeader.js
Jsyro: fixed

python-backend/tests/mines/resources/test_mine_verified_status_resource.py
```diff
@@ -0,0 +1,92 @@
...
+
+# GET
+def test_get_all_mine_verified_status(test_client, auth_headers, setup_info):
+    get_resp = test_client.get('/mines/verified-status', headers=auth_headers['full_auth_header'])
+    get_data = json.loads(get_resp.data.decode())
+    assert get_resp.status_code == 200, str(get_resp.response)
+    assert len(get_data['healthy']) == len(
+        [x for x in setup_info['mines'] if x.verified_status.healthy_ind == True])
+    assert len(get_data['unhealthy']) == len(
+        [x for x in setup_info['mines'] if x.verified_status.healthy_ind == False])
```
NWCalvank: It looks like these `len` values should be consistent between test runs. Any reason not to just hardcode the expected value here? The concern I have is just having a bit too much logic in the tests.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -379,3 +379,49 @@ export const fetchMineDocuments = (mineGuid) => (dispatch) => {
...
+      notification.error({
+        message: err.response ? err.response.data.error.message : String.ERROR,
+        duration: 10,
+      });
+      dispatch(error(reducerTypes.GET_MINE_VERIFIED_STATUS));
+      dispatch(hideLoading());
+    });
+};
+
+export const setMineVerifiedStatus = (mine_guid, healthy) => (dispatch) => {
```
NWCalvank: I think it would be beneficial for our action creators to follow a standardized input pattern going forward. I'd be in-favour of `actionCreatorName(urlParams: object, payload: object)` because that is the pattern used by axios already, and it is fully able to support 0, 1, or n urlParams and 0, 1, or n payload properties.

Thoughts? @mmumby ? I would see this as a fix-forward thing.

I think this works well with our use of `queryString` in the API endpoint constant functions, which have been gradually converted from accepting urlParam strings to param objects.

mmumby: If I am understanding this correctly, I believe I had an opposite suggestion in the comment above.

I'm all for standardization, @NWCalvank. 
I guess if we all agree on a standard put/post pattern, and using `queryString` in our API consts, than you can disregard the above comment @Jsyro

NWCalvank: Your suggestion is a reasonable refactor for the code as it currently is, but I think we should keep ourselves open to the idea that we may want to add more parameters to this endpoint in the future, which a params object more easily supports. Hope that makes sense :) 

And, as mentioned, it allows all of the endpoints to have the same argument pattern, which makes for less context-switching when moving to and from endpoints.

mmumby: Agreed! makes sense.

frontend/src/components/admin/AdminDashboard.js
Jsyro: this uses local state to store this list, but regardless, sorted on save, not on use.

frontend/src/components/admin/AdminDashboard.js
Jsyro: changed.

python-backend/app/api/mines/mine/models/mine_verified_status.py
Jsyro: changed.

python-backend/tests/mines/resources/test_mine_verified_status_resource.py
Jsyro: I don't disagree, but I want the test to pass regardless of if someone adds a few items to the test data... I think it's readable even if it is a bit more logic than it could be. @james-hollinger thoughts?

frontend/src/components/admin/AdminVerifiedMinesList.js
Jsyro: ah i see, ok, avoid having the extra member that could be confusing later. changed.

frontend/src/constants/API.js
Jsyro: ok, changed.

frontend/src/components/admin/AdminDashboard.js
Jsyro: Thanks for finding that, changed.

frontend/src/actionCreators/mineActionCreator.js
Jsyro: I've made the action creator accept generic payload, I think that get's us to where we want to be, let me know if you want any more changes.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -405,12 +405,12 @@ export const fetchMineVerifiedStatuses = (user_id) => (dispatch) => {
...
 };
 
-export const setMineVerifiedStatus = (mine_guid, healthy) => (dispatch) => {
+export const setMineVerifiedStatus = (mine_guid, payload) => (dispatch) => {
   dispatch(request(reducerTypes.SET_MINE_VERIFIED_STATUS));
   return axios
     .put(
       `${ENVIRONMENT.apiUrl}${API.MINE_VERIFIED_STATUS(mine_guid)}`,
-      { healthy },
+      { ...payload },
```
NWCalvank: The `{}` and `...` cancel each other out. `payload` on its own should work fine here

PR #619, opened 2019-03-25
Mds 1620

thegentlemanphysicist:
This implements the delete functionality.  Tests being written now.  Will need some help on front end tests.


frontend/src/components/parties/PartyProfile.js
```diff
@@ -150,24 +167,41 @@ export class PartyProfile extends Component {
...
+                <AuthorizationWrapper permission={Permission.ADMIN}>
+                  <Popconfirm
+                    placement="bottom"
+                    title={`Are you sure you want to delete the party '${formatedName}'?  Doing so will permanently
+                     remove the party and all associated roles.`}
+                    onConfirm={this.deleteParty}
+                    okText="Yes"
+                    cancelText="No"
+                  >
+                    <Button type="ghost" style={{ color: "#BC2929", borderColor: "#BC2929" }}>
```
mmumby: Okay for this button, (following up from our chat on slack)
- update the type to be "type=danger" as this is a built-in type from ant design that we haven't customized/utilized yet. 
- move all styling to themeOverride.scss, follow a similar pattern as `.ant-btn-secondary` for this new button, the class will be `.ant-btn-danger`
- For the icon, move the hex color into a constant in `constants/styles.js`

If you need further clarification let me know!

thegentlemanphysicist: done and pushed

frontend/src/components/parties/PartyProfile.js
```diff
@@ -103,6 +106,20 @@ export class PartyProfile extends Component {
...
 
+  deleteParty = () => {
+    const { id } = this.props.match.params;
+    this.props.deleteParty(id).then(() => {
+      const page = String.DEFAULT_PAGE;
+      const per_page = String.DEFAULT_PER_PAGE;
+      this.props.history.push(
+        router.CONTACT_HOME_PAGE.dynamicRoute({
+          page,
+          per_page,
```
NWCalvank: Small thing, but the assignment on lines 112-113 is a bit unnecessary. May as well just pass in the values here:
```javascript
router.CONTACT_HOME_PAGE.dynamicRoute({
  page: String.DEFAULT_PAGE,
  per_page: String.DEFAULT_PER_PAGE,
})
```

thegentlemanphysicist: fixed any way

PR #621, opened 2019-03-25
Add logstash

FreshGyan:
- Adds logstash to the pipeline.


elastic/logstash/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+# Add logstash config files
+COPY dictionaries /app/logstash/dictionaries
+COPY pipeline/ /app/logstash/pipeline/
+COPY config/ /app/logstash/config/
+
+# Ensure Logstash gets a UTF-8 locale by default.
+ENV LANG='en_US.UTF-8' LC_ALL='en_US.UTF-8'
+
+# Update permissions to be accessible by randomly generated user ID
```
itchison: What User Id is randomly generated?

elastic/logstash/dictionaries/action.yml
```diff
@@ -0,0 +1,4 @@
+GET .*/api/mines/[0-9A-Fa-f]{8}(?:-[0-9A-Fa-f]{4}){3}-[0-9A-Fa-f]{12}: Mine View
```
itchison: We're going to want to add a lot more items here. We should add a reminder to ourselves to talk about this in the readme and in general what events we want to capture.

elastic/logstash/pipeline/logstash.conf
```diff
@@ -0,0 +1,140 @@
+input {
+  file {
+    path => "/app/logstash/monitor/*.log"
```
itchison: Do we want access and error?

elastic/logstash/pipeline/logstash.conf
```diff
@@ -0,0 +1,140 @@
...
+    source => "message"
+    target => "[@metadata][fingerprint]"
+    method => "MURMUR3"
+  }
+
+
+  if [fileset][module] == "nginx" {
+    if [fileset][name] == "access" {
+      grok {
+        match => { "message" => ["%{IPORHOST:[nginx][access][remote_ip]} - %{DATA:[nginx][access][user_name]} \[%{HTTPDATE:[nginx][access][time]}\] \"%{WORD:[nginx][access][method]} %{DATA:[nginx][access][url]} HTTP/%{NUMBER:[nginx][access][http_version]}\" %{NUMBER:[nginx][access][response_code]} %{NUMBER:[nginx][access][body_sent][bytes]} \"%{DATA:[nginx][access][referrer]}\" \"Bearer %{DATA:[@metadata][jwt]}\" \"%{DATA:[nginx][access][agent]}\" \"%{IPORHOST:[nginx][access][client_ip]}\" \"%{GREEDYDATA:[nginx][access][payload]}\""] }
```
itchison: We have a ticket right now (MDS-1619) that will be adding groups. Not sure if this will bring that JWT item in.

openshift/tools/logstash.bc.json
```diff
@@ -0,0 +1,145 @@
...
+        "labels": {
+          "app": "${NAME}${SUFFIX}",
+          "buildconfig": "${NAME}${SUFFIX}"
+        }
+      },
+      "spec": {
+        "source": {
+          "type": "Git",
+          "git": {
+            "uri": "${SOURCE_REPOSITORY_URL}",
```
itchison: Are we just building the latest version here in the secret?

elastic/logstash/Dockerfile
FreshGyan: We never know. Openshift basically generated a random ID like `10002342` and it can be different for each pod and each envt.

elastic/logstash/dictionaries/action.yml
FreshGyan: Defs!

elastic/logstash/pipeline/logstash.conf
FreshGyan: Currently they're both coming in

openshift/tools/logstash.bc.json
FreshGyan: Yeah! But our builds only get executed if you change anything inside the `elastic/logstash` directory

PR #624, opened 2019-03-26
Feature/mds 1589 add response and model to swagger

JustinMacaulay:
This branch takes the applications endpoint and transitions it to following the flask-restplus standards.
- uses @api.model() to enforce the return type.
- errors raise werkzeug exceptions.
- Swagger documentation is at a point where it is useful.


frontend/src/utils/helpers.js
```diff
@@ -49,6 +49,13 @@ export const formatDate = (dateString) =>
   dateString !== "None" &&
   moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");
 
+// Function to format an API date string to human readable
+export const formatFullDateTime = (dateString) =>
+  dateString &&
+  dateString !== "9999-12-31" &&
+  dateString !== "None" &&
+  moment(dateString).format("MMM DD YYYY");
```
mmumby: How does this dateString differ than the one above?

JustinMacaulay: It is in a UTC full format so YYYY-MM-DDTHH:mm:SSZ

With this change all dates will be returned in this way so i didn't want to modify the currently working version and just added this to bridge the gap for now.

mmumby: Thanks for the update!

python-backend/app/api/applications/resources/application_status_code.py
```diff
@@ -9,11 +9,22 @@
...
+    @api.doc(
+        description=
+        'This endpoint returns a list of all possible document status codes and thier descriptions.'
+    )
     @requires_role_mine_view
     def get(self):
         application_status_codes = ApplicationStatusCode.find_all_active_application_status_code()
         if application_status_codes:
-            result = [x.json() for x in application_status_codes]
+            result = application_status_codes
```
nayr974: This is so much nicer

python-backend/app/api/applications/resources/application.py
```diff
@@ -0,0 +1,154 @@
...
+        if not application:
+            raise NotFound('Application not found')
+
+        return application
+
+    @api.expect(parser)
+    @api.doc(
+        description=
+        'This endpoint updates an application for the mine specified using the form-data passed.',
+        responses={
```
nayr974: This is purely for documentation correct, what are the expected error responses? Do we need to include the 200 response code here as well?

python-backend/app/api/applications/resources/application.py
```diff
@@ -0,0 +1,154 @@
...
+
+        mine = Mine.find_by_mine_guid(data.get('mine_guid'))
+
+        if not mine:
+            raise NotFound('There was no mine found with the provided mine_guid.')
+
+        application_no, application_status_code, received_date = list(
+            map(data.get, ['application_no', 'application_status_code', 'received_date']))
+
+        if not application_no or not application_status_code or not received_date:
```
nayr974: If this type of validation logic gets more convoluted in our resources or if different calls validate the same parameters, we should define a pattern for a validation function that they call.

python-backend/app/api/applications/resources/application.py
JustinMacaulay: The marshal_with handles the 200 or 201 or whatever the success response. The responses section is only for the error responses as far as I could tell.

PR #626, opened 2019-03-27
Task/mds 1596 common add button

nayr974:


frontend/src/components/common/AddButton.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @constant AddButton  - Globally styled add button
+ */
+
+const propTypes = {};
+
+const AddButton = (props) => (
+  <Button type="primary" className="full-mobile" {...props}>
+    <Icon type="plus" theme="outlined" style={{ fontSize: "18px", paddingRight: "5px" }} />
```
mmumby: My only suggestion is to use a class for the icon styling, so all button icons are the same, 
I believe a lot of the buttons are using `className="padding-small--right icon-sm"`

which is `padding-right: 5px` `font-size: 20px`

Happy to have the font-size be `18px` if that's preferred, but would like all the button icons to be the same!

nayr974: I originally had 18 because it lined the icon nicely inside the button. What I've now done is added display:inline-flex; and align-items: center; to the ant-btn override styles which will vertically align the content so everything is in the middle, then updated AddButton to use those two classes. 👍

mmumby: Awesome!!

PR #627, opened 2019-03-27
Bug/MDS-1658 Minespace dates not autopopulating

james-hollinger:
Currently, users experienced incidents where the upload date was not being set correctly (see MDS-1658).

This bug was due to the update logic comparing the old expected document to the new document in the store. If the user clicked submit right as the document was added but before the store was updated, the logic sees no difference in the expected document and does not update the date. This error would also happen if a user uploads a document, navigates away from the page without pressing submit, then later opens the upload modal and clicks submit. Since the "old" and "new" expected documents both have a file attached, the date is not updated.  All of the above also applies to deleting uploads.

Logic has been fixed. FilePond has also been updated for the public frontend for improved user experience.


james-hollinger:
Everything should still work with IE. I checked briefly myself manually.

NWCalvank:
@james-hollinger So long as the app still loads (so that it can show the warning banner for IE), we're good. Filepond was causing the app to throw a runtime error a while back that meant it crashed immediately. So, we definitely don't need full feature support in IE. Just to make sure that's clear

PR #628, opened 2019-03-27
Task/use store missing on put endpoint-parties

Jsyro:
First cut on store_missing put updates. If all the reqparser keys match the model attributes, then we can just loop through them. Quite a few formatting changes.


python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -32,26 +54,31 @@ class PartyResource(Resource, UserMixin, ErrorMixin):
...
     parser.add_argument(
         'address_line_2',
         type=str,
+        trim=True,
         store_missing=False,
         help='The second address line of the party address. Ex: 1234 Foo Road')
     parser.add_argument(
         'city',
         type=str,
+        trim=True,
```
nayr974: I didn't know you could set these to trim. Cool!

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -146,28 +171,13 @@ def put(self, party_guid):
...
-            existing_party.address_line_2 = data.get(
-                'address_line_2') if 'address_line_2' in data else existing_party.address_line_2
-            existing_party.city = data.get('city') if 'city' in data else existing_party.city
-            existing_party.sub_division_code = data.get(
-                'sub_division_code'
-            ) if 'sub_division_code' in data else existing_party.sub_division_code
-            existing_party.post_code = data.get(
-                'post_code') if 'post_code' in data else existing_party.post_code
+            current_app.logger.info(f'Updating {existing_party} with {data}')
+            for key, value in data.items():
```
nayr974: 💯

python-backend/app/scheduled_jobs/ETL_jobs.py
```diff
@@ -11,7 +11,7 @@ def _schedule_ETL_jobs(app):
 @register_apm
 def _run_ETL():
     with sched.app.app_context():
-        job_running = cache.get(ETL())
+        job_running = cache.get(ETL)
```
nayr974: Was this change intended? Does it have an effect?

Jsyro: @JustinMacaulay  requested this change, ETL isn't a function... so that would have been bad, it is unrelated to my changes.

PR #630, opened 2019-03-28
Task/store missing on put permit and permit amendments

Jsyro:
2nd round of resource clean up, put needs some exceptions for related fields to make sure they can't be changed.


Jsyro:
This PR did pass the pipeline...  but github isn't getting the update... force merge? 

image

PR #632, opened 2019-03-29
MDS-1526 changed the display format of the documents in minespace.

JustinMacaulay:
Changed the way documents are displayed in mine space, now separated by bullet points.

### DIFF
--- a/frontend-public/src/components/dashboard/mineInfo/MineInfoList.js
+++ b/frontend-public/src/components/dashboard/mineInfo/MineInfoList.js
@@ -62,9 +62,7 @@ const columns = [
   {
     title: "Received Date",
     dataIndex: "receivedDate",
-    render: (text, record) => (
-      <div title="Received Date">{record.doc.received_date || "-"}</div>
-    ),
+    render: (text, record) => <div title="Received Date">{record.doc.received_date || "-"}</div>,
   },
   {
     title: "Status",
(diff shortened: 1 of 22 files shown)


frontend-public/src/components/dashboard/mineInfo/MineInfoList.js
```diff
@@ -62,9 +62,7 @@ const columns = [
   {
     title: "Received Date",
     dataIndex: "receivedDate",
-    render: (text, record) => (
-      <div title="Received Date">{record.doc.received_date || "-"}</div>
-    ),
+    render: (text, record) => <div title="Received Date">{record.doc.received_date || "-"}</div>,
```
NWCalvank: I'm fairly sure our autoformatting will always wrap JSX from an implicit return in parentheses. Did you intentionally make this change? I'm curious why it would happen otherwise

JustinMacaulay: No, I just saved the file and pushed it. This was auto formatting. I only modified the document section explicitly.

NWCalvank: Very odd. The other changes look good, so I'll approve. I don't like that your formatted did something different. We should keep an eye on this to make sure you're not going to end up with big diffs in the future based strictly on your formatting being different. Everyone's should be exactly the same automatically.

nayr974: is there a different rule-set applied to the public site?

PR #638, opened 2019-04-01
add random sleep.

Jsyro:
trying to prevent APM errors that aren't legitmate, should reduce errors from 50% to <1%.


python-backend/app/scheduled_jobs/ETL_jobs.py
```diff
@@ -10,6 +12,8 @@ def _schedule_ETL_jobs(app):
 
 @register_apm
 def _run_ETL():
+    #try to desynchronize the two pods.
+    sleep(randint(10, 10000))
```
nayr974: Python's sleep is in seconds, was this range intended?

PR #641, opened 2019-04-03
Conditionally Run Frontend Tests pre-push

NWCalvank:
A quick solution built using on git version `2.19.1` on MacOS

This implements the following logic:
Check if remote exists.
If it doesn't, run the tests. If it does exist, compare local changes to remote branch and see if there are frontend changes.
If there are frontend changes, run tests. If there aren't frontend changes, exit with a success status code (allowing you to chain additional commands)

Caveats
This uses the `git` CLI output and parses it as a stdin. This assumes a lot about the git version, terminal type (bash, fish, whatever else), and standard formatting of the output.

This is largely just a discussion/starting point for what a solution _might_ look like.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -114,6 +114,9 @@
     "format:write": "prettier --config .prettierrc.json --write",
     "format:write-all": "prettier --config .prettierrc.json src/*.js --write && prettier --config .prettierrc.json src/**/*.js --write && prettier --config .prettierrc.json src/**/**/*.js --write",
     "test": "jest --env=jsdom --runInBand",
+    "test:conditional": "ls && npm run test:conditional-unix || npm run test:conditional-windows",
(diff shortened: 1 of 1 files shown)


FreshGyan:
I like where this is going! I was playing around briefly with a similar idea and wanted to see what you thought.

https://github.com/bcgov/mds/compare/testhook?expand=1

Basically i was thinking we have two problems, one being when to run "commit or "push" and second being only running it for frontend directory and not entire project.

So i just started off with commits and leveraged the lint-staged plugin you added. So since that only runs on js file changes, I added a jest parameter to only run tests that are relevant to the changed files. It runs relatively faster than running whole suite and avoids the whole mess of checking remote branch and what not.

Also played a little with https://www.npmjs.com/package/lint-prepush
But it didn't really respect the `.js` rule, unless i was configuring it wrong 🤷‍♂️ 

This also seemed like a decent configurable pre-push commit file, we can configure and limit to scope of frontend, I haven't had a chance to test it though
https://coderwall.com/p/l5ddcq/lint-code-before-git-pushing

NWCalvank:
@FreshGyan Yours is the obvious solution, which I like and had considered. The reason why I stuck with a pre-commit hook in my solution was that I was already getting feedback that people weren't stoked on the 20 seconds it takes for the tests to run before pushing, so moving the tests to the commit was unlikely to make team members very happy (even if the tests run faster).

The key assumption being, though, that people commit more frequently than they push. Perhaps this isn't a valid assumption -- particularly for a project that gets squash merged.

I'll take a look at the lint-prepush library, as configuring frontend builds is in my wheelhouse.

NWCalvank:
Closing this because a new approach has been taken here: https://github.com/bcgov/mds/pull/655

PR #642, opened 2019-04-03
Bug/mds 1760

thegentlemanphysicist:
When switching between the org and person type contacts, names (first, last, and party) are all cleared in the store, preventing them from breaking search function.


frontend/src/components/Forms/AdvancedContactSearchForm.js
```diff
@@ -11,6 +11,7 @@ import * as Strings from "@/constants/strings";
 const propTypes = {
   handleSubmit: PropTypes.func.isRequired,
   handleSearch: PropTypes.func.isRequired,
+  handleContactTypeChange: PropTypes.func.isRequired,
```
NWCalvank: Could you rename this to describe what it actually does? Something like `clearPartyNameFields` or `resetPartyNameFields`. I think it could be unnecessarily confusing to have a prop with this name and a method with this name in the same file. The method is more of the 'handler' than the prop.

frontend/src/components/dashboard/contactsHomePage/ContactHomePage.js
```diff
@@ -142,6 +145,12 @@ export class ContactHomePage extends Component {
     );
   };
 
+  handleContactTypeChange = () => {
+    this.props.change(FORM.CONTACT_ADVANCED_SEARCH, `party_name`, null);
```
NWCalvank: These don't need to be in backticks. Normal strings will do

frontend/src/components/Forms/AdvancedContactSearchForm.js
thegentlemanphysicist: done

frontend/src/components/dashboard/contactsHomePage/ContactHomePage.js
thegentlemanphysicist: done

PR #644, opened 2019-04-03
Refactor Axios interceptors to handle error toast messages

james-hollinger:
Error toast messages are now handled globally by an Axios interceptor.

the CustomAxios() method accepts a parameter to override the default error toast message.

No parameter, undefined, or "default" results in the default toast message
If a string is provided it will override the toast message
A non-undefined falsey value will suppresses the toast message entirely

I have also moved the dispatch(hideLoading()) calls into a finally block to reduce code repetition.


frontend/src/customAxios.js
```diff
@@ -0,0 +1,35 @@
+import axios from "axios";
+import { notification } from "antd";
+import * as String from "@/constants/strings";
+
+const UNAUTHORIZED = 401;
+const MAINTENANCE = 503;
+
+const CustomAxios = (errorToastMessage) => {
+  const instance = axios.create();
```
NWCalvank: Any reason why we wouldn't set the `baseURL` here so that we can remove `ENVIRONMENT.apiUrl` from every API call? And same question about request headers.

Everything else here looks good to me

james-hollinger: Ticket scope. There is a good argument to be made to make all these changes at once but the scope of work was only set to cover the response interceptors.

PR #645, opened 2019-04-04
MDS-1599 & MDS-1656 Feature/ Support Create Variances & Upload Documents (backend)

NWCalvank:


python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -0,0 +1,135 @@
...
+    @requires_role_mine_view
+    @api.marshal_with(variance_model, code=200, envelope='records')
+    def get(self, mine_guid):
+        try:
+            variances = Variance.find_by_mine_guid(mine_guid)
+        except DBAPIError:
+            raise BadRequest('Invalid mine_guid')
+
+        if variances is None:
+            raise BadRequest('Unable to fetch variances')
```
Jsyro: Do we want to throw an error, seems like we should just return an empty list?

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -0,0 +1,135 @@
...
+            'variance_id': 'ID of the variance to fetch'
+        }
+    )
+    @requires_role_mine_view
+    @api.marshal_with(variance_model, code=200)
+    def get(self, mine_guid, variance_id):
+        try:
+            variance = Variance.find_by_variance_id(variance_id)
+        except DBAPIError:
+            raise BadRequest('Invalid variance_id')
```
Jsyro: we used to try catch the query helpers... and now that i'm seeing this come up..... do you think this is better? 

```    @classmethod
        try:
            uuid.UUID(_id, version=4)
            return cls.query.filter_by(mine_guid=_id).filter_by(deleted_ind=False).first()
        except ValueError:
            return None``` 

This also stops the db from having to throw an exception and the python catches it... i'm open for either. just wanted you thought.

File: python-backend/app/api/mines/variances/resources/variance.py (hunk 1 above)
Comment (NWCalvank) [reply python-backend/app/api/mines/variances/resources/variance.py]: It'll be an empty list if there were no variances on the mine. If the query returns `None` then something went wrong with the query, which is why I figured 'Unable to fetch variances' made the most sense.

Not married to this perspective, but it seems logical to me.

File: python-backend/app/api/mines/variances/resources/variance.py (hunk 2 above)
Comment (NWCalvank) [reply python-backend/app/api/mines/variances/resources/variance.py]: Yes, I think that approach is better. I think safely returning a `None` and then doing the `None` check in the resource makes plenty of sense

File: python-backend/app/api/mines/variances/models/variance.py [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+    issue_date = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    received_date = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
+    expiry_date = db.Column(db.DateTime,
+                            nullable=False,
+                            default=datetime.strptime('9999-12-31', '%Y-%m-%d'))
+
+    related_documents = db.relationship(
+        'VarianceDocument',
+        backref="variance",
+        lazy='joined')
```
Jsyro: `lazy="joined"` means it is lazy loaded, given how tightly the documents are used with the variance, we might get preformance gains out of `lazy="select` meaning it's joined on select.

python-backend/app/api/mines/variances/resources/variance.py
Jsyro: oh right, you didn't check falsy, you check for None... yea that check doesn't hurt, disregard.

python-backend/app/api/mines/variances/resources/variance.py
Jsyro: Yea, ok, let's try and re-establish that pattern of using the model search helpers to check for guid formatting.

python-backend/app/api/mines/variances/models/variance.py
nayr974: Other way around :)

nayr974: https://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.lazy

Jsyro: well never mind then.....

python-backend/app/api/documents/variances/resources/variance.py
```diff
@@ -0,0 +1,66 @@
...
+        except DBAPIError:
+            raise BadRequest('Invalid variance_id')
+
+        if records is None:
+            records = []
+
+        return records
+
+
+class VarianceDocumentResource(Resource, UserMixin, ErrorMixin):
```
Jsyro: just making this visible after a verbal conversation yesterday, this resource should be unnecessary, Variances should directly have a set of documents (variances.documents) after leveraging the `secondary` attribute of the relationship on the variance model. Let me know if you have questions.

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -0,0 +1,238 @@
...
+        if variance is None:
+            raise NotFound('Variance not found')
+        if mine_document is None:
+            raise NotFound('Mine document not found')
+
+        variance.documents.remove(mine_document)
+        variance.save()
+
+        # FIXME: This response is inconsistent with the others
+        return {'status': 204, 'message': 'The document was removed succesfully'}
```
Jsyro: can this be
```return ('',204)``` 

204 means 'success but no response body'

File: python-backend/app/api/mines/variances/resources/variance.py [hunk 6]
```diff
@@ -0,0 +1,238 @@
...
+            return metadata
+
+        for key_value in request_metadata.split(","):
+            (key, value) = key_value.split(" ")
+            metadata[key] = base64.b64decode(value).decode("utf-8")
+
+        return metadata
+
+
+class VarianceUploadedDocumentsResource(Resource, UserMixin, ErrorMixin):
```
Jsyro: can you pull out the two upload document resources into another file? i'm guessing you wanted to reuse the swagger model, but this is a bit too much for one file.

python-backend/app/api/mines/variances/resources/variance.py
NWCalvank: I tried that before based on your comment on the original PR, and it just blew up. I'll revisit it now

python-backend/app/api/mines/variances/resources/variance.py
NWCalvank: I actually think this is better since these resources don't have associated models anyway. They're all associated with the Variance model now, and so the variance endpoints live in one place: the resources/variance file.

And it only puts the file length a little over 200 lines.

But I don't care that much, so I can split it if there's a good argument for it

python-backend/app/api/mines/variances/resources/variance.py
Jsyro: Odd, we are doing that elsewhere.... wonder if flask-restplus wants something else? also try ```return '', 204```

python-backend/app/api/mines/variances/resources/variance.py
Jsyro: Not worth any extra work, sounds good as is.

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -0,0 +1,237 @@
...
+        mine_doc = MineDocument(
+            mine_guid=mine_guid,
+            document_manager_guid=document_manager_guid,
+            document_name=document_name)
+
+        if not mine_doc:
+            raise BadRequest('Unable to register uploaded file as document')
+
+        variance.documents.append(mine_doc)
+        db.session.commit()
```
Jsyro: variance.save() should do the same thing but use the try/except we in that method.

PR #647, opened 2019-04-05
Feature/improving logstash pipeline

jlaplante333:
Includes SQL query, configuration settings, fixes for new index naming conventions dependent on environments.


elastic/logstash/pipeline/logstash.conf
```diff
@@ -15,126 +17,197 @@ filter {
...
-          gsub => [
-            "[nginx][access][payload]", "[\\]", ""
-          ]
+      if [mine][id] {
+            # lookup by mine id
+        jdbc_streaming {
+          jdbc_driver_library => "/app/logstash/jdbc/postgresql-42.2.5.jar"
+          jdbc_driver_class => "org.postgresql.Driver"
+          jdbc_connection_string => "${JDBC_CONNECTION_STRING}"
+          jdbc_user => "${JDBC_USER}"
```
itchison: Do we need the secret JSON BC/DC set up here?

elastic/logstash/pipeline/logstash.conf
```diff
@@ -15,126 +17,197 @@ filter {
...
     }
   }
 }
 
-
 output {
-
-  stdout { codec => rubydebug}
+  
+  # stdout {}
```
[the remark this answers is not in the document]
jlaplante333: No we do not need it! Thanks

elastic/logstash/pipeline/logstash.conf
jlaplante333: Please provide more info about what secret JSON BC/DC is. Thanks.

FreshGyan: Hi! So Openshift has a BuildConfig(BC) and DeploymentConfig(DC). Environment variables are runtime configuration so Deployment Config would need to be updated to support the newly added environment variables.

Current logstash deployment config is located under `openshift/tools/logstash.dc.json` The environment variables are defined around line 108 `https://github.com/bcgov/mds/blob/develop/openshift/tools/logstash.dc.json#L108`.
You would need to add 3 more blocks alongside the rest of the environment variables for the `JDBC_CONNECTION_STRING ` , `JDBC_USER ` and the `JDBC_PASSWORD ` 

Now since the code is public, we don't hardcode the database credentials into the deployment config and they come from openshift secrets.  You would need to define the json block so that the env variable can come from secrets and openshift will take care of the rest.

You can find a similar example here:
https://github.com/bcgov/mds/blob/develop/openshift/_python36.dc.json#L454
This one is a tool called flyway which also uses JDBC to connect to postgres so I assume the environment variables . will be similar to `PGUSER `, `FLYWAY_URL ` and `PGPASSWORD `.
Please ensure for `PGUSER` you are using `database-user` and not `database-admin-user`.

After the environment variable block has been updated, you will need to update the logstash deployment config template to add the name of the secret as a parameter like done here:
https://github.com/bcgov/mds/blob/develop/openshift/_python36.dc.json#L28

Once all of that is done, you just need to pass the value in from the `pipeline/config.groovy`, this file gets picked up by the pipeline to add parameters to the deployment configs.
The logstash deployment config is here `https://github.com/bcgov/mds/blob/develop/pipeline/config.groovy#L351`
You will need to add another parameter similar to here
https://github.com/bcgov/mds/blob/develop/pipeline/config.groovy#L312

Let me know if you get stuck or have any questions

elastic/logstash/pipeline/logstash.conf
```diff
@@ -15,126 +17,197 @@ filter {
...
 
-      if [nginx][access][payload] {
-        # remove \ from payload
-        mutate {
-          gsub => [
-            "[nginx][access][payload]", "[\\]", ""
-          ]
+      if [mine][id] {
+            # lookup by mine id
+        jdbc_streaming {
```
FreshGyan: Looking at the options for this plugin,  the default timeout for cache seems to be 5 seconds.  
https://www.elastic.co/guide/en/logstash/current/plugins-filters-jdbc_streaming.html#plugins-filters-jdbc_streaming-cache_expiration

I don't have . any strong feelings for what value this should be but maybe a minute or so? or longer?  mines don't change that much.
I don't know how many database hits we are talking so I may be wrong, @itchison any strong feelings?

elastic/logstash/pipeline/logstash.conf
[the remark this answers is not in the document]
FreshGyan: I think we should leave the output on stdout just to look at the pod in case it starts behaving abruptly?

Do you think it will be a significant performance hit or something @jlaplante333 ?

elastic/logstash/pipeline/logstash.conf
jlaplante333: Thank you Gyan.

jlaplante333: Changes have been made to dc json defintion and config.groovy for the pipeline

elastic/logstash/pipeline/logstash.conf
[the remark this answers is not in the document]
jlaplante333: I have reenabled it. Thanks.

elastic/logstash/pipeline/logstash.conf
```diff
@@ -1,11 +1,13 @@
 input {
+
+  # stdin {}
```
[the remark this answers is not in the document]
jlaplante333: Done

elastic/logstash/pipeline/logstash.conf
```diff
@@ -15,126 +17,197 @@ filter {
     method => "MURMUR3"
   }
 
+  # add environment
+  mutate {
+      add_field => { "[log][environment]" => "${ENVIRONMENT_SUFFIX}" }
+  }
+
+  mutate {
+    lowercase => [ "[log][environment]" ]
```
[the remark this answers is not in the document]
jlaplante333: Thanks

elastic/logstash/pipeline/logstash.conf
jlaplante333: I have changed the default timeout to 300 seconds/ 5 minutes. Thank you.

elastic/README.md
```diff
@@ -22,18 +25,14 @@ cd mds/elastic/logstash
...
-| ES_PASSWORD | elastic password   |
+Set up your deployment environemt with the existing Environment Variables
 
 Mount a pvc at the following path in the container:
 
 ```bash
 /app/logstash/monitor
 ```
 
+folder `app/logstash/data` should also be mounted to a persitant drive.
```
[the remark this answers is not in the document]
jlaplante333: Corrected. ;-)

jlaplante333:
Hello, all the conversations have been resolved and I have made the changes. Please review and if more changes are required for OpenShift, I will grant you edit access to the fork.

PR #648, opened 2019-04-05
Bug/mds 1667

thegentlemanphysicist:
This fixes the bug, but I'm not sure why.  Thoughts?


NWCalvank:
After a quick look at it, this seems to me like it's potentially related to the matrix producing an invalid transformation result for the mine pin SVG at certain viewport sizes: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform

I'm not sure if there's a better solution available, but I don't like that we don't have a proper handle on why this one works.

thegentlemanphysicist:
> After a quick look at it, this seems to me like it's potentially related to the matrix producing an invalid transformation result for the mine pin SVG at certain viewport sizes: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform
> 
> I'm not sure if there's a better solution available, but I don't like that we don't have a proper handle on why this one works.

I looked into this when I was hacking around on the ticket, even when I don't use a custom svg for the mine pin (using instead the default circle object) this dot still disappears on certain zoom levels

thegentlemanphysicist:
> > After a quick look at it, this seems to me like it's potentially related to the matrix producing an invalid transformation result for the mine pin SVG at certain viewport sizes: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform
> > I'm not sure if there's a better solution available, but I don't like that we don't have a proper handle on why this one works.
> > Looking further into this.
Maybe the problem is in how the transform is being processed by react-argis, but I don't think we have access to that feature.  It's handled under the hood by the <Map> component.

PR #649, opened 2019-04-05
Feature/ldap idir job

Jsyro:
trying again with fresh env.


migrations/sql/V2019.04.03__create_core_user_tables.sql
```diff
@@ -0,0 +1,63 @@
+CREATE TABLE IF NOT EXISTS core_user
```
FreshGyan: Is it just GitHub? The formatting seems a bit janky

python-backend/app/__init__.py
```diff
@@ -127,3 +122,14 @@ def default_error_handler(error):
                 'message': str(error)
             }
         }, getattr(error, 'code', 500)
+
+
+def register_scheduled_jobs(app):
```
FreshGyan: You don't have to refactor this as part of your PR but if we switch to using decorators instead of importing `_schedule_XXXX_jobs ` methods.
We can eliminate this whole function and every `_schedule_XXXX_jobs ` function in each scheduled taks

python-backend/requirements.txt
```diff
@@ -39,4 +39,5 @@ SQLAlchemy==1.2.10
 sqlalchemy-filters==0.8.0
 uwsgi==2.0.17.1
 uwsgitop==0.10
-Werkzeug==0.14.1
\ No newline at end of file
+Werkzeug==0.14.1
+ldap3==2.6.0
```
FreshGyan: Can you move this up to keep the list alphabetical?

python-backend/tests/users/resources/test_core_user_resource.py
```diff
@@ -0,0 +1,50 @@
...
+from app.extensions import db
+
+
+@pytest.fixture(scope="function")
+def setup_info(test_client):
+    cu = CoreUser.create(TEST_PARTY_PER_EMAIL_1, TEST_PARTY_PER_PHONE_EXT_1,
+                         TEST_PARTY_PER_PHONE_EXT_1)
+    cu.save()
+
+    yield dict(
```
FreshGyan: So if I understand it correctly, this runs before each test? Seeds the data, yields so that the test can run, and when that's done running it cleans up the record, is that correct?
Also thank you for moving towards a more isolated cleaner testing pattern, i really like this approach

migrations/sql/V2019.04.03__create_core_user_tables.sql
Jsyro: it's line wrapping cause the file has lots of tabs in it, it looks bad on github.

python-backend/app/__init__.py
Jsyro: I don't mind seeing it all in one place, but if it continues to grow this might be a good option.

python-backend/tests/users/resources/test_core_user_resource.py
Jsyro: correct, and this pattern is going to be made obsolete by James' work, but this is what we have been doing in the mean time.

python-backend/app/api/users/core/resources/core_user.py
```diff
@@ -0,0 +1,97 @@
...
+        if not core_users:
+            raise NotFound('No users found.')
+
+        return core_users
+
+
+class CoreUserResource(Resource, UserMixin):
+    parser = reqparse.RequestParser()
+    parser.add_argument(
+        'email', type=str, help='Users email address.', store_missing=False, location="form")
```
Jsyro: Note that this location parameter propogates to swagger.

python-backend/requirements.txt
Jsyro: Done

python-backend/app/api/users/core/resources/core_user.py
JustinMacaulay: Also, now that we are using separate parsers for the PUT and POST the required attribute is also really helpful for the swagger documentation.

PR #652, opened 2019-04-08
MDS-1656 Feature/ Create Variances & Upload/View Documents

NWCalvank:
@mmumby did the vast majority of the work on this PR. I've really just taken the final steps in hooking up the endpoints for document upload/view.

I also updated the action creators to use `CustomAxios`, which was added after Meredith left for her week off.


frontend/src/components/Forms/variances/AddVarianceForm.js
```diff
@@ -0,0 +1,145 @@
...
+    remove(this.state.uploadedFiles, { document_manager_guid: fileItem.serverId });
+    change("uploadedFiles", this.state.uploadedFiles);
+  };
+
+  render() {
+    return (
+      <Form
+        layout="vertical"
+        onSubmit={this.props.handleSubmit(this.props.onSubmit(this.state.documentNameGuidMap))}
+      >
```
mmumby: isn't `this.props.handleSubmit` and `this.props.onSubmit` the same function? 

would either of these work? 
```
onSubmit={this.props.handleSubmit(this.state.documentNameGuidMap)}
```
OR
```
onSubmit={this.props.onSubmit(this.state.documentNameGuidMap)}
```

NWCalvank: They're not! If I understand this correctly, it's some of the most confusing magic I've ever seen:
https://github.com/erikras/redux-form/issues/1629#issuecomment-334025455

Basically, you define your own `onSubmit`, which is your **actual** event handler. `handleSubmit` is magically provided to your component by Redux Forms. In order to use your handler (`onSubmit` in this case), you pass it as an argument to `handleSubmit`.

PR #654, opened 2019-04-08
Feature/incident reporting

Jsyro:
backend CRU for mine_incidents. Also reqparsing trim can be set class wide, not argument wide.


Jsyro:
Still business unknowns that need to be resolved, I don't want to write tests until the endpoints are firm.

PR #659, opened 2019-04-09
MDS-1686 Refactor Mine resource to follow restplus standards

JustinMacaulay:
Im sorry about how large this PR is, it touches a lot of files since the mine object is used in most areas of our code. This will need thorough regression testing and major testing in general.

- adds a new api_model for the mines which defines all the return values for endpoints.
- modified all frontend files to expect a mine_guid instead of a guid.
- added a new resource for getting the mines to display on the map.
- Broke the mine resource into two separate resources
- fixed frontend and backend tests.


frontend/src/tests/components/admin/__snapshots__/NewMinespaceUser.spec.js.snap
```diff
@@ -64,15 +64,15 @@ ShallowWrapper {
             Array [
               Object {
                 "label": "New Mine-BLAH6194",
-                "value": "New Mine~fc72863d-83e8-46ba-90f9-87b0ed78823f",
+                "value": "New Mine~undefined",
```
mmumby: The value for the `mine_guid` in this snapshot is being read as `undefined` - double check it is passed into the component correctly, it shouldn't be undefined.

NWCalvank: Meredith with the eagle eyes! 👀

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -23,114 +23,119 @@
...
         help='Any additional notes to be added to the mine.',
         trim=True,
-        store_missing=False)
-    parser.add_argument(
-        'tenure_number_id',
-        type=int,
-        help='Tenure number for the mine.',
-        trim=True,
-        store_missing=False)
+        store_missing=False,
```
Jsyro: I'm starting to waver on the store_missing flag on posts.... without it it will be defaulted to null, which is what the .get() method will do anyways. I feel like i have advocated the other way. but the store_missing is only needed on the put (where the fields supplied actually matter), i think store_missing=False, should be removed from the ListResources.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -332,14 +317,14 @@ def put(self, mine_no_or_guid):
         if 'mine_region' in data:
             mine.mine_region = data['mine_region']
         mine.save()
-
+        current_app.logger.error(f'made it here {tenure}')
```
Jsyro: debug line.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -332,14 +317,14 @@ def put(self, mine_no_or_guid):
...
         # Tenure validation
         if tenure:
             tenure_exists = MineralTenureXref.find_by_tenure(tenure)
+            current_app.logger.error(f'made it here {tenure} and {tenure_exists}')
```
Jsyro: same here.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -380,5 +365,61 @@ def get(self):
             mines = Mine.find_by_name_no_permit(search_term)
         else:
             mines = Mine.find_by_mine_name(name_search)
-        result = list(map(lambda x: {**x.json_by_name(), **x.json_by_location()}, mines))
+        result = list(
```
Jsyro: Are we leaving this one cause it's different? wondering why this doesn't have a model and envelopes etc.

python-backend/app/api/mines/status/models/mine_status.py
```diff
@@ -55,18 +49,27 @@ def create_mine_status_labels_list(self):
...
+    def __repr__(self):
+        return '<MineStatus %r>' % self.mine_status_guid
+
+    def validate_status_code_exists(self, mine_status_xref, mine_status_code, code_or_description):
+        try:
+            return mine_status_xref[mine_status_code][code_or_description]
+        except KeyError:
+            return None
+
+    # def json(self, show_mgr=True):
```
Jsyro: remove commented code?

python-backend/app/api/mines/namespace/mines.py
```diff
@@ -21,7 +22,10 @@
 
 api = Namespace('mines', description='Mine related operations')
 
-api.add_resource(MineResource, '', '/<string:mine_no_or_guid>')
+api.add_resource(MineResource, '/<string:mine_no_or_guid>')
+api.add_resource(MineListResource, '')
+api.add_resource(MineMapResource, '/map-list')
```
NWCalvank: I don't think this is an appropriate endpoint for our API to have. The API knows nothing about the map on the frontend. It's probably more reasonable for this to be something like `(MineBasicInfoResource, '/basic')`

Not sure if we're already referring to something as `BasicInfo` elsewhere. I'm not committed to that name if we are -- but I think it serves the purpose of illustrating my point

python-backend/app/api/applications/resources/application.py
```diff
@@ -22,24 +22,35 @@
 class ApplicationListResource(Resource, UserMixin):
     parser = reqparse.RequestParser()
     parser.add_argument(
-        'application_no', type=str, required=True, help='Number of the application being added.')
+        'application_no',
+        type=str,
+        required=True,
+        help='Number of the application being added.',
+        location='form')
```
NWCalvank: What's the story behind these application changes? Seems unrelated to the rest of the PR

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
```diff
@@ -80,7 +80,7 @@ def put(self, expected_document_guid):
             if not mine_doc:
                 return self.create_error_payload(404, 'Mine Document not found'), 404
 
-            expected_document.mine_documents.append(mine_doc)
+            expected_document.related_documents.append(mine_doc)
```
NWCalvank: Not sure why expected_document is being updated in this PR, but the line(s) below should be updated to use `save()` as per a recent chat with @Jsyro

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -23,114 +23,119 @@
...
+            'current_page': pagination_details.page_number,
+            'total_pages': pagination_details.num_pages,
+            'items_per_page': pagination_details.page_size,
+            'total': pagination_details.total_results,
+        }
+
+    @api.expect(parser)
+    @api.doc(
+        description=
+        'This endpoint creates a new mine with the information provided and returns it. If an error occurs the appropriate response is returned.'
```
NWCalvank: I'd like us to avoid using "This endpoint" at the start of every description. An active, user-action-focused voice is much more direct and concise. "Create a new mine."

The "with the information provided" is implied by the nature of a REST API, imo. "and returns it" should be obvious from the rest of the documentation for the endpoint.

And I think it's reasonable for someone to expect an error to be returned if an error occurs. Adding that to the description is unnecessarily wordy

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -23,114 +23,119 @@
...
-            'mine_no_or_guid':
-            'Mine number or guid. If not provided a paginated list of mines will be returned.'
-        })
+            'per_page': 'The number of results to be returned per page.',
+            'page': 'the current page number to be displayed.',
+            'search': 'The search term.',
+            'commodity': 'A commodity to filter the mine list by.',
+            'status': 'A mine status to filter the mine list by.',
+            'tenure': 'A mine tenure type to filter the mine list by.',
+            'region': 'A mine region to filter the mine list by.',
```
NWCalvank: The grammar nerd in me is really bothered by the prepositions at the end of these statements

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -23,114 +23,119 @@
...
+        required=True,
+        location='json')
 
     @api.doc(
         params={
-            'mine_no_or_guid':
-            'Mine number or guid. If not provided a paginated list of mines will be returned.'
-        })
+            'per_page': 'The number of results to be returned per page.',
+            'page': 'the current page number to be displayed.',
```
NWCalvank: Missing capitalization

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -23,114 +23,119 @@
...
+            'search': 'The search term.',
+            'commodity': 'A commodity to filter the mine list by.',
+            'status': 'A mine status to filter the mine list by.',
+            'tenure': 'A mine tenure type to filter the mine list by.',
+            'region': 'A mine region to filter the mine list by.',
+            'major': 'True or false, filters the mine list by major mines or regional mines.',
+            'tsf': 'True or false, filter the mine list by mines with or without a TSF.',
```
NWCalvank: We don't mention the data type in any of the other params here. Is there a reason why we'd say "True or false?" And if we do, we should probably say "Boolean."

python-backend/app/api/mines/mine/resources/mine_map.py
```diff
@@ -0,0 +1,74 @@
...
+        # Below caches the mine map response object in redis with a timeout.
+        # Generating and jsonifying the map data takes 4-7 seconds with 50,000 points,
+        # so caching seems justified.
+        #
+        # TODO: Use some custom representation of this data vs JSON. The
+        # json string is massive (with 50,000 points: 16mb uncompressed, 2.5mb compressed).
+        # A quick test using delimented data brings this down to ~1mb compressed.
+        map_result = cache.get(MINE_MAP_CACHE)
+        last_modified = cache.get(MINE_MAP_CACHE + '_LAST_MODIFIED')
+        if not map_result:
```
NWCalvank: Flagging this for later. I think this piece of the API knows far too much about how it's being consumed. It's location data, but there's nothing (from what I can tell) that forces it to be used by a map.

This should probably be more fully decoupled from the frontend use case at some point. Thoughts @Jsyro @nayr974 ?

python-backend/app/api/mines/status/models/mine_status.py
```diff
@@ -55,18 +49,27 @@ def create_mine_status_labels_list(self):
...
+    #     status_labels_list = self.create_mine_status_labels_list()
+    #     return {
+    #         'mine_status_guid': str(self.mine_status_guid),
+    #         'mine_guid': str(self.mine_guid),
+    #         'mine_status_xref_guid': str(self.mine_status_xref.mine_status_xref_guid),
+    #         'status_values': status_values_list,
+    #         'status_labels': status_labels_list,
+    #         'effective_date': self.effective_date.isoformat(),
+    #         'expiry_date': self.expiry_date.isoformat()
+    #     }
```
NWCalvank: Comment to remove? Why would the `mine_status` json function be changed/removed here?

python-backend/tests/auth/test_expected_auth.py
```diff
@@ -50,34 +51,50 @@
...
+     (RequiredDocumentResource, "get", [MINE_VIEW]),
      (PermitAmendmentResource, "post", [MINE_CREATE]),
      (PermitAmendmentResource, "put", [MINE_CREATE]),
      (PermitAmendmentResource, "delete", [MINE_ADMIN]),
      (PermitAmendmentDocumentResource, "post", [MINE_CREATE]),
      (PermitAmendmentDocumentResource, "put", [MINE_CREATE]),
      (PermitAmendmentDocumentResource, "delete", [MINE_CREATE]),
-     (RequiredDocumentResource, "get", [MINE_VIEW]), (MinespaceUserResource, 'get', [MINE_ADMIN]),
-     (MinespaceUserResource, 'post', [MINE_ADMIN]), (MinespaceUserResource, 'delete', [MINE_ADMIN]),
+     (RequiredDocumentResource, "get", [MINE_VIEW]),
```
NWCalvank: This doesn't look quite alphabetical.

frontend/src/components/mine/MineDashboard.js
```diff
@@ -123,7 +125,8 @@ export class MineDashboard extends Component {
 
   loadMineData(id) {
     this.props.fetchMineRecordById(id).then(() => {
-      this.props.fetchApplications({ mine_guid: this.props.mines[id].guid });
+      this.props.fetchApplications({ mine_guid: this.props.mines[id].mine_guid });
+      this.props.fetchPermits({ mine_guid: this.props.mines[id].mine_guid });
```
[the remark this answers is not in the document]
JustinMacaulay: We were fetching the permits on there own for the permits tab so I brought it out to the dashboard so that the permits could be fetched for the summary tab and remove most of theyre information from the mine object.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -156,9 +161,15 @@ export class Dashboard extends Component {
...
-    this.props.fetchMineRecords(params).then(() => {
-      this.setState({ mineList: true });
-    });
+    if (remainingParams.map) {
+      this.props.fetchMineRecordsForMap().then(() => {
+        this.setState({ mineList: true });
+      });
+    } else {
+      this.props.fetchMineRecords(params).then(() => {
+        this.setState({ mineList: true });
```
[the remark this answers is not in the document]
mmumby: Good catch! 
From what I can see, the `map=true` param is still useful, just in a different way than it was being used before. (ie frontend check vs backend) I can see how it will be confusing going forward. 

Better approaches:
Remove that param from the url, only use the `this.fetchDataFromUrl` to update the mineList using `this.props.fetchMineRecords`

`this.props.fetchMineRecordsForMap` could be run in the background on `componentDidMount()`, and stored in a different const in the mineReducer, - The benefit here, is that the user will not have to wait for all the mine information to be fetched when they navigate to the map view.

frontend/src/tests/components/admin/__snapshots__/NewMinespaceUser.spec.js.snap
JustinMacaulay: Missed one part of the mock data update. Fixed and good catch!

frontend/src/tests/reducers/minespaceReducer.spec.js
```diff
@@ -51,7 +51,7 @@ describe("mineReducer", () => {
         mine_status: [],
         mine_tailings_storage_facility: [],
```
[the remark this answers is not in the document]
JustinMacaulay: Yes and changed.

python-backend/app/api/applications/resources/application.py
JustinMacaulay: This was me updating the previous endpoints I had worked on to bring them inline with some new information we discovered.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -179,6 +179,21 @@ export const fetchMineRecords = (params) => (dispatch) => {
     .finally(() => dispatch(hideLoading()));
 };
 
+export const fetchMineRecordsForMap = () => (dispatch) => {
+  dispatch(request(reducerTypes.GET_MINE_RECORDS));
```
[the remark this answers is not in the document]
mmumby: correct, it _should_ be different.

However, it appears he is using the same action as `fetchMineRecords` and the data is being kept in the same spot. So in this situation, the reducer being the same is fine. 


I did recommend in a comment on `Dashboard.js` to make this its own action and store the data in a different const than the `mineReducer`, the benefit being that this can be fitched when the component mounts vs when the user navigates to the tab thus delaying loading time. If that switch is made, then this would need to have its own reducerType and actionType.

python-backend/app/api/documents/expected/resources/expected_document_uploads.py
JustinMacaulay: A lot of these random changes are due to the fact that what the frontend is expecting is named differently than what is on the models. I had to change around a lot of references in other classes due to this.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: True. I will remove them from the parser for the posts.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: removed.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: Also removed.

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: This one is different as it is returning an object with mixed values from several models. I tried to get it to work with response marshalling but was unable.

python-backend/app/api/mines/mine/resources/mine_map.py
```diff
@@ -0,0 +1,74 @@
...
+    'longitude': fields.String,
+})
+
+mine_map_list = api.model(
+    'MineMap', {
+        'mine_guid': fields.String,
+        'mine_name': fields.String,
+        'mine_no': fields.String,
+        'mine_location': fields.Nested(mine_location)
+    })
```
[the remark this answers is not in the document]
JustinMacaulay: Forgot to move these ones. Thanks!

python-backend/app/api/mines/mine_api_models.py
```diff
@@ -0,0 +1,117 @@
...
+        'verified_status': fields.Nested(MINE_VERIFIED)
+    })
+
+MINE = api.inherit(
+    'Mine', MINES, {
+        'mine_location': fields.Nested(MINE_LOCATION),
+        'mine_expected_documents': fields.List(fields.Nested(MINE_EXPECTED_DOCUMENT)),
+    })
+
+MINE_LIST_MODEL = api.model(
```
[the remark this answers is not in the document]
JustinMacaulay: I added model to all of them and moved all the models to the constants model file.

python-backend/tests/mines/resources/test_mine_list_resource.py
```diff
@@ -6,39 +6,28 @@ def test_get_mines(test_client, auth_headers):
     get_resp = test_client.get('/mines', headers=auth_headers['full_auth_header'])
     get_data = json.loads(get_resp.data.decode())
     assert get_resp.status_code == 200
-    assert get_data['mines'][0]['guid'] == TEST_MINE_GUID
+    assert get_data['mines'][0]['mine_guid'] == TEST_MINE_GUID
 
 
 def test_get_mines_with_map_true(test_client, auth_headers):
     get_resp = test_client.get('/mines?map=true', headers=auth_headers['full_auth_header'])
```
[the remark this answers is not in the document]
JustinMacaulay: Removed this test.

python-backend/app/api/mines/mine/resources/mine_map.py
nayr974: Primarily naming and comments at this point, correct?

NWCalvank: Yeah, I don't think it's any functional changes at this point, which is good -- but generalizing the naming could help us steer clear of adding map-specific functionality in the future.

frontend/src/components/mine/MineDashboard.js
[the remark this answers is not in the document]
NWCalvank: Cool. Marking as resolved -- we can carefully confirm in manual testing that this doesn't prevent permits from being loaded on any views

python-backend/app/api/applications/resources/application.py
NWCalvank: Okay. I'm marking it as resolved. With a mammoth PR like this, it's probably best to split out these changes if they're not 100% related to the primary PR goal

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -179,6 +179,21 @@ export const fetchMineRecords = (params) => (dispatch) => {
...
+  dispatch(showLoading());
+  return CustomAxios()
+    .get(ENVIRONMENT.apiUrl + API.MINE_MAP_LIST, createRequestHeader())
+    .then((response) => {
+      dispatch(success(reducerTypes.GET_MINE_RECORDS));
+      dispatch(mineActions.storeMineList(response.data));
+      dispatch(hideLoading());
+      return response;
+    })
+    .catch(() => dispatch(error(reducerTypes.GET_MINE_RECORD)))
```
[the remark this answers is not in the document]
JustinMacaulay: Its also that way in the fetchMineRecords which I based this action creator off of. Should that also be changed?

python-backend/app/api/mines/mine/resources/mine.py
JustinMacaulay: I have modified these descriptions.

python-backend/app/api/mines/status/models/mine_status.py
JustinMacaulay: If its no longer used by any returns. I am not sure about that though so I have it commented out for now and would like some more rigorous testing.

NWCalvank: If the intention is to remove it (assuming that nothing blows up in manual testing) then we can just remove it. We have a git history of it if we ever need to get it back

frontend/src/actionCreators/mineActionCreator.js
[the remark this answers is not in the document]
NWCalvank: I'll defer to @mmumby on that. If we haven't had to update `fetchMineRecords` then this probably doesn't cause any real problems, but I would expect the `error` to match the `success`

python-backend/app/api/mines/mine_api_models.py
```diff
@@ -0,0 +1,117 @@
...
+    'MineList', {
+        'mines': fields.List(fields.Nested(MINES)),
+        'current_page': fields.Integer,
+        'total_pages': fields.Integer,
+        'items_per_page': fields.Integer,
+        'total': fields.Integer,
+    })
```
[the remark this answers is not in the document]
JustinMacaulay: Yes, this is 100% a way you could handle this. I only ran into this return in one place though.

frontend/package-lock.json
```diff
@@ -1191,6 +1191,7 @@
       "version": "2.6.1",
       "resolved": "https://registry.npmjs.org/async/-/async-2.6.1.tgz",
       "integrity": "sha512-fNEiL2+AZt6AlAw/29Cr0UDe4sRAHCpEHh54WMz+Bb7QfNcFw4h3loofyJpLeQs4Yx7yuqu/2dLgM5hKOs6HlQ==",
+      "optional": true,
```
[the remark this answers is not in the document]
JustinMacaulay: I ran NPM CI so I'm not sure where this came from.

python-backend/tests/auth/test_expected_auth.py
JustinMacaulay: I put it in alphabetical order now.

python-backend/app/api/mines/namespace/mines.py
nayr974: mines/coordinates?

frontend/src/actionCreators/mineActionCreator.js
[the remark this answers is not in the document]
mmumby: Nathan is correct. 

It doesn't break anything, This is used more for logging purposes, but the request, success, and error should all be the same ReducerType.

python-backend/app/api/mines/mine_api_models.py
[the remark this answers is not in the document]
NWCalvank: Nice. I'll mark this as resolved since it's not direct feedback for this PR. I think we could benefit from this pattern in the future, though. Thanks, Justin

frontend/package-lock.json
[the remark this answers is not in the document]
NWCalvank: Very strange. Maybe you have a different version of `npm`? This shouldn't be a big deal, so I'll mark this as resolved. I just wanted to highlight this in case you were using the wrong command

python-backend/app/api/mines/namespace/mines.py
NWCalvank: I'd like to see this changed, but I might just split it into a separate tech debt ticket that deals with removing references to the map from the backend since those references exist in a few places

python-backend/app/api/mines/status/models/mine_status.py
JustinMacaulay: Un-commented as this function was used in other resources.

PR #669, opened 2019-04-15
Feature/incident reporting

Jsyro:
Full slice PR, with the new restplus patterns. FE has Create and Read only. writing unit tests now .


python-backend/app/api/mines/incidents/resources/mine_incident_followup_types.py
```diff
@@ -0,0 +1,26 @@
...
+from datetime import datetime
+from werkzeug.exceptions import BadRequest, NotFound, InternalServerError
+
+from app.extensions import api
+from app.api.utils.resources_mixins import UserMixin
+from app.api.utils.access_decorators import requires_role_mine_view, requires_role_mine_create
+
+from ..models.mine_incident_followup_type import MineIncidentFollowupType
+
+mine_incident_followup_type_model = api.model(
```
JustinMacaulay: Can you please move these to the mine_api_models file.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -0,0 +1,174 @@
...
+        'followup_type_code': fields.String,
+        'followup_inspection_no': fields.String,
+        'closing_report_summary': fields.String
+    })
+
+
+class MineIncidentListResource(Resource, UserMixin):
+    parser = reqparse.RequestParser(trim=True)
+    #required
+    parser.add_argument(
```
JustinMacaulay: For the required fields is there a reason they don't have required=true?

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -0,0 +1,174 @@
...
+        'closing_report_summary',
+        help='Report from mine in reaction to the incident',
+        location='json',
+        type=str,
+        store_missing=False)
+
+    @api.marshal_with(mine_incident_model, code=200)
+    @requires_role_mine_view
+    def get(self, mine_incident_guid, mine_guid=None):
```
JustinMacaulay: If the mine_guid is not used can you please remove it.

python-backend/app/api/mines/incidents/resources/mine_incident_followup_types.py
Jsyro: Moved.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: Yea there were two paths to get here, but we discussed and agreed there should only be one.

frontend/src/components/modalContent/AddIncidentModal.js
```diff
@@ -0,0 +1,25 @@
...
+  followupActionOptions: PropTypes.arrayOf(CustomPropTypes.dropdownListItem).isRequired,
+};
+
+const defaultProps = {
+  title: "",
+};
+
+export const AddIncidentModal = (props) => (
+  <div>
+    <AddIncidentForm {...props} />
```
NWCalvank: Best to pass in props explicitly

frontend/src/components/common/RenderDateTime.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
NWCalvank: This is a new file?

frontend/src/components/mine/Incidents/MineIncidents.js
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  openMineIncidentModal = (event) => {
+    event.preventDefault();
+    this.props.openModal({
+      props: {
+        onSubmit: this.handleAddMineIncident,
+        title: ModalContent.ADD_INCIDENT(this.props.mine.mine_name),
+        mineGuid: this.props.mine.mine_guid,
+        followupActionOptions: createDropDownList(
```
NWCalvank: It looks like you're already doing this in `frontend/src/selectors/staticContentSelectors.js`. If it's in the Redux Store, you should be able to just access it from there instead of regenerating the dropdown list every time (this is a fairly slow operation for JS)

frontend/src/utils/helpers.js
```diff
@@ -37,7 +37,6 @@ export const createItemIdsArray = (array, idField) => array.map((item) => item[i
 
 export const createDropDownList = (array, labelField, valueField) =>
   array.map((item) => ({ value: item[valueField], label: item[labelField] }));
-
 // Function to create a hash given an array of values and labels
```
NWCalvank: Was this an accident? Seems like an odd change. Obviously not a big deal either way

migrations/sql/V2019.04.08.09.32__create_incident_reports.sql
```diff
@@ -0,0 +1,48 @@
...
+    update_timestamp                                timestamp with time zone DEFAULT now()   NOT NULL
+);
+ALTER TABLE mine_incident_followup_type OWNER TO mds;
+COMMENT ON TABLE mine_incident_followup_type IS 'lookup table for the types of EMPR reactions to a given mine_incident.';
+
+CREATE TABLE IF NOT EXISTS mine_incident
+(
+    mine_incident_id                serial                      PRIMARY KEY,
+    mine_incident_id_year           integer                     NOT NULL,
+    mine_incident_guid              uuid DEFAULT gen_random_uuid()   NOT NULL,
```
NWCalvank: What's the reason for a guid and an id?

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -0,0 +1,159 @@
...
+    @api.expect(parser)
+    @api.doc(description='creates a new incident for the mine')
+    @api.marshal_with(MINE_INCIDENT_MODEL, code=201)
+    @requires_role_mine_create
+    def post(self, mine_guid):
+        mine = Mine.find_by_mine_guid(mine_guid)
+        if not mine:
+            raise NotFound('Mine not found')
+
+        data = self.parser.parse_args()
```
NWCalvank: If this raises an exception from the `required=True` args, the inconsistent error response will be returned.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -0,0 +1,159 @@
...
+        incident = MineIncident.create(
+            mine,
+            data['incident_timestamp'],
+            data['incident_description'],
+            followup_type_code=data['followup_type_code'] or 'UND',
+            followup_inspection_no=data['followup_inspection_no'],
+            reported_timestamp=data['reported_timestamp'],
+            reported_by=data['reported_by'],
+            reported_by_role=data['reported_by_role'],
```
NWCalvank: I get that this works because `store_missing=False` isn't set, but I'm curious why you'd prefer this over `.get()`

python-backend/app/api/mines/namespace/mines.py
```diff
@@ -49,3 +51,12 @@
 api.add_resource(
     VarianceUploadedDocumentsResource,
     '/<string:mine_guid>/variances/<string:variance_id>/documents/<string:mine_document_guid>')
+
+api.add_resource(MineIncidentListResource, '/<string:mine_guid>/incidents')
+
+api.add_resource(MineIncidentResource, '/<string:mine_guid>/incidents/<string:mine_incident_guid>',
+                 '/incidents/<string:mine_incident_guid>')
```
NWCalvank: Two endpoints to one resource? That's a pattern we're avoiding now, no? What's the reason for implementing it this way?

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -10,80 +10,66 @@
...
+        'note',
+        type=str,
+        store_missing=False,
+        help='A note to include on the variance. Limited to 300 characters.')
+    parser.add_argument(
+        'issue_date', store_missing=False, help='The date on which the variance was issued.')
+    parser.add_argument(
+        'received_date', store_missing=False, help='The date on which the variance was received.')
+    parser.add_argument(
+        'expiry_date', store_missing=False, help='The date on which the variance expires.')
```
NWCalvank: Autoformatter, I assume?

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
```diff
@@ -11,7 +11,7 @@
 
 
 class MinespaceUserMineResource(Resource, UserMixin, ErrorMixin):
-    parser = reqparse.RequestParser()
+    parser = reqparse.RequestParser(trim=True)
```
NWCalvank: Possibly pedantic, but I think all of these `trim=True` resource changes would have made a nice, separate PR. Then any discussion around it wouldn't need to be mixed in with incidents, and the incidents PR would be considerably smaller

frontend/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -0,0 +1,62 @@
...
+  transformRowData = (incidents, actions) =>
+    incidents.sort(this.sortByDateOrID).map((incident) => ({
+      key: incident.incident_id,
+      mine_incident_report_no: incident.mine_incident_report_no,
+      incident_timestamp: formatDate(incident.incident_timestamp),
+      reported_timestamp: formatDate(incident.reported_timestamp),
+      reported_by: incident.reported_by,
+      followup_action: actions.find(
+        (x) => x.mine_incident_followup_type_code === incident.followup_type_code
+      ).description,
```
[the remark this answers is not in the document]
Jsyro: Fair point on the possible access. I also considered the hashing step, but the number of items is 4 in this list, so I don't think the overhead of setting up the hash it worth it.

frontend/src/components/mine/Incidents/MineIncidents.js
Jsyro: I need to access the object list, and the dropdown converted list, so I stopped using the selector one, cause the form doesn't access the store. 

I'll delete it from staticContentSelectors.js

frontend/src/components/common/RenderDateTime.js
Jsyro: Yup, we only had RenderDate, so it's a separate component for DateTime.

frontend/src/components/modalContent/AddIncidentModal.js
Jsyro: Yea It's not that long a list.

frontend/src/utils/helpers.js
Jsyro: Accident.

frontend/src/components/common/RenderDateTime.js
NWCalvank: Gotcha. I didn't realize that this was slightly different from RenderDate. It looks like this is the difference?
```javascript
      showTime={{ format: "HH:mm" }}
      format="YYYY-MM-DD HH:mm"
      style={{ width: "100%" }}
```

My inclination would be to pass in `format` in the props and default to `YYYY-MM-DD` if it's not provided. Or something like a `minutes` boolean to check whether or not to show minutes.

Is there a reason why you preferred to make a new component instead of adding an item to props?

Jsyro: There is some  minuta about the props that we should deal with centrally instead of having everyone figure out... i'm not opposed to just flagging, but it was the easiest way to debug and get it started.

migrations/sql/V2019.04.08.09.32__create_incident_reports.sql
Jsyro: We never want to expose the serial id, used for FK's, anything that needs to be uniquely identified from the outside needs a guid.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: I think I recommended the reverse when both the POST and PUT were in the same resource, but the values have to get defaulted at somepoint, I liked the idea of nulling it as high as possible in the reqparser, then defaulting in the create helper. it also saves some boiler plate.

python-backend/app/api/mines/namespace/mines.py
Jsyro: You are right, meant to remove that, thanks.

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
Jsyro: Fair point, that also would have made it more visible. Differentiating based on trailing or leading whitespace seems like a bad idea, but I didn't exactly have a conversation. We can have one though.

python-backend/app/api/mines/variances/resources/variance.py
Jsyro: Yea i think so .

python-backend/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: This exists in lots of places, when we go to our customReqParser, this will get caught everywhere.

frontend/src/components/common/RenderDateTime.js
NWCalvank: Not sure what you mean by dealing with props centrally. They're only relevant to a specific component. Maybe I'm misunderstanding you?

Anyway, if you default to the existing behaviour and just add the `minutes` boolean/flag then no consumers of this component are effected. That's my personal preference here, given that it's almost entirely copy/paste of the `RenderDate` component and the different caused by the minutes flag is quite minimal.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
NWCalvank: But it's only relevant on resources where we've added `required=True`, so I figured it was worth mentioning

python-backend/app/api/mines/incidents/resources/mine_incidents.py
NWCalvank: I dig it. Marking as resolved

python-backend/app/api/users/minespace/resources/minespace_user_mine.py
NWCalvank: I think it's a good change to have made. My only point was it might be easier to keep an eye on something fairly global like this if it has its own PR. For the future 😄

Marking as resolved

migrations/sql/V2019.04.08.09.32__create_incident_reports.sql
NWCalvank: Makes sense 👍

frontend/src/components/mine/Incidents/MineIncidents.js
mmumby: Transforming `followupActionOptions` into a dropdown would be best done in the selector. You can utilize `createSelector` for this, there are a lot of examples in staticContentSelectors.js that do the exact same thing, 


```
export const getDropdownfollowupActionOptions = createSelector(
  [followupActionOptions],
  (options) => createDropDownList(options, "description", "mine_incident_followup_type_code")
);
```
you can import `getDropdownfollowupActionOptions` like you do `followupActionOptions` and pass it into the form the same way.

Jsyro: I mean I could do it through a selector, but I already have the list of things I want. Does it make sense to have two props, one for the transformed list and one for the normal list, just so i can hand one of them to a child? just making sure i get it.

frontend/src/components/common/RenderDateTime.js
Jsyro: Not sure this is the way you would do it, so let me know if the flag is ok.

frontend/src/components/mine/Incidents/MineIncidents.js
mmumby: Yeah, redux state transformations should occur in the selectors.

frontend/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+  {
+    title: "Reported By",
+    dataIndex: "reported_by",
+  },
+  {
+    title: "EMPR Action",
+    dataIndex: "followup_action",
+    render: (action) => (
+      <div title="followup_action">{action ? action.description : "error..."}</div>
```
NWCalvank: I think we have error string constants. @mmumby ?

mmumby: yeah there are constants, with this being a table I'm thinking the null view should be "N/A" or "-" which is consistent with the other tables. We don't use "error..." anywhere else.

Jsyro: This should never happen, but we didn't want the app to crash if it did.... so is there more discussion we want to have to handle and show subtle errors? or should we just keep it simple and make it '-'?

Jsyro: We already have the three letter code, i'm just going to display that if something goes wrong.

PR #670, opened 2019-04-15
MDS-1418 & MDS-1687 Feature/Sort Contacts Table & Refactor Party Resources for Restplus

NWCalvank:
- Add sorting by `party_name`
- Simplify contact search implementation on Frontend
- Display Contact names as `lastname, first_name` or `party_name`
- Split `PartyResource` into `PartyResource` and `PartyListResource`
- Backwards-compatibility with existing `mine_party_appt` requirements for `party_instance.json()`
- Proper API documentation (at least largely complete compared to before)
- Proper party marshalling
- Split party address into a separate `address` table, with a `party_address_xref` secondary table for managing the relationship(s)
- Remove `parties/search` endpoint
- Fix tests that weren't actually testing anything
- Update tests for new endpoints


NWCalvank:
Merging latest develop appears to have broken my build. Closing this PR to nuke everything.

NWCalvank:
🤞 ✅

migrations/sql/V2019.04.12.13.45__create_address_table.sql
```diff
@@ -0,0 +1,86 @@
...
+);
+
+ALTER TABLE address OWNER TO mds;
+
+CREATE SEQUENCE address_id_seq
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
```
Jsyro: _serial_ as the column type is a short hand for exactly this. 

`    address_id        serial   PRIMARY KEY,` will handle the sequence.

python-backend/app/api/parties/party/models/address.py
```diff
@@ -0,0 +1,68 @@
...
+            'post_code': self.post_code
+        }
+
+    @classmethod
+    def create(
+            cls,
+            address_type_code=None,
+            email=None,
+            first_name=None,
+            phone_ext=None,
```
Jsyro: unused parameters (email, phone, first name)

python-backend/app/api/parties/party/models/party.py
```diff
@@ -69,15 +65,7 @@ def json(self, show_mgr=True, relationships=[]):
...
-            'address': [{
-                'suite_no': self.suite_no,
-                'address_line_1': self.address_line_1,
-                'address_line_2': self.address_line_2,
-                'city': self.city,
-                'sub_division_code': self.sub_division_code,
-                'post_code': self.post_code,
-                'address_type_code': self.address_type_code
-            }]
+            'address': self.address[0].json() if len(self.address) > 0 else [{}]
```
Jsyro: empty lists are falsy, don't need to do a length check. should this return empty list? or list with empty object.

python-backend/app/api/parties/party/models/party_address.py
```diff
@@ -0,0 +1,17 @@
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.schema import FetchedValue
+
+from ....utils.models_mixins import Base
+from app.extensions import db
+
+class PartyAddress(Base):
```
Jsyro: xref suffix just for consistency?

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -54,113 +51,59 @@ class PartyResource(Resource, UserMixin, ErrorMixin):
...
-            sub_division_code=data.get('sub_division_code'),
-            post_code=data.get('post_code'))
+    @api.marshal_with(PARTY, code=200)
+    def get(self, party_guid):
+        try:
+            uuid.UUID(party_guid)
+        except ValueError:
+            raise BadRequest('Invalid Party guid')
 
+        party = Party.find_by_party_guid(party_guid)
```
Jsyro: I think we should handle the formatting of the guid in the find helper. find_by_party_guid should return None or throw an error if it is malformed... thoughts?

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -171,11 +114,26 @@ def put(self, party_guid):
...
+        # This condition should never pass, as every party should be
+        # associated with an address, but it's been added for runtime safety,
+        # in case the API is circumvented and a party is created without an
+        # address
+        if len(existing_party.address) == 0:
+            address = Address.create()
+            existing_party.address.append(address)
+
+        for key, value in data.items():
+            setattr(existing_party.address[0], key, value)
```
Jsyro: should this exclude party_type_code? I don't think we want to all the API to change an Org to a Person or visa versa? Always check the parser arguments to see if there is a value we need to protect.

migrations/sql/V2019.04.12.13.45__create_address_table.sql
NWCalvank: Yeah, I think I saw that in one of your other PRs. I just copied this format from our big, initial migration file. If we'd like to use serial instead in future migrations, I can update this one so that it can be used as a proper point of reference. Otherwise, I don't think it should really matter

python-backend/app/api/parties/party/models/party.py
NWCalvank: I'm strongly against coercing an empty list. In python, an empty list is false. In Javascript, an empty array is truthy. Considering those are the two big languages in this repo, I think we should avoid that conflicting edge case

As far as the return value, the entire `json` method is all for backwards compatibility. Since every party had exactly one address, and it was in the form of an object as the first item in an array, I went with this format `[{}]`. I updated the frontend to handle an empty array where I noticed it, but this was the safer choice, imo.

python-backend/app/api/parties/party/models/party_address.py
NWCalvank: I didn't use an `Xref` suffix in the Variances class, so it wouldn't have been consistent with that. I, personally, think that this is plenty clear (party + address = PartyAddress) and the `Xref` is a bit of info about the db leaking into the application layer that probably isn't needed.

But I'm also fine with renaming it if we think it helps.

migrations/sql/V2019.04.12.13.45__create_address_table.sql
Jsyro: I don't think it matters, that base schema was generated by pg_dump, so it is as verbose as it can be. the name of the sequence has to be perfect to it to play along with sqlalchemy (I think), but you have named it properly.

python-backend/app/api/parties/party/resources/party_resource.py
NWCalvank: Oh, yeah. Definitely. I did that for Variances. Leaving it here was an accident. Thanks!

python-backend/app/api/parties/party/models/party_address.py
Jsyro: lots of the other xref tables have xref suffix on the class name. I think it should be as close to the table name as possible.

python-backend/app/api/parties/party/resources/party_resource.py
NWCalvank: Valid point

python-backend/app/api/parties/party/models/party.py
Jsyro: We have run across this a few times, but the backend should be treated as a separate application that is trying to be the best version of itself, not to accommodate or make it easier for the current FE.

NWCalvank: Yes. I agree. Perhaps I wasn't clear enough... the `json` method is strictly for backwards compatibility within the backend itself (mine_party_appt uses it). This will be removed as soon as that dependency is gone. The reason for keeping an empty object as the first item in the list is to avoid a potential runtime exception if I missed some reference to the first item in the address response because it was previously always present as the first item in the list.

I don't see how this has any relation to accommodating the frontend. It's backwards compatibility with the existing API & with the other parts of the backend codebase.

Jsyro: Ah you have handled the frontend to handle empty array, ok great. If the JSON method is to be removed later, than all is good. Thanks for clarifying.

frontend/src/constants/API.js
```diff
@@ -4,11 +4,8 @@ import queryString from "query-string";
 export const MINE = "/mines";
 export const MINE_LIST = "/mines";
 export const MINE_LIST_QUERY = (params) => `/mines${params}`;
+export const PARTIES_LIST_QUERY = (params = {}) => `/parties?${queryString.stringify(params)}`;
```
thegentlemanphysicist: I'm unfamiliar with this function type.  Does it mean:  If no param provided use '{}' as a default?

frontend/src/components/parties/PartyProfile.js
```diff
@@ -221,7 +221,7 @@ export class PartyProfile extends Component {
                 {parties.phone_no} {parties.phone_ext ? `x${parties.phone_ext}` : ""}
               </p>
             </div>
-            <Address address={parties.address[0]} />
+            <Address address={parties.address[0] || {}} />
```
thegentlemanphysicist: Javascript question:  Does this mean: If parties.address[0] is truthy, use it else use '{}'

frontend/src/components/dashboard/contactsHomePage/ContactList.js
```diff
@@ -15,15 +15,27 @@ import CustomPropTypes from "@/customPropTypes";
...
 
 const columns = [
   {
     title: "Contact Name",
     dataIndex: "name",
-    render: (text, record) => (
-      <Link to={router.PARTY_PROFILE.dynamicRoute(record.key)}>{text}</Link>
+    sortField: "party_name",
+    render: ([firstName = "", lastName = ""], record) => (
+      <Link to={router.PARTY_PROFILE.dynamicRoute(record.key)}>{` ${lastName}${
```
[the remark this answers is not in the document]
NWCalvank: Actually, I can't. The formatter (Prettier) decides how to break the lines. I could remove the implicit return and make an intermediary variable if it's too hard to follow.

python-backend/app/api/parties/party/models/address.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Address(Base):
+    __tablename__ = "address"
+    address_id = db.Column(db.Integer, primary_key=True)
+    suite_no = db.Column(db.String, nullable=True)
+    address_line_1 = db.Column(db.String, nullable=True)
+    address_line_2 = db.Column(db.String, nullable=True)
+    city = db.Column(db.String, nullable=True)
+    sub_division_code = db.Column(db.String,
+                                  # FIXME: Workaround for code-first tests
```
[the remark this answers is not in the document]
NWCalvank: It might not be needed now with James' test changes. At the time of writing, it is because the test DB is generated using SQLAlchemy models (instead of our migration files). The teardown kept blowing up because of the foreign key constraint between addresses and the `sub_division_code` table. To fix this, I instructed the teardown to just cascade the delete to all of its dependents (instead of blowing up).

frontend/src/constants/API.js
NWCalvank: Yep. It's just a default parameter value

frontend/src/components/parties/PartyProfile.js
NWCalvank: Yep!

python-backend/app/api/parties/party/models/address.py
```diff
@@ -0,0 +1,64 @@
...
+            save=True):
+        address = cls(
+            suite_no=suite_no,
+            address_line_1=address_line_1,
+            address_line_2=address_line_2,
+            city=city,
+            sub_division_code=sub_division_code,
+            post_code=post_code)
+        if save:
+            address.save(commit=False)
```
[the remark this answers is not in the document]
NWCalvank: It has to do with managing the DB session. You stage a bunch of changes and then can commit them or rollback. I'm not 100% clear on why we explicitly don't commit on save

python-backend/app/api/parties/party/models/address.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+    @validates('post_code')
+    def validate_post_code(self, key, post_code):
+        if post_code and len(post_code) > 6:
+            raise AssertionError('post_code must not exceed 6 characters.')
+        validPostalCode = re.compile(r"\s*([a-zA-Z]\s*\d\s*){3}$")
+        if post_code and not validPostalCode.match(post_code):
+            raise AssertionError('Invalid post_code format.')
+        return post_code
```
Jsyro: Do we allow addresses with no values? filled should there be a required subset to create an address?

python-backend/app/api/parties/party/resources/party_resource.py
Jsyro: This loop is duplicated above, but you have added the protection to type code. thanks!

NWCalvank: One look is for the `Party` instance. This one is for the associated `PartyAddress` instance. I've added the protection to the one above because it lives on `Party`.

python-backend/app/api/parties/party/models/address.py
NWCalvank: All address fields are nullable. No subset is currently required here, in the db, or on the frontend.

python-backend/app/api/parties/party/resources/party_resource.py
Jsyro: address has an address_id field, we don't serve it, but someone writing manual requests might be able to cause a real problem in editing the PK. we should protect against that. you are right though, i miss read the issue two loops.

python-backend/app/api/parties/party/models/address.py
Jsyro: Ok.

python-backend/app/api/parties/party/resources/party_resource.py
NWCalvank: Maybe I'm wrong, but the parser doesn't know about `address_id`, so we don't protect against it.

Jsyro: Wow, didn't have enough coffee today apparently, yup you are right. thanks.

python-backend/app/api/parties/party/models/address.py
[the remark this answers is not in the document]
thegentlemanphysicist: neat

frontend/src/components/dashboard/contactsHomePage/ContactList.js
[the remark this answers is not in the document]
thegentlemanphysicist: I thought that might be the case.  :P I'm not picky about it.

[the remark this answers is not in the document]
NWCalvank: I did do the refactor for readability 😄

PR #671, opened 2019-04-15
Adding the GW WMSLayer

itchison:
I just want to see it running on our servers to see if this works.


itchison:
Nope - did not work!

PR #672, opened 2019-04-15
Tests now use factories

james-hollinger:
Tests now use factories. Tests also use transactional sessions. If you include the db_session fixture in your test, anything done to the session (including deletes and commits) will be rolled back after the test. 

The whole test suite now takes ~20s to run instead of 100s (on my machine). Likely further speed improvements are possible by only loading the code tables once.

setup_data is dead. Long live setup_data.


python-backend/tests/mines/resources/test_mine_basicinfo_resource.py
```diff
@@ -12,25 +15,24 @@ def test_get_mines_basic_info_empty(test_client, auth_headers):
...
 
-def test_get_mines_basic_info_multiple(test_client, auth_headers):
-    new_mine_guid = 'b6e1c212-aa7d-4f30-8c37-f7e3837be561'
-    mine_identity2 = Mine(
-        mine_guid=uuid.UUID(new_mine_guid),
-        mine_no="Bzzz1234",
-        mine_name="test_mine_name",
-        **DUMMY_USER_KWARGS)
-    mine_identity2.save()
+def test_get_mines_basic_info_multiple(test_client, db_session, auth_headers):
```
NWCalvank: Am I missing something here? Why are these called `test_get_` if they're for posts?

Actually, I don't understand this these three tests at all.

Jsyro: This is an endpoint that exists to turn a list of mine_guids into their basicinfo (name, lat/long, number). 

GET actions aren't supposed to have a payload, and REPORT isn't widely supported, so POST fit best.

NWCalvank: Why would this not just be a GET to something like
`/mines/basic-info?mine_guid=guid1,guid2,guid3`?

or (if feasible for the use-case) a series of GETs to something like this, which seems preferable to me:
`/mines/<mine_guid>/basic-info`?

or a series of GETs to something like:
`/mines/<mine_guid>?basic=true`?

Maybe it's just me with this opinion, but the POST seems misleading/confusing here. It's just filtered list of basic-info mines. Also, this is definitely OOS for this PR. I just noticed these tests and didn't understand them.

PR #677, opened 2019-04-17
Update python packages as per GitHub recommendations

FreshGyan:
Audited all our python packages and upgraded minor versions without breaking changes.

The only problem is requests library is still stuck on Urrllib 1.24 which has a major vulnerability that snyk hates. So for now it throws this a dependancy warning, Not an ERROR.
The fix is underway and we can upgrade our requests library to that when it gets updated with the fix.
https://github.com/kennethreitz/requests/pull/5063

Besides that doesn't seem to cause any issues.


FreshGyan:
@NWCalvank defs needs more manual testing. I did a quick rundown and didnt seem to cause much issues

python-backend/requirements.txt
```diff
@@ -11,34 +7,25 @@ Flask-APScheduler==1.11.0
...
-python-editor==1.0.3
-pytz==2018.5
 redis==3.0.1
-requests==2.20.0
+requests==2.21.0
 six==1.11.0
-SQLAlchemy==1.2.10
+SQLAlchemy==1.3.3
 sqlalchemy-filters==0.8.0
+urllib3==1.25.1
```
FreshGyan: Needed by requests,  some other package is forcing an older version of urllib to be installed and requests seemed incompatible with that

docker-images/python-36/Dockerfile
```diff
@@ -13,5 +13,5 @@ RUN curl -sLo /tmp/sonar-scanner-cli.zip https://dl.bintray.com/sonarsource/Sona
 COPY requirements.txt ${APP_ROOT}/src
 RUN source /opt/app-root/etc/scl_enable && \
     set -x && \
-       pip install -U pip==18.1 setuptools wheel && \
+    pip install -U pip setuptools wheel && \
```
FreshGyan: I had to freeze pip version a while back because they introduced some bug.  Seems to be happy and fixed now

python-backend/app/api/mines/mine/resources/mine_map.py
```diff
@@ -36,11 +36,9 @@ def get(self):
             last_modified = datetime.utcnow()
 
             # jsonify then store in cache
-            current_app.logger.error(str(records))
```
FreshGyan: Seemed un-necessary to log an error if the data in cache is not found

FreshGyan:
@Jsyro fixed! Sqlalchemy was deprecating the way it does things and I think one of the other packages implicitly installed a newer version. So i upgraded ours to support that

PR #679, opened 2019-04-18
Feature/mds 1555

mmumby:
Frontend changes for compliance view: 
- Updated styling of cards
- Created a reusable fixed width/height card for key compliance data
- added ability to filter

backend changes:
- added count for inspections in past 12 months
- added count for inspection since April 1st of the current fiscal year


frontend/src/components/mine/MineDashboard.js
```diff
@@ -141,8 +161,74 @@ export class MineDashboard extends Component {
...
+    const orderNo = params.order_no ? order.order_no.includes(params.order_no) : true;
+    const reportNoString = order.report_no.toString();
+    const reportNo = params.report_no ? reportNoString.includes(params.report_no) : true;
+    const violation =
+      params.violation.length !== 0 ? params.violation.includes(order.violation) : true;
+    return overdue && inspector && date && orderNo && reportNo && violation;
+  };
+
+  handleComplianceFilter = (values) => {
+    if (isEmpty(values)) {
```
NWCalvank: How about we refactor this to isolate the difference?
```javascript
const { violation, ...rest } = values;
const urlifiedViolation = violation ? { violation: violation.join(",") } : {};
const params = { ...violation, ...rest };

this.props.history.push(
  router.MINE_SUMMARY.dynamicRoute(
    this.props.match.params.id,
    this.props.match.params.activeTab,
    params
  )
);
```

frontend/src/constants/routes.js
```diff
@@ -38,7 +38,10 @@ export const CONTACT_HOME_PAGE = {
 
 export const MINE_SUMMARY = {
   route: "/dashboard/:id/:activeTab",
-  dynamicRoute: (id, activeTab = "summary") => `/dashboard/${id}/${activeTab}`,
+  dynamicRoute: (id, activeTab = "summary", filterParams) =>
```
NWCalvank: Even though it's a bit of a pain right now, I think this should be updated to accept an object of params instead of adding another argument to the function. `CONTACT_HOME_PAGE` and `MINE_HOME_PAGE` above do this well in the `.dynamicRoute()`.

I think we're getting dangerously close to unmanageable overloading of this function by adding a third argument and a client-side specific route to the server-side route building logic.

Happy to chat about it if you have any questions

frontend/src/customPropTypes/compliance.js
```diff
@@ -0,0 +1,34 @@
+import { PropTypes } from "prop-types";
+
+// This file is anticipated to have multiple exports
+// eslint-disable-next-line import/prefer-default-export
```
NWCalvank: This shouldn't be needed in a file with multiple exports

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -18,6 +18,12 @@ export const {
   getIncidentFollowupActionOptions,
 } = staticContentReducer;
 
+// removes all expired compliance codes from the array
+export const getCurrentComplianceCodes = createSelector(
+  [getComplianceCodes],
+  (codes) => codes.filter((code) => new Date(code.expiry_date) > new Date())
+);
```
NWCalvank: We should chat with someone from business to find out if we should be excluding expired compliance codes at the API level

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -103,27 +109,49 @@ export const getDropdownIncidentFollowupActionOptions = createSelector(
...
-  [getComplianceCodes],
+  [getCurrentComplianceCodes],
   (codes) =>
     codes
       .filter(({ article_act_code }) => article_act_code === "HSRCM")
       .map((code) => {
-        const composedLabel = `${code.section}.${code.sub_section}.${code.paragraph} - ${
-          code.description
-        }`;
+        const composedLabel = code.sub_paragraph
```
NWCalvank: This could be cleaned up a little to just handle the sub_paragraph directly:
```javascript
`${code.section}.${code.sub_section}.${code.paragraph}${code.sub_paragraph ? `.${code.sub_paragraph}` : ''} - ${code.description}`
```
Or move it outside of the string if the double-interpolation is tough to parse:
```javascript
const subParagraph = code.sub_paragraph ? `.${code.sub_paragraph}` : ''
const composedLabel = `${code.section}.${code.sub_section}.${code.paragraph}${subParagraph} - ${code.description}`
```

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -103,27 +109,49 @@ export const getDropdownIncidentFollowupActionOptions = createSelector(
...
+  [getCurrentComplianceCodes],
+  (codes) =>
+    codes
+      .filter(({ article_act_code }) => article_act_code === "HSRCM")
+      .map((code) => {
+        const composedValue = code.sub_paragraph
+          ? `${code.section}.${code.sub_section}.${code.paragraph}.${code.sub_paragraph}`
+          : `${code.section}.${code.sub_section}.${code.paragraph}`;
+        const composedLabel = code.sub_paragraph
+          ? `${code.section}.${code.sub_section}.${code.paragraph}.${code.sub_paragraph} - ${
```
NWCalvank: Same comment here about managing sub_paragraph directly

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -103,27 +109,49 @@ export const getDropdownIncidentFollowupActionOptions = createSelector(
...
+  [getCurrentComplianceCodes],
   (codes) =>
     codes
       .filter(({ article_act_code }) => article_act_code === "HSRCM")
       .reduce((map, code) => {
-        const composedValue = `${code.section}.${code.sub_section}.${code.paragraph} - ${
-          code.description
-        }`;
+        const composedValue = code.sub_paragraph
+          ? `${code.section}.${code.sub_section}.${code.paragraph}.${code.sub_paragraph} - ${
```
NWCalvank: Also here. And it may be worth it to just pull the `composed*` function out of the exported functions so that they can all share it (assuming they all use the same logic)

frontend/src/styles/components/MineCompliance.scss
```diff
@@ -0,0 +1,68 @@
...
+  }
+  .ant-row-flex {
+    min-height: 100px;
+  }
+}
+
+// needed to set a fixed width to the cards on the compliance Tab,
+// opted to create a custom class vs interfering with the default style of cards that do not need a set width
+// cards are full responsive, regardless of the amount on view.
+.compliace--card {
```
NWCalvank: Spellcheck "compliance" 😄

python-backend/app/api/services/NRIS_service.py
```diff
@@ -157,16 +180,22 @@ def _process_NRIS_data(data, mine_no):
...
+        'num_inspections_since_april': num_inspections_since_april,
         'num_open_orders': num_open_orders,
         'num_overdue_orders': num_overdue_orders,
         'advisories': advisories,
         'warnings': warnings,
         'section_35_orders': section_35_orders,
         'open_orders': open_orders_list,
     }
-    cache.set(NRIS_COMPLIANCE_DATA(mine_no), overview, timeout=TIMEOUT_24_HOURS)
+    cache.set(NRIS_COMPLIANCE_DATA(mine_no), overview, timeout=TIMEOUT_12_HOURS)
```
NWCalvank: Was this timeout intentionally changed?

frontend/src/components/mine/Compliance/MineComplianceCard.js
```diff
@@ -0,0 +1,37 @@
...
+  <div className="compliace--card">
+    <div className="compliace--card--head">
+      <div className="compliace--card--head--content">
+        <span className="info-display">
+          {props.icon && <img src={props.icon} alt="icon" className="padding-small--right" />}
+          {props.content}
+        </span>
+      </div>
+    </div>
+    <div className="compliace--card--body">
```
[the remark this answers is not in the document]
mmumby: fixed. spell checker didn't catch that one.. :(

frontend/src/utils/helpers.js
```diff
@@ -91,3 +91,17 @@ export const normalizePhone = (value, previousValue) => {
...
+  // new Date is index 0, Jan = 0
+  const fiscalYear = new Date(currentYear, 3, 1); // April 1, currentYear
+  if (today > fiscalYear) {
+    return currentYear;
+  }
+  // return the previous year if between the months of Jan-March
+  return currentYear - 1;
+};
+
+export const returnBoolean = (param) => param && param === "true";
```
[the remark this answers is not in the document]
mmumby: nooooo, nothing to do with our API. 

I had this in mineDashboard as a class method, but decided to move it out.
When filtering the open_orders, one drop down option has true/false values but they were returned as strings after being pushed to the URL, so I needed to transform them back into booleans. 

Have any alternate suggestions?

frontend/src/components/mine/MineDashboard.js
mmumby: yeah this works! will update it

frontend/src/constants/routes.js
NWCalvank: Follow-up: I just realized this doesn't build an API route. It's another clientside route. My first paragraph still applies. The second one does not

frontend/src/customPropTypes/compliance.js
mmumby: i'll remove

frontend/src/selectors/staticContentSelectors.js
mmumby: Yes, agreed. Chatted with Aaron and he was happy to remove the expired, but has no preference if its client or server side.

frontend/src/styles/components/MineCompliance.scss
mmumby: fixed.

python-backend/app/api/services/NRIS_service.py
mmumby: oops. will change it back to 24.

frontend/src/utils/helpers.js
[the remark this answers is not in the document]
NWCalvank: It should be functionally equivalent to `someVariableFromTheUrl === "true"`, right?

The case when that wouldn't be true is if `param` (in this case) is a falsey value other than `false`, which would cause it to short-circuit and return. Such as if `param = ""`, then `returnBoolean` would return `""`.

Hopefully you're not relying on that behaviour, given the name of the function? So, I'd say we can probably just scrap this function and use `===` directly. Thoughts?

frontend/src/constants/routes.js
mmumby: This route is following similar logic to `CONTACT_HOME_PAGE` and `MINE_HOME
_PAGE`
The `filerParams` are an object and being handled the same way, the only difference with this route is that has an `id` whereas the others do not.

frontend/src/selectors/staticContentSelectors.js
NWCalvank: Okay. It's really just a matter of whether or not the expired (or future) values should be "exposed" on principle. Would we want to provide that data to an external consumer of the API? If not, we should move it to the server. This would be my preference, considering how we handle `active_ind` currently (filtering by it before returning data from the API). Mind you, I was told that the codes were being treated differently, so I'm not sure if we need to confirm that with someone.

It's probably OOS for this PR. Not a big job to add the logic to the backend and then just remove this filter in the future.

frontend/src/utils/helpers.js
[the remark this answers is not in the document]
mmumby: yep, happy to do that!

frontend/src/constants/routes.js
NWCalvank: Chatted about this IRL. Marking as resolved

frontend/src/components/mine/MineDashboard.js
```diff
@@ -141,8 +161,74 @@ export class MineDashboard extends Component {
...
+        : queryString.stringify(initialSearchValues);
+      this.renderDataFromURL(correctParams);
+    }
   }
 
+  componentWillUnmount() {
+    this.setState({ complianceFilterParams: initialSearchValues });
+  }
+
+  format = (param) => (param ? param.split(",").filter((x) => x) : []);
```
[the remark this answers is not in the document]
mmumby: moved the function to helpers,

frontend/src/components/mine/MineDashboard.js
mmumby: @NWCalvank So, this requires a bit of a bigger change, due to the way `componentWillReceiveProps` is currently set up.
Is this more of a suggestion rather than a must do?

frontend/src/components/mine/MineDashboard.js
```diff
@@ -141,8 +161,74 @@ export class MineDashboard extends Component {
...
+      open_orders.filter((order) => this.handleFiltering(order, formattedParams));
+
+    this.setState({
+      filteredOrders,
+      complianceFilterParams: formattedParams,
+    });
+  };
+
+  handleFiltering = (order, params) => {
+    const overdue = params.overdue === null ? true : order.overdue === params.overdue;
```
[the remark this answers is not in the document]
mmumby: done

frontend/src/utils/helpers.js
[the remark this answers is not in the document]
mmumby: removed!

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -102,31 +108,47 @@ export const getDropdownIncidentFollowupActionOptions = createSelector(
...
 
+const formatComplianceCodeValueOrLabel = (code, showDescription) => {
+  const { section, sub_section, paragraph, sub_paragraph, description } = code;
+  const formattedSubSection = sub_section ? `.${sub_section}` : "";
+  const formattedParagraph = paragraph ? `.${paragraph}` : "";
+  const formattedSubParagraph = sub_paragraph !== null ? `.${sub_paragraph}` : "";
+  const formattedDescription = showDescription ? ` - ${description}` : "";
+
+  return `${section}${formattedSubSection}${formattedParagraph}${formattedSubParagraph}${formattedDescription}`;
+};
```
NWCalvank: This is a lot better! 😄

PR #680, opened 2019-04-22
upgrade flyway version

FreshGyan:
- Update flyway version to accommodate for new retry parameter
- Improves deploy times by not having to spin up different lifecycle hooks
- Change deployment strategy to rolling instead of recreate to swap environments with less downtime


openshift/_python36.dc.json
```diff
@@ -542,6 +542,10 @@
                   {
                     "name": "FLYWAY_OUT_OF_ORDER",
                     "value": "true"
+                  },
+                  {
+                    "name": "FLYWAY_CONNECT_RETRIES",
+                    "value": "300"
```
itchison: Is 300 reasonable? Seems like a really high number considering we used to loop 98 times above.

FreshGyan: Yeah I try to imagine that the deployment should never take more than 5 minutes.
Basically the above loop was, ping postgres upto 90 times and each time you fail wait 5 seconds.  so worst case scenario it would go upto 450 seconds and if all that failed, the deployment itself is supposed to retry from scratch.

Since the check has been moved into flyway now, I figured reducing the time might actually be beneficial and if the postgres pod isnt up for 5 minutes,  something's probably wrong

itchison: Ah ok. It'll try to connect 300 times waiting one second between each now vs 90 times waiting 5 seconds between.

PR #684, opened 2019-04-22
Feature/incident report edit

Jsyro:
allow edit of incident report.


frontend/src/components/common/NullScreen.js
```diff
@@ -120,6 +121,12 @@ const NullScreen = (props) => (
         <p>Create a contact using the menu above</p>
       </div>
     )}
+    {props.type === "incidents" && (
```
nayr974: You should add to and use NullScreen component instead of having No incidents found here.

frontend/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -45,23 +72,25 @@ export class MineIncidentTable extends Component {
       followup_action: actions.find(
         (x) => x.mine_incident_followup_type_code === incident.followup_type_code
       ),
+      ...funcs,
```
nayr974: We've been moving away from spread operators (mainly for ...props but I don't see why other uses would be different) and are preferring to explicitly set properties unless they are required to be dynamic. Is there good reason not to pass in the two functions as separate parameters?

Jsyro: thought it looked cleaner, but i suppose it hides things. changed.

frontend/src/components/common/NullScreen.js
Jsyro: This is the nullscreen component, but right now we pass the nullscreen the thing it is showing null for... I think this nullscreen should be looked, because this pattern is not sustainable, nor that helpful.

nayr974: lol derp

PR #690, opened 2019-04-23
MDS-1682 Task/Add Variances tests

NWCalvank:
There should be no functional changes other than the standardized error message and a 400 status code being replaced with a 404.

Changes
- Added Variance factories
- Updated Variance resources to use CustomReqparser
- Some variances marshalling cleanup
- Properly use `required=True` to check for missing params


python-backend/app/api/mines/custom_reqparser.py
```diff
@@ -0,0 +1,18 @@
...
+        parser_instance = reqparse.RequestParser(trim=True)
+        self.parser = parser_instance
+        self.add_argument = parser_instance.add_argument
+
+    def parse_args(self):
+        try:
+            data = self.parser.parse_args()
+        except BadRequest:
+            raise BadRequest(DEFAULT_MISSING_REQUIRED)
+        return data
```
Jsyro: Should there be one CustomReqparser that lives in utils folder? eventually we will need to package that utils folder for reuse anyways.

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -104,21 +85,18 @@ class VarianceResource(Resource, UserMixin, ErrorMixin):
...
 
         return variance
 
 
 class VarianceDocumentUploadResource(Resource, UserMixin, ErrorMixin):
-    parser = reqparse.RequestParser(trim=True)
-    parser.add_argument('mine_document_guid', type=str)
-    parser.add_argument('document_manager_guid', type=str)
-    parser.add_argument('document_name', type=str)
+    parser = CustomReqparser()
```
Jsyro: Is this used? 

the post parses arguments (but there are none). 
do we like the parser arguments being added in the function? I feel like that's going to get awkward and harder to find things. I'd rather see the resources parser have all it's arguments defined in one spot.

python-backend/app/api/mines/custom_reqparser.py
NWCalvank: Yes, I think so. But I don't understand what rules we have for our microservices. My only previous experience with microservices used Go and kept each service in a separate repo (which Go could easily require in), so I'm not clear on where I can/can't put shared code.

I figured I had to put an instance in each namespace at this point.

python-backend/app/api/mines/variances/resources/variance.py
NWCalvank: Well, I can define them all here and then redefine/override them in the methods that have different required args?

PR #691, opened 2019-04-24
Feature/MDS-1473 Download Progress

james-hollinger:
This PR changes the file download flow to pull up the download modal before the file is downloaded, instead of after. This eliminates the several second wait between clicking on the link and the user getting feedback.

This change required implementing an endpoint for the frontend to hit without auth headers (you cannot pass headers into a new tab/window). The frontend first hits an authenticated endpoint that issues a short lived one-use token for accessing a single file from an unsecured endpoint. Minespace has also been updated accordingly.

New test has been added to test file downloads. DocumentMangerFactory now has a path_root Param that lets you point it to an actual file.

Core filepond has also been updated to the same version Minspace was updated to.


frontend-public/src/utils/actionlessNetworkCalls.js
```diff
@@ -2,21 +2,22 @@ import axios from "axios";
...
   axios
-    .get(url, { responseType: "arraybuffer", ...createRequestHeader() })
+    .get(
+      `${ENVIRONMENT.apiUrl + DOCUMENT_MANAGER_TOKEN_GET_URL(docManagerGuid)}`,
+      createRequestHeader()
+    )
     .then((response) => {
-      fileDownload(response.data, filename);
+      const token = { token: response.data.token_guid };
+      window.location = `${ENVIRONMENT.apiUrl + DOCUMENT_MANAGER_FILE_GET_URL(token)}`;
```
NWCalvank: I'm concerned about us setting `window.location`, as I think it has effects on our router behaviour.

It looks like this is to trigger the document to download? I'd suggest confirming that the back button still works after downloading a document.

frontend-public/package.json
```diff
@@ -17,7 +17,6 @@
     "filepond-plugin-file-validate-type": "^1.2.1",
     "filepond-polyfill": "^1.0.2",
     "hoist-non-react-statics": "^3.2.1",
-    "js-file-download": "^0.4.4",
```
NWCalvank: Neither of your package-lock.json files appear to have been updated. This will cause problems with `npm ci`. If you changed the package.json directly, you'll need to run `npm install` to get the package-lock in sync

frontend-public/src/utils/actionlessNetworkCalls.js
```diff
@@ -2,21 +2,22 @@ import axios from "axios";
...
-
-const downloadFileFromDocumentManager = (docManagerGuid, filename) => {
-  if (!docManagerGuid || !filename) {
-    throw new Error("Must provide both docManagerGuid and filename");
+const downloadFileFromDocumentManager = (docManagerGuid) => {
+  if (!docManagerGuid) {
+    throw new Error("Must provide docManagerGuid");
   }
 
   // TODO: Update url when Document Manager moves to its own microservice.
```
NWCalvank: I don't think this file should have any references to the backend architecture.

frontend/package.json
```diff
@@ -14,13 +14,12 @@
     "enzyme-adapter-react-16": "^1.1.1",
     "express": "^4.16.3",
     "express-cache-controller": "^1.1.0",
-    "filepond": "^3.5.0",
+    "filepond": "4.3.0",
```
NWCalvank: Same comment about the package-lock here. Just make sure they're in sync 😄

python-backend/tests/document_manager/resources/test_document_manager_resource.py
```diff
@@ -1,26 +1,59 @@
...
-    get_resp = test_client.get(
-        f'/document-manager/1234', headers=auth_headers['full_auth_header'])
+def test_get_file_with_guid(test_client, db_session):
+    get_resp = test_client.get(f'/document-manager/1234')
     get_data = json.loads(get_resp.data.decode())
 
-    assert get_resp.status_code == 404
-    assert get_data['error']['status'] == 404
+    assert get_resp.status_code == 500
+    assert get_data['error']['status'] == 500
```
NWCalvank: What's the reason for making this a 500?

frontend-public/src/utils/actionlessNetworkCalls.js
james-hollinger: It doesn't actually navigate away, surprisingly. Since the backend response sets a download header, the browser magically understands that its supposed to download the file.

This is the first time iv'e ever seen where the browser's builtin behavior is 100% what we wanted.

python-backend/tests/document_manager/resources/test_document_manager_resource.py
james-hollinger: This is now a 500 because the test is hitting an API path that doesn't exist. The download path used to be 'document-manager/GUID' but now its 'document-manager?token=token'

frontend/package.json
james-hollinger: Thanks 👍

frontend-public/package.json
james-hollinger: Thanks 👍

python-backend/tests/document_manager/resources/test_document_manager_resource.py
NWCalvank: Oh, okay. Any thoughts on when it would make sense to remove tests for deprecated endpoints?

I took a quick look at the spec (https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), and it looks like we might want code 410 instead? But I'm not super picky about that either way

frontend-public/src/utils/actionlessNetworkCalls.js
NWCalvank: Fair enough! If it's doing what we want and not causing any issues with React router then I'm happy. I'll mark this as resolved

frontend-public/src/utils/actionlessNetworkCalls.js
```diff
@@ -2,21 +2,22 @@ import axios from "axios";
...
+import { DOCUMENT_MANAGER_FILE_GET_URL, DOCUMENT_MANAGER_TOKEN_GET_URL } from "@/constants/API";
 
-const fileDownload = require("js-file-download");
-
-const downloadFileFromDocumentManager = (docManagerGuid, filename) => {
-  if (!docManagerGuid || !filename) {
-    throw new Error("Must provide both docManagerGuid and filename");
+const downloadFileFromDocumentManager = (docManagerGuid) => {
+  if (!docManagerGuid) {
+    throw new Error("Must provide docManagerGuid");
```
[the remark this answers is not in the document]
james-hollinger: No toast is displayed,m this is shown in the console if the method is called incorrectly.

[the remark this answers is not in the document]
NWCalvank: Oh, yes. My mistake. I thought this was part of the Promise chain. Marking as resolved. Thanks!

python-backend/tests/document_manager/resources/test_document_manager_resource.py
james-hollinger: I've deleted the test because it has no real value.

PR #692, opened 2019-04-24
Feature/mds 1686 continue refactoring mine resources 2

JustinMacaulay:
NGINX and Logstash pods were failing so trying to remake.

This is a refactor of the rest of the resources in the mine subfolder of mines. Mine commodity and mine disturbance were left out due the them needing a much larger re-work. Compliance, Location, Region, tailings, and status all still require refactoring as well.

I believe all comments have been addressed in the previous PR.


frontend/src/tests/components/Forms/__snapshots__/MineRecordForm.spec.js.snap
```diff
@@ -202,26 +202,7 @@ ShallowWrapper {
...
-          "label": "Placer",
-          "value": "PLR",
-        },
-        Object {
-          "label": "BC Land",
-          "value": "BCL",
-        },
-      ]
-    }
+    mineTenureTypes={undefined}
```
Jsyro: this doesn't look good. Probably want to check on this one.

python-backend/app/api/mines/mine/resources/mine_type.py
```diff
@@ -1,59 +1,59 @@
...
-                mine_guid, mine_tenure_type_code, add_to_session=False)
-            mine_type.save()
-        except exc.IntegrityError as e:
-            self.raise_error(400, 'Error: Unable to create mine_type.')
 
-        return mine_type.json()
-
-    @api.expect(parser)
+class MineTypeResource(Resource, UserMixin):
+    @api.doc(description='Deletes the mine type provided.')
```
Jsyro: code=204 for swagger docs ?

JustinMacaulay: I have added in a api.response() so there is a code displaying now.

frontend/src/tests/components/Forms/__snapshots__/MineRecordForm.spec.js.snap
JustinMacaulay: Fixed, as well as in a few other places.

frontend/src/reducers/staticContentReducer.js
```diff
@@ -37,7 +38,7 @@ const staticContentReducer = (state = initialState, action) => {
     case actionTypes.STORE_TENURE_TYPES:
       return {
         ...state,
-        mineTenureTypes: action.payload.options,
+        mineTenureTypes: createDropDownList(action.payload, "description", "mine_tenure_type_code"),
```
mmumby: whats the purpose of this change?

JustinMacaulay: This was outlined in developers by Jason yesterday in an effort to decouple the front end from the back end.

mmumby: Follow up: I don't see any changes to the selectors, the `getMineTenureTypes` selector was used to create several others, including a dropdown.

JustinMacaulay: Im not sure I understand what you mean?

mmumby: This is the opposite of what we decided in the developers channel. 
data transformations should occur in the selectors, not when they are being saved in the store.

mmumby: The data returned`getMineTenureTypes` was used to create 4 other selectors, including dropdown, a hash object, conditional rendering for commodity/disturbances.

JustinMacaulay: I agree. However, this was the smallest change that would allow for the swagger refactor to work while also de-coupling the backend from the front end in this instance. The rest of the selector issues can be batch fixed as this isn't the only instance of this.

python-backend/app/api/mines/mine/resources/mine_tenure_type_code.py
```diff
@@ -1,11 +1,16 @@
...
 class MineTenureTypeCodeResource(Resource, UserMixin, ErrorMixin):
-    @api.doc(params={})
+    @api.doc(description='Returns all the mineral tenure type codes.')
     @requires_role_mine_view
+    @api.marshal_with(MINE_TENURE_TYPE_CODE_MODEL, code=200)
     def get(self):
-        return { 'options': MineTenureTypeCode.all_options() }
+        mineral_tenure_type_codes = MineTenureTypeCode.find_all()
+
+        return mineral_tenure_type_codes
```
NWCalvank: We've agreed to put a `records` envelope around our list responses, correct? Please add that here.

Also, I'm not sure that `.find_all()` is actually buying us anything over just `.query.all()`, so I'm curious if you have a reason for adding that method.

Finally, if you've removed the use of `.all_options()`, could you please remove that method from the model? thanks!

frontend/src/reducers/staticContentReducer.js
NWCalvank: If I understand this all correctly, I see Justin's point here. `action.payload.options` used to return an array of objects with `label`s and `value`s, correct? And so, in order to maintain backwards compatibility, `createDropDownList` is added here to format the new API response into the "old" format that the frontend uses.

With that said, Meredith is also correct. This is the opposite of how we want to do our data transformation. We want to store it as-is and then format it on extraction from the store. This should be as simple as updating the selector to do what you've done here instead of formatting it in the reducer.

Also, Justin, I've left a comment about the response that you're providing from this endpoint. Please take a look at that feedback alongside this comment.

Does that all make sense to everyone?

Jsyro: This is an effort to bring the backend up to standard, but the FE expected the value in the store to be the correct form for the dropdown, this is the smallest change to fix the backend without breaking the frontend. I think someone needs to take a pass these outstanding issues to bring them in line. But I don't think we want Justin going down all these rabbit holes in this PR/Ticket.

mmumby: ah, yes. Thanks, @NWCalvank for chiming in.
My point is specifically about the location of the transformation.
Happy with the decoupling of the BE & FE.

mmumby: @Jsyro I'm all for bringing up the backend standard, but not at the expense of the frontend standards we have for the reducers/selectors. 

Please store the data as is, and make the transformation in the correct location.

nayr974: Not the hill I would die on given the scope of changes Justin has been working on, and it can be updated in another branch as this code doesn't get chiseled in stone :) That said, this should be a one liner in the same file. Justin, if you don't think making this change will cause a lot of regression testing for you I would make it. Otherwise, we can note it with a comment or in our backlog.

python-backend/app/api/mines/mine/resources/mine_tenure_type_code.py
JustinMacaulay: I have removed the all_options function and renamed the find_all to find_all_active. As for why I did it as a model class method as far as i know we don't really have a standard on where we put the actual DB query code. Half is in the models and then some of it lives in the resources.

frontend/src/reducers/staticContentReducer.js
mmumby: I thought this was a breaking change until Nathans comment. I see now that it doesn't break anything, so disregard my previous comments. 

I'm very pleased with the work Justin is doing, but I'd also to avoid adding unnecessary tickets into the backlog, in order to decouple the BE and FE there will need to be changed made to both and I'd like to see the standards being met for both. 

That is all, not blocking this PR, happy coding.

python-backend/app/api/utils/models_mixins.py
```diff
@@ -51,9 +52,11 @@ def save(self, commit=True):
         if commit:
             try:
                 db.session.commit()
-            except DBAPIError:
+            except SQLAlchemyError as e:
+                current_app.logger.error(
+                    f'When trying to save {self} an exception was thrown by the database {e}')
                 db.session.rollback()
-                raise
+                raise InternalServerError(f'Could not save {self.__class__.__name__}')
```
Jsyro: Are there existing try Exceptions that are looking DBAPIError's?

PR #693, opened 2019-04-24
MDS-1418 Feature/Add sorting to compliance table

NWCalvank:
All data is available to the client, so I swapped this component over to using the `pagination` provided on Ant Design tables out-of-the-box. This allows the pagination and sorting to work together without a need for us to manage it manually.

### DIFF
--- a/frontend/src/styles/components/Tables.scss
+++ b/frontend/src/styles/components/Tables.scss
@@ -81,3 +81,30 @@ span.ant-table-column-sorter {
   right: 4px;
   top: 35%;
 }
+
+// Centered Pagination
+$pagination-width-1: 112px;
+$pagination-width-2: 152px;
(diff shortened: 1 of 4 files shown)


frontend/src/styles/components/Tables.scss
```diff
@@ -81,3 +81,30 @@ span.ant-table-column-sorter {
...
+.center-pagination {
+  @for $i from 1 through length($pagination-widths) {
+    &.page-count-#{$i} {
+      .ant-table-pagination.ant-pagination {
+        $pagination-width: nth($pagination-widths, $i);
+        margin-right: unquote("calc( 50% - #{$pagination-width} / 2 )");
+      }
+    }
+  }
+}
```
mmumby: This is some serious css.. too bad there wasn't an easier way.

PR #694, opened 2019-04-25
Refactor deployment configurations

FreshGyan:
Break out the single config.groovy into environment specific files.

- Improves readability among files
- Provides more control over individual environments
- Introduced a bit of duplication but I think that's worth the benefits
- Completely removed logstash, metabase, database-backup, and schemaspy from dev to reduce extra PVC and resource overhead


FreshGyan:
@nayr974 unfortunately not without doing some script trickery. The ocp-cd-pipeline expects all the configuration to be present in a single file. The filetype it expects is a configSlurper format which only allows for objects and strings etc.

Instead of duplication I like to think of it as separate definitions for each environment ;) 😅 
It would be nice to move all build/deploy templates in some common file and just importing them :(

PR #696, opened 2019-04-25
Features/MDS-962, MDS-1655, MDS-1739, MDS-1736; Bugs/MDS-1801, MDS-1850 - Add Permittee in Permit Workflow / Update Add Quick Party Workflow

nayr974:
This PR wraps up a number of tickets that were related, and includes:

- Adding two new components PartySelectField and AddPartyComponentWrapper that handle party search and select behavior for forms. They provide a mechanism for adding new parties that does not require any form to be otherwise concerned with parties, party actions, the party reducer, or this behavior.

- Refactoring to use the above components.

- A Permittee field when adding permits and permit amendments. If the permittee changes, the currently active permittee is "ended" and a new permittee role for the selected party is added.

- Permit table updates based on user feedback.

- Bug fixes for permits and parties.


frontend/src/components/common/PartySelectField.js
```diff
@@ -0,0 +1,187 @@
...
+
+import { Icon, Divider, AutoComplete } from "antd";
+import { Field } from "redux-form";
+import RenderLargeSelect from "./RenderLargeSelect";
+
+import { fetchParties, setAddPartyFormState } from "@/actionCreators/partiesActionCreator";
+import { getRawParties, getLastCreatedParty } from "@/selectors/partiesSelectors";
+import { createItemMap, createItemIdsArray } from "@/utils/helpers";
+
+// Set organization to only see organizations or set people to only see people. Displays both people and organizations by default.
```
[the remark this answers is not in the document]
nayr974: I was intending to refer to the props below. Would it make more sense to put those inline with those two props?

frontend/src/components/common/PartySelectField.js
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  validOption = (value) =>
+    this.state.partyDataSource.find((opt) => opt.key === value)
+      ? undefined
+      : `"Invalid ${this.props.partyLabel}`;
+
+  render = () => (
+    <Field
+      {...this.props}
```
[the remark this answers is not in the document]
nayr974: It's to pass any desired <Field/> prop through on the form level, of which there are a lot and I thought including them would just clutter this class. The intention is to have <PartySelectField/> behave as any one would expect any other <Field/> would. 

I really wanted this to be used as a component={renderConfig.PartySelect} but the complexity of AutoComplete's own internal behavior made using a controlled value not seem to work with <Field/>'s event hooks. I'll revisit that at some point because it should work, and then this wouldn't be needed.

frontend/src/components/common/PartySelectField.js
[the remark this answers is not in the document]
NWCalvank: Either inline or on the line right above the relevant props, instead of above the entire props block, would make more sense to me, yes. I thought this might have been an outdated comment

frontend/src/components/common/PartySelectField.js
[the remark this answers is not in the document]
NWCalvank: Perfect. I figured that the reason was going to be something like that. I'll mark this as resolved

frontend/src/components/common/PartySelectField.js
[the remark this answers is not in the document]
nayr974: I'll move them 👍

frontend/src/components/common/wrappers/ModalWrapper.js
```diff
@@ -86,11 +86,12 @@ export class ModalWrapper extends Component {
             zIndex: 100,
           }}
         />
-        {ChildComponent && (
-          <ChildComponent
+        {this.props.content && (
+          <AddPartyComponentWrapper
             closeModal={this.props.closeModal}
```
mmumby: It seems like only a small subset of models deal with parties, and thus need to be concerned with `AddPartyComponentWrapper`, what's the purpose of wrapping every modal in this check?

mmumby: Maybe its the naming that's throwing me off, this is more so a CarouselWrapper? Where the second form could be anything, but in this situation/point in time it's an `AddPartyFrom`?

nayr974: Good thing to notice :) Primarily, I thought the usability trade-off was worth putting it here. Any form can use PartySelectField and have its behavior available without additional code. Another pro is that nobody has to examine other uses of it in the code base and selectively copy what they determine to be the pieces of code required in different files, and that there is less code duplication in the code-base. Parties are pretty core to the application and the assumption is that a reasonable number of new forms will include dealing with parties.

nayr974: Currently the second form is AddPartyForm only, and some of the logic is specific to adding parties, but if we have need for that to be generic it could probably be modified to be generic.

mmumby: ah okay, Thanks for the explanation, I pulled down the branch and played around with it, it's a very smooth transition.

Could you add a comment to briefly explain what's happening in this file? I can see how this would be confusing for future devs.

nayr974: For sure. I'll go though and add some extra comments. 👍

python-backend/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -104,6 +104,13 @@ def find_by_party_guid(cls, _id):
         except ValueError:
             return None
 
+    @classmethod
+    def find_by_permit_guid(cls, _id):
+        try:
+            return cls.find_by(permit_guid=_id)
+        except ValueError:
```
JustinMacaulay: This try catch for the value error can be remove. The global handler will catch this.

python-backend/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -142,13 +149,19 @@ def find_manager_history_by_mine_no(cls, mine_no):
         return records, 200, 'OK'
 
     @classmethod
-    def find_by(cls, mine_guid=None, party_guid=None, mine_party_appt_type_codes=None):
+    def find_by(cls,
+                mine_guid=None,
+                party_guid=None,
+                mine_party_appt_type_codes=None,
+                permit_guid=None):
         try:
```
JustinMacaulay: Same as above.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
```diff
@@ -134,6 +166,8 @@ def put(self, permit_guid=None, permit_amendment_guid=None):
             else:
                 setattr(pa, key, value)
 
+        pa.save()
```
JustinMacaulay: Should this be in the else block or should the save in the if block be removed?

python-backend/app/api/parties/party_appt/models/mine_party_appt.py
nayr974: Removed

python-backend/app/api/parties/party_appt/models/mine_party_appt.py
nayr974: Removed

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
nayr974: Changed

frontend/src/components/common/wrappers/ModalWrapper.js
nayr974: Added extra comments where I thought things might not be intuitive.

frontend/src/components/common/PartySelectField.js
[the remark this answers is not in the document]
nayr974: Moved

PR #697, opened 2019-04-26
MDS-1828 Feature/Add new variance columns & business rules to database

NWCalvank:


NWCalvank:
Setting this as a WIP. Forgot about the need to update the model + tests in order to get DB changes merged.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
+CREATE TABLE variance_application_status_code (
+    variance_application_status_code character varying(3)                   NOT NULL,
```
Jsyro: This should be Primary Key.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
...
+    description                      character varying(100)                 NOT NULL,
+    active_ind                       boolean                  DEFAULT true  NOT NULL,
+    create_user                      character varying(60)                  NOT NULL,
+    create_timestamp                 timestamp with time zone DEFAULT now() NOT NULL,
+    update_user                      character varying(60)                  NOT NULL,
+    update_timestamp                 timestamp with time zone DEFAULT now() NOT NULL,
+
+    UNIQUE(variance_application_status_code)
```
Jsyro: not necessary if status_code is PK.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
...
+    description,
+    create_user,
+    update_user
+)
+VALUES
+    ('REV', 'In Review', 'system-mds', 'system-mds'),
+    ('NAP', 'Not Applicable', 'system-mds', 'system-mds'),
+    ('APP', 'Approved', 'system-mds', 'system-mds'),
+    ('DEN', 'Denied', 'system-mds', 'system-mds');
+
```
Jsyro: I think this should be moved to seed data unless needed otherwise.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
...
+
+ALTER TABLE variance
+ADD COLUMN variance_application_status_code character varying(3) DEFAULT 'REV' NOT NULL,
```
Jsyro: hmmm, do we want variances and variance applications do be in the same table? I know one precedes the other..... I would defer to @klefler or @itchison on their recommendation.

NWCalvank: The decision to put them in one table was the result of multiple conversations that included Kristin.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
...
+        NEW.variance_no := max_mine_variance_no + 1;
+        RETURN NEW;
+    END;
+$$LANGUAGE plpgsql;
+
+CREATE TRIGGER new_variance_no
+  BEFORE INSERT
+  ON variance
+  FOR EACH ROW
+  EXECUTE PROCEDURE add_incremented_variance_no();
```
Jsyro: Is all of this just for an autoincrementing column. i think we can do this as a sequence, just as it was done for serial id's.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
nayr974: As long as the record stays intact and we don't create a second record for the same variance that is in a different status, having them in one table is preferable IMO.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: It increments on a per-mine basis. It's not a global sequence

NWCalvank:
@nayr974 I'll make the time to do this properly the first time. If we think there's enough risk with the current implementation and that ORM-based alternatives are better then I'll use one of those

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: Sure. What's the proposed benefit of having seed data in that single file?

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: It was just explained to be incrementing per mine. Which was a requirement asked for mine-incidents, but we asked if a globally incrementing integer was fine and Aaron was fine with that.  Not sure if this is a different case but is feels like a similar situation.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: Sounds like there was good discussion, and there won't be any duplication, thanks Nathan.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: We have talked about wanting to reuse that seed data file for the test db, which may or may not be necessary given our testing overhaul @james-hollinger can chime in. 

Also means we can just change the existing seed_data file if we need to add an option to this. (removing an option would be tricky). also keeps all the type/code data in one place.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
```diff
@@ -0,0 +1,105 @@
...
+
+ALTER TABLE variance
+ADD CONSTRAINT variance_status_variance_application_status_code_fkey
+    FOREIGN KEY (variance_application_status_code)
+    REFERENCES variance_application_status_code(variance_application_status_code)
+    DEFERRABLE INITIALLY DEFERRED,
+ADD CONSTRAINT variance_inspector_core_user_fkey
+    FOREIGN KEY (inspector_id) REFERENCES core_user(core_user_id) DEFERRABLE INITIALLY DEFERRED,
+
+-- Business Rule Constraints
```
Jsyro: are we building these business rules at the db level? I thought we wanted to keep this in the python.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: The reason I ask is two-fold:
1. I seem to remember there was a time before launch when we did a full database teardown to resolve an issue with database seeds + versioning that wouldn't have happened if the seeds were done in in versioned migration files. I could be misremembering this.

2. I'm not convinced of the benefit of maintaining a large and growing seed data file. For example, the combination of commodities and compliance codes would have made for a substantial seed data file on their own.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: This seems appropriate to me given that the applications and variances exist in the same table. It's a few small rules strictly pertaining to the presence/absence of certain values to maintain data integrity.

What are your concerns with this implementation?

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
james-hollinger: While not all of our static data is in one place, it really should be. We shouldn't have to trawl through 30 files just to see what the default data in our db's code tables are.

As well, if its all in one place we could theoretically drop and recreate the test db easier for testing in the future. The test db shouldn't be defined by constants files, it should be getting the same static data our app is getting.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: If they need different rules, then it feels like they should be different tables.... I'm not strongly against it, but we have no other constraints like this in the system, so just poking abit.... 

In theory we could write these as python @validates methods.... which are much more testable and we control them, we are currently running into hiccups when trying to catch db errors globally. @JustinMacaulay do we think we are going to able to do that?

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: 1. I don't recall seed data being the issue, but i'm not confident that it wasn't either. 

One large files that contains the seed data for the tables, separate from the table definitions themselves seems like it would be preferred over having data and structure distributed through dozens or hundreds of sql files.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: Well, we should have the model-level validation anyway. Relying on database errors would be a bad pattern, in my opinion.

We have `NOT NULL` constraints in the database for a reason. These are just `NOT NULL` for an approved variance.

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
NWCalvank: Okay. I'll concede

migrations/sql/V2019.04.25.11.30__add_variance_applications_metadata.sql
Jsyro: Okay. I'll concede

PR #698, opened 2019-04-29
MDS-1828 Feature/ Variance API Changes

NWCalvank:


python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -107,85 +76,44 @@ def get(self, mine_guid, variance_id):
...
-        self.parser.add_argument('document_manager_guid', type=str, required=True)
-
         variance = Variance.find_by_variance_id(variance_id)
+        if variance is None:
+            raise NotFound('Unable to fetch variance')
 
-        data = self.parser.parse_args()
-        document_name = data.get('document_name')
-        document_manager_guid = data.get('document_manager_guid')
+        data = VarianceResource.parser.parse_args()
```
Jsyro: any reason why this isn't self?

python-backend/app/api/mines/variances/resources/variance_document_upload.py
```diff
@@ -0,0 +1,96 @@
...
+        variance.save()
+        return variance
+
+    def _parse_request_metadata(self):
+        request_metadata = request.headers.get("Upload-Metadata")
+        metadata = {}
+        if not request_metadata:
+            return metadata
+
+        for key_value in request_metadata.split(","):
```
Jsyro: `key_value` looks odd, is this equal to `key, value`?

python-backend/app/api/mines/variances/resources/variance.py
```diff
@@ -107,85 +76,44 @@ def get(self, mine_guid, variance_id):
         return variance
```
Jsyro: This is related to the URL design discussion, the mine_guid is in the URL but isn't used to respected or used. I may comment on this after that discussion.

python-backend/app/api/mines/variances/resources/variance_document_upload.py
NWCalvank: I don't believe so. Based on the next line, it looks like it's a string in the form of "{key} {value}". I didn't look at the exact implementation logic here. It was just copied from `python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py` and then worked as-expected.

Jsyro: oh, i looked the next reference and missed it, that make sense. thanks.

python-backend/app/api/mines/mine_api_models.py
```diff
@@ -171,9 +171,19 @@ def format(self, value):
     'Variance', {
         'variance_id': fields.Integer,
```
nayr974: Really hate to not have noticed this before given all the discussion, but the integer PK shouldn't be exposed by our API and there should be a separate GUID that our API exposes. Integer PKs should be used internally by the DB for FKs, and the API can use them internally but not expose them to consumers. We had previously had discussion as a team led by Julian and settled on this pattern due to the efficiencies of integer PKs, but complications for database maintenance if we use an auto-incremented column as our global reference to an entity. I'd recommend merging this PR if it's blocking MD, then writing a migration script to add the GUID column and swap which is being exposed in a separate PR.

python-backend/app/api/mines/variances/resources/variance_list.py
```diff
@@ -0,0 +1,99 @@
+from werkzeug.exceptions import BadRequest
```
Jsyro: I see this differently, do we want the list and singular resources to be in the same file or separate files? I see this already done inconsistently. I'm on the fence but think i'd lean to having them on the same file cause they both function to serve one purpose.

Jsyro: I get the documents being separate. but just variance and variance_list.

python-backend/app/api/mines/variances/resources/variance_uploaded_documents.py
```diff
@@ -0,0 +1,30 @@
...
+    @api.doc(description='Delete a document from a variance.')
+    @requires_any_of([MINE_CREATE, MINESPACE_PROPONENT])
+    def delete(self, mine_guid, variance_id, mine_document_guid):
+        variance = Variance.find_by_mine_guid_and_variance_id(mine_guid, variance_id)
+        mine_document = MineDocument.find_by_mine_document_guid(mine_document_guid)
+
+        if variance is None:
+            raise NotFound('Variance not found.')
+        if mine_document is None:
+            raise NotFound('Mine document not found.')
```
Jsyro: there is no check to ensure that that document is actually attached to that variance. I believe if that was the case, it would either crash or do nothing (probably do nothing). Just thinking about our URL integrity discussion in this example.

python-backend/app/api/mines/mine_api_models.py
NWCalvank: Sure thing. I agree with that concern. I'd like to get this merged and open a new PR to make the guid changes

python-backend/app/api/mines/variances/resources/variance_uploaded_documents.py
NWCalvank: I had actually written a new lookup class method on `MineDocument` and used it here. But since `variance.documents.remove(mine_document)` won't find the document if it's not associated with the variance, I didn't worry about it at this time in an effort to limit the scope of this PR.

However, I did not test what the response looks like if an incorrect mine_document_guid is provided.

python-backend/app/api/mines/variances/models/variance.py
```diff
@@ -68,22 +73,28 @@ def create(
 
     @classmethod
     def find_by_mine_guid(cls, mine_guid):
-        try:
-            uuid.UUID(str(mine_guid), version=4)
-            return cls.query.filter_by(mine_guid=mine_guid).all()
-        except ValueError:
-            return None
+        cls.validate_guid(mine_guid, INVALID_MINE_GUID)
```
james-hollinger: This actually changes the functionality of this method, throwing an error if the guid is invalid rather than returning None. Are we good with this?

NWCalvank: That was an intentional change. I guess there could be an argument to keep the `None` check + error raising at every invocation of this lookup, but I think it's valuable to catch the invalid GUID in the model and avoid duplicating that `None` check everywhere.

Not a big deal for me to revert this change if we prefer the original way that I was doing this.

If we wanted to suppress the error for some reason, we could always try/except at the callsite. It seems to me that that would be the odd use case, and we'd usually want to throw for the invalid GUID.

This does raise the point that I should remove any existing `None` checks.

python-backend/app/api/mines/variances/resources/variance_list.py
NWCalvank: Hey Jason, I actually split these because you had requested it in my original Variances PR and since I was splitting out documents, I decided to split everything. I guess your opinion has changed?

PR #700, opened 2019-04-29
MDS-1817 add V1 to API

JustinMacaulay:
Adding an optional v1 to the api.


nginx/conf.d/api.core.conf.tmpl
```diff
@@ -9,6 +9,15 @@ location ${PATH_PREFIX}/api/ {
                 return 503;
         }
 
+        proxy_pass ${API_SERVICE_URL}${PATH_PREFIX}/api/;
+        include /opt/app-root/etc/nginx.conf.d/proxy.params;
+}
+
+location ${PATH_PREFIX}/api/v1 {
```
FreshGyan: You may have to add a trailing slash to v1

JustinMacaulay: Added.

nayr974:
> 
> 
> Perhaps I'm missing something, but this would require an update to /frontend and /frontend-public, right?

No, if you look at the full file the original /api proxy persists, and /api/v1 is added. Our FE will directly call /api, and external users will call the /v1 proxy so if we need to manage their requests differently that option is open to us.

NWCalvank:
@nayr974 So, we're not treating our FE as an external consumer? That's antithetical to our other design decisions, no? It doesn't seem to me that the FE should be special.

itchison:
@NWCalvank @nayr974  We certainly can switch the Front End to use the /v1/ if we want to. We have an immediate need to give the NRIS team a versioned API so we decided to do a quick and dirty faux versioning.

For this PR we're deferring the "how we're going to version our API" so that we can discuss it deeper without the pressure of this immediate need.

nayr974:
> @NWCalvank @nayr974 We certainly can switch the Front End to use the /v1/ if we want to. We have an immediate need to give the NRIS team a versioned API so we decided to do a quick and dirty faux versioning.
> 
> For this PR we're deferring the "how we're going to version our API" so that we can discuss it deeper without the pressure of this immediate need.

We certainly can. My thought is that since the same team is publishing FE and API through at the same time through the same mechanisms, we may actually want to just reference /api directly instead of having some additional (admittedly fairly trivial) overhead of adding /v2/ in the future if we actually have different logic for third parties due to some breaking changes.

PR #701, opened 2019-04-29
MDS-1854 Update status filtering on Compliance Table

FreshGyan:
- Needs Unit tests updated
- Needs some business logic validation

Haven't done FE/BE in a while so just opening up one a bit early to get feedback

### DIFF
--- a/frontend/src/components/mine/MineDashboard.js
+++ b/frontend/src/components/mine/MineDashboard.js
@@ -199,7 +199,7 @@ export class MineDashboard extends Component {
   handleFiltering = (order, params) => {
     // convert string to boolean before passing it into a filter check
     const order_status =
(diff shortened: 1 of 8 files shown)


frontend/src/components/mine/MineDashboard.js
```diff
@@ -198,8 +198,8 @@ export class MineDashboard extends Component {
 
   handleFiltering = (order, params) => {
     // convert string to boolean before passing it into a filter check
-    const parsedOverdue = params.overdue === "" ? "" : JSON.parse(params.overdue);
-    const overdue = params.overdue === "" || order.overdue === parsedOverdue;
+    const order_status =
+      params.order_status === "" || order.order_status.toString().includes(params.order_status);
```
mmumby: isn't `order_status` already a string?

PR #706, opened 2019-04-30
Feature/mds 1822

thegentlemanphysicist:
Toggle added to mine dashboard page map.


frontend/src/components/maps/MineMap.js
```diff
@@ -51,7 +51,10 @@ class MineMap extends Component {
   handleLoadMap = (map, view) => {
     if (!this.props.mine) {
       this.renderWidgets(view);
+    } else {
+      this.renderBasemapToggle(view);
```
FreshGyan: I know the ticket says only the small map but since it's just a toggle can you double check that's all they really want? I see no harm in exposing the option on both maps 🤷‍♂

PR #711, opened 2019-05-02
Infrastructure/nris testing

Jsyro:
build configs and deploy configs at the microservice level. 

Python image needs to install OS packages for cxOracle to work... for later.


Jsyro:
What we have so far: 
mds-nris-backend pod that 

- Has a connection to the NRIS Oracle database through cx_Oracle
- Has a connection to the mds_postgres container with an nris user that has access to a new schema named 'nris' 
- Uses Flask-Migrate (alembic) to write code first migrations to the nris schema

Concerns: 
- the nris postgres user's password is hard coded and in source control. This will be addressed when we establish the proper way to initalize the database that doesn't rely on flyway
- It has a docker compose that spins up an empty database locally, or the main docker-compose can spin it up to work with a database that has mutiple schemas, more like openshift. 
- cx_Oracle relies on OS dependencies that we are satisfying by loading and unziping ~100mb of zipped data, openshift hasn't been great about preserving that base image. (~20 minutes builds)
- No deployment config for TEST or PROD, will only be stood up for DEV stage. 
- No unit testing yet. 
- No Sonarqube integration yet

Jsyro:
Also no JWT based auth yet

microservices/nris_api/backend/.env-example
```diff
@@ -0,0 +1,21 @@
+export DB_HOST=postgres
+export DB_USER=
```
itchison: Should these be blank? They aren't for MDS.

docker-images/python-36-oracle/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+USER 0
+
+RUN yum -y install libaio
+
+# yum install removed for openshift build
+# pkgs build-essential libaio1 python-minimal - don't exist for RHEL 
+# pkgs curl unzip git - are all already installed on python RHEL image
+
+RUN unzip /tmp/oracle/instantclient-basic-linux.x64-19.3.0.0.0.zip -d ${APP_ROOT}/oracle_bin && \
```
itchison: Have you seen any license information on these zip files and if we're allowed to copy them into github?

microservices/nris_api/backend/app/extensions.py
```diff
@@ -0,0 +1,28 @@
...
+from flask_sqlalchemy import SQLAlchemy
+from flask_apscheduler import APScheduler
+from flask_migrate import Migrate, MigrateCommand
+from flask import current_app
+from elasticapm.contrib.flask import ElasticAPM
+
+from .config import Config
+from .helper import Api
+
+apm = ElasticAPM()
```
itchison: Is the APM complete?

microservices/nris_api/backend/.env-example
Jsyro: It's the .env-example file, and those environment variables are used by the nris_api to connect to the mds_core postgres database with the nris user.

microservices/nris_api/backend/app/extensions.py
Jsyro: local doesn't hook up to APM, depends on if the elastic env variables are being set in the deployment config properly, which i have not validated. Should something be set/changed to differentiate this traffic from the core_api traffic?

microservices/nris_api/backend/requirements.txt
```diff
@@ -0,0 +1,45 @@
+alembic==1.0.0
```
FreshGyan: I know these started off as a copy of the python backend requirements. I did an audit on those and more than half requirements we were able to get rid of and upgrade the rest.
Can you remove the ones you're not using from here?

microservices/nris_api/backend/run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -exv -o pipefail
+
+coverage erase
+coverage run --branch --source=app -m pytest
+coverage xml -i
+
+/opt/app-root/sonar-scanner-cli/bin/sonar-scanner
```
FreshGyan: Does this actually run? I thought you didnt install sonar-scanner yet

microservices/nris_api/docker-compose.yaml
```diff
@@ -0,0 +1,45 @@
...
+  # Only for Local, openshift will share one db pod for all MDS services
+  nris_database:
+    restart: always
+    container_name: nris_postgres
+    build:
+      context: postgres
+      dockerfile: Dockerfile.dev
+    environment:
+      - POSTGRES_USER=nris
+      - POSTGRES_PASSWORD=VsPJAuMJHu3IVMEv
```
FreshGyan: Is this just for local? Usually local db's have passwords like password or something easy so you dont have to remember it to copy/paste etc.
Just a personal preference :)

Also im hoping this password isnt used anywhere 😅

microservices/nris_api/openshift/_python36_oracle.bc.json
```diff
@@ -0,0 +1,233 @@
...
+            "ref": "${SOURCE_REPOSITORY_REF}"
+          },
+          "contextDir": "${SOURCE_CONTEXT_DIR}"
+        },
+        "strategy": {
+          "type": "Source",
+          "sourceStrategy": {
+            "from": {
+              "kind": "ImageStreamTag",
+              "name": "${NAME}-cxoracle-base:${VERSION}"
```
FreshGyan: Since you're already passing the NAME variable i dont think you need the cxoracle in there 🤷‍♂ 
If it looks more or less the same as the exisiting bc we avoid confusion during debugging.

microservices/nris_api/openshift/_python36_oracle.dc.json
```diff
@@ -0,0 +1,444 @@
...
+        "annotations": {
+          "description": "Defines how to deploy the application server"
+        }
+      },
+      "spec": {
+        "strategy": {
+          "type": "Rolling",
+          "rollingParams": {
+            "post": {
+              "failurePolicy": "ignore",
```
FreshGyan: Don't you want to abort the deployment if the migration fails? If it silently ignores it we'll never know.

Also for Pre/Post hook. You may know better how alembic functions with the code but just wanted to point it out.

Coz with pre the behaviour is ->
The old app runs while the new migrations gets executed, if it works then new app gets scaled and old apps gets killed. Im not sure if alembic will complain about running migrations on a database while something is connected to it 🤷‍♂ 

and with post ->
The old app gets killed, the new one scales and then the migration is run. Now if the migration fails, you're still running new version of code and that may not work as expected.

docker-compose.yml
```diff
@@ -107,6 +107,24 @@ services:
     networks:
       - mds
 
+  ####################### NRIS_BACKEND Definition #######################
+  nris_backend:
+    restart: always
+    container_name: nris_python
+    build:
+      context: microservices/nris_api/backend
+      dockerfile: Dockerfile
```
NWCalvank: I'm fairly confident that this is redundant since you've added the context

microservices/nris_api/openshift/_python36_oracle.dc.json
```diff
@@ -0,0 +1,444 @@
...
+                "kind": "ImageStreamTag",
+                "name": "${NAME}:${VERSION}"
+              }
+            }
+          },
+          {
+            "type": "ConfigChange"
+          }
+        ],
+        "replicas": "${{REPLICA_MIN}}",
```
FreshGyan: Just in case you're curious, the double {{ bracket templates input as a number whereas the single { bracket templates as a string.

microservices/nris_api/openshift/_python36_oracle.dc.json
```diff
@@ -0,0 +1,444 @@
...
+                        "name": "${MDS_NRIS_ODB_CREDENTIALS}",
+                        "key": "db-host"
+                      }
+                    }
+                  },
+                  {
+                    "name": "NRIS_DB_SERVICENAME",
+                    "valueFrom": {
+                      "secretKeyRef": {
+                        "name": "${MDS_NRIS_ODB_CREDENTIALS}",
```
FreshGyan: Dunno if MDS is redundant here 🤷‍♂

microservices/nris_api/openshift/_python36_oracle.dc.json
```diff
@@ -0,0 +1,444 @@
...
+                    "name": "NRIS_DB_PORT",
+                    "value": "1521"
+                  },
+                  {
+                    "name": "ELASTIC_ENABLED",
+                    "value": "${ELASTIC_ENABLED}"
+                  },
+                  {
+                    "name": "ELASTIC_SERVICE_NAME",
+                    "value": "${ELASTIC_SERVICE_NAME}"
```
FreshGyan: As for the question with APM above, im fairly certain the service name is what differentiates different application on APM.
@itchison may know better tho

docker-compose.yml
```diff
@@ -107,6 +107,24 @@ services:
...
+    ports:
+      - 5500:5500
+    volumes:
+      - ./microservices/nris_api/backend/:/app/
+    depends_on:
+      - flyway
+    # Export environment variables from .env into the container environment
+    entrypoint: ["bash", "-c", "grep -v '^#' .env | xargs ; flask run"]
```
NWCalvank: `entrypoint` can be a bash file, which might be easier to maintain than this command?

Is there a reason not to just COPY an .env file with the relevant args into the container and then `. .env` or `source .env` if necessary when initializing the container? That could all be managed in the Dockerfile instead of in the compose file

microservices/nris_api/backend/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN mkdir /app
+WORKDIR /app
+
+# Install the requirements
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Run the server
+EXPOSE 5500
+CMD [ "flask", "run"]
```
NWCalvank: You execute `flask run` here and in docker-compose? As far as I can tell, that will cause it to run twice. Is that intentional?

openshift/postgresql.dc.json
```diff
@@ -303,6 +336,24 @@
                       }
                     }
                   },
+                  {
```
FreshGyan: You don't need to inject the secrets as environment variable to the database container

openshift/postgresql.dc.json
```diff
@@ -172,6 +186,25 @@
         "oracle-fdw-schema": null
       }
     },
+    {
```
FreshGyan: You can probably move this to the python deployment config until you get a separate postgres deployment config

pipeline/config-build.groovy
```diff
@@ -43,6 +43,17 @@ app {
                             'SOURCE_REPOSITORY_URL': "${app.git.uri}"
                     ]
                 ],
+                // bc for NRIS API Backend
```
FreshGyan: unused comment?

microservices/nris_api/backend/README.md
```diff
@@ -0,0 +1,81 @@
...
+    |-- utils (Contains utility shared across modules)
+|-- tests (Unit/Integration tests for the application)
+|-- app.sh (Shell script used by the python OpenShift s2i image to run the application)
+|-- Dockerfile (Dockerfile for running the application locally using Docker)
+|-- requirements.txt (Libraries required by the project)
+```
+
+## Pre-requisites and Installation
+
+The application assumes you already have a working postgres DB with the required schema and tables and have the connection details in the .env file.
```
NWCalvank: Why is that assumption made? Is that because the microservice is (currently) meant to leverage the single MDS database that already exists?

If that's the case, I'm not sure about the phrasing here. It seems to indicate that you need postgres running on the host.

pipeline/config-dev.groovy
```diff
@@ -178,6 +179,32 @@ app {
...
+                ],
+                [
+                    'file':'microservices/nris_api/openshift/_python36_oracle.dc.json',
+                    'params':[
+                            'NAME':"mds-nris-backend",
+                            'SUFFIX': "${vars.deployment.suffix}",
+                            'VERSION':"${app.deployment.version}",
+                            'CPU_REQUEST':"${vars.resources.python.cpu_request}",
+                            'CPU_LIMIT':"${vars.resources.python.cpu_limit}",
+                            'MEMORY_REQUEST':"${vars.resources.python.memory_request}",
```
FreshGyan: We may want to tweak the memory/cpu request and pod resources when this reaches test and prod.
Since its much more lightweight it probably can get away with less resources than our API

microservices/nris_api/backend/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+To see the list of all click commands, checkout `register_commands` method under the `__init__.py` file.
+
+## Testing
+
+The application uses `pytest` to run the tests and `coverage` to output the results. The testing structure is based around [flask testing documentation.](http://flask.pocoo.org/docs/1.0/testing/)
+
+To run tests, use the following command:
+```
+coverage run --branch --source=app -m pytest
```
NWCalvank: This command works both when the app is running on the host AND when running in Docker?

microservices/nris_api/backend/app/helper.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Api(BaseApi):
+    """
+    Monkey patch Swagger API to return HTTPS URLs
+    """
+
+    @property
+    def specs_url(self):
+        scheme = 'http' if '5500' in self.base_url else 'https'
```
NWCalvank: Is this the same solution we're using in CORE? Seems like this would be configurable directly in Swagger

microservices/nris_api/backend/app/routes.py
```diff
@@ -0,0 +1,28 @@
+from app.utils.logger import get_logger
```
NWCalvank: This file is proposing a different route-management pattern than what is used in CORE at the moment? Is there a reason for this, or this is just filler code while the deployment is being set up initially?

microservices/nris_api/postgres/Dockerfile.dev
```diff
@@ -0,0 +1,7 @@
+FROM postgres:9.6
+
+# Copy only migration scripts for creation
+RUN echo "CREATE USER postgres" >> /docker-entrypoint-initdb.d/01-create-user.sql
```
NWCalvank: How/when/where is this directory being created? And is this command just meant to create & fill that sql file with the command string?

If so, why not just have that file COPY'd with the content already in it? That seems like the simpler solution to me.

microservices/nris_api/postgres/Dockerfile.dev
```diff
@@ -0,0 +1,7 @@
...
+
+# Set the working directory to container entrypoint
+WORKDIR /docker-entrypoint-initdb.d
```
NWCalvank: Is this file complete? What's the reason for setting WORKDIR as the last line of the file?

docker-compose.yml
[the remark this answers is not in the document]
Jsyro: I actually tried it without and it didn't work, i can try it again, i knew it was weird.

docker-images/python-36-oracle/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN mkdir -p /tmp/oracle
+COPY oracle/ /tmp/oracle
+
+RUN mkdir -p ${APP_ROOT}/oracle_bin
+
+USER 0
+
+RUN yum -y install libaio
+
+# yum install removed for openshift build
```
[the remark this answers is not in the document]
Jsyro: removed.

docker-images/python-36-oracle/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN unzip /tmp/oracle/instantclient-basic-linux.x64-19.3.0.0.0.zip -d ${APP_ROOT}/oracle_bin && \
+    unzip /tmp/oracle/instantclient-sdk-linux.x64-19.3.0.0.0.zip -d ${APP_ROOT}/oracle_bin && \
+    mv ${APP_ROOT}/oracle_bin/instantclient_19_3 ${APP_ROOT}/oracle_bin/instantclient 
+
+RUN echo '${APP_ROOT}/oracle_bin/instantclient/' | tee -a /etc/ld.so.conf.d/oracle_instant_client.conf && ldconfig && \
+    rm -rf /tmp/oracle/instantclient-basic-linux.x64-19.3.0.0.0.zip /tmp/oracle/instantclient-sdk-linux.x64-19.3.0.0.0.zip && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN ls -la ${APP_ROOT}/oracle_bin
```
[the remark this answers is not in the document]
Jsyro: removed.

microservices/nris_api/backend/.env-dev-local-keycloak
```diff
@@ -0,0 +1,19 @@
+export DB_HOST=oracle
```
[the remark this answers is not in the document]
Jsyro: updated.

microservices/nris_api/backend/.env-example
```diff
@@ -0,0 +1,21 @@
+export DB_HOST=postgres
```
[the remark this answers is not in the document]
Jsyro: updated.

microservices/nris_api/backend/README.md
```diff
@@ -0,0 +1,81 @@
+# Backend API
```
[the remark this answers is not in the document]
Jsyro: also added flask upgrade command to readme.

microservices/nris_api/backend/run_tests.sh
Jsyro: no it does not. building on it next PR.

microservices/nris_api/docker-compose.yaml
Jsyro: no it's not just for local, its the same password coded into the DC until we can dynamically create that pw, because this user is create by the same flyway script that runs on openshift, they are the same credentials for now .

microservices/nris_api/openshift/_python36_oracle.bc.json
Jsyro: seems reasonable, removed `cxoracle`.

microservices/nris_api/openshift/_python36_oracle.dc.json
Jsyro: yea pre sounds better, i suppose that makes sense 'pre' means, before it's used, not before it's deployed. and yes, ignore was because i wanted to test other parts of the dc as well. it's now retry.

microservices/nris_api/openshift/_python36_oracle.dc.json
Jsyro: APM is going to be coming in later, i believe you are correct and i plan on tackling that next PR.

openshift/postgresql.dc.json
Jsyro: not currently, however i was hoping that it would be a bash script on the postgres container that would create the nris user from the secret, if that is not the long term solution then i suppose this could be removed.

pipeline/config-build.groovy
Jsyro: removed.

pipeline/config-dev.groovy
Jsyro: Probably true, good point, not sure i'm worried about that level in the PR, let me know if you want those changes now.

docker-compose.yml
Jsyro: hmmm that's an artifact from copy-paste... but it does smell, let me take a look.

microservices/nris_api/backend/Dockerfile
Jsyro: Same as above, i'll take a look .

microservices/nris_api/backend/README.md
Jsyro: Copy paste from existing readme, it's saying that the backend does nothing to stand up a database, that needs to be separate and you need to provide this service those credentials. Thats how i read it. Do you want some changes to it?

microservices/nris_api/backend/app/helper.py
Jsyro: Yup, same monkey patch in CORE, i didn't look it into it.

microservices/nris_api/backend/app/routes.py
Jsyro: Not sure it's a different route management, can you please clarify. We have to access the application logger through different means if it's the scheduler running the app or a request... so i was hoping to abstract that complexity so it's easier to get logging in from the start.

microservices/nris_api/postgres/Dockerfile.dev
Jsyro: the postgres s2i looks for that specific files and runs it i think? and we use env variables to change the superuser, but use this to still make sure there is a `postgres` user in the db.... i think that's right. @FreshGyan?

docker-compose.yml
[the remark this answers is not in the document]
Jsyro: it worked without it... cool

microservices/nris_api/backend/migrations/versions/72047fbe36af_.py
```diff
@@ -0,0 +1,35 @@
+"""empty message
```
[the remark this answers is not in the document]
Jsyro: added upgrade command to readme, i'll put the migrate command somewhere too.

microservices/nris_api/backend/requirements.txt
Jsyro: How about this, i'm gonna comment out a bunch when we start working on this a bunch, and then we will uncomment if they are needed or delete them if it never comes up. I don't really want to spend overhead time just testing this (there are no tests or things that use these libraries, ever though they will in the next two weeks.

docker-compose.yml
pdittaro: Any reason that we aren't using the `env-file` parameter for docker-compose? 
https://docs.docker.com/compose/compose-file/#env_file

microservices/nris_api/backend/Dockerfile
pdittaro: Passing 'CMD' in docker-compose overwrites the CMD in the dockerfile, not an addendum. Only one CMD per container.

microservices/nris_api/backend/migrations/versions/72047fbe36af_.py
```diff
@@ -0,0 +1,35 @@
...
+    sa.Column('output_val', sa.Integer(), nullable=False),
+    sa.Column('exec_timestamp', sa.DateTime(), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('input_val')
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
pdittaro: Its also kind of strange to see an ORM so enthusiastically telling you to overwrite their generated code. 

Maybe we can talk about getting some docstring consistency in the py projects?

microservices/nris_api/postgres/Dockerfile.dev
pdittaro: I think setting workdir also sets the context for any CMD or ENTRYPOINT commands passed into the container at runtime. E.g. `docker run container_name relative_file_path` , when the container executes its entrypoint script it will look at the `relative_file_path` command passed to it in the workdir directory.

openshift/postgresql.dc.json
FreshGyan: Hmm 🤔 Okay yeah that makes sense. Lets ship that and we can then find out if we use it later or not 👍

pipeline/config-dev.groovy
FreshGyan: Eh i mean dev is good. Can deal with that later when it gets to test/prod 💯

microservices/nris_api/backend/app/helper.py
FreshGyan: yeah I had to monkey patch this in core mostly because of how swagger is built without reverse proxying in mind :(

microservices/nris_api/postgres/Dockerfile.dev
FreshGyan: This is only for local development so s2i hasnt played a role yet.

However this was added because by default the docker image that comes from dockerhub creates the superuser with whatever username you provide. i.e. mds
But the docker image that comes from redhat already comes with a superuser named postgres.

The flyway migrations use the postgres user and thus that user is just created so that locally we dont have failing migrations.



The reason i went with just echoing the contents instead of COPY is because then i'd have to grab a copy of the entrypoint file that's supplied by postgres and then add my command and overwrite the one in the container and that just seemed like too much work to last year me 😅

microservices/nris_api/postgres/Dockerfile.dev
FreshGyan: It may work without the WORKDIR directive as well. I think I had errors and then I ended up putting that in, but that was a while back and im not certain if this is needed or not anymore

microservices/nris_api/backend/app/routes.py
Jsyro: are you referring to `backend/app/nris` being equal to `python-backend/app/api`, but it's `backend/app/utils` instead of `python-backend/app/api/utils`, if so that's a good eye, but there are multiple branches based on these changes, so i'd have to ask the team for a full stop when i make that change and merge, so I agree utils should be under `backend/app/nris`, but i don't think that change can be made right now.

microservices/nris_api/backend/README.md
Jsyro: yes. `./run_tests.sh` runs the tests in the same way in local docker, and openshift. may should also work on local native.

microservices/nris_api/backend/README.md
NWCalvank: My only concern is that it seems to apply when you're running the app on the host, but not when it's running in Docker. Certainly not a blocker for this PR, as it's copy/pasted. There's likely a need to have a full ticket to address the current state of all of our READMEs.

I'll mark this as resolved

microservices/nris_api/backend/app/routes.py
NWCalvank: I may just be misunderstanding the purpose of this file 😅 

By looking at it, I thought this would be replacing an `/api/model/namespace/model.py` file by putting all of the routes for this microservice into one `routes` file.

If that's not correct then just disregard my comment. If it is correct, my question is the benefit proposition of having this different pattern in the microservice vs the cognitive overhead of differing patterns.

Jsyro: I see where the confusion lies, in the core backend we currently have a function called register_routes, it's getting quite long with all the namespace imports and exception handlers, so i made a file called routes just holds that domain of initialization. 

Does that make sense. I am not suggesting we have a folder that contains all of the resource files.

NWCalvank: Yep. All good 👍 I'll mark this as resolved

PR #712, opened 2019-05-02
MDS-1868 Feature/ Add created_at to variance documents

NWCalvank:
The `uploaded_at` value is a timestamp of when the xref record was created. Interpreted as business logic, this is the moment when the document became associated with the variance.

- Adds `uploaded_at`
- Renames `VarianceDocument` to `VarianceDocumentXref`, which was a requested change in an earlier PR

### DIFF
--- a/python-backend/app/api/documents/variances/models/variance.py
+++ b/python-backend/app/api/documents/variances/models/variance.py
@@ -15,7 +15,7 @@ class VarianceDocumentXref(Base):
     variance_id = db.Column(db.Integer,
                             db.ForeignKey('variance.variance_id'),
                             server_default=FetchedValue())
-    uploaded_at = db.Column(db.DateTime, server_default=FetchedValue())
+    created_at = db.Column(db.DateTime, server_default=FetchedValue())
 
     mine_document = db.relationship('MineDocument', lazy='joined')
 
--- a/python-backend/app/api/documents/variances/models/variance.py
+++ b/python-backend/app/api/documents/variances/models/variance.py
@@ -1,9 +1,11 @@
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
+from sqlalchemy.ext.associationproxy import association_proxy
 
 from ....utils.models_mixins import Base
 from app.extensions import db
 
+# TODO: Rename to VarianceDocumentXref
 class VarianceDocument(Base):
     __tablename__ = "variance_document_xref"
     variance_document_xref_guid = db.Column(UUID(as_uuid=True),
@@ -14,6 +16,13 @@ class VarianceDocument(Base):
     variance_id = db.Column(db.Integer,
                             db.ForeignKey('variance.variance_id'),
                             server_default=FetchedValue())
+    uploaded_at = db.Column(db.DateTime, server_default=FetchedValue())
+
+    mine_document = db.relationship('MineDocument', lazy='joined')
+
+    mine_guid = association_proxy('mine_document', 'mine_guid')
+    document_manager_guid = association_proxy('mine_document', 'document_manager_guid')
+    document_name = association_proxy('mine_document', 'document_name')
 
     def __repr__(self):
         return '<VarianceDocument %r>' % self.variance_document_xref_guid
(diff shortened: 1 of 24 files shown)


python-backend/app/api/documents/variances/models/variance.py
```diff
@@ -14,6 +15,13 @@ class VarianceDocument(Base):
     variance_id = db.Column(db.Integer,
                             db.ForeignKey('variance.variance_id'),
                             server_default=FetchedValue())
+    uploaded_at = db.Column(db.DateTime, server_default=FetchedValue())
```
Jsyro: should uploaded_at be a column on the mine_document table.

python-backend/app/api/documents/variances/models/variance.py
```diff
@@ -14,6 +15,13 @@ class VarianceDocument(Base):
...
+
+    mine_document = db.relationship('MineDocument', lazy='joined')
+
+    mine_guid = association_proxy('mine_document', 'mine_guid')
+    document_manager_guid = association_proxy('mine_document', 'document_manager_guid')
+    document_name = association_proxy('mine_document', 'document_name')
```
Jsyro: haven't seen `assocation_proxy` yet, but this is cool to see, however you are already loading the mine_document relationship, so it's saving you `mine_document.` however that could be very cool to use with the swagger models (show related attributes as being on the object.

python-backend/app/api/mines/variances/models/variance.py
```diff
@@ -39,7 +39,8 @@ class Variance(AuditMixin, Base):
     received_date = db.Column(db.DateTime, nullable=False)
     expiry_date = db.Column(db.DateTime)
 
-    documents = db.relationship('MineDocument', lazy='joined', secondary='variance_document_xref')
+    documents = db.relationship('VarianceDocumentXref', lazy='joined')
```
Jsyro: I don't think this attribute is necessary .

python-backend/app/api/documents/variances/models/variance.py
NWCalvank: I'm not totally sure what you're suggesting here. I'm using association proxy here to add those attributes to the variance model so that the marshalling works properly. It references the mine_document relationship

python-backend/app/api/mines/variances/models/variance.py
NWCalvank: This attribute is necessary because I want to marshal the `VarianceDocumentXref` model now, not the `MineDocument` model because `MineDocument` doesn't have the `uploaded_at` attribute.

As I mentioned in the opening PR comment, the business case for `uploaded_at` is when the mine_document becomes associated with the variance. So, it needs to (as far as I can tell) live on `variance_document_xref`.

Does that make sense?

Jsyro: if that's the case then the column title is a bit unclear. how about `attached_at` or `created_at`

python-backend/app/api/documents/variances/models/variance.py
Jsyro: Maybe i'm not understanding out assocation_proxy works, i thought it makes an attribute on a related object look like it's on this object. the variance_document swagger model in the mine_api_models doesn't have any of these three. Where are these coming back through marshalling?

python-backend/app/api/mines/variances/models/variance.py
NWCalvank: `created_at` is probably a better name because it reflects both the implementation details and the business use case.

The reason for `uploaded_at` is/was that the business wants "the time the document was uploaded to the variance." I think that `uploaded_at` reflects that use case. It's not the businesses fault that we allow document uploads to a non-existent variance. Which is a process/user flow that I think should be changed and is currently being investigated by the UX team.

Point being, if you don't feel good about `uploaded_at`, I can rename it to `created_at`, but I also think the concern with `uploaded_at` is a symptom of weird implementation details that don't necessarily reflect the end-user's goal(s) / use case.

python-backend/app/api/documents/variances/models/variance.py
NWCalvank: It does have these. It inherits from the mine document swagger model

python-backend/app/api/mines/variances/models/variance.py
Jsyro: documents may live in the document manager before they are added to the api, but the record in mine_documents shouldn't exist until the variance does. I haven't looked at the variance implementation, i'm referring to the permit/permit amendment document workflow. the odd thing is that `VarianceDocumentXref`s aren't 'uploaded', so `created_at` seems like a better name.

PR #714, opened 2019-05-03
Update create_data command to use test factories for data generation

nayr974:


python-backend/app/commands.py
```diff
@@ -1,60 +1,33 @@
...
+from app.api.constants import (PERMIT_STATUS_CODE, MINE_OPERATION_STATUS,
+                               MINE_OPERATION_STATUS_REASON, MINE_OPERATION_STATUS_SUB_REASON,
+                               MINE_REGION_OPTIONS)
+from app.api.mines.mine.models.mine import Mine
+from app.api.mines.mine.models.mineral_tenure_xref import MineralTenureXref
+from app.api.mines.mine.models.mine_tenure_type_code import MineTenureTypeCode
+from app.api.parties.party.models.party import Party
+from app.api.parties.party.models.party_type_code import PartyTypeCode
+from app.api.permits.permit.models.permit import Permit
+from app.api.mines.mine.models.mine_verified_status import MineVerifiedStatus
```
Jsyro: Can some of these imports be removed?

nayr974: i dunno, i guess so, if i have to

nayr974: removed 👍

PR #715, opened 2019-05-03
Feature/mds 1668

thegentlemanphysicist:
Adding open map in new page funtionality.

### DIFF
--- a/frontend/src/components/dashboard/minesHomePage/Dashboard.js
+++ b/frontend/src/components/dashboard/minesHomePage/Dashboard.js
@@ -98,8 +98,8 @@ export class Dashboard extends Component {
     this.handleMineSearchDebounced = debounce(this.handleMineSearch, 1000);
     this.state = {
       mineList: false,
-      lat: String.DEFAULT_LAT,
-      long: String.DEFAULT_LONG,
+      lat: Number(String.DEFAULT_LAT),
+      long: Number(String.DEFAULT_LONG),
       zoom: 6,
       showCoordinates: false,
       mineName: null,
@@ -154,7 +154,12 @@ export class Dashboard extends Component {
 
   componentWillUnmount() {
     this.handleMineSearchDebounced.cancel();
-    this.setState({ params: {}, lat: String.DEFAULT_LAT, long: String.DEFAULT_LONG, zoom: 6 });
+    this.setState({
+      params: {},
+      lat: Number(String.DEFAULT_LAT),
+      long: Number(String.DEFAULT_LONG),
+      zoom: 6,
+    });
   }
 
   renderDataFromURL = (params) => {
@@ -196,9 +201,9 @@ export class Dashboard extends Component {
     // set the lat, long, zoom, and blinking to true
     if (format(lat)[0] && format(long)[0]) {
       this.setState({
-        lat: format(lat)[0],
-        long: format(long)[0],
-        zoom: format(zoom)[0] ? format(zoom)[0] : 6,
+        lat: Number(format(lat)[0]),
+        long: Number(format(long)[0]),
+        zoom: format(zoom)[0] ? Number(format(zoom)[0]) : 6,
       });
       this.handleNavitationFromMine();
     }
@@ -238,8 +243,8 @@ export class Dashboard extends Component {
         });
       } else {
         this.setState({
-          lat: String.DEFAULT_LAT,
-          long: String.DEFAULT_LONG,
+          lat: Number(String.DEFAULT_LAT),
+          long: Number(String.DEFAULT_LONG),
           zoom: 6,
           showCoordinates: false,
         });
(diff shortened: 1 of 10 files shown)


frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -189,6 +198,15 @@ export class Dashboard extends Component {
         this.setState({ mineList: true });
       });
     }
+    // set the lat, long, zoom, and blinking to true
```
NWCalvank: I don't see anything about blinking in the following code? Since it's clear that `setState` is setting lat/long/zoom, I'd suggest removing this comment.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -248,9 +268,26 @@ export class Dashboard extends Component {
     }
   };
 
+  handleNavitationFromMine = () => {
+    // TODO: spent 4 hours looking for a solution to not hardcoding this scroll value. Need to find a dynamic way of scroling the screen to this location.
```
NWCalvank: I know this isn't your comment, but could you update this to something like `FIXME: scroll dynamically instead of by a hardcoded value`?

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -235,6 +254,7 @@ export class Dashboard extends Component {
       this.setState({
         lat: Number(value.latitude),
         long: Number(value.longitude),
+        zoom: 14,
```
NWCalvank: Can you assign these 14s and 6s to a `const` at the top of the file so that they're not just magic numbers

frontend/src/components/mine/MineHeader.js
```diff
@@ -249,6 +252,17 @@ export class MineHeader extends Component {
...
 
+    const mapRoute = this.props.mine.mine_location
+      ? route.MINE_HOME_PAGE.mapRoute(
+          null,
+          null,
+          null,
+          this.props.mine.mine_location.latitude,
+          this.props.mine.mine_location.longitude,
+          12
+        )
```
NWCalvank: This is far too many positional arguments. 3 is generally the acceptable max, imo. If you want to pass in more arguments, I'd recommend passing in an object so that the arguments can be passed in by name.

frontend/src/constants/routes.js
```diff
@@ -22,9 +22,12 @@ export const MINE_HOME_PAGE = {
...
     `/dashboard/mines/?${queryString.stringify({ page, per_page, ...params }, { sort: false })}`,
-  mapRoute: (page, perPage, search = null) => {
-    const searchParam = search ? `&search=${search}` : "";
-    return `/dashboard/mines?page=${page}&per_page=${perPage}${searchParam}&map=true`;
+  mapRoute: (page, perPage, search = null, lat = null, long = null, zoom = null) => {
+    let searchParam = search ? `&search=${search}` : "";
+    searchParam += lat ? `&lat=${lat}` : "";
+    searchParam += long ? `&long=${long}` : "";
+    searchParam += zoom ? `&zoom=${zoom}` : "";
+    return `/dashboard/mines?${searchParam}&map=true`;
```
NWCalvank: Positional arguments need to be updated, as mentioned in my last comment. An use `queryString` instead of manually concatenating all of the query string params together. You can look at `dynamicRoute` for an example of how to do this

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -207,38 +229,32 @@ export class Dashboard extends Component {
    */
   handleCoordinateSearch = (value) => {
     if (typeof value === "string") {
+      // Case: searching my mine name
       const newVal = value.split(",");
```
mmumby: This comment looks out of place, either remove or make it more clear

frontend/src/components/dashboard/minesHomePage/Dashboard.js
mmumby: looks like `handleNavitationFromMine` contains all the scroll logic, would be beneficial to rename this function to `handleScroll` or something similar, and invoke it in every situation there is scroll logic to dry up the code a bit.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
thegentlemanphysicist: done

frontend/src/components/dashboard/minesHomePage/Dashboard.js
thegentlemanphysicist: done

frontend/src/components/dashboard/minesHomePage/Dashboard.js
thegentlemanphysicist: Done

frontend/src/components/mine/MineHeader.js
thegentlemanphysicist: done

frontend/src/constants/routes.js
thegentlemanphysicist: done

frontend/src/components/dashboard/minesHomePage/Dashboard.js
thegentlemanphysicist: removed

PR #719, opened 2019-05-07
MineList row bugfix

FreshGyan:


frontend/src/components/dashboard/minesHomePage/MineList.js
```diff
@@ -181,22 +181,26 @@ const applySortIndicator = (_columns, field, dir) =>
...
-      props.mineTenureHash,
-      props.mineCommodityOptionsHash
-    )}
-    locale={{ emptyText: <NullScreen type="no-results" /> }}
-    onChange={handleTableChange(props.handleMineSearch)}
-  />
-);
+export const MineList = (props) => {
+  const tableColumns = applySortIndicator(columns, props.sortField, props.sortDir);
+  const tableData = transformRowData(
```
NWCalvank: I think these should have a comment asking it not to be refactored. Since this should behave exactly the same as the original code (but somehow doesn't), I see this being "cleaned up" in the future and a regression introduced

PR #723, opened 2019-05-08
Task/ Update Variance API sad paths & add test coverage

NWCalvank:
This is some general clean-up of the variances endpoints.

**Note:** All references to `variance_id` will need to be replaced with `variance_guid` once MDS-1664 is ready to be merged in.


python-backend/app/api/mines/variances/resources/variance_document_upload.py
```diff
@@ -51,7 +50,7 @@ def post(self, mine_guid, variance_guid):
         description='Associate an uploaded file with a variance.',
         params={
             'mine_guid': 'guid for the mine with which the variance is associated',
-            'variance_guid': 'ID for the variance to which the document should be associated'
+            'variance_guid': 'GUID for the variance to which the document should be associated'
         })
     @api.marshal_with(VARIANCE_MODEL, code=200)
     @requires_any_of([MINE_CREATE, MINESPACE_PROPONENT])
```
james-hollinger: Could we move the parser up out of the PUT method to the top of the class?

NWCalvank: As per the discussion around its behaviour across multiple API requests, I'm in favour of each method creating its own parser instance.

I can move it to the class if there's a good benefit proposition. At the moment, I only see a potential downside of having a stateful resource.

PR #728, opened 2019-05-09
Miscellaneous updates

FreshGyan:
- Minor driver upgrades
- More resources for ZAP so it runs in less time


FreshGyan:
@NWCalvank Mostly copy pasting. It just caught my eye and I realized there is no reason to pull the image over and over, especially since it takes relatively longer and our environments are already isolated

PR #730, opened 2019-05-13
Correct Race Condition in Functional Tests

NWCalvank:
Opening this PR for the DEV environment within which to test my changes


NWCalvank:
I broke the pipeline by restarting it to check for consecutive successful runs. Attempting to successfully clean things up now that all seems to be well.

PR #732, opened 2019-05-15
Feature / MDS-1678; MDS-1679; General Search

nayr974:
This PR implements search generally for Core across multiple entities.

The search is based on individual text searches across many single columns, using an ILIKE filter to find matches that contains the search term (%term%), and then ranks them using PostgreSQL' trigram module (pg_trgm) for similarity to the search string. That ranking is then modified using some simple scoring  multipliers in the API to score exact matches, strings that start with the search term, then aggregate scores for each result across its multiple fields.


frontend/src/components/common/LinkButton.js
```diff
@@ -0,0 +1,36 @@
...
+    style={props.style}
+  >
+    {props.children}
+  </a>
+);
+
+LinkButton.propTypes = propTypes;
+LinkButton.defaultProps = defaultProps;
+
+export default LinkButton;
```
mmumby: This is a great component addition!

python-backend/app/api/search/search/resources/search.py
```diff
@@ -0,0 +1,132 @@
...
+    @api.marshal_with(SEARCH_RESULT_RETURN_MODEL, 200)
+    def get(self):
+        search_results = []
+        app = current_app._get_current_object()
+
+        search_term = request.args.get('search_term', None, type=str)
+        search_types = request.args.get('search_types', None, type=str)
+        search_types = search_types.split(',') if search_types else search_targets.keys()
+
+        reg_exp = regex.compile(r'\'.*?\' | ".*?" | \S+ ', regex.VERBOSE)
```
Jsyro: comment or helpful name on what this regex is supposed to do.

python-backend/app/api/search/search/resources/search.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if getattr(item, value_field).lower().startswith(search_term.lower()):
+        score_multiplier = score_multiplier * 3
+
+    if getattr(item, value_field).lower() == search_term.lower():
+        score_multiplier = score_multiplier * 10
+
+    search_results.append(
+        SearchResult(
+            getattr(item, 'score') * score_multiplier, type, {
```
Jsyro: i don't believe there is any reason to use getattr unless you have defined a default value (3rd parameter), i think these should be item.value_field.

python-backend/app/api/search/search/resources/search.py
```diff
@@ -0,0 +1,132 @@
...
+                        func.similarity(column, term).label('score'),
+                        *type_config['entities']).filter(column.ilike(f'%{term}%'))
+                    if type_config['has_deleted_ind']:
+                        similarity = similarity.filter_by(deleted_ind=False)
+                    similarity = similarity.order_by(desc(func.similarity(column, term))).all()
+                    [
+                        append_result(search_results, search_term, type, item,
+                                      type_config['id_field'], type_config['value_field'],
+                                      type_config['score_multiplier']) for item in similarity
+                    ]
```
Jsyro: weird usage of list comprehension syntax, just loop with `for item in similarity`?

python-backend/app/api/search/search/resources/simple_search.py
```diff
@@ -0,0 +1,94 @@
...
+                        func.similarity(column, term).label('score'),
+                        *type_config['entities']).filter(column.ilike(f'%{term}%'))
+                    if type_config['has_deleted_ind']:
+                        similarity = similarity.filter_by(deleted_ind=False)
+                    similarity = similarity.order_by(desc(func.similarity(column, term))).all()
+                    [
+                        append_result(search_results, search_term, type, item,
+                                      type_config['id_field'], type_config['value_field'],
+                                      type_config['score_multiplier']) for item in similarity
+                    ]
```
Jsyro: same execute_search and append_result?

Jsyro: possibly move simple search, search, and search options into the same file, to share these methods.

python-backend/tests/search/resource/test_search_resource.py
```diff
@@ -0,0 +1,84 @@
...
+        f'/search?search_term={str(party.first_name)}', headers=auth_headers['full_auth_header'])
+    get_data = json.loads(get_resp.data.decode())
+    parties = get_data['search_results']['party']
+    assert len(parties) == 1
+    assert party.first_name in parties[0]['result']['name']
+    assert uuid.UUID(parties[0]['result']['party_guid']) == party.party_guid
+    assert len([
+        key for key, value in get_data['search_results'].items()
+        if key is not 'party' and len(value) is 0
+    ]) == 4
```
Jsyro: why 4? should this be dynamic? maybe a comment?

Jsyro:
Backend looks good, 3 small comments. Great work,

python-backend/app/api/search/search/resources/search.py
nayr974: Added

python-backend/app/api/search/search/resources/search.py
nayr974: Changed

python-backend/tests/search/resource/test_search_resource.py
nayr974: @JustinMacaulay thoughts?

python-backend/app/api/search/search/resources/simple_search.py
nayr974: Moved these to the helper class.

frontend/src/components/common/LinkButton.js
nayr974: I agree! It seemed justified to wrap up that block that was being used all over.

frontend/src/actionCreators/searchActionCreator.js
```diff
@@ -0,0 +1,62 @@
...
+      dispatch(searchActions.storeSearchResults(response.data));
+      dispatch(hideLoading("modal"));
+    })
+    .catch(() => dispatch(error(reducerTypes.GET_SEARCH_RESULTS)))
+    .finally(() => dispatch(hideLoading("modal")));
+};
+
+export const fetchSearchBarResults = (searchTerm) => (dispatch) => {
+  dispatch(request(reducerTypes.GET_SEARCH_BAR_RESULTS));
+  dispatch(showLoading("modal"));
```
mmumby: Does this loading bar make sense? (in any of these actionCreators) it is modal specific loading, and I don't think you are using models for the search.

frontend/src/components/search/ContactResultsTable.js
```diff
@@ -0,0 +1,121 @@
...
+        showHeader={false}
+        pagination={false}
+        columns={columns}
+        dataSource={props.searchResults}
+      />
+      {props.showAdvancedLookup && (
+        <Link
+          style={{ float: "right", fontSize: "1.25rem" }}
+          className="padding-large--left"
+          to={router.CONTACT_HOME_PAGE.dynamicRoute(parseQuery(props.query))}
```
mmumby: `router.CONTACT_HOME_PAGE.dynamicRoute` expects `page` and `per_page` params, I don't see these being set anywhere? Does this behave as expected?

frontend/src/components/search/ContactResultsTable.js
```diff
@@ -0,0 +1,121 @@
...
+  query: PropTypes.string.isRequired,
+  partyRelationshipTypeHash: PropTypes.objectOf(PropTypes.strings).isRequired,
+  showAdvancedLookup: PropTypes.bool.isRequired,
+};
+
+const defaultProps = {};
+
+const parseQuery = (query) => {
+  const emailExp = /@/;
+  const phoneExp = /^\d+(-\d*)*$/;
```
mmumby: We have email and phone RegEx's in `Validate.js`
Can we import those and use them in this component? ( I know we spoke in the past about making our email regex the barebones version, so the validate.js can be updated to use the above regEx for email)
- There are also existing checks in that file for phone and email.

frontend/src/components/search/MineResultsTable.js
```diff
@@ -0,0 +1,98 @@
...
+        showHeader={false}
+        pagination={false}
+        columns={columns}
+        dataSource={props.searchResults}
+      />
+      {props.showAdvancedLookup && (
+        <Link
+          style={{ float: "right", fontSize: "1.25rem" }}
+          className="padding-large--left"
+          to={router.MINE_HOME_PAGE.dynamicRoute({ search: props.query })}
```
mmumby: the same question regarding `page` and `per_page` with this route, does this work as expected?

frontend/src/components/search/PermitResultsTable.js
```diff
@@ -0,0 +1,74 @@
...
+    {
+      title: "Permit Guid",
+      dataIndex: "permit_guid",
+      key: "permit_guid",
+      render: (text, record) => [
+        <Row>
+          <Col span={24}>
+            <Link
+              to={router.MINE_SUMMARY.dynamicRoute(record.mine_guid, "permit")}
+              style={{ fontSize: "1.25rem" }}
```
mmumby: Are these links supposed to be larger than other links inside tables? 
Links should all be the same size, I fear we are going to run into more inconsistencies across components.

frontend/src/components/search/SearchBar.js
```diff
@@ -0,0 +1,138 @@
...
+        const newSearchTermHistory = prevState.searchTermHistory.slice(
+          Math.max(prevState.searchTermHistory.length - 2, 0)
+        );
+        newSearchTermHistory.push(searchTerm);
+        return {
+          isSelected: false,
+          searchTermHistory: newSearchTermHistory,
+        };
+      });
+      this.clearSearchBar(false);
```
mmumby: `this.clearSearchBar()` can be updated to accept a param, if the param isn't passed in it is implied to be false. 

```
clearSearchBar = (isSelected) => {
    this.props.clearSearchBarResults();
    this.setState({
      isSelected,
      searchTerm: "",
    });
  };
```

frontend/src/components/search/SearchResults.js
```diff
@@ -0,0 +1,287 @@
...
+                {`${
+                  type_filter ? this.props.searchOptionsHash[type_filter] : "Search results"
+                } for ${results}`}
+              </h1>
+              <div>
+                {type_filter ? (
+                  <LinkButton
+                    key="all"
+                    onClick={() => {
+                      this.props.history.push(`/search?q=${this.state.params.q}`);
```
mmumby: We don't use `this.props.history.push()` anywhere else in the app, 
for consistency, the route should have a dynamic route that we pass the param into

```
export const SEARCH_RESULTS = {
  route: "/search",
dynamicRoute: (q, t) => t ? `/search?q={q}&t={t}` : `/search?q={q}`
  component: SearchResults,
};
```

frontend/src/actionCreators/searchActionCreator.js
nayr974: I didn't realize that's how it worked. I'll change that 👍

frontend/src/components/search/PermitResultsTable.js
nayr974: I thought the wire-frames depicted them larger then the other text, although double checking now they don't. Comparing, I do think it's easier to scan the results with the names being larger but I'll change it for consistency.

frontend/src/components/search/ContactResultsTable.js
nayr974: Sure, that makes sense!

frontend/src/components/search/ContactResultsTable.js
nayr974: It does work, although I notice the page count looks wrong. The contact and mine page should probably be responsible for setting its default parameters if hit. I'll see if that's quick to do.

frontend/src/components/search/SearchResults.js
nayr974: this.props.history.push is used in a few places, although in conjunction with a dynamic route. I'll update it to be in line with the others.

frontend/src/actionCreators/searchActionCreator.js
nayr974: Changed!

frontend/src/components/search/ContactResultsTable.js
nayr974: Added page and per_page params!

frontend/src/components/search/ContactResultsTable.js
nayr974: Changed!

frontend/src/components/search/MineResultsTable.js
nayr974: Added page and per_page params :)

frontend/src/components/search/PermitResultsTable.js
nayr974: Changed

frontend/src/components/search/SearchResults.js
nayr974: Updated 👍

frontend/src/components/search/MineResultsTable.js
mmumby: Thanks, Ryan!

PR #737, opened 2019-05-16
Task/ Move OHSC & Union indicators from variances to mines

NWCalvank:
Backend:
- Move ohsc_ind and union_ind from variance to mine
- Remove redundant factory code

Frontend:
- Remove ohsc_ind and union_ind fields from form(s)
- Remove inaccurate required indicator on upload files in variance modal


NWCalvank:
Received feedback that we actually don't want the columns to be nullable. Setting this to WIP until that change is complete

python-backend/tests/factories.py
```diff
@@ -430,7 +428,6 @@ class Meta:
     email = factory.Faker('email')
     phone_no = factory.Faker('numerify', text='###-###-####')
     last_logon = TODAY
-    idir_user_detail = factory.SubFactory('tests.factories.IdirUserDetailFactory')
 
     @factory.post_generation
     def idir_user_detail(obj, create, extracted, **kwargs):
```
james-hollinger: idir_user_detail is one-to-one with core_user, this @factory.post_generation is unnecessary. idir_user_detail should be a RelatedFactory here, and there should be a corresponding core_user = factory.SubFactory on IdirUserDetailsFactory.

Currently, if you tried to call IdirUserDetailFactory(), it would fail because factory.SelfAttribute won't work like that.

NWCalvank: Oh, I thought you just said that `RelatedFactory` is unreleased, but I see it was `RelatedFactoryList`.

And yes, I originally wrote core_user = factory.SubFactory on IdirUserDetailsFactory, but it was causing problems when being called via the CoreUser factory. I can revisit it, though that's really not the purpose of this PR.

python-backend/tests/factories.py
```diff
@@ -430,20 +428,16 @@ class Meta:
...
-        IdirUserDetailFactory.create(core_user_id=obj.core_user_id, **kwargs)
+    idir_user_detail = factory.RelatedFactory('tests.factories.IdirUserDetailFactory', 'core_user')
 
 
 class IdirUserDetailFactory(BaseFactory):
     class Meta:
         model = IdirUserDetail
 
+    class Params:
+        core_user = factory.SubFactory('tests.factories.CoreUserFactory')
```
james-hollinger: You only need to reference factories like this if they are declared later in the file, as the binding happens when the file is loaded. CoreUserFactory is declared already, so this line can just be `core_user = factory.SubFactory(CoreUserFactory)`

james-hollinger: Not a big deal, however. This is more for a heads-up on slightly nicer looking style; it doesn't change the functionality at all.

PR #740, opened 2019-05-16
MDS-1764 Add dangerous occurrence determination to incident report

james-hollinger:
Adds a dropdown to select if an incident is a dangerous occurrence, and if so, requires the user to select one or more applicable sub-paragraphs of code section 1.7.3 (dangerous occurrences).

Also adds basic support for tooltips on the multi-select component. The component now accepts items in the form of {value: <>, label: <>, tooltip: <>}


frontend/src/components/Forms/AddIncidentForm.js
```diff
@@ -1,126 +1,181 @@
...
-        </Button>
-      </Popconfirm>
-      <Button className="full-mobile" type="primary" htmlType="submit" disabled={props.submitting}>
-        {props.title}
-      </Button>
-    </div>
-  </Form>
-);
+class AddIncidentForm extends Component {
+  constructor(props) {
```
NWCalvank: I don't believe this constructor method is required

frontend/src/components/Forms/AddIncidentForm.js
```diff
@@ -1,126 +1,181 @@
...
+                  id="dangerous_occurrence_subparagraph_ids"
+                  name="dangerous_occurrence_subparagraph_ids"
+                  label="Which sub-paragraph(s) of the code apply to this dangerous occurrence?*"
+                  placeholder="Please choose one or more"
+                  component={renderConfig.MULTI_SELECT}
+                  data={this.props.doSubparagraphOptions}
+                  validate={[this.validateDoSubparagraphs]}
+                />
+              </Form.Item>
+            ) : null}
```
NWCalvank: If this is only being rendered when `this.state.doDetermination === "DO"`, I would guess that the custom validator is not needed. I'd expect that `required` would work just fine?

frontend/src/tests/components/common/__snapshots__/RenderMultiSelect.spec.js.snap
```diff
@@ -68,8 +68,12 @@ ShallowWrapper {
           showSearch={false}
           transitionName="slide-up"
         >
-          <Option />
-          <Option />
+          <Option
+            title={undefined}
```
NWCalvank: Please either confirm that this is an optional field or update the test data to populate this value

migrations/sql/V2019.05.02.06.01__add_dangerous_occurrence_code_subsections.sql
```diff
@@ -0,0 +1,59 @@
...
+    ('HSRCM', '1', '7', '3', '(7)', 'Stoppage of ventilation', 'Unplanned stoppage of the main underground ventilation system', '1970-01-01', '9999-12-31', 'system-mds', 'system-mds'),
+    ('HSRCM', '1', '7', '3', '(8)', 'Out of control vehicle', 'A mine vehicle going out of control', '1970-01-01', '9999-12-31', 'system-mds', 'system-mds'),
+    ('HSRCM', '1', '7', '3', '(9)', 'Fire', 'Outbreak of fire if it endangers persons or threatens or damages equipment and all underground fires', '1970-01-01', '9999-12-31', 'system-mds', 'system-mds'),
+    ('HSRCM', '1', '7', '3', '(10)', 'Electrical equipment failure', 'Electrical equipment failure or incident that causes or threatens to cause injury to persons or damage to equipment or property', '1970-01-01', '9999-12-31', 'system-mds', 'system-mds'),
+    ('HSRCM', '1', '7', '3', '(11)', 'Any other unusual accident', 'Any other unusual accident or unexpected event which had the potential to result in serious injury', '1970-01-01', '9999-12-31', 'system-mds', 'system-mds')
```
NWCalvank: Oh man... I missed a 9999-12-31 column 😞

migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -403,6 +403,20 @@ VALUES
...
+    mine_incident_determination_type_code,
+    description                          ,
+    display_order                        ,
+    create_user                          ,
+    update_user
+)
+VALUES
+    ('PEN', 'Pending determination', 10, 'system-mds', 'system-mds'),
+    ('DO', 'This was a dangerous occurrence', 20, 'system-mds', 'system-mds'),
+    ('NDO', 'This was not a dangerous occurrence', 30, 'system-mds', 'system-mds')
```
NWCalvank: Curious on @klefler's thoughts here. I'd find it odd to have a mix of 2 and 3-character codes in a single table

python-backend/app/api/mines/incidents/resources/mine_incident_determination_types.py
```diff
@@ -0,0 +1,15 @@
...
+
+from app.extensions import api
+from app.api.utils.access_decorators import requires_role_mine_view
+
+from ..models.mine_incident_determination_type import MineIncidentDeterminationType
+from ...mine_api_models import MINE_INCIDENT_DETERMINATION_TYPE_MODEL
+
+
+class MineIncidentDeterminationTypeResource(Resource):
+    @api.marshal_with(MINE_INCIDENT_DETERMINATION_TYPE_MODEL, envelope='options', code=200, as_list=True)
```
NWCalvank: I know we have a number of endpoints that return code table dropdown options as `options`, but I believe we've agreed to move away from that pattern. New/refactored endpoints are using `records`, which can be standard across all list endpoints

migrations/sql/V2019.05.02.06.01__add_dangerous_occurrence_code_subsections.sql
Jsyro: I think this is db defaulted? so i don't think this script should define the effective and expiry dates. just let the defaults do their thing.

migrations/sql/V2019.05.02.06.01__add_dangerous_occurrence_code_subsections.sql
```diff
@@ -0,0 +1,59 @@
...
+    mine_incident_determination_type_code   character varying(3) PRIMARY KEY,
+    description                             character varying(100) NOT NULL,
+    display_order                           integer NOT NULL,
+    active_ind                              boolean NOT NULL DEFAULT 'true',
+
+    create_user                             character varying(60)                    NOT NULL,
+    create_timestamp                        timestamp with time zone DEFAULT now()   NOT NULL,
+    update_user                             character varying(60)                    NOT NULL,
+    update_timestamp                        timestamp with time zone DEFAULT now()   NOT NULL
+);
```
Jsyro: Out of curiosity, was this discussed as being implemented as a nullable boolean?

frontend/src/components/Forms/AddIncidentForm.js
james-hollinger: The value of the determination dropdown is put in the state and is used to conditionally render the sub-paragraph component. The value comes in through the initialValues. The state has to be set here in the constructor, else when the window opens and the determination is "DO", the state is empty and the component is not rendered.

python-backend/app/api/mines/compliance/models/compliance_article.py
```diff
@@ -13,8 +13,14 @@ class ComplianceArticle(AuditMixin, Base):
     paragraph = db.Column(db.String, nullable=True)
     sub_paragraph = db.Column(db.String, nullable=True)
     description = db.Column(db.String, nullable=False)
+    long_description = db.Column(db.String, nullable=False)
     effective_date = db.Column(db.DateTime, nullable=False, default=datetime.utcnow)
-    expiry_date = db.Column(db.DateTime, nullable=False, default=datetime.strptime('9999-12-31', '%Y-%m-%d'))
+    expiry_date = db.Column(
+        db.DateTime, nullable=False, default=datetime.strptime('9999-12-31', '%Y-%m-%d'))
```
Jsyro: i think this is the default that @NWCalvank is trying to hunt down. This is now a nullable column with no default right?

NWCalvank: It's not. I missed this table while trying to find all of the 9999-12-31 columns.

I updated: mine_location, permit_status_code, party_type_code, party, mineral_tenure_xref, mine_status, mine_region_code, and article_act_code. But I missed compliance_article, so the db default for it is still 9999-12-31

frontend/src/tests/components/common/__snapshots__/RenderMultiSelect.spec.js.snap
james-hollinger: This is an optional field

frontend/src/components/Forms/AddIncidentForm.js
james-hollinger: Good point. The only issue is the value from a multi-select is an array, so `required` wont work. `minLength` technically works but its expecting a string so the error message references characters specifically. So a custom validator is still needed but I can remove the "DO" conditional.

Of note there are currently no other multi-selects that require at least one option to be selected.

python-backend/app/api/mines/incidents/resources/mine_incident_determination_types.py
james-hollinger: Thanks, fixed 👍

migrations/sql/V2019.05.02.06.01__add_dangerous_occurrence_code_subsections.sql
```diff
@@ -0,0 +1,59 @@
+ALTER TABLE compliance_article ADD COLUMN long_description text;
```
klefler: Apologies for being late to the party... if there is still an opportunity, can you please change the data type from 'text' to 'varchar' and provide a max length?

migrations/sql/afterMigrate__1__seed_data.sql
klefler: Thanks @NWCalvank, I think I'm ok with this.

migrations/sql/V2019.05.02.06.01__add_dangerous_occurrence_code_subsections.sql
pdittaro: What max length will be required?

klefler: Depends on how much we think this column will contain... I just checked quickly, and the longest long_description value being added for this release is about 310 characters. We could set the max for 400 characters... but could set it at 500 for extra wiggle room.

james-hollinger: Column is now a varchar 500

PR #742, opened 2019-05-16
Fixes to docker and associated

pdittaro:
Replaced entrypoint with env-file attribute in compose.
Moved folder copy into container after dependency install for faster builds.


frontend/Dockerfile
```diff
@@ -5,10 +5,12 @@ RUN mkdir /app
 WORKDIR /app
 
 # Install app dependencies
-COPY . /app/
+COPY package.json package-lock.json ./
```
NWCalvank: This is typically abbreviated to package*.json

frontend/Dockerfile
```diff
@@ -5,10 +5,12 @@ RUN mkdir /app
...
 
 RUN npm ci
 
+COPY . .
```
NWCalvank: I would expect this to overwrite the node_modules, which were just created with `npm ci`. Is that not the case?

I think everyone will have a node_modules folder on the host because of the developer tooling (eslint, prettier, etc).

pdittaro: Yeah, `COPY` should respect the .dockerignore file - I haven't run into any issues with it recently. I think there may be an issue *if* the Dockerfile isn't in the same directory as the .dockerignore? 

This should substantially speed up building the docker images locally though. Might be good to compare.

frontend/Dockerfile
pdittaro: I mean in the sense that it would make more sense at a glance, rather than having to think through a wildcard.

pdittaro:
I guess a large caveat on this PR is that people will need to update their .env files after.

PR #744, opened 2019-05-21
nginx pod needs new vars.

Jsyro:


pipeline/config-prod.groovy
```diff
@@ -346,6 +347,10 @@ environments {
                     HOST = "http://mds-python-backend${vars.deployment.suffix}:5000"
                     PATH = "/api"
                 }
+                'mds-nris-backend' { 
+                    HOST = "http://mds-nris-backend${vars.deployment.suffix}:5500"
+                    PATH = "/nris_api"
```
pdittaro: This URL may not be the most desirable. We should think about how we want to name the paths of any microservices. For instance, its fine to just have `/api` like the other backend. I'm not sure of the convention here, but its nothing major.

PR #745, opened 2019-05-21
testing suite for nris api

Jsyro:
tests now run. other items will come from Paul in separate PR's


Jsyro:
Obsolete merged into PR 743.

PR #746, opened 2019-05-21
Feature/mds 1611

thegentlemanphysicist:
There is now a 'Status Since' field on the the front mine create and edit form.  
The value is displayed on the mine header.
Editing the status changes this date to the current date
The default is unknown
Migration needed to add a table to the backend


frontend/src/components/mine/MineHeader.js
```diff
@@ -117,6 +126,11 @@ export class MineHeader extends Component {
     this.setState({ menuVisible: false });
   };
 
+  handleClearStatusDate = () => {
+    const date = new Date();
+    this.props.change(FORM.MINE_RECORD, "status_date", date);
+  };
+
```
mmumby: This can occur inside the form component, inside the if/else statement

frontend/src/components/mine/MineHeader.js
```diff
@@ -328,24 +343,36 @@ export class MineHeader extends Component {
...
+              </div>
+
+              <div className="inline-flex padding-small">
+                <p className="field-title">Status Since </p>
+
+                {this.props.mine.mine_status[0].status_date ? (
+                  formatDate(this.props.mine.mine_status[0].status_date)
+                ) : (
+                  <p>Not Entered</p>
+                )}
```
mmumby: For consistency, this should be 'N/A'

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -104,6 +115,12 @@ export class MineRecordForm extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
```
mmumby: If you want to set the state because of changing props, use `componentWillReceiveProps` instead.

checking `this.props` against `nextProps`
There is already a pattern in place in this component that you can follow, inside the if/else you can call directly with the correct input`this.props.change(FORM.MINE_RECORD, "status_date", null)`

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -316,6 +333,21 @@ export class MineRecordForm extends Component {
...
         </Row>
+
+        <Row gutter={16}>
+          <Col>
+            <Form.Item>
+              <Field
+                id="status_date"
+                name="status_date"
+                label="Date of Status Change"
+                placeholder="Unknown"
```
mmumby: If this is a date field, the placeholder should be "yyyy-mm-dd"

frontend/src/components/mine/MineHeader.js
thegentlemanphysicist: I asked Aaron, this was the wording he chose.

frontend/src/components/Forms/MineRecordForm.js
thegentlemanphysicist: done

frontend/src/components/Forms/MineRecordForm.js
thegentlemanphysicist: Done

frontend/src/components/mine/MineHeader.js
thegentlemanphysicist: I have questions about implementation for this.  Can we chat when you get in?

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -27,8 +35,10 @@ const propTypes = {
   handleSubmit: PropTypes.func.isRequired,
   closeModal: PropTypes.func.isRequired,
   change: PropTypes.func.isRequired,
+  initialValues: PropTypes.objectOf(),
   handleDelete: PropTypes.func.isRequired,
```
mmumby: This should be throwing a warning in the console, `PropTypes.objectOf()` cannot be empty, you could create a customProp for the initialValues and use that instead.

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -410,10 +448,12 @@ export class MineRecordForm extends Component {
 
 MineRecordForm.propTypes = propTypes;
 MineRecordForm.defaultProps = defaultProps;
+const selector = formValueSelector(FORM.MINE_RECORD);
```
mmumby: We're now accessing form values in two different ways, either use `formValueSelector` or `getFormValues`, I think `formValueSelector` is actually better as it returns only the value we want and not the entire object.

frontend/src/components/Forms/MineRecordForm.js
thegentlemanphysicist: done

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -421,9 +427,18 @@ def _mine_operation_code_processor(mine_status, index):
         return None
 
 
-def _mine_status_processor(mine_status, mine):
+def _mine_status_processor(mine_status, status_date, mine):
+
     if not mine_status:
-        return mine.mine_status
+        existing_status_date = mine.mine_status[0].status_date if mine.mine_status else None
+        if status_date is existing_status_date:
```
Jsyro: I don't think this `is` reference will work as you intend. `is` evaluates that the references are the same, but `mine.mine_status[0].status_data` and `data.get('status_date')` will never be the same object.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -436,17 +451,19 @@ def _mine_status_processor(mine_status, mine):
...
 
         existing_status.expiry_date = datetime.today()
         existing_status.active_ind = False
         existing_status.save()
-
-    new_status = MineStatus(mine_status_xref_guid=mine_status_xref.mine_status_xref_guid)
+    if not status_date:
+        new_status = MineStatus(mine_status_xref_guid=mine_status_xref.mine_status_xref_guid)
+    else:
+        new_status = MineStatus(mine_status_xref_guid=mine_status_xref.mine_status_xref_guid, status_date=status_date)
```
Jsyro: not sure why this conditional is here. status_date is nullable.

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -436,17 +451,19 @@ def _mine_status_processor(mine_status, mine):
 
     existing_status = mine.mine_status[0] if mine.mine_status else None
     if existing_status:
-        if existing_status.mine_status_xref_guid == mine_status_xref.mine_status_xref_guid:
+        if existing_status.mine_status_xref_guid == mine_status_xref.mine_status_xref_guid \
+                and status_date == existing_status.status_date:
```
Jsyro: Just checking, a change to only the status_date will cause it to create a new row in the db? or should it update in place. We do want effective dating on this table, but i'm just wondering if this was intended

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: good catch, removed

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: fixed

python-backend/app/api/mines/mine/resources/mine.py
Jsyro: Discussed offline that this was intended.

PR #748, opened 2019-05-22
Add configurable database users

FreshGyan:
- Added functionality to define environment variable templated .sql files for migrations.
- Apparently 100 gets executed before 1 so renamed files so that permission assignment happens at the end at 99
- Reworked the postgres deployment template to follow variable conventions like the rest of our deployment templates.

I debated making this as a post hook to the database but it didn't make sense to redeploy your database anytime you want to add a new user. So decided to keep schema level changes in the migration flow.

Suggested flow:
- Any files that need environment variable substitution end in .tmpl
- A bash script runs and converts the .tmpl files into regular files while substituting the values
- Runs migrations on normal and test database
- After every migration re-assign permissions to capture new tables and such


migrations/sql/afterMigrate__99__assign_permissions.sql
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+Allow logstash user to READ anything on the mds schema
+**/
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO logstash;
+
+/**
+Allow metabase user to READ anything on the mds schema
+**/
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO metabase;
```
Jsyro: i think @itchison wants metabase to be able to cut across the schemas, so it will need to have access the NRIS schema as well.

openshift/postgresql.dc.json
```diff
@@ -359,7 +252,7 @@
                     "valueFrom": {
                       "secretKeyRef": {
                         "key": "oracle-fdw-name",
-                        "name": "${DATABASE_SERVICE_NAME}"
+                        "name": "${FDW_SECRET_NAME}"
```
pdittaro: Since this variable is `FDW_NAME`, having the secret key reference be named `FDW_SECRET_NAME` might cause some confusion (in that it references the name of the secret, not the FDW name). As `FDW_SECRET_NAME` contains other values like the schema name and password, suggest that it be renamed to `FDW_SECRET`

openshift/_python36.dc.json
```diff
@@ -210,7 +215,7 @@
                 "command": [
                   "bash",
                   "-c",
-                  "cd $FLYWAY_HOME && ./flyway migrate && ./flyway -url=jdbc:postgresql://$FLYWAY_DB_HOST/mds_test migrate"
+                  "cd $FLYWAY_HOME/scripts && ./run_migrations.sh"
```
pdittaro: Not required but since you're only running one command now you could maybe simplify the command to just `$FLYWAY_HOME/run_migrations.sh`. If you need to make reference to a script location within a script you can use `SCRIPTDIR="$(dirname "$0")"`.

openshift/postgresql.dc.json
FreshGyan: Oh I see the confusion. But the FDW name is actually coming in from the secret.

So we go to the secret and fetch the name from in there and pass it to the env variable FDW_NAME that goes to the container. 
We're not passing the FDW_NAME in the template, but fetching it from the secret

openshift/_python36.dc.json
FreshGyan: TIL 🙌

migrations/scripts/run_migrations.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+set -exv -o pipefail
+
+# Change directory to Flyway home
+cd $FLYWAY_HOME
+
+# Substitute environment variables in template files
+export DOLAR='$'
```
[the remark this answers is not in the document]
FreshGyan: removed! I was testing something and forgot twas there

migrations/sql/afterMigrate__99__assign_permissions.sql
FreshGyan: 💯

PR #749, opened 2019-05-23
Feature/mds 1918

mmumby:
- Added a `isViewOnly` param to the `ModalWrapper` to disable the popConfirm when closing the modal
- Added ability to remove documents from variances


frontend/src/actionCreators/varianceActionCreator.js
```diff
@@ -60,6 +60,20 @@ export const fetchVariancesByMine = ({ mineGuid }) => (dispatch) => {
...
 
+export const fetchVariancesById = (mineGuid, varianceGuid) => (dispatch) => {
+  dispatch(request(reducerTypes.GET_VARIANCE));
+  dispatch(showLoading("modal"));
+  return CustomAxios(Strings.ERROR)
+    .get(ENVIRONMENT.apiUrl + API.VARIANCE(mineGuid, varianceGuid), createRequestHeader())
+    .then((response) => {
+      dispatch(success(reducerTypes.GET_VARIANCE));
+      dispatch(varianceActions.storeVariance(response.data));
+      dispatch(hideLoading("modal"));
```
thegentlemanphysicist: do we need to dispatch hide loading modal here if the 'finally' block does it?

frontend/src/components/common/wrappers/ModalWrapper.js
```diff
@@ -43,13 +45,6 @@ export class ModalWrapper extends Component {
     this.props.closeModal();
   };
 
-  renderTitle = () => (
```
thegentlemanphysicist: Does this modal no longer have a title?

frontend/src/components/modalContent/EditVarianceModal.js
```diff
@@ -1,32 +1,95 @@
...
+  variance: {},
+};
+
+export class EditVarianceModal extends Component {
+  state = { isLoaded: false };
+
+  componentDidMount() {
+    this.props
+      .fetchVariancesById(this.props.mineGuid, this.props.varianceGuid)
+      .then(this.setState({ isLoaded: true }));
```
thegentlemanphysicist: do we need to put catch statements into these chains

frontend/src/components/modalContent/EditVarianceModal.js
```diff
@@ -1,32 +1,95 @@
...
+
+  // handling delete functionality inside the modal, so the data can be updated properly.
+  handleRemoveDocument = (event, documentGuid) => {
+    event.preventDefault();
+    this.props
+      .removeDocumentFromVariance(this.props.mineGuid, this.props.varianceGuid, documentGuid)
+      .then(() => {
+        this.props
+          .fetchVariancesById(this.props.mineGuid, this.props.varianceGuid)
+          .then(this.setState({ isLoaded: true }));
```
thegentlemanphysicist: same as above, catching errors?

frontend/src/reducers/varianceReducer.js
```diff
@@ -12,11 +13,17 @@ const varianceReducer = (state = initialState, action) => {
...
+        ...state,
+        variance: action.payload,
+      };
     default:
       return state;
   }
 };
 
 export const getMineVariances = (state) => state[VARIANCES].mineVariances;
+export const getVariance = (state) => state[VARIANCES].variance;
```
thegentlemanphysicist: Purely for my own education:  Why is VARIANCES in all caps?   I s the re a reason for that convention?

frontend/src/components/common/wrappers/ModalWrapper.js
mmumby: This block of code wasn't being used, I think it was there originally but wasn't removed when the component was refactored.

frontend/src/actionCreators/varianceActionCreator.js
mmumby: good catch

frontend/src/reducers/varianceReducer.js
mmumby: Redux naming conventions, reducerTypes, and actionTypes are `SCREAMING_SNAKE_CASE`

frontend/src/components/modalContent/EditVarianceModal.js
mmumby: no, we're handling the error in the `actionCreator`, I will only set the state if successful, but in saying that, I should add a loading view to the modal to handle this better

mmumby: ended up removing the `.then()` it wasn't doing much without needing a new loading view (which should be implemented site wide anyways)

frontend/src/actionCreators/varianceActionCreator.js
```diff
@@ -54,14 +52,26 @@ export const fetchVariancesByMine = ({ mineGuid }) => (dispatch) => {
     .then((response) => {
       dispatch(success(reducerTypes.GET_MINE_VARIANCES));
       dispatch(varianceActions.storeVariances(response.data));
-      dispatch(hideLoading());
     })
     .catch(() => dispatch(error(reducerTypes.GET_MINE_VARIANCES)))
     .finally(() => dispatch(hideLoading()));
 };
 
+export const fetchVariancesById = (mineGuid, varianceGuid) => (dispatch) => {
```
Jsyro: `fetchVarianceById`? single? or does it do something else.

frontend/src/reducers/varianceReducer.js
```diff
@@ -12,11 +13,17 @@ const varianceReducer = (state = initialState, action) => {
         ...state,
         mineVariances: action.payload.records,
       };
+    case actionTypes.STORE_VARIANCE:
+      return {
+        ...state,
+        variance: action.payload,
```
Jsyro: Just curious if this the way to do this? I'm guessing this is the super strict redux pattern? but wouldn't it make more sense to add it to some collection then have a selector that is able to pull out a specific variance? Cause if a separate component called 'getVarianceById` it will overwrite this value? seems like a lots of work to set up a place in the store for a single item (likely from a set), and we might want to load multiple specific ones at some point. 

Tailings Expected documents has a similar workflow, but it reloads from the whole mine.

openshift/scripts/database-dump.sh
```diff
@@ -76,7 +76,7 @@ SAVE_PATH=$( dirname ${SAVE_TO} )
 mkdir -p ${SAVE_PATH}
 oc exec ${POD_DB} -n ${PROJECT} -- /bin/bash -c \
 	'pg_dump -U ${POSTGRESQL_USER} -d ${POSTGRESQL_DATABASE} -Fc \
-	--column-inserts --data-only -T mms.* -T public.etl* -T public.flyway* > /tmp/'${SAVE_FILE}
+	--column-inserts --data-only -T mms.* -T public.etl* -T nris.* -T public.flyway* > /tmp/'${SAVE_FILE}
```
Jsyro: Thanks for adding this, it solves a permission issue and we don't want to back it up anyhow.

frontend/src/actionCreators/varianceActionCreator.js
mmumby: Correct, it is fetching a single variance.

aaand good catch! will remove the `s`

frontend/src/reducers/varianceReducer.js
mmumby: There is a `mineVariance` array stored in the state as well, which houses all the variances attached to a mine.

Previously I was able to access the individual variances in the array without needing `fetchVarianceById` by mapping through the array and passing in the entire variance into the modal when it was opened.
That works great until you want to update the variance inside the modal without closing the modal.

Does that make sense? I can't think of a situation where the user would be updating the store in a different part of the application that would interfere with this functionality.

PR #751, opened 2019-05-24
Feature/mds 1841

mmumby:
- Added "my dashboard" view to Test
- updated FE subscription feature, and exposed in Test
- Update Nav styling - removed icons, changed the color of dropdown


frontend/src/components/dashboard/customHomePage/MineSubscriptionTable.js
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * @class MineSubscriptionTable is a user specific table of mines they have subscribed to with the ability to unsubscribe
+ *
+ */
+
+const propTypes = {
+  handleUnSubscribe: PropTypes.func.isRequired,
+  subscribedMines: PropTypes.arrayOf(CustomPropTypes.mine).isRequired,
+  mineRegionHash: objectOf(string).isRequired,
```
nayr974: If we're going to start importing specific named PropTypes functions to avoid prefixing  PropTypes.objectOf, can we please make it consistent and import func, arrayOf, etc as well? I'd prefer we keep the PropTypes prefix given how widely we have that in our code-base, or refactor it everywhere as a seperate PR.

frontend/src/components/dashboard/customHomePage/MineSubscriptionTable.js
```diff
@@ -0,0 +1,177 @@
...
+      commodityHash: mineCommodityHash,
+      tenure: mine.mine_type.length ? mine.mine_type : null,
+      tenureHash: mineTenureHash,
+      tsf: mine.mine_tailings_storage_facility
+        ? mine.mine_tailings_storage_facility.length
+        : Strings.EMPTY_FIELD,
+    }));
+
+  render() {
+    const columns = [
```
nayr974: At a glance this looks very close to the table definition on the mine lookup. I'm wondering if they could be shared, and the extra column for subscribe/unsubscribe appended to the columns. You might want a subscribe column on the Mine Lookup anyways. Not a request, just a thought.

frontend/src/actionCreators/mineActionCreator.js
```diff
@@ -338,35 +338,33 @@ export const setMineVerifiedStatus = (mine_guid, payload) => (dispatch) => {
...
   return CustomAxios()
     .delete(ENVIRONMENT.apiUrl + API.SUBSCRIPTION(mineGuid), createRequestHeader())
     .then(() => {
       notification.success({
-        message: "Successfully unsubscribed",
+        message: `Successfully unsubscribed ${mineName}`,
         duration: 10,
       });
       dispatch(success(reducerTypes.UNSUBSCRIBE));
-      dispatch(hideLoading());
```
thegentlemanphysicist: bit by bit

frontend/src/components/navigation/NavBar.js
```diff
@@ -52,21 +41,16 @@ const defaultProps = {
...
-    </Menu>
-  );
-
   componentDidMount() {
     this.props.fetchMineVerifiedStatuses(`idir\\${this.props.userInfo.preferred_username}`);
   }
 
+  handleLogout = () => {
+    this.props.keycloak.logout();
+    localStorage.removeItem("jwt");
```
thegentlemanphysicist: I didn't know we could do this

frontend/src/components/dashboard/customHomePage/MineSubscriptionTable.js
mmumby: Agreed, overlooked that. I also prefer to keep the PropTypes prefix. Will update!

frontend/src/components/dashboard/customHomePage/MineSubscriptionTable.js
mmumby: They are very similar, the main difference being `MineList` expects the array of mineIds and then maps through that to access the mine via `mine[id]`

In this situation, I am passing in an array of subscribedMines and mapping through that. 
Super trivial, but requires a bit of a refactor. 
I'll chat with Aaron/ designers to see if they want the ability to subscribe/unsub on the main mine list, and refactor at that time. 

I think that would be a good enhancement!

PR #752, opened 2019-05-24
Infrastructure / NRIS Microservice and ETL

nayr974:
This PR creates a NRIS service to expose NRIS data as a RESTful service, and a scheduled ETL job.
- The ETL job connects directly to the NRIS Oracle database, then processes and inserts this data into a new relational structure on a NRIS schema of our MDS database.
- The NRIS service provides a few basic endpoints and some parameters on which to filter NRIS data.


microservices/nris_api/backend/app/__init__.py
```diff
@@ -51,3 +56,10 @@ def register_extensions(app):
     Compress(app)
 
     return None
+
+
+def register_scheduled_jobs(app):
+    if app.config['ENVIRONMENT_NAME'] in ['test', 'prod']:
```
FreshGyan: I remember SonarQube complaining about these two nested Ifs as a code smell, maybe combine the two ifs?

microservices/nris_api/backend/app/commands.py
```diff
@@ -1,15 +1,29 @@
...
+    @app.cli.command()
+    def _clean_nris_raw_data():
+        clean_nris_xml_import()
+
+    @app.cli.command()
+    def _run_nris_etl():
+        etl_nris_data()
+
+    @sched.app.cli.command()
+    def run_nightly_nris_etl():
```
FreshGyan: Is `run_nightly_nris_etl` identical to `_run_nris_etl ` and exists just to test scheduler?

microservices/nris_api/backend/app/commands.py
```diff
@@ -1,15 +1,29 @@
...
+from app.nris.models.nris_raw_data import NRISRawData
+from app.nris.etl.nris_etl import import_nris_xml, etl_nris_data, clean_nris_data, clean_nris_xml_import
+from app.nris.scheduled_jobs import nris_jobs
 
 
 def register_commands(app):
     @app.cli.command()
-    def _test_oracle_db():
-        cursor = oracle_db.cursor()
+    def _get_nris_raw_data():
```
FreshGyan: Do these actually give you data on the console somewhere? The naming suggests im getting the data as output, but im assuming im really just importing the data into the database right?

microservices/nris_api/backend/app/nris/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+from os.path import dirname, basename, isfile, join
+import glob
+modules = glob.glob(join(dirname(__file__), "*.py"))
+__all__ = [basename(f)[:-3] for f in modules if isfile(f) and not f.endswith('__init__.py')]
```
FreshGyan: what this do?

microservices/nris_api/backend/app/nris/resources/document.py
```diff
@@ -0,0 +1,34 @@
...
+from flask_restplus import Resource
+
+from werkzeug.exceptions import BadRequest, NotFound, InternalServerError
+
+from app.extensions import api, db
+from app.nris.utils.access_decorators import requires_role_nris_view
+
+from app.nris.models.document import Document as Model, DOCUMENT_RESPONSE_MODEL as RESPONSE_MODEL
+
+module_path = 'documents'
```
FreshGyan: I like the new use of module_path 💯 Saves the init file overhead and repetition on routes

microservices/nris_api/backend/app/nris/resources/document.py
```diff
@@ -0,0 +1,34 @@
...
+filter_fields = ['document_type', 'file_name']
```
FreshGyan: Any reason this isn't scoped to the class and is global?

microservices/nris_api/backend/migrations/env.py
```diff
@@ -12,14 +13,14 @@
...
+schema_name = config.get_main_option('schema_name')
 
 # add your model's MetaData object here
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-from flask import current_app
 config.set_main_option('sqlalchemy.url',
-                       current_app.config.get('SQLALCHEMY_DATABASE_URI'))
+                       current_app.config['SQLALCHEMY_DATABASE_URI'])
```
FreshGyan: I thought we were using .get to avoid exception throwing. Is that not a pattern being followed anymore?

microservices/nris_api/backend/migrations/env.py
```diff
@@ -70,17 +71,26 @@ def process_revision_directives(context, revision, directives):
...
     try:
         with context.begin_transaction():
             context.run_migrations()
     finally:
         connection.close()
 
+
+def include_object(object, name, type_, reflected, compare_to):
+    # Prevent alembic suicide
+    return name != "alembic_version"
```
FreshGyan: What's this for 😂

microservices/nris_api/backend/app/nris/models/__init__.py
nayr974: :) I had actually just added comments for that. It scans for and imports all models, versus creating and maintaining hard references to every model.

microservices/nris_api/backend/app/commands.py
nayr974: It imports the data into the database. Get to me doesn't suggest console output, but I prefer import either way so I'll update the name.

microservices/nris_api/backend/app/commands.py
nayr974: One exists for the scheduler, one exists for manual execution or testing.

JustinMacaulay: The above first clears the data and re pulls it, if that succeeds then it truncates the old data and runs the ETL. The _run_nris_etl just runs the ETL it doesn't truncate the tables.

microservices/nris_api/backend/migrations/env.py
nayr974: alembic_version is the versioning table alembic uses to track which migrations have been executed. However, it tries to delete it next migration because there is no matching model.

microservices/nris_api/backend/app/nris/resources/document.py
nayr974: Primarily to group "configuration", although I could argue that even though today only the List resources use the filter_fields, both resources should be able to use them.

microservices/nris_api/backend/app/commands.py
pdittaro: I agree with Gyan on this one - especially if you're having it as a cli command. Import is a little more clear I think.

microservices/nris_api/backend/app/commands.py
pdittaro: ```run_nightly_nris_etl``` isn't just for testing, I would say. It lets us use openshift cron to spin up an ETL container with a custom command, rather than having to have our service running all the time.

microservices/nris_api/backend/app/commands.py
nayr974: Renamed.

microservices/nris_api/backend/app/nris/resources/document.py
nayr974: @Jsyro 🥇

microservices/nris_api/backend/migrations/env.py
pdittaro: I was having issues with the config and changed some settings around. I'll see if this still works with the .get. Having an object named config on the App object and a file named config in the app module is kind of confusing... Maybe we can rename the config file to settings?

microservices/nris_api/backend/app/nris/scheduled_jobs/nris_jobs.py
```diff
@@ -0,0 +1,48 @@
...
+from flask import current_app
+from random import randint
+from time import sleep
+import sys
+import cx_Oracle
+
+# the schedule of these jobs is set using server time (UTC)
+
+
+def _schedule_NRIS_ETL_jobs(app):
```
FreshGyan: I raised this during the implementation in core api, and just raising it again that we can replace this whole method and registering it in init file by using decorators provided by apscheduler
https://github.com/viniciuschiele/flask-apscheduler/blob/master/examples/decorated.py

microservices/nris_api/backend/migrations/env.py
Jsyro: the .get pattern was about parsing/receiving form data, and in this case. we definitely want the application to crash if `SQLALCHEMY_DATABASE_URL` isn't set in the app config.

microservices/nris_api/backend/migrations/env.py
pdittaro: We were working around being stuck with using the same schema name as the username with SqlAlchemy. Alembic and SqlAlchemy weren't playing well with each other when we just used the 'table_args' parameter to specify the schema in the BaseModel object (Alembic was recreating FKs on every migrate). 

This line was part of code on the Alembic side to work with explicit and/or multiple schemas specified in a config file. 

We weren't able to finish it off from the SqlAlchemy side however. It might be worth keeping this around if we want to deal with multiple schemas, but I think we'll have to think through it a little further.

microservices/nris_api/backend/app/__init__.py
nayr974: Combined

PR #754, opened 2019-05-26
Add fields to match MDS API to MWSL API

FreshGyan:
This PR adds the following new data to our API responses:
- UTM converted fields for the lat/long
- site_access_description which has been ETL'ed from MMS
- Status description of the permits.

The above things were asked by NRIS to add to our API so that they can switch over from using MWSL.

Some other miscellaneous unused imports removed


python-backend/app/api/permits/permit/models/permit.py
```diff
@@ -1,8 +1,5 @@
-from datetime import datetime
```
thegentlemanphysicist: Why is date time no longer being used?

Makefile
```diff
@@ -96,7 +96,7 @@ cache:
 
 database-build:
 	@echo "+\n++ Performing postgres build ...\n+"
-	@docker-compose build postgres
+	@docker-compose build postgres flyway
```
thegentlemanphysicist: What is this doing?

migrations/sql/R__stored_procedure_for_mine_information.sql
```diff
@@ -352,11 +366,20 @@ CREATE OR REPLACE FUNCTION transfer_mine_information() RETURNS void AS $$
             INSERT INTO ETL_LOCATION(
                 mine_guid           ,
                 mine_no             ,
+                mine_location_description,
                 latitude            ,
                 longitude           )
             SELECT
                 new.mine_guid,
                 new.mine_no  ,
+                COALESCE(
```
thegentlemanphysicist: What does 'Coalesce' do?

Makefile
FreshGyan: Makes sure our flyway docker images are rebuilt everytime you rebuild the database so you can get the latest migrations.
Earlier we used to volume mount but i made some change a while back that broke that behaviour :/

migrations/sql/R__stored_procedure_for_mine_information.sql
FreshGyan: Basically a wrapper that says, execute this SQL and if it returns NULL then do something else, if you dont do anything else, imma just return NULL.
Since in my case I want it to return NULL if the query doesnt resolve, I left it at that.
http://www.postgresqltutorial.com/postgresql-coalesce/

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -414,7 +414,9 @@ def get(self):
                     'mine_name': x.mine_name,
                     'mine_no': x.mine_no,
                     'latitude': str(x.mine_location.latitude) if x.mine_location else '',
-                    'longitude': str(x.mine_location.longitude) if x.mine_location else '', },
+                    'longitude': str(x.mine_location.longitude) if x.mine_location else '',
+                    'mine_location_description': x.mine_location.mine_location_description,
+                    },
```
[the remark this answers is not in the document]
FreshGyan: Yes, yes there should be done 😅

python-backend/app/api/permits/permit/models/permit.py
FreshGyan: I'm not sure when the change was made, but it was just a unused import at this point so i just did cleanup.
Functionally im not sure when/why the usage for it was removed

migrations/sql/R__stored_procedure_for_mine_information.sql
thegentlemanphysicist: thanks for the link

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -414,7 +414,9 @@ def get(self):
                     'mine_name': x.mine_name,
                     'mine_no': x.mine_no,
                     'latitude': str(x.mine_location.latitude) if x.mine_location else '',
-                    'longitude': str(x.mine_location.longitude) if x.mine_location else '', },
+                    'longitude': str(x.mine_location.longitude) if x.mine_location else '',
+                    'mine_location_description': x.mine_location.mine_location_description if x.mine_location else '',
+                    },
```
NWCalvank: This resource should have been updated to use the new marshalling, which removes the need for these type coercions + null checks. I'm not sure why it wasn't updated with the other resources in this file. I think it needs to be prioritized ASAP unless it was intentionally skipped. @JustinMacaulay ?

I'm concerned that we're returning an empty string for lat/long instead of `None`/`null`. That will make switching to marshalling either unnecessarily difficult, a breaking API change (switch to returning `null`), or a db update to make lat/lon default to an empty string (which seems like the worst option to me).

migrations/sql/R__stored_procedure_for_mine_information.sql
```diff
@@ -199,6 +200,7 @@ CREATE OR REPLACE FUNCTION transfer_mine_information() RETURNS void AS $$
             CREATE TEMP TABLE IF NOT EXISTS pmt_now (
                 lat_dec   numeric(11,7),
                 lon_dec   numeric(11,7),
+                mine_location_description    varchar,
```
NWCalvank: Character count on `varchar`?

FreshGyan: the description is a free range text input where people put notes on how to get to a mine. I didnt see any business requirement on length so I left it open since varchar will only take as much space as needed and nothing more

NWCalvank: Works for me. Marking as resolved

PR #759, opened 2019-05-29
Trim and Upgrade NRIS requirements.txt

pdittaro:
Removed unnecessary and non-top-level requirements from requirements.txt. Upgraded remaining requirements to latest.


pdittaro:
Confirmed ETL works and tests pass.

microservices/nris_api/backend/app/commands.py
```diff
@@ -2,26 +2,34 @@
```
FreshGyan: I see that you removed click as dependancy but are importing click on top of this file. Is python not throwing an error on that?

pdittaro: Click is included with Flask now as a nested dependency. I tried to only include top level

pdittaro:
Thanks @FreshGyan 
- For the command line prints: They're really only interactive messages to be shown on the console when the command line is being used - not sent to the logger. I agree though, not sure how its handled in all cases, might be good to be explicit and wrap our `print()`s in something to make sure its getting to sysout only when we want it.
- Removed the click import and the commands still work. Good point about explicit imports having named dependencies in requirements.

PR #761, opened 2019-05-29
NRIS scheduled job lock and sequence reset

pdittaro:
The scheduling half of MDS-1969.

Using a lock on redis in order to prevent more than one pod from doing the ETL job. Implementation mirrors Core. Also resetting the sequence on the NRIS Raw Data table.


pdittaro:
Changed register_apm decorator to have optional parameter 'name'. Decorate using `@register_apm(name="")` or `@register_apm()` rather than `@register_apm

microservices/nris_api/backend/app/nris/scheduled_jobs/nris_jobs.py
```diff
@@ -1,25 +1,35 @@
...
+    if job_running is None:
+        cache.set(NRIS_JOB_PREFIX + NRIS_ETL_JOB,
+                  'True', timeout=TIMEOUT_12_HOURS)
+        _run_nris_etl()
+        cache.delete(NRIS_JOB_PREFIX + NRIS_ETL_JOB)
+
+
+@register_apm("NRIS ETL Job")
+def _run_nris_etl():
+    print("hello")
```
Jsyro: debug statement should be removed.

microservices/nris_api/backend/app/nris/scheduled_jobs/nris_jobs.py
```diff
@@ -31,19 +41,13 @@ def run_nightly_nris_etl():
             sched.app.logger.info('XML Import completed')
             # TODO: Insert update into status table
 
+            clean_nris_etl_data()
+            etl_nris_data()
+            sched.app.logger.info('NRIS ETL Completed!')
```
Jsyro: i believe this will crash if we try to run this job via flask cli.. maybe add some safety around this? I tried to abstract it with a get_logger util, maybe use that?

pdittaro: I don't think it should since the command explicitly runs with the sched.app context?

pdittaro: You're right thought that we do need to abstract the logger out of the app context, lets create an item for that.

JustinMacaulay: This should be fine. I agree though, having the logger abstracted would be nice.

microservices/nris_api/backend/app/nris/scheduled_jobs/nris_jobs.py
pdittaro: Removed.

PR #762, opened 2019-05-29
Remove un-necessary request logging

FreshGyan:
Just trying to reduce the noise in nginx logs to stop kibana from being flooded with useless health checks logs.

Also reverting the uwsgi error logs back to be thrown at stdout and disabling request logs because they're un-necessary


pdittaro:
Are you disabling all nginx logging? I agree that we have to trim down the health checks but we don't want to lose everything. Aren't we getting IP address logging from nginx?

Still, agreed that we want to clean up the APM quite a bit.

FreshGyan:
@pdittaro Only disabling logging for the `/nginx_status` endpoint. That endpoint is only accessible by localhost and the pathfinder team has some tool (sysdig i think?) that keeps trying to access that endpoint and since its not localhost nginx throws an error. 

That endpoint is there just so we can curl localhost/nginx_status from the pod and see if its getting overwhelmed or not.

All the other endpoints should work exactly the same

PR #763, opened 2019-05-30
use our date format, not iso format.

Jsyro:
NRIS_API date formats should be our format, as ISO format is not easy to parse.


Jsyro:
Don't think we are doing this.

PR #764, opened 2019-05-31
Feature/user nris api in core

Jsyro:
moved compliance tab to use new endpoint which leverages the new nris_api. Once smoke checked i will delete the old code in this PR.


python-backend/app/api/services/NRIS_API_service.py
```diff
@@ -0,0 +1,160 @@
...
+import uuid
+import requests
+import json
+import functools
+
+from dateutil.relativedelta import relativedelta
+from datetime import datetime, timedelta
+from flask import request, current_app
+from flask_restplus import Resource, reqparse
+from werkzeug.datastructures import FileStorage
```
pdittaro: The reliance on the werkzeug datastructures might be an issue. I can't load uwsgi and werkzeug onto my Windows dev environment. It would be nice to limit dependencies on the hosting platform.

pdittaro: Are the werkzeug dependencies even being used in this service?

Jsyro: removed, and other unused imports.

python-backend/app/api/mines/compliance/resources/compliance.py
```diff
@@ -1,34 +1,95 @@
...
+        'last_inspection':
+        fields.DateTime,
+        'last_inspector':
+        fields.String,
+        'num_open_orders':
+        fields.Integer,
+        'num_overdue_orders':
+        fields.Integer,
+        'all_time':
+        fields.Nested({
```
JustinMacaulay: Why not define a model for this instead of copy pasting the same thing?

python-backend/app/api/utils/apm.py
```diff
@@ -4,25 +4,36 @@
...
-            client = Client(current_app.config['ELASTIC_APM'])
+            config = current_app.config['ELASTIC_APM']
         elif sched.app:
-            client = Client(sched.app.app_context().app.config['ELASTIC_APM'])
+            config = sched.app.app_context().app.config['ELASTIC_APM']
 
-        #ensure client was created properly
-        if client:
+        client = Client(config)
+        if client and config.get('SECRET_TOKEN'):
```
pdittaro: We've already got a mechanism in place for this :)

microservices/nris_api/backend/app/nris/scheduled_jobs/nris_jobs.py
```diff
@@ -11,19 +11,21 @@
...
     job_running = cache.get(NRIS_JOB_PREFIX + NRIS_ETL_JOB)
     if job_running is None:
-        cache.set(NRIS_JOB_PREFIX + NRIS_ETL_JOB,
-                  'True', timeout=TIMEOUT_12_HOURS)
-        _run_nris_etl()
-        cache.delete(NRIS_JOB_PREFIX + NRIS_ETL_JOB)
+        try:
+            cache.set(NRIS_JOB_PREFIX + NRIS_ETL_JOB, 'True', timeout=TIMEOUT_12_HOURS)
+            _run_nris_etl()
+        finally:
```
pdittaro: Using a finally is probably the right approach here, really caused us some heartache with debugging otherwise.

PR #765, opened 2019-06-03
Feature/mds 1660

mmumby:
- added ability to add a variance application
- edit variance applications by adding or removing documents
- viewing an approved variance

Note: This feature is still hidden from users.


frontend-public/src/components/Forms/variances/AddVarianceForm.js
```diff
@@ -0,0 +1,111 @@
...
+      documentNameGuidMap: {
+        [document_manager_guid]: documentName,
+        ...documentNameGuidMap,
+      },
+    }));
+    change("uploadedFiles", this.state.uploadedFiles);
+  };
+
+  onRemoveFile = (fileItem) => {
+    remove(this.state.uploadedFiles, { document_manager_guid: fileItem.serverId });
```
NWCalvank: I know there's a bug with removing files, so I pulled this up in the lodash docs, and this isn't how `remove` works: https://lodash.com/docs/#remove

It accepts a predicate function as the second argument. Since it sounded like `onRemoveFile` is never invoked anyway, that likely doesn't fix anything, but it might save you some future debugging time.

frontend-public/src/components/Forms/variances/AddVarianceForm.js
```diff
@@ -0,0 +1,111 @@
...
+        <Field
+          id="note"
+          name="note"
+          label="Description"
+          component={renderConfig.AUTO_SIZE_FIELD}
+          validate={[maxLength(300)]}
+        />
+
+        <Form.Item label="Attached Files">
+          <p> Please upload all the required documents here for the variance application</p>
```
NWCalvank: Unintentional leading whitespace?

frontend-public/src/components/Forms/variances/EditVarianceForm.js
```diff
@@ -0,0 +1,109 @@
...
+      <Form layout="vertical" onSubmit={(event) => this.handleSubmit(event)}>
+        <VarianceDetails
+          mineName={this.props.mineName}
+          variance={this.props.variance}
+          removeDocument={this.props.removeDocument}
+          varianceStatusOptionsHash={this.props.varianceStatusOptionsHash}
+          complianceCodesHash={this.props.complianceCodesHash}
+        />
+        <Form.Item label="Attached Files">
+          <p> Please upload all the required documents here for the variance application</p>
```
NWCalvank: Unintentional leading whitespace?

frontend-public/src/components/common/DocumentTable.js
```diff
@@ -1,25 +1,28 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
-import { Table } from "antd";
+import { Table, Icon, Popconfirm, Button } from "antd";
 import CustomPropTypes from "@/customPropTypes";
 import { formatDate } from "@/utils/helpers";
 import downloadFileFromDocumentManager from "@/utils/actionlessNetworkCalls";
-import * as Strings from "@/constants/strings";
+import * as String from "@/constants/strings";
```
NWCalvank: What's the reason for switching this to `String`? I'd argue that `Strings` is preferable, as we shouldn't be modifying JS constructors imo

frontend-public/src/components/dashboard/mine/variances/VarianceTable.js
```diff
@@ -15,144 +15,208 @@ const propTypes = {
...
+        <div title="Code Section" style={this.errorStyle(record.isOverdue)}>
+          {text}
+        </div>
+      ),
+    },
+    {
+      title: "Submission Date",
+      dataIndex: "received_date",
+      // className on the column - applies to the entire column when full size
+      //  This attribute has no effect on the responsive view, same class is added to the div to handle responsive views
```
NWCalvank: Could you wrap this long comment line? 😃

frontend-public/src/components/dashboard/mine/variances/Variances.js
```diff
@@ -49,6 +68,80 @@ export class Variances extends Component {
     this.props.fetchVarianceStatusOptions();
   }
 
+  handlAddDocuments = (files, varianceGuid) => {
+    Promise.all(
```
NWCalvank: The promise isn't being returned, so this function will always return `undefined`, and the rest of the app will not be able to wait for this async operation to complete

frontend-public/src/components/dashboard/mine/variances/Variances.js
```diff
@@ -49,6 +68,80 @@ export class Variances extends Component {
     this.props.fetchVarianceStatusOptions();
   }
 
+  handlAddDocuments = (files, varianceGuid) => {
```
NWCalvank: Should be `handleAddDocuments` 😃

frontend-public/src/components/dashboard/mine/variances/Variances.js
```diff
@@ -49,6 +68,80 @@ export class Variances extends Component {
...
+        })
+      )
+    );
+  };
+
+  handleAddVariances = (files) => (values) => {
+    const received_date = values.received_date
+      ? values.received_date
+      : moment().format("YYYY-MM-DD");
+    const newValues = { received_date, ...values };
```
NWCalvank: If `received_date` exists on `values` then I think this is just going to overwrite `received_date` with `values.received_date`, so `newValues` will be equal to `values` in that case

frontend-public/src/components/dashboard/mine/variances/Variances.js
```diff
@@ -49,6 +68,80 @@ export class Variances extends Component {
...
+    return this.props
+      .createVariance(this.props.mine.mine_guid, this.props.mine.mine_name, newValues)
+      .then(async ({ data: { variance_guid } }) => {
+        await this.handlAddDocuments(files, variance_guid);
```
NWCalvank: This is where the issue of `handlAddDocuments` not returning a Promise would come up. It'll be a race condition against the following actions

frontend-public/src/components/modalContent/variances/EditVarianceModal.js
```diff
@@ -0,0 +1,94 @@
...
+export class EditVarianceModal extends Component {
+  state = { isLoaded: false };
+
+  componentDidMount() {
+    this.props.fetchVarianceById(this.props.mineGuid, this.props.varianceGuid).then(() => {
+      this.setState({ isLoaded: true });
+    });
+  }
+
+  // handling delete functionality inside the modal, so the data can be updated properly.
```
NWCalvank: This comment probably makes sense if you know what the alternative approach(es) could be, but coming across the code for the first time, it doesn't tell me anything.

frontend-public/src/styles/components/Loading.scss
```diff
@@ -14,3 +14,17 @@
     @include center(true, true);
   }
 }
+
+.loading-screen--small {
+  min-height: 500px;
+  width: 100%;
+  #spinner {
+    // offset size of spinner to center
+    left: 45% !important;
```
NWCalvank: Can you explain why it needs to be offset by 5%?

frontend-public/src/components/Forms/variances/AddVarianceForm.js
mmumby: Thank you! I was looking into the docs for remove just now

frontend-public/src/styles/components/Loading.scss
mmumby: The size of the spinner is based on `font-size` not on `width/height`, and because of that when you increase the `font-size` the icon does not re-center. 
I was forcing it to re-center here based on the larger `font-size`

frontend-public/src/components/modalContent/variances/EditVarianceModal.js
mmumby: removed!

frontend-public/src/components/dashboard/mine/variances/Variances.js
mmumby: Ah good catch, that should actually just be set to the current date, received_date will never be filled in by the user. 

updated it to be: 
```
const received_date = moment().format("YYYY-MM-DD");
```

frontend-public/src/components/dashboard/mine/variances/Variances.js
mmumby: Put that there just to see if you're paying attention haha

frontend-public/src/components/dashboard/mine/variances/Variances.js
mmumby: ahh, this explains so much.

frontend-public/src/components/dashboard/mine/variances/Variances.js
mmumby: Returned it, everything is working as expected now!

frontend-public/src/components/dashboard/mine/variances/VarianceTable.js
mmumby: done!

frontend-public/src/components/common/DocumentTable.js
mmumby: oops, meant to do the opposite for the exact reason. my bad.

frontend-public/src/components/Forms/variances/EditVarianceForm.js
mmumby: removed.

frontend-public/src/components/Forms/variances/AddVarianceForm.js
mmumby: removed

frontend-public/src/actionCreators/varianceActionCreator.js
```diff
@@ -9,28 +9,29 @@ import { ENVIRONMENT } from "@/constants/environment";
 import { createRequestHeader } from "@/utils/RequestHeaders";
 import CustomAxios from "@/utils/customAxios";
 
-export const createVariance = ({ mineGuid }, payload) => (dispatch) => {
+export const createVariance = (mineGuid, mineName, payload) => (dispatch) => {
```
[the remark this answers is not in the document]
mmumby: Switched it mainly because we aren't following this pattern in any of our actionCreators.
Happy to chat about the benefits of passing an object vs positional params though!

thegentlemanphysicist:
The document files are not inline with the label "Documents"  on small screens
image

PR #766, opened 2019-06-03
Bug/ MDS-1965

mmumby:
`MineVarianceTable.js` is used for both applications and approved variances, depending on which is displayed different columns will appear on the UI. 

The `className` attribute attached to the column array only takes effect on the web version, this PR adds the same class to the rendered `div` to conditionally renders the correct content when responsive.

### DIFF
--- a/frontend/src/components/mine/Variances/MineVarianceTable.js
+++ b/frontend/src/components/mine/Variances/MineVarianceTable.js
@@ -32,6 +32,8 @@ const defaultProps = {
 export class MineVarianceTable extends Component {
   errorStyle = (isOverdue) => (isOverdue ? { color: errorRed } : {});
 
+  hideColumn = (isApplication) => (isApplication ? "column-hide" : "");
+
   handleOpenModal = (event, isEditable, variance) => {
     event.preventDefault();
     if (isEditable) {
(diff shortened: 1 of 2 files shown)


frontend/src/components/mine/Variances/MineVarianceTable.js
```diff
@@ -32,6 +32,8 @@ const defaultProps = {
 export class MineVarianceTable extends Component {
   errorStyle = (isOverdue) => (isOverdue ? { color: errorRed } : {});
 
+  hideColumn = (isApplication) => (isApplication ? "column-hide" : "");
```
NWCalvank: Since this is a pure function, it doesn't need to be a method on the class (ie. it doesn't need to be this-aware). For simplicity, I'd suggest moving it out of the class. The same can be said for `errorStyle`

frontend/src/components/mine/Variances/MineVarianceTable.js
```diff
@@ -83,16 +85,24 @@ export class MineVarianceTable extends Component {
       {
         title: "Submission Date",
         dataIndex: "received_date",
-        className: !this.props.isApplication ? "column-hide" : "",
-        render: (text) => <div title="Submission Date">{text}</div>,
+        className: this.hideColumn(!this.props.isApplication),
+        render: (text) => (
+          <div className={this.hideColumn(!this.props.isApplication)} title="Submission Date">
```
NWCalvank: Is this `className` needed? It looks like if the entire column is hidden then this cell should be automatically hidden as well

mmumby: yeah, unfortunately, the outer `className` only applies when the table is full view and hides the entire column. 

But when the table is responsive that class doesn't apply, so the same class was added to the inner div to fix the behavior when responsive. 

If the outer class is removed and only the inner div has the styling, then the header will be visible when in full view but the content will not.

PR #768, opened 2019-06-04
MDS-1803 refactor mine status

JustinMacaulay:
This is the first part of the ticket to refactor the Mine Status to no longer use a constant.
-  Changed the backend to return the list of possible Mine status' instead of a pre-formatted constant.
-  Added a selector to the front end to format the Status information in the way the drop downs expect.


frontend/src/selectors/staticContentSelectors.js
```diff
@@ -192,3 +192,43 @@ export const getVarianceStatusOptionsHash = createSelector(
...
+const itemInList = (child, itemToAdd) => {
+  let item = child.find((x) => x.value === itemToAdd.code);
+
+  if (item === undefined) {
+    item = { value: itemToAdd.code, label: itemToAdd.label, children: [] };
+    child.push(item);
+  }
+
+  return item;
+};
```
NWCalvank: Could we refactor this to be more declarative?

I'm also concerned that we are mutating `child` here. If we want to add a new item to `child`, we should almost certainly be returning a new `child` array instead.

Lastly, if `child` is an array, it should probably be `children`? Or `list`, given that the function is called `itemInList`. I'll look more at what this function is being used for, but it's raising some red flags for me

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -192,3 +192,43 @@ export const getVarianceStatusOptionsHash = createSelector(
...
+      });
+      if (status.mine_operation_status_sub_reason.mine_operation_status_sub_reason_code !== null) {
+        itemInList(reason.children, {
+          code: status.mine_operation_status_sub_reason.mine_operation_status_sub_reason_code,
+          label: status.mine_operation_status_sub_reason.description,
+        });
+      }
+    }
+  });
+  return statusOptions;
```
NWCalvank: So, I think I understand what's happening here...

We are adding an object (called `code`) to `statusOptions` and then conditionally adding a `reason` property to that object, which has a `children` array property that may contain an object.

If that's right, let me provide some suggestions...

We always add `code` to `statusOptions`, unless an object with the same `code` property already exists in the array. So, I'd suggest initializing `statusOptions = data`. If there are duplicates, you can add a uniqueness filter based on the `code` property. This removes lines 211-214.

Next, since we're not adding/removing elements from the `statusOptions` array anymore, you can just return `statusOptions.map(({ children }) => { // add children + reason here })`

I imagine the whole thing will look something like...
```javascript
const transformMineStatus = (statusOptions) =>
  statusOptions.map((status) => {
    const code = status.mine_operation_status.mine_operation_status_code;
    const label = status.mine_operation_status.description;
    const reason = status.mine_operation_status_sub_reason;
    const subReason = status.mine_operation_status_sub_reason.mine_operation_status_sub_reason_code;

    const reasonChildren = subReason === null ? {} : {
      children: [
        {
          code: subReason.mine_operation_status_sub_reason_code,
          label: subReason.description,
        },
      ],
    };

    const codeChildren = reason === null ? {} : {
      children: [
        {
          code: reason.mine_operation_status_sub_reason_code,
          label: reason.description,
          ...reasonChildren,
        },
      ],
    };

    return {
      code,
      label,
      ...codeChildren,
    };
  });
```

This much more directly shows the output date shape and removes the need for mutations. `itemInList` isn't needed in this version.

If the separate `codeChildren` and `reasonChildren` variables make it confusing, you could just return those nested objects directly instead of pulling them out into variables. That would make the returned data shape maximally explicit. But I think breaking them out makes it a bit nicer to work with.

python-backend/app/api/mines/status/models/mine_status_xref.py
```diff
@@ -48,6 +48,10 @@ def json(self):
             if self.mine_operation_status_sub_reason_code else {},
         }
 
+    @classmethod
+    def active_status_codes(cls):
```
NWCalvank: Small naming suggestion. Since we already know we're dealing with status codes (based on the class), this can probably just be `active` or `active_records`. It would be nice to have a consistent method name across all/most of our classes for accessing active records, imo

python-backend/app/api/mines/status/resources/status.py
```diff
@@ -1,18 +1,29 @@
...
-    def get(self, mine_status_guid=None):
-        if mine_status_guid:
-            mine_status = MineStatus.find_by_mine_status_guid(mine_status_guid)
-            return mine_status.json() if mine_status else self.create_error_payload(
-                404, 'Mine Status not found'), 404
-        else:
-            return {'options': MINE_STATUS_OPTIONS}
+    def get(self, mine_status_guid):
+        if not mine_status_guid:
+            raise BadRequest('A mine guid must be provided.')
```
NWCalvank: Does MineStatusResource have multiple routes to this GET? If not, this method will never be called without a `mine_status_guid` url param

python-backend/app/api/mines/status/resources/status.py
```diff
@@ -1,18 +1,29 @@
...
+        mine_status = MineStatus.find_by_mine_status_guid(mine_status_guid)
+        if not mine_status:
+            raise NotFound('Mine Status not found')
+
+        return mine_status.json()
```
NWCalvank: Do we have MineStatus flagged for being updated to the new patterns and removing this json method?

python-backend/tests/status/resources/test_mine_status_resource.py
```diff
@@ -1,19 +1,17 @@
...
 
-from app.api.constants import MINE_STATUS_OPTIONS
-
 
 # GET
 def test_get_mine_status_option(test_client, db_session, auth_headers):
     get_resp = test_client.get('/mines/status', headers=auth_headers['full_auth_header'])
     get_data = json.loads(get_resp.data.decode())
-    assert get_data == {'options': MINE_STATUS_OPTIONS}
+    assert len(get_data.get('records')) == 14
```
NWCalvank: Not that it should really matter, since the test will fail either way, but I think directly accessing the key is acceptable in the tests (ie. `get_data['records']`). This helps to also verify that the `records` attribute is returned in the response object

python-backend/tests/status/resources/test_mine_status_resource.py
```diff
@@ -1,19 +1,17 @@
...
     assert get_resp.status_code == 200
 
 
 def test_get_mine_status_not_found(test_client, db_session, auth_headers):
     get_resp = test_client.get(
         '/mines/status/' + str(uuid.uuid4()), headers=auth_headers['full_auth_header'])
     get_data = json.loads(get_resp.data.decode())
-    assert get_data['error']['message'] == 'Mine Status not found'
+    assert get_data['message'] == '404 Not Found: Mine Status not found'
```
NWCalvank: I think we've moved toward a case-insensitive substring match on errors now, making it modestly useful but more flexible. Not saying either one is right or wrong, but I think that's how our newer tests have been written

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -192,3 +192,43 @@ export const getVarianceStatusOptionsHash = createSelector(
   [getDropdownVarianceStatusOptions],
   createLabelHash
 );
+
+const itemInList = (child, itemToAdd) => {
```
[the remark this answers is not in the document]
JustinMacaulay: Fair, I will change that.

python-backend/app/api/mines/status/resources/status.py
JustinMacaulay: It is flagged for refactoring.

python-backend/app/api/mines/status/resources/status.py
JustinMacaulay: Yes, there is one route to the MineStatusResource which requires a mine_status_guid and one route to the list resource which has no path params.

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -25,7 +25,7 @@ import {
   getConditionalCommodityOptions,
   getDisturbanceOptionHash,
   getCommodityOptionHash,
-  getMineStatusOptions,
+  getMineStatusDropdownOptions,
```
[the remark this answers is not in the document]
JustinMacaulay: I added it as this function because I didn't remove the getMineStatusOptions selector. That selector still exists and will get the list of all of the acceptable status triplicates. I named it DropDownOptions because this is specifically getting information formatted only for use by the antd cascader.

python-backend/app/api/mines/status/resources/status.py
NWCalvank: Yeah, but if they're different resources then this `get` will always have the `mine_status_guid`, no? Otherwise, it'll go to the other `get` method?

frontend/src/components/Forms/MineRecordForm.js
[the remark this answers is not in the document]
NWCalvank: Okay. That makes sense. Maybe I'm missing something, but why does the diff show you removed `getMineStatusOptions` then?

python-backend/app/api/mines/status/models/mine_status_xref.py
JustinMacaulay: Changed to just be active.

python-backend/app/api/mines/status/resources/status.py
JustinMacaulay: True, the error catch is unnecessary, I will remove it.

NWCalvank: Cool, thanks. I just took a look, and `find_by_mine_status_guid` doesn't check that `mine_status_guid` is a valid guid, so replacing the presence check with a uuid validation probably makes the most sense 😃 I can't remember if an invalid uuid on a SELECT query causes a runtime error

frontend/src/selectors/staticContentSelectors.js
nayr974: Another option could be to group and map and group and map:
```javascript
let result = records.groupBy((status) => status.mine_operation_status.mine_operation_status_code)
.map((statusCodeGroup) => ({
	value: statusCodeGroup[0].mine_operation_status.mine_operation_status_code,
	label: statusCodeGroup[0].mine_operation_status.description,
	children: statusCodeGroup.groupBy((status) => status.mine_operation_status_reason.mine_operation_status_reason_code)
		.map((statusReasonGroup) => ({
			value:statusReasonGroup[0].mine_operation_status_reason.mine_operation_status_reason_code,
			label: statusReasonGroup[0].mine_operation_status_reason.description,
			children: statusReasonGroup.groupBy((status) => status.mine_operation_status_sub_reason_code)
				.map((statusSubReasonGroup) => ({
					value:statusSubReasonGroup[0].mine_operation_status_sub_reason_code, 
					label: statusSubReasonGroup[0].description
					}))
  }))
}));
```

frontend/src/selectors/staticContentSelectors.js
JustinMacaulay: This function ha been removed.

frontend/src/selectors/staticContentSelectors.js
JustinMacaulay: Ryan worked to modify this function into something much more obvious and more focused on using groupby, filter, and map.

python-backend/tests/status/resources/test_mine_status_resource.py
JustinMacaulay: Changed

python-backend/tests/status/resources/test_mine_status_resource.py
JustinMacaulay: I haven't heard or seen anything on that, but I changed this to just null check the message instead.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -472,7 +472,7 @@ const mapStateToProps = (state) => ({
   mineRegionHash: getMineRegionHash(state),
   mineTenureHash: getMineTenureTypesHash(state),
   mineCommodityOptionsHash: getCommodityOptionHash(state),
-  mineStatusOptions: getMineStatusOptions(state),
+  mineStatusOptions: getMineStatusDropdownOptions(state),
```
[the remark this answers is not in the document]
JustinMacaulay: These have been changed to match that.

frontend/src/selectors/staticContentSelectors.js
mmumby: The new implementation is much easier to follow. Thank you, Justin and Ryan, for making that change.

PR #770, opened 2019-06-04
MDS-1885 Feature/GET all Variances (API changes only)

NWCalvank:
Opening this for discussion. There are two things of note:
1. GET variances agnostic to their mine is conceptually a decoupling of the mine-to-variance relationship. I've separated the resources associated with the `/variances` namespace into their own directory with the variance model, and I've renamed the mines-variances models to be prefixed with `Mine`.

How do we feel about this pattern for top-level namespaces?

I originally thought that this would mean that `/variances` (and, by extension) every top-level namespace would need to be its own micro-service, but this pattern does not work due to the FK constraint between a mine and its variances. This relationship seems to be of primary importance, and I'm not sure that splitting each namespace into its own micro-service would be worth giving-up that constraint.

If we want to split each namespace into its own micro-service, we will need to remove the cross-schema FKs, figure out the new best pattern for validating these relationships/running queries, and add `/model/model_guid` endpoint(s) to every namespace, regardless of their "business" relationships (in this case, the business logic that "mines have variances").

2. The utils test breaks any HTTP requests in tests that run after that test. I consider this fix OOS for this PR. At the time of opening this PR, I have commented the test out, but I am open to renaming the directory to something like `zzz_utils` to make that test run last. I'm also open to someone else opening a PR against this branch to fix that test if they know the solution. I'm not clear on everything that test is doing/covering.


python-backend/tests/utils/include/test_user_info.py
```diff
@@ -5,9 +5,10 @@
 from tests.constants import VIEW_ONLY_AUTH_CLAIMS, TOKEN_HEADER
 
 
-def test_get_user_info(test_client):
-    User._test_mode = False
+# FIXME: This test breaks all future http requests, presumably from the patch?
```
thegentlemanphysicist: Is there a ticket made for this fix?  May be a good idea to have a jira ticket to track it.

python-backend/app/api/mines/variances/resources/variance_list.py
```diff
@@ -2,19 +2,20 @@
 from flask_restplus import Resource, fields
 from werkzeug.exceptions import BadRequest, NotFound
 
-from ..models.variance import Variance
 from ....utils.access_decorators import requires_any_of, MINE_VIEW, MINE_CREATE, MINESPACE_PROPONENT
 from ...mine.models.mine import Mine
 from ....utils.resources_mixins import UserMixin, ErrorMixin
 from app.api.utils.custom_reqparser import CustomReqparser
 from app.api.mines.mine_api_models import VARIANCE_MODEL
+from app.api.variances.models.variance import Variance
```
thegentlemanphysicist: Why are there two conventions for imports?  '...' vs "app.api.etc".  Is this part of the microservice-afication?

python-backend/tests/utils/include/test_user_info.py
NWCalvank: No ticket at this point. I've just opened this as a discussion that should happen. If this PR gets merged without having resolved this issue, I'll create a ticket.

python-backend/app/api/mines/variances/resources/variance_list.py
NWCalvank: It doesn't look like there's any pattern that's been followed when deciding between absolute vs. relative imports. Not sure it's been discussed (and we don't have a linter monitoring this sort of thing).

I chose an absolute import here because `Variance` is outside of the `app/api/mines` directory, and anything over 2-3 levels up is probably too many directories to traverse with relative imports. But that's just my personal opinion.

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -0,0 +1,39 @@
...
+
+from app.extensions import api
+
+from ..models.variance import Variance
+from ..response_models import PAGINATED_VARIANCE_LIST
+from ...utils.access_decorators import requires_any_of, MINE_VIEW
+from ...utils.resources_mixins import UserMixin, ErrorMixin
+
+PAGE_DEFAULT = 1
+PER_PAGE_DEFAULT = 25
```
nayr974: Not suggesting you make this change, but these should probably be a common and shared constant as other endpoints hard-code 25. I'll keep in mind to do so if I'm working on something where it makes sense to do so. We have some continued back-end refactoring in our tech backlog.

NWCalvank: Definitely agree with that. Thanks for raising it 👍

PR #773, opened 2019-06-05
MDS-1607 Feature/Add logout page

NWCalvank:


frontend/src/tests/components/common/LogoutScreen.spec.js
```diff
@@ -0,0 +1,21 @@
...
+const setupProps = () => {
+  props.logoutUser = jest.fn();
+  props.keycloak = {};
+};
+
+beforeEach(() => {
+  setupProps();
+});
+
+describe("LogoutScreen", () => {
```
mmumby: The test should have the same file name as the component being tested.

NWCalvank: Oh, I missed updating that! It was originally called LogoutScreen. Will update 👍

PR #777, opened 2019-06-06
MDS-1846 Feature/Add Metabase Dashboard

NWCalvank:
I've also consolidated the active menu button ids. I'd recommend moving this into an `--active` modifier class instead of an id, but dealing with the change in specificity of classes vs. ids was OOS for this small bit of tech debt removal.


frontend/src/components/navigation/NavBar.js
```diff
@@ -42,6 +42,8 @@ export class NavBar extends Component {
     this.props.fetchMineVerifiedStatuses(`idir\\${this.props.userInfo.preferred_username}`);
   }
 
+  ifActiveButton = (route) => (includes(this.props.activeButton, route) ? "active-menu-btn" : "");
```
mmumby: This is a much better way of doing it 😅

NWCalvank:
Closing PR to burn down the environment. Build timed out.

frontend/package.json
```diff
@@ -20,6 +20,7 @@
     "filepond-polyfill": "^1.0.2",
     "hoist-non-react-statics": "^3.2.1",
     "jest-localstorage-mock": "^2.2.0",
+    "jsonwebtoken": "^8.5.1",
```
FreshGyan: Do you still need this package?

NWCalvank: Oh, not anymore. That logic/dependency was moved to the python-backend

python-backend/requirements.txt
```diff
@@ -29,3 +29,4 @@ sqlalchemy-filters==0.8.0
 uwsgi==2.0.17.1
 uwsgitop==0.10
 Werkzeug==0.14.1
+PyJWT==1.7.1
```
Jsyro: Is this used, I don't see it imported anywhere.

NWCalvank: It's imported as `jwt`

PR #784, opened 2019-06-10
Minespace content changes and document download placeholder

pdittaro:


pdittaro:
Same as before, couldn't get OC to build properly though.

PR #790, opened 2019-06-11
Bug/missed data for mine status description

JustinMacaulay:
Migration to fix the bad insert for the mine status description.

### DIFF
--- a/python-backend/tests/status/resources/test_mine_status_resource.py
+++ b/python-backend/tests/status/resources/test_mine_status_resource.py
@@ -1,11 +1,12 @@
 import json, uuid
+from app.api.mines.status.models.mine_status_xref import MineStatusXref
 
 
 # GET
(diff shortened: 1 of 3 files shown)


python-backend/tests/status/resources/test_mine_status_resource.py
```diff
@@ -1,11 +1,12 @@
 import json, uuid
+from app.api.mines.status.models.mine_status_xref import MineStatusXref
 
 
 # GET
 def test_get_mine_status_option(test_client, db_session, auth_headers):
     get_resp = test_client.get('/mines/status', headers=auth_headers['full_auth_header'])
     get_data = json.loads(get_resp.data.decode())
-    assert len(get_data['records']) == 14
+    assert len(get_data['records']) == len(MineStatusXref.active())
```
nayr974: much better than a hard-coded number of record

PR #791, opened 2019-06-11
MDS-2029 Feature/ Add filtering by application status code to GET /variances

NWCalvank:


python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -18,22 +19,47 @@ class VarianceResource(Resource, UserMixin, ErrorMixin):
         description='Get a list of variances.',
         params={
             'page': f'The page number of paginated records to return. Default: {PAGE_DEFAULT}',
-            'per_page': f'The number of records to return per page. Default: {PER_PAGE_DEFAULT}'
+            'per_page': f'The number of records to return per page. Default: {PER_PAGE_DEFAULT}',
+            'variance_application_status_code':
+            'Comma-separated list of code statuses to include in results. Default: All status codes.',
```
thegentlemanphysicist: is there no point to using an f-string when we don't add a {variable} ?

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -18,22 +19,47 @@ class VarianceResource(Resource, UserMixin, ErrorMixin):
...
+    def _apply_filters_and_pagination(self,
+                                      page_number=PAGE_DEFAULT,
+                                      page_size=PER_PAGE_DEFAULT,
+                                      application_status=None):
+        status_filter_values = list(map(
+            lambda x: x.variance_application_status_code,
+            VarianceApplicationStatusCode.active()))
+
+        if application_status is not None:
+            status_filter_values = application_status.split(',')
```
thegentlemanphysicist: Is it possible to have multiple application statuses or is this used in searching for all variances in a list of application statuses?

python-backend/tests/status_code_gen.py
```diff
@@ -95,3 +96,11 @@ def SampleDangerousOccurrenceSubparagraphs(num):
         db.session.query(ComplianceArticle).filter(
             ComplianceArticle.section == '1', ComplianceArticle.sub_section == '7',
             ComplianceArticle.paragraph == '3', ComplianceArticle.sub_paragraph != None).all(), num)
+
+def RandomVarianceApplicationStatusCode():
+    return random.choice([
+        x.variance_application_status_code
+        for x in filter(
+            lambda x: x.variance_application_status_code not in ['APP', 'DEN'],
+            VarianceApplicationStatusCode.active())
```
thegentlemanphysicist: Is this setting the status to active when the variance is not app or denied?

python-backend/app/api/variances/resources/variance_resource.py
NWCalvank: That's correct 😄 the `f` allows for string interpolation, like backticks in JS

python-backend/app/api/variances/resources/variance_resource.py
NWCalvank: This searches across all variances and only returns the ones with the requested application status(es). If no statuses were provided, it filters with all statuses -- ie. no variances will be filtered out.

python-backend/tests/status_code_gen.py
NWCalvank: This is just using the `.active()` method, which is a custom method on the `VarianceApplicationStatusCode` model, to run a query for all active variance application status code options. Then, it filters 'APP' and 'DEC' out of the list, returning all the other codes.

The reason for filtering these out is that approved and denied variances have requirements for other column values. This is part of the tests factory code, so you can see how it's used and how APP & DEC are handled in `factories.py`

python-backend/app/api/variances/resources/variance_resource.py
thegentlemanphysicist: cool

PR #792, opened 2019-06-12
Bug/mds 1964

thegentlemanphysicist:
This addresses the buggyness of the status date.  I also made a quick fix on the backend since put and post were supposed to accept and empty status_date in the original ticket.  The edgecases I looked at were:
Edit Modal, Newly created mine (Status not null)
    • Edit does not override null date if status not changed
    • Edit does not override a date if status not changed
    • Date can be changed to new date
    • Date can no longer be set to null.  (there is one edge case, but the user has to work at it and it no longer causes a 500 error)
Edit Modal, Imported mine
    • Editing mine with status null -> status !null updates date to today
    • Editing mine with null status allows date to be set in the past
Create Modal
    • Default create sets status to today’s date (handled in component mount)
    • Can create with a date in the past
    • Filling out non-status fields does not override the status-date
    • can’t create date in future


frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -175,6 +173,15 @@ export class MineRecordForm extends Component {
...
   );
 
+  toggleStatusDate = () =>
+    this.setState((prevState) => ({ showStatusDate: !prevState.showStatusDate }));
+
+  // When the status changes, set the status date to current date.
+  onStatusChange = () => {
+    const date = new Date();
+    this.props.change("status_date", date);
+  };
```
mmumby: This is much better 🙌🏼

frontend/src/components/Forms/MineRecordForm.js
```diff
@@ -49,24 +48,31 @@ const propTypes = {
...
+  isCreated: PropTypes.bool,
 };
 
 const defaultProps = {
   title: "",
   currentMineTypes: [],
   mine_types: [],
-  mineStatus: [],
+  isCreated: false,
 };
```
mmumby: I think we can do better with this constant, I know I initially named it, my bad. 
I think `isNewRecord` or `isNewMine` or something to that effect would be better.

frontend/src/components/dashboard/minesHomePage/Dashboard.js
```diff
@@ -450,7 +451,7 @@ export class Dashboard extends Component {
               <AuthorizationWrapper permission={Permission.ADMIN}>
                 <AddButton
                   onClick={(event) =>
-                    this.openModal(event, this.handleSubmit, ModalContent.CREATE_MINE_RECORD)
+                    this.openModal(event, this.handleSubmit, ModalContent.CREATE_MINE_RECORD, true)
                   }
```
mmumby: the `true` here isn't needed. 

Instead, you can just initiate the prop inside the openModal function like so: 
```
openModal(event, onSubmit, title) {
    const handleDelete = () => {};
    event.preventDefault();
    this.props.openModal({
      props: {
        handleDelete,
        onSubmit,
        title,
        isCreated: true,
      },
      content: modalConfig.MINE_RECORD,
    });
```

PR #796, opened 2019-06-13
PVC storage upgrade

FreshGyan:
Switching gluster storage from file to block. The speed difference is significant.
Also switching database backups to the nfs-backup storage type PVC because pathfinder team backs it up constantly

Workflow:

- [ ] Get code review
- [ ] Backup data from TEST and PROD
- [ ] Merge code
- [ ] Wait till code reaches TEST
- [ ] Copy data over from old drives to new one
- [ ] Merge data to PROD
- [ ] Copy data over from old drives to new one


pipeline/config-prod.groovy
```diff
@@ -273,7 +273,7 @@ environments {
         vars {
             DB_PVC_SIZE = '50Gi'
             DOCUMENT_PVC_SIZE = '50Gi'
```
Jsyro: should we take this change to increase document storage pvc? how much room do we have left right now.

openshift/_python36.dc.json
```diff
@@ -192,14 +192,14 @@
...
+        "name": "mds-document-block-storage${SUFFIX}",
         "annotations": {
-          "volume.beta.kubernetes.io/storage-class": "gluster-file",
+          "volume.beta.kubernetes.io/storage-class": "gluster-block",
           "volume.beta.kubernetes.io/storage-provisioner": "kubernetes.io/glusterfs"
         }
       },
       "spec": {
-        "accessModes": ["ReadWriteMany"],
+        "accessModes": ["ReadWriteOnce"],
```
Jsyro: we have more than one backend pod. I don't think we can do this for the backend (we also loose rolling deploys.

openshift/_nginx.dc.json
```diff
@@ -88,14 +88,14 @@
...
+        "name": "mds-logs-block${SUFFIX}",
         "annotations": {
-          "volume.beta.kubernetes.io/storage-class": "gluster-file",
+          "volume.beta.kubernetes.io/storage-class": "gluster-block",
           "volume.beta.kubernetes.io/storage-provisioner": "kubernetes.io/glusterfs"
         }
       },
       "spec": {
-        "accessModes": ["ReadWriteMany"],
+        "accessModes": ["ReadWriteOnce"],
```
Jsyro: same here, we have nginx on high availability. i think this needs to be RWX.

pipeline/config-prod.groovy
itchison: We have used 5.5GB of the 50GB. Not a pressing concern but more of a mid-term issue.

PR #798, opened 2019-06-13
Feature/mds 2055

mmumby:
Added additional fields to `AddVarianceFrom` and `EditVarianceForm` to capture: 
- Categories to associate with documents
- if associated parties have been notified

- added all redux for fetching/storing the `varianceDocumentCategoryOptions` and all associated tests


frontend/src/components/Forms/variances/AddVarianceForm.js
```diff
@@ -50,6 +51,7 @@ export class AddVarianceForm extends Component {
   };
 
   render() {
+    const filesUploaded = this.state.uploadedFiles.length >= 1;
```
NWCalvank: Logically, this will be exactly the same, but I'm so used to seeing `> 0` that this looks weird 😆

PR #799, opened 2019-06-14
tone down dev resources

FreshGyan:
- Reducing Dev resources to allocate more capacity for more Pull Requests
- Less resources also keeps pathfinder team happier and keeps our deploys faster
- Made UWSGI threads configurable, to test better before the bigger release


microservices/nris_api/openshift/_python36_oracle.dc.json
```diff
@@ -69,6 +69,14 @@
       "name": "MEMORY_REQUEST",
       "value": "768Mi"
     },
+    {
+      "name": "UWSGI_PROCESSES",
+      "value": "2"
+    },
+    {
+      "name": "UWSGI_THREADS",
+      "value": "4"
```
NWCalvank: Were these intentionally flipped? It was 4 processes and 2 threads before

functional-tests/src/test/resources/GebConfig.groovy
```diff
@@ -154,8 +154,8 @@ reportOnTestFailureOnly = false //true
 //in seconds
 //If unspecified, the values of 5 for timeout and 0.1 for retryInterval.
 waiting {
-    timeout = 30
-    retryInterval = 1
+    timeout = 60
+    retryInterval = 2
```
NWCalvank: 12X the default timeout? Is that a symptom of our tests being in a dire situation or of the defaults being unrealistically low?

microservices/nris_api/openshift/_python36_oracle.dc.json
FreshGyan: Yes! I did some more digging and it turns out the flow is as follows:
Since python is single threaded, the more processes you have, the more copies of your program it creates and thus requires more CPU. Basically each fork gets its own GIL and own state and everything.
For dev it seemed okay to work with one processes so i toned that down. But for Test and PROD I left em at 2

I kept threads to 2 since threads take more memory and incoming requests can still be served "multi-threaded"

functional-tests/src/test/resources/GebConfig.groovy
FreshGyan: I was doing some testing thinking that now that DEV will be relatively slower it may make more sense to increase timeouts on functional tests given how rapidly they run and jump from one scenario to next

FreshGyan: Im not too too certain these are _needed_ but not sure if I should turn them back down.

@thegentlemanphysicist any strong feelings?

microservices/nris_api/openshift/_python36_oracle.dc.json
NWCalvank: Makes sense. Thanks for the explanation 👍

PR #801, opened 2019-06-14
Geb/mds 2039

thegentlemanphysicist:
This now allows tests to run and pass on nonempty database.  Mostly by changing names so they are first in alphabetical order.  It also deletes test generated data more cleanly without deleting user generated data

The tests now run in a full database.
The permit tests have not failed due to out of bounds exceptions for many runs.


thegentlemanphysicist:
> If it's the latter, I wonder if adding a search step would be all that's needed to make it work.
You are correct, a search would make this work.  However, that would require a refresh of the page and each refresh of the page increases the chance of a test failing because the loading screen doesn't clear.  I figured this was the least test-run-time intensive approach.

functional-tests/src/test/groovy/data/data_deletion.sql
```diff
@@ -12,26 +16,49 @@ BEGIN
...
     DELETE FROM mine_status WHERE create_user LIKE IDIR_USER;
     DELETE FROM mine_tailings_storage_facility WHERE create_user LIKE IDIR_USER;
-    DELETE FROM mine_document WHERE create_user LIKE IDIR_USER;
     DELETE FROM mine_expected_document_xref
     WHERE exp_document_guid = ANY(
-    SELECT exp_document_guid
-    FROM mine_expected_document
-    WHERE create_user LIKE IDIR_USER);
+        SELECT exp_document_guid
+        FROM mine_expected_document
```
NWCalvank: Not sure what the goal is of this `WHERE`. It looks like this will just delete all records from `mine_expected_document_xref`? The original condition was limited to documents created by the `IDIR_USER`

functional-tests/src/test/groovy/FrontendCore/spec/Dashboard.groovy
```diff
@@ -71,22 +69,22 @@ class  DashboardSpec extends GebReportingSpec {
     }
 
     def "Scenario: User can view a mine"(){
-        def viewMineName = firstMineName
+        def SearchedDashboard = new Dashboard(url: "dashboard/mines?page=1&per_page=25&search=MINETEST")
```
NWCalvank: Nice 👍 

Do we have a constant for our test mine name? If we could use that here instead of a hardcoded string, that would probably be best

functional-tests/src/test/groovy/FrontendCore/spec/Contacts.groovy
```diff
@@ -62,13 +62,15 @@ class  Contacts extends GebReportingSpec {
         and: "I click on the contact page link."
         waitFor() { createContactButton.click() }
 
-        and: "I fill out the first form."
+        and: "I fill out the create contacts forms."
         createContactForm.createContactFormOne(input1)
         createContactForm.createContactFormTwo(input2)
-        waitFor {toastMessage != null}
+        waitFor {toastMessage != null}
```
NWCalvank: It's a good practice to auto-trim trailing whitespace. This can usually be done on-save by your editor. Maybe check your settings and see if you can toggle that option on.

functional-tests/src/test/groovy/data/data_deletion.sql
thegentlemanphysicist: good catch O_o

functional-tests/src/test/groovy/FrontendCore/spec/Contacts.groovy
thegentlemanphysicist: weird the setting is turned on.

functional-tests/src/test/groovy/data/data_deletion.sql
thegentlemanphysicist: fixed

functional-tests/src/test/groovy/FrontendCore/spec/Dashboard.groovy
thegentlemanphysicist: fixed

functional-tests/src/test/groovy/FrontendCore/spec/Contacts.groovy
thegentlemanphysicist: fixed

functional-tests/src/test/groovy/FrontendCore/spec/Dashboard.groovy
thegentlemanphysicist: Now running through the pipeline, fingers crossed

PR #809, opened 2019-06-17
MDS-1958 Bug/ Fix error regression in overlapping Mine Manager dates

NWCalvank:
This fixes a regression where the `mine_party_appt` date overlap db error was no longer being caught at the application layer. The result was that a generic (and unhelpful) error was returned to the client.

Extra Context:
When the `try/except` block was added inside of the `.save()` that re-throws an `InternalServerError`, the error type changed (so the resource-level catch no longer worked) and the original error was swallowed by the `save()` method.

The real heart of the issue is that SQLAlchemy errors are not being caught by the `api.errorhandler` methods, which catch above the resource level (instead of below, like this `save()` method catch). I'm going to look further into properly catching the SQLAlchemy errors at the highest level such that the `save()` catch can always re-throw the original error after handling the database rollback, instead of conditionally doing so, as is done in this PR.

I'm opening this PR regardless because the bug fix continues to be valuable in the meantime.


thegentlemanphysicist:
I noticed a bug on this, if a user adds a mine-manager, then the the mine manger is deleted by Aaron (delete button).  We still can't add a new mine manager that overlaps  with the previous manager's dates.

thegentlemanphysicist:
Turns out the bug was an existing problem. Now in jira as MDS-2082

NWCalvank:
If we're happy with PR #813 then this one is unnecessary, as it was a make-due until a global error handling solution was found.

NWCalvank:
Global SQLAlchemy error handlers have been merged. This PR is obsolete.

PR #812, opened 2019-06-18
Feature/mds 1703 attach supporting docs for do

JustinMacaulay:
Added the document workflow for dangerous occurrences.


frontend/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -59,6 +61,56 @@ const columns = (props) => [
...
+    title: "Initial Report Documents",
+    dataIndex: "initialDocuments",
+    width: 200,
+    render: (text, record) => (
+      <div title="Initial Report Documents">
+        {record.docs.length === 0 ? (
+          <span>--</span>
+        ) : (
+          record.docs
+            .filter((file) => file.mine_incident_document_type_code === "INI")
```
pdittaro: Could we create a helper function for this, since its the same across both columns?

frontend/src/components/modalContent/AddIncidentModal.js
```diff
@@ -64,11 +66,24 @@ const actionVerb = (newIncident) => {
...
+  onRemoveFile
+) => [
   {
     title: "Initial Report",
     content: (
-      <AddIncidentReportingForm initialValues={props.initialValues} inspectors={props.inspectors} />
+      <AddIncidentReportingForm
+        initialValues={props.initialValues}
+        inspectors={props.inspectors}
+        onFileLoad={onFileLoad}
```
pdittaro: Why is this prop passed here? We don't do file upload in this form?

frontend/src/components/modalContent/AddIncidentModal.js
```diff
@@ -153,11 +182,17 @@ const StepForms = (props, next, prev, handleIncidentSubmit) => [
...
+      : [],
+  };
 
   handleIncidentSubmit = () => {
     this.props.onSubmit({
       ...this.props.addIncidentFormValues,
+      updated_documents: this.state.uploadedFiles,
     });
     // TODO: Catch error
     this.close();
```
pdittaro: Are the documents orphaned if they cancel upload? I know we talked about running a cron to deal with orphaned documents.

pipeline/config-dev.groovy
```diff
@@ -235,28 +235,28 @@ environments {
             }
             resources {
                 node {
-                    cpu_request = "50m"
-                    cpu_limit = "100m"
-                    memory_request = "256Mi"
-                    memory_limit = "384Mi"
-                    replica_min = 1
-                    replica_max = 1
+                    cpu_request = "100m"
```
pdittaro: Was this fine tuning done as part of this PR, or is this in conflict with changes that @FreshGyan  was doing earlier this week?

frontend/src/components/modalContent/AddIncidentModal.js
nayr974: Removed 👍

frontend/src/components/modalContent/AddIncidentModal.js
nayr974: Documents are orphaned, yes. Orphaned documents are a system concern and occur in many places, and we really do need to deal with them.

pipeline/config-dev.groovy
nayr974: This was done for Brian's demo, I think the last of which is today. I'll revert the changes after.

frontend/src/components/mine/Incidents/MineIncidentTable.js
nayr974: I'm not sure it's useful to create a helper function for a single filter. It's one line and the intent is self describing. That said, maybe defining a constant for our INI/FIN codes would be a good idea?

pdittaro: This isn't one line, its around 15 and its repeated half the file down but checking against a different document code

PR #815, opened 2019-06-19
MDS-1554 Fix/ Add frontend date validation for mine contacts

NWCalvank:
Adds form validation to Add/Edit mine party appoints. This prevents a user from being able to submit the form, only to have the backend reject the request due to a date conflict.

The message includes the type of contact being added as well as the conflicting party, on a per-field basis.

~Note~
~There is a considerable chunk of duplication here. I'm in two minds about it, and I will play with moving it to the shared parent component while I garner feedback from the team. Essentially, the forms already handle their validation independently, and I'm not sure that we should assume that this date validation is actually a pattern worth abstracting quite yet. On the flip side.. it's a fairly large function to duplicate.~

Edit: This PR is good-to-go and ready for review 🎉


pdittaro:
There is a similar issue with the Incident date validation, if thats worth anything. I'll take a look at how you implemented this and we can review tomorrow?

frontend/src/components/Forms/PartyRelationships/AddPartyRelationshipForm.js
```diff
@@ -23,13 +28,58 @@ const defaultProps = {
...
+  const toDate = (dateString) => moment(dateString, "YYYY-MM-DD").toDate();
+  const allAppointments = [...existingAppointments, newAppt].sort((a, b) =>
+    toDate(a.start_date) > toDate(b.start_date) ? 1 : -1
+  );
+
+  const errorMessages = {};
+  for (let i = 0; i < allAppointments.length - 1; i += 1) {
+    const current = allAppointments[i];
+    const next = allAppointments[i + 1];
+
```
thegentlemanphysicist: If bad data is somehow imported into the system (two mine managers overlap, but not the current one being added) , won't this fail?

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
```diff
@@ -21,13 +26,58 @@ const defaultProps = {
   mine: {},
 };
 
-const validate = (values) => {
+const checkDatesForOverlap = (values, props) => {
```
thegentlemanphysicist: If this is the same function as above can we make a single helper function?

NWCalvank:
I appreciate the feedback from reviewers. I should have put the WIP on this, as this is not the final version of the code. I'll take your feedback into consideration as I'm making my changes

frontend/src/components/Forms/PartyRelationships/AddPartyRelationshipForm.js
NWCalvank: This constraint exists at the database level at the moment. We can't simultaneously enforce "good" data while ignoring "bad" data without severely making a mess of our validations

frontend/src/components/Forms/PartyRelationships/EditPartyRelationshipForm.js
NWCalvank: Please see the opening comment on this PR

frontend/src/components/modalContent/EditPartyRelationshipModal.js
```diff
@@ -8,10 +8,15 @@ const propTypes = {
   handleChange: PropTypes.func.isRequired,
   closeModal: PropTypes.func.isRequired,
   title: PropTypes.string.isRequired,
+  partyRelationships: PropTypes.arrayOf(CustomPropTypes.partyRelationship),
   partyRelationship: CustomPropTypes.partyRelationship.isRequired,
   mine: CustomPropTypes.mine.isRequired,
 };
 
+const defaultProps = {
```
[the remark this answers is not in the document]
NWCalvank: This is because I'm not marking partyRelationships as required here, so I need to provide a default value.

frontend/src/components/Forms/PartyRelationships/AddPartyRelationshipForm.js
```diff
@@ -23,13 +28,58 @@ const defaultProps = {
...
+  const newAppt = { start_date: null, end_date: null, ...values };
+
+  // Sort all appointments into ascending order by start_date
+  const toDate = (dateString) => moment(dateString, "YYYY-MM-DD").toDate();
+  const allAppointments = [...existingAppointments, newAppt].sort((a, b) =>
+    toDate(a.start_date) > toDate(b.start_date) ? 1 : -1
+  );
+
+  const errorMessages = {};
+  for (let i = 0; i < allAppointments.length - 1; i += 1) {
```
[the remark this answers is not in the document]
NWCalvank: I'm using that here because I want to exit the loop one element early. This could likely be done with a `forEach`, but this seemed like the more explicit way to handle it.

pdittaro:
This looks clean, and a good implementation of form validation outside of the validators on specific fields. There are two unit tests failing, otherwise good to go I would think.

frontend/src/tests/utils/Validate.spec.js
```diff
@@ -216,4 +217,237 @@ describe("Validate class", () => {
       expect(validateIncidentDate(reportedDate)(value)).toBeTruthy();
     });
   });
+
+  describe("`validateDateRanges` function", () => {
```
mmumby: 🔥

PR #817, opened 2019-06-20
Feature/MDS 1976 

mmumby:
- Added paginated table for variances on "My Dashboard" 
- Only variances "In-review" or "waiting for decision" are visible at this time
- Variances are visible and editable by every user with create permissions.


frontend/src/components/mine/Variances/MineVarianceTable.js
```diff
@@ -50,6 +61,8 @@ export class MineVarianceTable extends Component {
     variances.map((variance) => ({
       key: variance.variance_guid,
       variance,
+      mineName: variance.mine_name || null,
+      mineGuid: variance.mine_guid || null,
```
NWCalvank: Is this to convert empty strings and undefined to null?

mmumby: `MineVarianceTable` is being used to display the results from `/variances` and `/mines/guid/variances`. The former includes the `mine_name` and `mine_guid` int he payload. 

I was trying to avoid the error "cannot read mine_name of undefined"

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
```diff
@@ -51,14 +98,104 @@ export class CustomHomePage extends Component {
...
+      },
+      content: modalConfig.VIEW_VARIANCE,
+      isViewOnly: true,
+    });
+  };
+
+  handleFilterChange = (pagination, filters) => {
+    const { status } = filters;
+    this.setState({ variancesLoaded: false });
+    const params = { ...this.state.params, variance_application_status_code: [...status], page: 1 };
```
NWCalvank: `[...status]` is this spreading an array inside of an array? If so, that's a noop. You can just use `status` instead

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
```diff
@@ -51,14 +98,104 @@ export class CustomHomePage extends Component {
...
+            {
+              document_manager_guid,
+              document_name,
+              variance_document_category_code,
+            }
+          )
+        )
+      );
+      this.props.closeModal();
+      this.props.fetchVariances(this.state.params).then(() => {
```
NWCalvank: I could be wrong, but this makes an API request to get the latest version of the variances, correct?

When a document is associated with a variance, the updated variance should be returned. Which means that the response from each `addDocumentToVariance` should be the latest version of the variance. That would make the following call to `fetchVariances` unnecessary.

Let me know if I'm wrong about either of these things or if you have any questions

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -55,7 +56,7 @@ def _apply_filters_and_pagination(self,
             status_filter_values = application_status.split(',')
 
         filtered_query = apply_filters(
-            Variance.query,
+            Variance.query.order_by(desc(Variance.received_date)),
```
NWCalvank: Oh, we should add to the Swagger docs that the sort order is `received_date`

python-backend/app/api/variances/models/variance.py
```diff
@@ -45,6 +45,10 @@ class Variance(AuditMixin, Base):
     mine_documents = db.relationship(
         'MineDocument', lazy='joined', secondary='variance_document_xref')
     inspector = db.relationship('Party', lazy='joined', foreign_keys=[inspector_party_guid])
+    mine = db.relationship('Mine', lazy='joined')
+
+    mine_name = association_proxy('mine', 'mine_name')
```
NWCalvank: I don't love having `mine_name` on the variance response, but I can see why it might be necessary in practice. What was the reason for needing to add it?

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
mmumby: removed!

python-backend/app/api/variances/resources/variance_resource.py
mmumby: Can that be done in this PR?

python-backend/app/api/variances/models/variance.py
mmumby: When viewing the table of variances sitewide, it is necessary to know what mine the variance belongs to, so we're displaying the `mine_name` and a link to the mine on the table view of variances.

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
mmumby: The call `fetchVariances` will update the `variances` redux state with the updated variance information. 
The response returned from `addDocumentToVariance` could, in theory, update the singular `variance` state, but the `fetchVariances` is necessary to update the table view of all `variances`.

NWCalvank: I'm not clear on why updating the singular variance state is insufficient. But I know this is the current pattern across the app, so it's probably a bit OOS anyway.

python-backend/app/api/variances/models/variance.py
NWCalvank: In-theory, that's what the `mine_guid` is for, but I can see why making a separate API call to fetch the mine name isn't ideal.

I think it might be worth coming up with a pattern for conditionally requesting these related fields. Similar to how we are/were treating nested response objects. But that's a larger team discussion. Marking this as resolved.

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
mmumby: Updating the singular 'variance' will have no effect on the same variance record that's contained in the `variances` list. 

The state for `variance` is updated every time an edit modal is opened, whereas the 'variances` state is only updated when the page is initially loaded.
 If `fetchVariances` is not called every time a record is updated than the table view will never update to reflect the change in data (unless a route change or hard refresh occurs)

Is that more clear?

python-backend/app/api/variances/resources/variance_resource.py
mmumby: added!

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
NWCalvank: Yep

PR #818, opened 2019-06-20
Task/import do's from mms

Jsyro:
Table updates to support some new columns we want to preserve.  and type code for dummy record.
ETL shows process of converting mms.mmocc with some related dumps as csv files. Could pull in more tables, but local development doesn't have FDW to MMS.
 
Plan it to run this offline and get postgres to get a insert script and run that script.

this will need to go in after other mine_incident PR as this sql builds on that.


migrations/sql/V2019.09.19.02.53__add_historical_mine_incident_followup_type_code.sql
```diff
@@ -0,0 +1,18 @@
...
+    active_ind,
+	create_user,
+	update_user
+	)
+VALUES
+('HUK', 'Historical - Unknown', 40, 'F', 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
+ALTER TABLE mine_incident ADD COLUMN mms_min_acc_no character varying(9); 
+ALTER TABLE mine_incident ADD COLUMN mms_recp_cd character varying(3);
```
klefler: This column is no longer used in MMS... not sure what value it has. What data are you trying to capture?

klefler: As discussed with Jason... we want INSP_CD instead of RECP_CD.

migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -393,13 +393,15 @@ INSERT INTO mine_incident_followup_investigation_type
...
 	update_user
 	)
 VALUES
-	('MIU', 'Yes - MIU Investigation', 10, 'system-mds', 'system-mds'),
-	('INS', 'Yes - Inspector Investigation', 20, 'system-mds', 'system-mds'),
-	('NO', 'No', 30, 'system-mds', 'system-mds')
+	('MIU', 'Yes - MIU Investigation', 10, 'T', 'system-mds', 'system-mds'),
+	('INS', 'Yes - Inspector Investigation', 20, 'T', 'system-mds', 'system-mds'),
+	('NO', 'No', 30, 'T', 'system-mds', 'system-mds'),
+    ('HUK', 'Historical - Unknown', 40, 'F', 'system-mds', 'system-mds')
```
nayr974: The front-end needs to be modified to not display this in the list of options.

migrations/sql/V2019.09.19.02.53__add_historical_mine_incident_followup_type_code.sql
```diff
@@ -0,0 +1,16 @@
...
+	display_order,
+    active_ind,
+	create_user,
+	update_user
+	)
+VALUES
+('HUK', 'Historical - Unknown', 40, 'F', 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
+ALTER TABLE mine_incident ADD COLUMN mms_min_acc_no character varying(9);
```
nayr974: I disagree with verbatim putting mms column names into our data structure. We should have an ETL_INCIDENTS table if we need to persist key relationships, and if we need to transform data we should have our own column names.

python-backend/app/scripts/incident_import.py
```diff
@@ -0,0 +1,197 @@
+import psycopg2
```
nayr974: PYTHON FOR DATA TRANSFORMATION. GREAT USE CASE FOR PYTHON!!! 💯 💯 💯

migrations/sql/afterMigrate__1__seed_data.sql
Jsyro: does it? Aaron is fine if for this week we allow it to be a fully fledged option with no special behaviour, there will be a ticket to lock it down.

nayr974: It should be a one line change, .filter((x) => x.active_ind=true). Might be more work due to overhead to put it off, but if you want to separate it into another ticket, fine.

PR #819, opened 2019-06-20
Fix regional mines ETL updated_at issue

FreshGyan:
Only update the record if any of the data has changed and the last person to change that data was the automated migration.

NOTE: I'll revert the postgres increased resources after getting code review so that I can test any changes asked.

### DIFF
--- a/migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
+++ b/migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
@@ -773,6 +773,14 @@ CREATE OR REPLACE FUNCTION transfer_permit_permitee_information() RETURNS void A
                 party_type_code  = etl.party_type
             FROM ETL_PERMIT etl
             WHERE party.party_guid = etl.party_guid
+            AND (
+                party.first_name != etl.first_name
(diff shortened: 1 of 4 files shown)


migrations/sql/R__stored_procedure_for_permit_and_permitee_information.sql
```diff
@@ -583,7 +583,7 @@ CREATE OR REPLACE FUNCTION transfer_permit_permitee_information() RETURNS void A
             insert_amendment_row integer;
             total_amendment_row  integer;
         BEGIN
-            RAISE NOTICE '.. Step 2 of 4: Update permit info';
+            RAISE NOTICE '.. Step 2 of 4: Update permit info in MDS';
```
itchison: We have a large change to permit and permittee - I think you should exclude this whole file from your PR until that work goes in.

FreshGyan: @itchison Done :D

FreshGyan: I reverted my changes but the file is still showing due to some non-harmful spacing

PR #820, opened 2019-06-20
Reduce Dev resources

FreshGyan:
Started off by reducing the DEV PR resources to allocate more PRs and faster deploys. Did some other misc updates. I was able to go through and do everything without waiting too long, but would appreciate manual testing/input

- Reduced the amount of CPU/Memory allocated for each envt. 
- Made Unit test pods customizable for less resource consumption
- Disabled APM for both backends on DEV to free up worker threads and the fact that no one looks at that data and we're just flooding the APM for nothing
- Installed pm2 for better process management and improved performance for FEs. pm2 is to node what uwsgi is to flask (We can probably customize it and play further as needed)
- Updated the Functional Test base image latest to have newer firefox/chrome etc
- Made functional tests a bit more *patient*
- Updated gradle wrapper version to get relatively faster builds and smaller jvm print
- Adjusted bddstack build and runtime pod for better performance
- Made threads and processes for UWSGI customizable so we can play with that more in later iterations.
- Fixed APM init code block


pipeline/config-prod.groovy
```diff
@@ -301,6 +305,8 @@ environments {
                     cpu_limit = "400m"
                     memory_request = "1.5Gi"
                     memory_limit = "3Gi"
+                    uwsgi_threads = 2
+                    uwsgi_processes = 4
                     replica_min = 2
                     replica_max = 4
```
Jsyro: how do we feel about more smaller pods? like 3-6(0r 8) instead of 2-4? with less resources each. Celcio made a good point that we can't really say HA unless we have 3, cause we might lose one pod at any given time.

FreshGyan: That's a good idea! something we can move towards possibly 🤷‍♂ Definitely has its pros/cons and needs some manual testing. I would argue that having 2 pods still makes you HA just as much as 3 would but im sure he had a good argument as well.
Also all our deploys and such will take longer the more pods we throw on there 🤔 

I didnt alter any test/prod resources in this PR just to keep the scope limited but maybe we can look into that more?

PR #824, opened 2019-06-21
Bug/mds 2037

thegentlemanphysicist:
fix the clipping on the party modal


frontend/src/components/Forms/parties/AddFullPartyForm.js
```diff
@@ -26,7 +26,9 @@ export const AddFullPartyForm = (props) => (
             </Radio.Group>
           </div>
           <Row gutter={16}>
-            <h5>Basic Details</h5>
+            <Col md={12} sm={12} xs={24}>
```
mmumby: if `md` and `sm` are the same values you can remove `sm`

PR #830, opened 2019-06-25
Bug/mds 1314

thegentlemanphysicist:
This fixes the same day party appointment bug.  The party role is now displayed on the day it expires, and a user can assign a party for a single day.
-Can update an assigned role date to start and end on the same date.
-Tiles displayed as active until end of day of expiration
-Correct permittee is displayed in the mine summary tab (permits section)
-On permits tab, only the active permittee is displayed on the permit table
BONUS!
This also fixes MDS-1080 the email verification now accepts any string  that has one or more characters before and after an @ symbol.


frontend/src/components/mine/ContactInfo/ViewPartyRelationships.js
```diff
@@ -184,10 +184,17 @@ export class ViewPartyRelationships extends Component {
     });
   };
 
+  // Since end date is stored at yyyy-mm-dd, comparing current Date() to
+  // the the start of the next day ensures appointments ending today are displayed.
+  getTomorrow = (end_date) => {
```
NWCalvank: Small naming thing, but this is really more of a `getNextDay` since it accepts a date and gives the day after that.

frontend/src/components/mine/Summary/MineSummary.js
```diff
@@ -70,8 +70,15 @@ const renderSummaryTSF = (tsf, partyRelationships) => (
...
+// Since end date is stored at yyyy-mm-dd, comparing current Date() to
+// the the start of the next day ensures appointments ending today are displayed.
+const getTomorrow = (end_date) => {
+  const dayAfterEndDate = new Date(end_date);
+  return dayAfterEndDate.setDate(dayAfterEndDate.getDate() + 1);
+};
+
 const isActive = (pr) =>
-  (!pr.end_date || Date.parse(pr.end_date) >= new Date()) &&
+  (!pr.end_date || getTomorrow(pr.end_date) > new Date()) &&
```
NWCalvank: Instead of using this custom function, we can likely leverage the existing `momentjs` library that we're using for something like: `moment(x.end_date).add(1, 'days') > new Date()`

Thoughts?

Docs:
https://momentjs.com/docs/#/parsing/string/
https://momentjs.com/docs/#/manipulating/add/

frontend/src/components/mine/ContactInfo/ViewPartyRelationships.js
thegentlemanphysicist: good point

frontend/src/components/mine/Summary/MineSummary.js
nayr974: I agree 👍 I was looking at this before lunch and was going to comment that getTomorrow seemed awkward.

thegentlemanphysicist: implemented with your suggestions

PR #831, opened 2019-06-25
Try new base image for Functional Tests

FreshGyan:


openshift/bddstack.bc.json
```diff
@@ -91,8 +121,7 @@
           "dockerStrategy": {
             "from": {
               "kind": "ImageStreamTag",
-              "namespace": "bcgov",
-              "name": "jenkins-slave-bddstack:v1-latest"
+              "name": "selenium-firefox:latest"
```
thegentlemanphysicist: why is this selenium now?

FreshGyan: Because instead of bcgov, we are pulling the firefox pre installed image published to DockerHub by selenium folks

ShellyXueHan: Dope! Once you have tested the image, we could start to push it to the BDDStack imagestream.

PR #833, opened 2019-06-25
MDS-1789 Task/ Remove 0 lat/lons from ETL process

NWCalvank:
Changes defaults from 0 to `NULL` when:
- Updating mine records in ETL_MINE
- Inserting new mine records into ETL_MINE
- Updating ETL_LOCATION from multiple sources
- Inserting into ETL_LOCATION from multiple sources

Keep in mind:
- ETL_MINE is the final fallback for lat/lon data when inserting into ETL_LOCATION
- ETL_LOCATION is used to populate `mine.latitude` and `mine.longitude`, so the final value that is upserted into ETL_LOCATION is the one that will reach CORE

### DIFF
--- a/migrations/sql/R__stored_procedure_for_mine_information.sql
+++ b/migrations/sql/R__stored_procedure_for_mine_information.sql
@@ -131,6 +131,7 @@ CREATE OR REPLACE FUNCTION transfer_mine_information() RETURNS void AS $$
             old_row         integer;
             new_row         integer;
             update_row      integer;
+            delete_row      integer;
         BEGIN
             RAISE NOTICE '.. Step 2 of 5: Update mine in MDS';
             SELECT count(*) FROM mine into old_row;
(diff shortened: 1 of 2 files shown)


NWCalvank:
I forgot to address a `NaN` issue on the FE now that `null` values could be returned. This is not ready to merge yet. A fix will be pushed to this PR soon

migrations/sql/R__stored_procedure_for_mine_information.sql
```diff
@@ -532,8 +556,18 @@ CREATE OR REPLACE FUNCTION transfer_mine_information() RETURNS void AS $$
...
             FROM new_record new;
             SELECT count(*) FROM mine_location into new_row;
+
+            -- Remove invalid rows that were updated or inserted
+            WITH deleted_rows AS (
+                DELETE FROM mine_location
+                WHERE latitude IS NULL OR longitude IS NULL
+                RETURNING 1
+            )
+            SELECT COUNT(*) FROM deleted_rows INTO delete_row;
```
NWCalvank: Backwards-compatibility: The API-level validation enforces that a `mine_location` record must have both a `latitude` and `longitude` value. If either value is missing, the POST is rejected. If neither value is provided, the record is not created.

To stay consistent with this behaviour, I am removing any records that were created with `null` lat and/or long values or were updated to include a `null` lat/long value.

FreshGyan: Hmm 🤔 Interesting point on enforcing API rules at the ETL level as well.
Currently if I run 
```
SELECT FROM mine_location
WHERE latitude IS NULL OR longitude IS NULL
```
I dont get anything but I dont see this causing any harm either

NWCalvank: Explained IRL.

For future reference:
This is consistency with our apps expectations, not with existing data. The app can't handle `null` lat/long values. The lat & long must both be present or the `mine_location` record is not created.

PR #836, opened 2019-06-26
Make Functional Tests patient

FreshGyan:


functional-tests/src/test/resources/GebConfig.groovy
```diff
@@ -154,8 +154,8 @@ reportOnTestFailureOnly = false //true
 //in seconds
 //If unspecified, the values of 5 for timeout and 0.1 for retryInterval.
 waiting {
-    timeout = 120
-    retryInterval = 1
+    timeout = 600
```
Jsyro: i feel like this is addressing a symptom and not a cause. Is this going to make them more likely to pass? Just wondering if the loading spinner is infinite or something else weird in the headless browser.

PR #839, opened 2019-06-27
Dependency Audit and Upgrade

pdittaro:
Upgrade of all dependencies in python-backend, nris-backend, frontend, and frontend-public.


frontend-public/webpack.parts.js
```diff
@@ -1,11 +1,11 @@
 const webpack = require("webpack");
 const autoprefixer = require("autoprefixer");
-const cssnano = require("cssnano")({ zindex: false });
```
pdittaro: @FreshGyan , did you implement this `zindex: false`? It was causing errors with webpack plugins.

frontend-public/webpack.parts.js
```diff
@@ -1,11 +1,11 @@
...
+const cssnano = require("cssnano");
 
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
-const UglifyWebpackPlugin = require("uglifyjs-webpack-plugin");
-const CleanWebpackPlugin = require("clean-webpack-plugin");
+const TerserPlugin = require("terser-webpack-plugin");
```
pdittaro: Switched from Uglify to Terser plugin. Mainly based on recommendations. Terser has support for ES6 as well, which is where I was running into webpack pipeline errors during the upgrade.

frontend/src/actionCreators/staticContentActionCreator.js
```diff
@@ -114,7 +114,7 @@ export const fetchMineIncidentFollowActionOptions = () => (dispatch) => {
       dispatch(success(reducerTypes.GET_MINE_INCIDENT_FOLLOWUP_ACTION_OPTIONS));
       dispatch(staticContentActions.storeMineIncidentFollowActionOptions(response.data));
     })
-    .catch(() => dispatch(error(reducerTypes.GET_MINE_INCIDENT_FOLLOW_ACTION_OPTIONS)));
+    .catch(() => dispatch(error(reducerTypes.GET_MINE_INCIDENT_FOLLOWUP_ACTION_OPTIONS)));
```
pdittaro: One thing I noticed is that now - after the updates - the compiler gets a lot more stringent on imports. Found a fair few typos that I fixed as part of this.

pdittaro:
Not sure what happened to the snapshots, I'll take a look again.

pdittaro:
Fixed Snapshots using enzyme serializer (https://github.com/facebook/jest/issues/7802). Trims down the info in the snapshots, but I think it makes them more legible and informative.

pdittaro:
@NWCalvank , after this we're left with 4 moderate vulnerabilities - but they're all in devDependencies so I don't think we need to worry about it.

frontend-public/webpack.parts.js
FreshGyan: Nope! It came with the boilerplate we were using, if its not making our css files sad. then im good

frontend/src/index.js
```diff
@@ -1,4 +1,3 @@
-import "babel-polyfill";
```
FreshGyan: That's for IE compatibility.  Is that not needed anymore?

frontend-public/webpack.parts.js
NWCalvank: I made this change in `/frontend`. It was blindly copied over to `/frontend-public`, but I see you've removed it from there as well?

It was done to address a bug where the loading bar went over-top of the toast message. This was caused by a bug in one of our libraries. Perhaps your upgrades have removed this bug? If not, please verify that you're not reintroducing the bug.

The original PR: https://github.com/bcgov/mds/pull/317

frontend-public/.babelrc
```diff
@@ -1,12 +1,13 @@
 {
-  "presets": ["env", "react", "stage-2"],
+  "presets": ["@babel/preset-env", "@babel/preset-react"],
   "plugins": [
     [
-      "transform-runtime",
+      "@babel/transform-runtime",
```
NWCalvank: Is the `@babel/` prefix just added here for verbosity? If memory serves, it's assumed and not needed

frontend-public/src/tests/HOC/__snapshots__/authenticationGuard.spec.js.snap
```diff
@@ -1,64 +1,8 @@
...
-      },
-    },
-    "attachTo": undefined,
-    "hydrateIn": undefined,
-  },
-}
+<Component
+  getUserInfoFromToken={[MockFunction]}
+  isAuthenticated={true}
+/>
```
NWCalvank: I'm concerned by the lack of content in this snap. It has almost no details about the state of the component

frontend-public/src/utils/customAxios.js
```diff
@@ -5,6 +5,10 @@ import * as Strings from "@/constants/strings";
 const UNAUTHORIZED = 401;
 const MAINTENANCE = 503;
 
+// https://stackoverflow.com/questions/39696007/axios-with-promise-prototype-finally-doesnt-work
+const promiseFinally = require("promise.prototype.finally");
+promiseFinally.shim();
```
NWCalvank: Why do we need this? The link is from 2016. The first example in Axios v0.19 docs use `finally`, so it should be fully-supported

frontend/src/HOC/AuthenticationGuard.js
```diff
@@ -39,21 +39,18 @@ export const AuthenticationGuard = (WrappedComponent) => {
     async keycloakInit() {
       // Initialize client
       const keycloak = Keycloak(KEYCLOAK);
-      await keycloak.init();
-
-      // Prompt for login using IDIR if not authenticated
-      if (!keycloak.authenticated) {
-        await keycloak.login({
+      await keycloak
```
NWCalvank: Doesn't look like the `await` here does anything 😄

frontend/webpack.config.js
```diff
@@ -33,7 +33,7 @@ const BUILD_FILE_NAMES = {
 
 const PATH_ALIASES = {
   "@": PATHS.src,
-  // Put your aliases here
+  "react-dom": "@hot-loader/react-dom",
```
NWCalvank: This is weird. It's definitely needed? What does it do?

frontend-public/webpack.parts.js
FreshGyan: Curious if u checked out the parallel plugin on it
https://www.npmjs.com/package/terser-webpack-plugin#parallel

frontend-public/webpack.parts.js
pdittaro: Thanks Nathan, I'll confirm whether the behaviour has been addressed.

frontend-public/.babelrc
pdittaro: I think this is an artifact of change with babel v7. The documentation still says that only presets prefixed with `babel-preset-` can be shorthanded. No mentions of `@babel/preset-`

frontend/webpack.config.js
pdittaro: Agreed, this is weird. This was to address this [react-hot-loader issue](https://github.com/gaearon/react-hot-loader/issues/1227) which was showing up in my console. I didn't think the PATH_ALIASES was the best fit, but I wasn't keen on creating a NON_PATH_ALIASES.

frontend-public/src/tests/HOC/__snapshots__/authenticationGuard.spec.js.snap
pdittaro: Were we using snapshots to describe the state of the component? I thought the intent was to ensure that any potential UI changes were identified. In that sense, wouldn't state be an implementation detail tested through unit and integration tests? Also, for this specific instance - isn't the state just the internals of the Enzyme renderer, rather than anything to do with the component?

I was tracing through the snapshots because of this [Jest issue](https://github.com/facebook/jest/issues/7802), wherein someone raises a similar issue about snapshotting the private methods of the Enzyme library. One of the jest contributors then suggests needing a proper serializer to flatten the component structure, ala enzyme-to-json. Though I'm not sure of the configurations of the enzyme-to-json, it may be able to provide us a bit more information.

frontend-public/src/utils/customAxios.js
pdittaro: Without this all of the unit tests failed for all of the action creators. Maybe we can raise this as a separate issue?

frontend/src/HOC/AuthenticationGuard.js
pdittaro: I'm inclined to agree with you :P. I'll remove it.

frontend-public/webpack.parts.js
pdittaro: Not yet, I'll take a look.

frontend/src/index.js
pdittaro: It looks like IE compatability is broken at the moment. I'll see if I can get it working again.

FreshGyan: Before you dive too deep into that, i'd recommend asking the business if they actually care.
And if it takes too long, focus on fixing everything else with this PR and ship it as long as it works in every other browser 👌 
Then look at fixing IE

frontend-public/src/tests/HOC/__snapshots__/authenticationGuard.spec.js.snap
NWCalvank: Apologies, I shouldn't have used the term "state" as what I really meant was the values of all values in the component based on the props that it received. We provide props when generating the snapshots.

Regardless, I think you're right anyway. I didn't look carefully enough at 1) Which component snap I was comparing -- authenticationGuard has very little content and 2) What the majority of the diff removal was showing.

It looks like the props are being properly passed into the snapshots. I'll mark this as resolved

frontend/webpack.config.js
NWCalvank: That link was really helpful, thanks.

Since this patch is a bit different than our regular path aliases, you could do something like this on line 72:
```javascript
resolve: {
      alias: {
         "react-dom": "@hot-loader/react-dom", // patch react-dom import
         ...PATH_ALIASES
      },
    },
```

That just keeps it out of our other aliases. But they do the same thing in the end, so not a big deal to me either way.

pdittaro: Sure, that makes sense - I'll move it out of the way.

frontend/src/HOC/AuthenticationGuard.js
pdittaro: To follow-up, removing an await within an async shouldn't change the behaviour...but it caused the application to get caught in a login loop. Changing the keycloakInit to a synchronous function fixes it, but I'm not sure if thats the behaviour we want.

frontend/src/index.js
pdittaro: If it was a somewhat graceful break I would agree, but this is just a whitescreen load. If I can't fix it I'll try and shim a compatibility error message in front. I asked the PO about it and he said to hold off bringing it in until we addressed it.

frontend-public/webpack.parts.js
pdittaro: Belay that, the options for Terser were made to be compatible from Uglify; I had copied over the parallel option into the configuration during first migration.

frontend/src/HOC/AuthenticationGuard.js
NWCalvank: Promises are non-blocking, so a "non-async" function is totally fine here, if that's your concern

frontend-public/src/utils/customAxios.js
NWCalvank: Oh, weird... Yeah, it can likely just be a separate issue. I don't think it's worth it to block the rest of these changes

frontend-public/webpack.parts.js
pdittaro: I can't replicate this issue on my machine.

NWCalvank: Okay. So long as it's not happening, I'm happy. Marking as resolved.

frontend-public/src/tests/components/__snapshots__/ReturnPage.spec.js.snap
```diff
@@ -1,78 +1,8 @@
...
-        ],
-      }
-    }
-    location={
-      Object {
-        "search": "type=login&code=blah",
-      }
-    }
-    redirect="/"
-    unAuthenticateUser={[MockFunction]}
```
NWCalvank: It does look like some of these have been lost. 🤔 There's no longer a reference to `unAuthenticateUser` nor `authenticateUser`, which are provided in the spec as dispatch props.

That may be acceptable, as dispatched actions are probably not coupled tightly enough to the view layer for us to care about them in our snapshots... But it makes me think that there may be no benefit to passing them in when we generate the snapshot. If removing the dispatch function mocks returns a runtime exception during test execution then I'm not sure how to feel.

Thoughts?

NWCalvank:
Once this is definitely code-complete, I'll do some manual testing. We may want to pull in @thegentlemanphysicist as well since this is touching the entire app. The code changes all look reasonable to me 👍 

I just have some uncertainty around the snapshots, but I think they're probably fine.

NWCalvank:
Bug report:
Clicking "Add permit amendment" causes the screen to just go white. The runtime error is `Cannot read property 'filter' of undefined`, which appears to occur at `PermitAmendmentForm.js:36`

pdittaro:
> Clicking "Add permit amendment" causes the screen to just go white. 

@NWCalvank  I think that this bug might have been fixed from develop, I can't replicate now.

pdittaro:
Refactored the date validator becuase I was running into recurring inconsistent behaviour with sorting null values in the original impl.

pdittaro:
Some of the icons are stretching in a weird way. The height and width css attributes don't match. I think previously when they were being served they were being served in such a way that they would not stretch, but now they are stretching. I'll investigate.
image

pdittaro:
Reopening to clear state

frontend-public/src/tests/components/__snapshots__/ReturnPage.spec.js.snap
pdittaro: I think the end result is the point though? The parameters that we pass in the test should result in a redirect to the root - at least as how I understand this test. And the snapshot should capture how that 'Redirect' component is rendered to the client.

If we wanted to capture the various 'outcome renders' that could occur with a component given different properties, then we could create seperate snapshot tests given seperate mocked properties for each case?

NWCalvank:
**Bug report**
While a modal is submitting, the purple loading bar is in the wrong spot. It is about halfway down the screen and overlaps with the modal itself. It should be at the top, under the nav bar.

FreshGyan:
@NWCalvank Totally agreed! If this is the only issue, I vote we ship this off and deal with the bug in a follow up task.
Its mostly a dev issue at the moment due to slower loading, but ideally in TEST/PROD it should not happen given we have higher resources and such.

I will however try to ship off a fix by EOD if I can

PR #840, opened 2019-06-28
Tweak FT's even more

FreshGyan:
- Turned off XVFB as it's not needed anymore
- Updated geb version
- Cleaned up dockerfile
- Enabled gradle daemon for faster builds
- Switched order that the tests run in.  i.e. Build and Run Minespace first! So the gradle daemon can intialize with a small overhead and then quickly run through Core tests
- Reduced build resources as needed
- Added a maximize option to the firefox driver (I dont know why but it makes them more stable)

~Ran tests 10 times in a row, one random failure, 9 blazing successes~ 10/10 ✅ 

Out of the 40-50ish tests, I ran! The TSF test failed twice :( Which I wasn't sure why but the rest of the runs worked great.

### DIFF
--- a/functional-tests/src/test/resources/GebConfig.groovy
+++ b/functional-tests/src/test/resources/GebConfig.groovy
@@ -90,19 +90,22 @@ environments {
 			profile.setPreference("browser.download.manager.showAlertOnComplete", false);
 			profile.setPreference("browser.download.manager.closeWhenDone", false);
 			profile.setPreference("browser.download.panel.shown", false);
+            options.setProfile(profile);
 
 			//To find the settings that need to be modified, visit the about:config firefox url
 			options.addPreference("browser.download.forbid_open_with", true);
 			//In order to allow auto saving of files the user must specify the specific MIME types in a comma separated
 			//string. (Cannot set all to true)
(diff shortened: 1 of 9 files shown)


functional-tests/src/test/resources/GebConfig.groovy
```diff
@@ -90,19 +90,22 @@ environments {
...
-			options.addArguments("--height=1080")
-			options.addArguments("--headless")
-			new FirefoxDriver(options);
+			options.addArguments("--width=1920");
+			options.addArguments("--height=1080");
+			options.addArguments("--headless");
+
+            FirefoxDriver firefoxDriver = new FirefoxDriver(options);
+            firefoxDriver.manage().window().maximize();
+            firefoxDriver;
```
NWCalvank: Looks like vertical alignment might be off here?

FreshGyan: I saw that but My IDE disagrees. I tried doing all sorts of re-tabbing, edit lines and everything but it's not helping

FreshGyan: Not sure if I need to force format something and how to do so 😅

PR #843, opened 2019-06-28
mds 2089 migrations and models for mine reports

JustinMacaulay:
Added a migration to add tables for code required reports.
Added SQL alchemy models for the tables added.


migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+    create_timestamp                  timestamp with time zone DEFAULT now()  NOT NULL   ,
+    update_user                       character varying(60)                   NOT NULL   ,
+    update_timestamp                  timestamp with time zone DEFAULT now()  NOT NULL
+);
+
+COMMENT ON TABLE mine_report_due_date_type is 'Types of due dates available for mine reports.';
+ALTER TABLE mine_report_due_date_type OWNER TO mds;
+
+CREATE TABLE mine_report_submission_status_code (
+    mine_report_submission_status_code  character varying(3)   PRIMARY KEY DEFAULT 'MIA',
```
klefler: Apologies if this is a dumb question, but should there be a default value on a Primary Key?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+    update_timestamp                    timestamp with time zone DEFAULT now()  NOT NULL
+);
+
+COMMENT ON TABLE mine_report_submission_status_code is 'All the possible available status codes for mine reports.';
+ALTER TABLE mine_report_submission_status_code OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_definition (
+    mine_report_definition_id            serial                                  PRIMARY KEY, 
+    mine_report_definition_guid          uuid                                    NOT NULL   ,
+    name                                 varchar(50)                             NOT NULL   ,
```
klefler: May want to consider changing this column name to something other than "name", i.e. report_name or something similar.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+    update_user                  character varying(60)                    NOT NULL            ,
+    update_timestamp             timestamp with time zone DEFAULT now()   NOT NULL            ,
+    
+    FOREIGN KEY (mine_report_definition_id) REFERENCES mine_report_definition(mine_report_definition_id) DEFERRABLE INITIALLY DEFERRED,
+    FOREIGN KEY (mine_guid) REFERENCES mine(mine_guid) DEFERRABLE INITIALLY DEFERRED,
+    FOREIGN KEY (permit_id) REFERENCES permit(permit_id) DEFERRABLE INITIALLY DEFERRED
+
+);
+
+COMMENT ON TABLE mine_report is 'A code required report for a specific mine or permit on a mine.';
```
klefler: I'm wondering if we should be more specific about what this table is... I am still a little foggy on it... but is it essentially a list of mandatory reports that must be submitted by a mine as required by the Health, Safety and Reclamation Code? We've been using the term "code required reports" and I'm thinking we should define, somewhere, what this means is a more non-business-jargon way...

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+ALTER TABLE mine_report OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_submission (
+    mine_report_submission_id             serial                                  NOT NULL PRIMARY KEY,
+    mine_report_submission_guid           uuid                                    NOT NULL            ,
+    mine_report_id                        integer                                 NOT NULL            ,
+    received_date                         timestamp                                                   ,
+    mine_report_submission_status_code    varchar(3)                              NOT NULL            ,
+    active_ind                            boolean DEFAULT true                    NOT NULL            ,
```
klefler: Not sure I understand the use of an "active" submission. Can a submission expire? Or is this covering a soft-delete of a report that has been submitted?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+    FOREIGN KEY (mine_report_id) REFERENCES mine_report(mine_report_id) DEFERRABLE INITIALLY DEFERRED ,
+    FOREIGN KEY (mine_report_submission_status_code) REFERENCES mine_report_submission_status_code(mine_report_submission_status_code) DEFERRABLE INITIALLY DEFERRED
+);
+
+COMMENT ON TABLE mine_report_submission is 'A file submission for a code required report.';
+ALTER TABLE mine_report_submission OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_document_xref
+(
+    mine_report_document_xref_guid uuid    DEFAULT gen_random_uuid() NOT NULL PRIMARY KEY,
```
klefler: We may not need a surrogate primary key here... the combination of the mine_document_guid + mine_report_submission_id should be enough to keep the record unique and can be used as the primary key.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+);
+
+COMMENT ON TABLE mine_report is 'A code required report for a specific mine or permit on a mine.';
+ALTER TABLE mine_report OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_submission (
+    mine_report_submission_id             serial                                  NOT NULL PRIMARY KEY,
+    mine_report_submission_guid           uuid                                    NOT NULL            ,
+    mine_report_id                        integer                                 NOT NULL            ,
+    received_date                         timestamp                                                   ,
```
klefler: We may want to describe what "received date" means, in a column comment. Is it the date the submission was made? Or when the ministry received the submission? Or are those one in the same?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+
+COMMENT ON TABLE mine_report_document_xref is 'Links a mine document to report submission.';
+ALTER TABLE mine_report_document_xref OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_comment (
+    mine_report_comment_id           serial                                  NOT NULL PRIMARY KEY,
+    mine_report_comment_guid         uuid                                    NOT NULL            ,
+    mine_report_id                   integer                                 NOT NULL            ,
+    mine_report_submission_id        integer                                                     ,
+    comment                          varchar(300)                            NOT NULL            ,
```
klefler: "comment" may be a reserved word in the database... should be changed to "something"_comment. I would suggest a name but I'm not sure if this is a mine_report comment or a report_submission_comment.... if it's both, what would be the generic name... also, if this is a mutually exclusive relationship, we may want a database constraint to allow a record to tie to either the mine_report OR the mine_report_submission, but not both. If that's the case.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
Jsyro: @JustinMacaulay Should this be in the backend not in the db?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
Jsyro: should this be deleted_ind?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+    description              character varying(100)                  NOT NULL   ,
+    display_order            smallint                                NOT NULL   ,
+    active_ind               boolean DEFAULT true                    NOT NULL   ,
+    create_user              character varying(60)                   NOT NULL   ,
+    create_timestamp         timestamp with time zone DEFAULT now()  NOT NULL   ,
+    update_user              character varying(60)                   NOT NULL   ,
+    update_timestamp         timestamp with time zone DEFAULT now()  NOT NULL
+);
+
+COMMENT ON TABLE mine_report_category is 'Categories available for mine reports..';
```
Jsyro: two periods, should this be more descriptive that this isn't in the code, just friendly groupings that some business logic/filtering might leverage.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+ALTER TABLE mine_report_category OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_category_xref (
+
+    mine_report_category_xref_guid  uuid DEFAULT gen_random_uuid()    NOT NULL PRIMARY KEY,
+    mine_report_definition_id       integer                           NOT NULL            ,
+    mine_report_category            varchar(3)                        NOT NULL            ,
```
Jsyro: does this need it's own PK? or just the two FK's?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+
+CREATE TABLE IF NOT EXISTS mine_report_comment (
+    mine_report_comment_id           serial                                  NOT NULL PRIMARY KEY,
+    mine_report_comment_guid         uuid                                    NOT NULL            ,
+    mine_report_id                   integer                                 NOT NULL            ,
+    mine_report_submission_id        integer                                                     ,
+    comment                          varchar(300)                            NOT NULL            ,
+    minespace_user_id                integer                                                     ,
+    core_user_id                     integer                                                     ,
+    comment_visibility               boolean                                 NOT NULL            ,
```
klefler: May want to name this column "comment_visible_ind". The "ind" keeps the boolean type consistent with most other booleans in the model. And because it sounds like there is some business logic applied to this column, there should probably be a column comment to describe what this column means.

klefler: Also may want a default value.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+ALTER TABLE mine_report_document_xref OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_comment (
+    mine_report_comment_id           serial                                  NOT NULL PRIMARY KEY,
+    mine_report_comment_guid         uuid                                    NOT NULL            ,
+    mine_report_id                   integer                                 NOT NULL            ,
+    mine_report_submission_id        integer                                                     ,
+    comment                          varchar(300)                            NOT NULL            ,
+    minespace_user_id                integer                                                     ,
+    core_user_id                     integer                                                     ,
```
klefler: re: minespace_user_id and core_user_id.... can a comment be related to both at the same time, or just one? I assume one comment record is related to one user type, and any additional commenting would be an additional record, which would be related to only one type of user? If so, we may want to constrain this in the database, to allow a single comment to be tied to either a minespace user OR a core user, but not both.

Unless I am misunderstanding this.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: The logic for this is that there can be several submissions for a report and only the most recent is the active submission. You don't want to lose the history or the previous submissions though.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,154 @@
...
+
+CREATE TABLE IF NOT EXISTS mine_report
+(
+    mine_report_id               serial                                   NOT NULL PRIMARY KEY,
+    mine_report_guid             uuid                                     NOT NULL            ,
+    mine_report_definition_id    integer                                  NOT NULL            ,
+    mine_guid                    uuid                                     NOT NULL            ,
+    permit_id                    integer                                                      ,
+    due_date                     timestamp                                NOT NULL            ,
+    submission_year              integer                                                      ,
```
klefler: Will this column be sized to 4 characters? Or should it be? Is this a free form field? Will there be any specific reporting off this column?   ....wondering if this column should be driven by a lookup table, or find a way to ensure the values are consistently entered.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
klefler: Can you please provide a couple of examples of "Categories" in the table comment?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: removed the default.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
klefler: Could this be managed through the submission_status_code?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: The way this was mad is that this is the date the submission was made in the system.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: Changed to report_name

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: This is used on one other table which is where I got this and it has no limits at the moment. I think we will need some good backend validation for it.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: I tried to make the comment more specific or more understandable.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: Jason S and I had talked about this briefly and looking at our previous xref tables they all seem to use a separate primary key. I am all for using the combo as the PK but do we want to keep it consistent?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: So the way this was thought up is that the comment will always link to at least a mine report. It can sometimes be linked to a mine_report submission and if it is linked to a submission than this comment should only show up on that submission and not the report.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: changed and added a default.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: In the comment or here? This table is more for the business to tag reports such as geotechnical, health and safety and so on.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
klefler: I took a quick look at the other xref tables and there is a bit of a mix... I am all for consistency and probably missed the boat with the other xref tables, although some of them are good candidates for a surrogate key (and probably not "true" xref tables). I'm happy with whatever you guys decide.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
klefler: I think just adding some examples within the table description...

"Categories available for mine reports, which are used to tag the reports with business terms. For example, Geotechnical, Health and Safety, etc."

...something like this.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: This was removed and planned to be handled through submission date.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: I added a constraint for this table to ensure that only one of these FK's can be populated at a time.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
JustinMacaulay: I added those to the comment.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,172 @@
...
+    description                         character varying(100)                  NOT NULL,
+    display_order                       smallint                                        ,
+    active_ind                          boolean DEFAULT true                    NOT NULL,
+    create_user                         character varying(60)                   NOT NULL,
+    create_timestamp                    timestamp with time zone DEFAULT now()  NOT NULL,
+    update_user                         character varying(60)                   NOT NULL,
+    update_timestamp                    timestamp with time zone DEFAULT now()  NOT NULL
+);
+
+COMMENT ON TABLE mine_report_submission_status_code is 'All the possible available status codes for mine reports.';
```
Jsyro: 'possible' and 'available'... not a blocker.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,173 @@
+CREATE TABLE mine_report_due_date_type (
```
itchison: @klefler - I don't think we should tackle this on this PR. However, I see some inconsistencies with "type" tables vs "code" tables. I also see this propagated to the PK column.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,173 @@
...
+    create_user                         character varying(60)                   NOT NULL,
+    create_timestamp                    timestamp with time zone DEFAULT now()  NOT NULL,
+    update_user                         character varying(60)                   NOT NULL,
+    update_timestamp                    timestamp with time zone DEFAULT now()  NOT NULL
+);
+
+COMMENT ON TABLE mine_report_submission_status_code is 'All the possible and available status codes for mine reports.';
+ALTER TABLE mine_report_submission_status_code OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS mine_report_definition (
```
itchison: Could this table exist?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
```diff
@@ -0,0 +1,173 @@
...
+CREATE TABLE IF NOT EXISTS mine_report
+(
+    mine_report_id               serial                                   NOT NULL PRIMARY KEY,
+    mine_report_guid             uuid UNIQUE                              NOT NULL            ,
+    mine_report_definition_id    integer                                  NOT NULL            ,
+    mine_guid                    uuid                                     NOT NULL            ,
+    permit_id                    integer                                                      ,
+    due_date                     timestamp                                NOT NULL            ,
+    received_date                timestamp                                                    ,
+    submission_year              smallint                                                     ,
```
itchison: Most submissions are done on fiscal year are they not? Should this be character so that "18/19" is accepted? Or does "19" imply "18/19"?

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
itchison: I dislike compounded keys. Highly prefer surrogates. Willing to talk offline about it.

migrations/sql/V2019.06.26.10.30__add_code_required_report_tables.sql
alaigov: We've talked about how this will be tricky because it's really a compliance _period,_ which may be a calendar year, or a rolling 5 year period. But I think most often the compliance year is a calendar year (eg TSF reports, though due on Mar 31 for some random reason, apply to the previous calendar year - if an event happens on Jan 30, it will be covered in the TSF report due Mar 31 of the following year.)

PR #844, opened 2019-07-02
Fixed broken date sorting on the front end

thegentlemanphysicist:
This fixes the date sorting in the Variance, Incident, and Compliance tables.
The compliance sorting must be tested on Test since that tab is not populated locally.
This is a purely frontend solution since none of these tables seem to be paginated.  
The strategy was to use moment on every date based sort function in the front end.

### DIFF
--- a/frontend/src/components/mine/Compliance/ComplianceOrdersTable.js
+++ b/frontend/src/components/mine/Compliance/ComplianceOrdersTable.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { Table } from "antd";
-
+import moment from "moment";
 import { RED_CLOCK } from "@/constants/assets";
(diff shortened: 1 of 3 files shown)


NWCalvank:
I'm pretty sure this was meant to address sorting on all columns, not just dates. I seem to remember there being an issue with sorting by strings that have decimals (code sections, maybe?)

thegentlemanphysicist:
> I'm pretty sure this was meant to address sorting on all columns, not just dates. I seem to remember there being an issue with sorting by strings that have decimals (code sections, maybe?)

You may be right... The date search was the broken one, while the code sorting wasn't implemented at all.  I will check with Aaron after lunch.

PR #846, opened 2019-07-03
Task / Permit Resource Refactor

nayr974:


python-backend/app/api/permits/namespace/permits.py
```diff
@@ -1,24 +1,26 @@
...
-                 '/amendments/<string:permit_amendment_guid>',
+
+api.add_resource(PermitAmendmentListResource, '/<string:permit_guid>/amendments')
+api.add_resource(PermitAmendmentResource,
                  '/<string:permit_guid>/amendments/<string:permit_amendment_guid>')
 
 api.add_resource(
-    PermitAmendmentDocumentResource,
+    PermitAmendmentDocumentListResource,
     '/amendments/documents',
```
Jsyro: is this how we want to have paths available to upload files for objects no yet created?

python-backend/app/api/permits/permit/models/permit.py
```diff
@@ -17,9 +17,7 @@ class Permit(AuditMixin, Base):
     permit_status_code = db.Column(
         db.String(2), db.ForeignKey('permit_status_code.permit_status_code'))
     permit_status_code_relationship = db.relationship(
-        'PermitStatusCode',
-        foreign_keys=[permit_status_code],
-        lazy='select')
+        'PermitStatusCode', foreign_keys=[permit_status_code], lazy='select')
```
Jsyro: didn't know about this foreign_keys attribute, do we want to use this going forward?

python-backend/app/api/permits/namespace/permits.py
```diff
@@ -1,24 +1,26 @@
...
 
-from ..permit.resources.permit import PermitResource
+from ..permit.resources.permit import PermitResource, PermitListResource
 from ..permit.resources.permit_status_code import PermitStatusCodeResource
-from ..permit_amendment.resources.permit_amendment import PermitAmendmentResource
-from ..permit_amendment.resources.permit_amendment_document import PermitAmendmentDocumentResource
+from ..permit_amendment.resources.permit_amendment import PermitAmendmentResource, PermitAmendmentListResource
+from ..permit_amendment.resources.permit_amendment_document import PermitAmendmentDocumentListResource, PermitAmendmentDocumentResource
 
 api = Namespace('permits', description='Permit related operations')
```
Jsyro: permits should be moved under mines.... this might be the time to move the routes. we can move the files on sprint planning day.

python-backend/app/api/permits/permit/resources/permit.py
```diff
@@ -118,24 +98,73 @@ def post(self, permit_guid=None):
...
+    def get(self, permit_guid):
+        permit = Permit.find_by_permit_guid_or_no(permit_guid)
+        if not permit:
+            raise NotFound('Permit not found.')
+        return permit
 
+    @api.doc(params={'permit_guid': 'Permit guid.'})
+    @requires_role_mine_create
+    @api.marshal_with(PERMIT_MODEL, code=200)
+    def put(self, permit_guid):
```
Jsyro: when this path changes, mine_guid will need to be checked against the permit

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
```diff
@@ -116,39 +86,78 @@ def post(self, permit_guid=None, permit_amendment_guid=None):
...
+    parser.add_argument('uploadedFiles', type=list, location='json', store_missing=False)
+
+    @api.doc(params={'permit_amendment_guid': 'Permit amendment guid.'})
+    @requires_role_mine_view
+    @api.marshal_with(PERMIT_AMENDMENT_MODEL, code=200)
+    def get(self, permit_guid=None, permit_amendment_guid=None):
+        permit_amendment = PermitAmendment.find_by_permit_amendment_guid(permit_amendment_guid)
+        if not permit_amendment:
+            raise NotFound("Permit Amendment not found.")
+        return permit_amendment
```
Jsyro: needs to be checked against the permit as well.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
```diff
@@ -64,71 +62,76 @@ def post(self, permit_guid=None, document_guid=None):
...
 
-        return permit_amendment.json()
 
+class PermitAmendmentDocumentResource(Resource, UserMixin):
     @requires_role_mine_create
-    def delete(self, permit_amendment_guid, document_guid=None, permit_guid=None):
-        if not document_guid:
-            return self.create_error_payload(400, 'must provide document_guid to be unlinked'), 400
+    @api.response(204, 'Successfully deleted.')
+    def delete(self, permit_amendment_guid, permit_amendment_document_guid=None, permit_guid=None):
```
Jsyro: should these be optional variables? looks like they should be required.

python-backend/app/api/permits/namespace/permits.py
nayr974: I couldn't think of a better path. Any suggestions? I think what we did for Incidents is a better approach overall, but that's a change in front-end behavior I'm not sure we necessarily want.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
nayr974: I can add that fairly easily, but the cost is a DB hit as the amendment does not have permit_guid to validate against itself. Worth it? I lean no, but I agree it would be better to validate.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
nayr974: Agree, changed to required here and in a few other places. Does flaskrestplus enforce these if the default is not None? Do we need to keep the conditional validation below?

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
Jsyro: That was part of the URL pattern discussion is that permit guid needs to matter if it's in the URL.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
Jsyro: if they are not defaulted, then all of the URL parameter defined i the path need to exist. correct no more None check.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
nayr974: I'm not totally sold on the cost/benefit of being strict, but I'll add it. Conceptually agree, but adds complexity and performance costs.

python-backend/app/api/permits/response_models.py
```diff
@@ -0,0 +1,49 @@
+from app.extensions import api
+from flask_restplus import fields
+
+PERMIT_AMENDMENT_DOCUMENT_MODEL = api.model(
```
JustinMacaulay: I think given how many things now live under the mine we may want to start breaking the API models up into separate files like this instead of having one large model file.

python-backend/app/api/permits/namespace/permits.py
nayr974: I just moved it under mines (not pushed yet) and yes it ends up being /mines/permits/amendments/documents

python-backend/app/api/permits/permit/models/permit.py
```diff
@@ -39,39 +38,25 @@ def __repr__(self):
...
     def find_by_permit_guid(cls, _id):
         return cls.query.filter_by(permit_guid=_id).first()
 
     @classmethod
     def find_by_mine_guid(cls, _id):
-        return cls.query.filter_by(mine_guid=_id)
+        return cls.query.filter_by(mine_guid=_id).all()
 
     @classmethod
     def find_by_permit_no(cls, _permit_no):
```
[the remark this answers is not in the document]
nayr974: Yes, it's used to query all permits for a mine by the mines/mine_guid/permits get

python-backend/app/api/permits/response_models.py
nayr974: I agree.

python-backend/app/api/permits/permit_amendment/resources/permit_amendment_document.py
nayr974: Updated and removed None check

python-backend/app/api/permits/permit_amendment/resources/permit_amendment.py
nayr974: Added. Don't like it as a pattern :)

python-backend/app/api/permits/permit/resources/permit.py
nayr974: Path changed, and mine_guid's are validated. After working with it I don't like this method of longer paths and strict validation. It adds cruft. The front-end components need to pass more things through, it adds extra SQL calls and overall complexity that I don't think will be offset by this notion of extra safety.

python-backend/app/api/permits/namespace/permits.py
nayr974: Moved routes and files under mines.

python-backend/app/api/permits/permit/models/permit.py
nayr974: We may want to, but it doesn't actually do anything in this use. I actually just removed it.

PR #847, opened 2019-07-03
Map cache is deleted when a mine is edited, map functional test is un…

thegentlemanphysicist:
This should fix the map funtional test.


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -387,7 +387,8 @@ def put(self, mine_no_or_guid):
             mine.mine_location = MineLocation(
                 latitude=data['latitude'], longitude=data['longitude'])
             mine.save()
-            cache.delete(MINE_MAP_CACHE)
+
+        cache.delete(MINE_MAP_CACHE)
```
nayr974: Why do we want to delete the map cache if the lat and long have not changed?

thegentlemanphysicist: The name of the mine changes, the name on the  pin should change, same with commodities, and permits, at this point it makes more sense to reload the pins

thegentlemanphysicist: also, I'm going to be closing this pr since it doesn't fix the geb test.  This was grouping two bug tickets

nayr974: The cost of rebuilding the cached copy is fairly high and there's only a few fields on the pins, I think it would make sense to add an extra conditional for the fields that are on the pin instead of doing this whenever a mine is edited, otherwise people adding notes and such will invalidate the cache when not needed.

PR #848, opened 2019-07-03
MDS-1175: Feature/mds Show PDF Preview

NWCalvank:
Tested on MacOS in
- Google Chrome
- Safari
- Firefox

**Feature**
Open PDF as a preview in a new tab, with the option to download. Non-PDF files are downloaded (or prompted, in the case of Firefox defaults or user-adjusted browser preferences).

**Caveat**
All browsers tested detected this new tab as a popup and blocked it on first click. The browser shows a notification that allows the users to "allow popups from this domain." We should include this as an example in our training material. Once the popups have been allowed, the tab will open with the preview, and all future previews open without the popup blocker warning.

### DIFF
--- a/frontend/src/components/common/DocumentTable.js
+++ b/frontend/src/components/common/DocumentTable.js
@@ -23,7 +23,7 @@ export class DocumentTable extends Component {
   transformRowData = (documents) =>
     documents.map((document) => ({
       key: document.mine_document_guid,
-      guid: document.document_manager_guid,
+      document_manager_guid: document.document_manager_guid,
       name: document.document_name,
       category: document.variance_document_category_code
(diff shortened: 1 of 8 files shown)


NWCalvank:
Download functional test isn't passing still. Marking this as WIP

NWCalvank:
Issues have been resolved.

Test failure was caused by a regression in the downloaded file name. This was the result of no longer setting the file as an attachment (which sets a `Content-disposition` header).

To fix this, I've used the new behaviour only for PDFs, which are previewed in a new tab and have the option of downloading. All other filetypes are automatically downloaded with their appropriate name set, as was the existing behaviour.

PR #849, opened 2019-07-04
Switch metabase database from H2 to Postgres

FreshGyan:
- Updated Metabase
- Cleaned up deployment template
- Added a postgres 9.6 container to the metabase pod
- Adjusted resources accordingly


FreshGyan:
@Jsyro My plan was to copy the h2 database files locally before deployment, then when it's done deploying, I copy it to the pod, run the transfer process and then delete it.
So we can save another merge to remove the code for old PVC mount and such.

PR #851, opened 2019-07-05
Feature/mds 2098

mmumby:
-Created a View modal for incidents
- updated POST/GET to include `recommendations` (A followup ticket has been created to address the additional work [MDS-2153](https://bcmines.atlassian.net/browse/MDS-2153)


frontend/src/components/mine/Incidents/MineIncidents.js
```diff
@@ -59,10 +59,18 @@ export class MineIncidents extends Component {
   }
 
   handleAddMineIncident = (values) => {
-    this.props.createMineIncident(this.props.mine.mine_guid, values).then(() => {
-      this.props.closeModal();
-      this.props.fetchMineIncidents(this.props.mine.mine_guid);
-    });
+    // set default fatalities & injuries to 0 if not set by user
```
NWCalvank: I think this line of code is self-documenting, and the comment isn't needed 🙂

frontend/src/components/mine/Incidents/MineIncidents.js
```diff
@@ -59,10 +59,18 @@ export class MineIncidents extends Component {
...
+    const { number_of_fatalities = 0, number_of_injuries = 0 } = values;
+    this.props
+      .createMineIncident(this.props.mine.mine_guid, {
+        number_of_fatalities,
+        number_of_injuries,
+        ...values,
```
NWCalvank: I'm not able to think of a case where this would be an issue, but to be a bit more explicit here and to avoid a potential overwrite of the defaulted values, you could pull out the remaining values when you assign the defaults.

```javascript
const { number_of_fatalities = 0, number_of_injuries = 0, ...otherValues } = values;
   this.props
     .createMineIncident(this.props.mine.mine_guid, {
        number_of_fatalities,
        number_of_injuries,
        ...otherValues,
      })
```

frontend/src/components/modalContent/ViewIncidentModal.js
```diff
@@ -0,0 +1,288 @@
...
+          </p>
+        </div>
+      </div>
+      <br />
+    </div>
+  );
+
+  renderInitialDocuments = () => {
+    const initialDocuments = this.props.incident.documents.filter(
+      ({ mine_incident_document_type_code }) => mine_incident_document_type_code === "INI"
```
NWCalvank: Thoughts on pros/cons of setting this type code to a constant? It smells a bit of a magic string here, but if it's hyper-specific to just this component then it might not make sense to use a constant.

frontend/src/components/modalContent/ViewIncidentModal.js
```diff
@@ -0,0 +1,288 @@
...
+          </div>
+        </div>
+      </div>
+      <br />
+    </div>
+  );
+
+  renderFinalDocuments = () => {
+    const finalDocuments = this.props.incident.documents.filter(
+      ({ mine_incident_document_type_code }) => mine_incident_document_type_code === "FIN"
```
NWCalvank: Same thoughts here.

I wonder if we may want to make an object of potential document types?
```javascript
const DOCUMENT_TYPES = {
  initial: "INI",
  final: "FIN",
};

const finalDocuments = this.props.incident.documents.filter(
  ({ mine_incident_document_type_code }) => mine_incident_document_type_code === DOCUMENT_TYPES.final
);
```

Compared to string constants, this might be the more reasonable way of maintaining our FE collection of valid codes from BE code tables 🤔 

Thoughts? Feelings?

frontend/src/components/modalContent/ViewIncidentModal.js
```diff
@@ -0,0 +1,288 @@
...
+  ];
+
+  render() {
+    return (
+      <div>
+        {this.renderInitialDetails()}
+        {this.renderIncidentDetails()}
+        {this.renderInitialDocuments()}
+        {this.renderFollowUpInformation()}
+        {this.renderFinalDocuments()}
```
NWCalvank: This is excellent

frontend/src/tests/components/mine/Incidents/__snapshots__/MineIncidentTable.spec.js.snap
```diff
@@ -374,6 +376,7 @@ ShallowWrapper {
                 "key": undefined,
                 "mine_incident_report_no": "2019-70",
                 "openMineIncidentModal": [MockFunction],
+                "openViewMineIncidentModal": undefined,
```
NWCalvank: Looks like mocks or props need to be updated

frontend/src/components/mine/Incidents/MineIncidents.js
mmumby: removed.

frontend/src/tests/components/mine/Incidents/__snapshots__/MineIncidentTable.spec.js.snap
mmumby: 100%, Haven't touched the tests yet.

frontend/src/components/modalContent/ViewIncidentModal.js
mmumby: Like them being in constants. Meant to update that. 

We have similar constants for Variances that live in `strings.js` since they are used in multiple places. I'll initiate the constants in this file for now, as they aren't being used elsewhere.

frontend/src/components/mine/Incidents/MineIncidents.js
mmumby: updated!

frontend/src/tests/mocks/dataMocks.js
```diff
@@ -2437,3 +2475,74 @@ export const VARIANCE_DOCUMENT_CATEGORY_OPTIONS_HASH = {
...
+    description: "This was a dangerous occurrence",
+    display_order: 20,
+    mine_incident_determination_type_code: "DO",
+  },
+  {
+    active_ind: true,
+    description: "This was not a dangerous occurrence",
+    display_order: 30,
+    mine_incident_determination_type_code: "NDO",
+  },
```
NWCalvank: `active_ind` and I believe `display_order` are not meant to be exposed by the API. I'm assuming you've added them here because they're being returned?

If they are, I'll update the follow-up backend ticket that I made to include properly filtering out inactive records & withholding these two fields

mmumby: They are indeed being returned, I can delete them from the mocks as they are not being used on the UI.

NWCalvank: Okay. I'll update the follow-up ticket to include this new info

PR #856, opened 2019-07-09
Bug/mds 1536

thegentlemanphysicist:
This clears the cache of map data on the edit of any mine's description. It then rebuilds it asyncronously in a newly spawned thread.


thegentlemanphysicist:
hmm, found a bug, working on it now :(

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -149,6 +154,7 @@ def post(self):
         if lat and lon:
             mine.mine_location = MineLocation(latitude=lat, longitude=lon)
             cache.delete(MINE_MAP_CACHE)
+            self.call_rebuild_map_cache()
```
FreshGyan: I'd recommend spawning the thread after the DB object has been saved/committed or the return statement to ensure the new cached values already exist in the DB.

Probably can throw in another conditonal there to look for all the fields that are displayed on the mine pin like Ryan suggested earlier

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -387,11 +393,26 @@ def put(self, mine_no_or_guid):
...
+
         return mine
 
+    def call_rebuild_map_cache(self):
+        app = current_app._get_current_object()
+        environ = request.environ
+
+        def do_work():
+            with app.request_context(environ):
+                mineMapResource = MineMapResource()
```
FreshGyan: What does the environ do here?

If you define rebuild_map_cache as a static method you may be able to skip the class init step.

Maybe more descriptive function names as well 😅 Alternatively you can just not call do_work and directly reference the rebuild_map_cache function in the thread initialization

python-backend/app/api/mines/mine/resources/mine_map.py
```diff
@@ -54,4 +45,17 @@ def get(self):
         response.last_modified = last_modified
         response.make_conditional(request)
 
-        return response
\ No newline at end of file
+        return response
+
+    def rebuild_map_cache(self):
```
FreshGyan: Just something to think about here coz I don't have an answer off the top of my head. But it may be worth it to keep the function scope to just get data in the expected format and then storing it into cache in parent/different step?
Not sure, just brain storming here

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: The environ is needed to pass in the request context.  My understanding is that this is how you give the thread access to the authentication information... as for more descriptive names how does 'do work faster' sound.

python-backend/app/api/mines/mine/resources/mine_map.py
thegentlemanphysicist: Not sure I understand... I'll chat when you get in.

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: Good catch about the commi/cache order.  Though putting a check in the post shouldn't be needed since any created mine will have a location. I will write one for the put method

python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -379,17 +388,25 @@ def put(self, mine_no_or_guid):
...
             mine.mine_location = MineLocation(
                 latitude=data['latitude'], longitude=data['longitude'])
+            refresh_cache = True
             mine.save()
-            cache.delete(MINE_MAP_CACHE)
-        # Status validation
+
         _mine_status_processor(data.get('mine_status'), data.get('status_date'), mine)
+
+        # If more fields are added to the map popup this refresh cache will need to be called for them as well
```
NWCalvank: I'd prefer if the backend doesn't discuss the frontend. The cache behaviour should be totally self-contained to the backend.

python-backend/app/api/mines/mine/resources/mine_map.py
```diff
@@ -54,4 +46,30 @@ def get(self):
...
+        thread.start()
+
+    @staticmethod
+    def rebuild_and_return_map_cache():
+        records = MineMapViewLocation.query.filter(MineMapViewLocation.latitude != None).all()
+        last_modified = datetime.utcnow()
+
+        # jsonify then store in cache
+        map_result = json.dumps({
+            'mines': list((map(lambda x: x.json_for_map(), records)))
```
NWCalvank: Just an FYI, our `.json*` methods are being deprecated. I'm surprised this one still exists.

thegentlemanphysicist: ah, yeah I literally moved the existing code to a new method.  Is it worth fixing it with this pr, or should I work on that in a separate one?

NWCalvank: It's fine as-is, I think. But we should be aware of any time we're leveraging these json methods. If I'm seeing it correctly, this was already being used, so it's not a big deal here

python-backend/app/api/mines/mine/resources/mine.py
thegentlemanphysicist: appropriately altered

PR #863, opened 2019-07-11
Centralized Scheduler

FreshGyan:
Digdag -> https://www.digdag.io/
My PR -> https://mds-digdag-dev-pr-863.pathfinder.gov.bc.ca/projects

This PR adds digdag and ports over the current ETL jobs from each envt and moves them over to this.
The process is basically this:
- Login into openshift
- Create a pod using the env and base image for relevant (python-backend/nris) pod
- Run the command
- Wait for pod to exit with a success/failure code

I also did some cleanup of removing the scheduler code and unused imports as I stumbled upon them.
Also scaled up the nris pods in test/prod to be 2 at min.

Task checklist

- [x] Add secrets to test and prod
- [x] Add config blocks for test and prod for deploy
- [x] Manually create service accounts for test and prod digdag
- [x] Disable ETLs for test and prod
- [x] Remove the config-dev block for digdag
- [ ] Merge the code
- [ ] Hope it works
- [ ] Celebrate if it does


python-backend/app/commands.py
```diff
@@ -1,25 +1,20 @@
...
 from tests.factories import MineFactory, MinePartyAppointmentFactory
 
 
 def register_commands(app):
+
     @app.cli.command()
     def import_idir():
-        from app.scheduled_jobs.IDIR_jobs import _import_empr_idir_users
-        _import_empr_idir_users()
+        from app.scheduled_jobs.IDIR_jobs import import_empr_idir_users
```
Jsyro: Did you preserve this only so it could be run via cli? Can it not be moved into scheduler?

pipeline/config-test.groovy
```diff
@@ -317,8 +333,8 @@ environments {
                     memory_limit = "1Gi"
                     uwsgi_threads = 2
                     uwsgi_processes = 4
-                    replica_min = 1
-                    replica_max = 1
+                    replica_min = 2
+                    replica_max = 4
```
Jsyro: Are we trying to use more, but, smaller pods?

python-backend/app/commands.py
Jsyro: I see, we have removed all the schedulers and preserved the cli commands, and the central scheduler kicks up an instance of the pod and runs the cli command. maybe rename the module to be cli jobs?

FreshGyan: Yes! CLI seemed to be the cleanest way of running this and it also made sense as sort of a pseudo pattern.
-> Where we have a scheduled-jobs dir in each project(backend/nris) that contains the definition of the job, since it may need access to models and internal helper methods.
-> Then a cli invocation in commands file, which may have more logging and what not to invoke the functions in the scheduled jobs.
-> Then in scheduler we just spin up the container with a different entrypoint to run the job

Whereas moving the code to scheduler may end up introducing duplication with requirements.txt and code which i wasnt able to think of doing in a clean way

pipeline/config-test.groovy
FreshGyan: Aitchison asked me increase pods to 2 at min, I didnt try to shuffle the cpu and memory yet, wanted to do that for all envts with more pods and play around a bit more in a diff PR

python-backend/app/commands.py
FreshGyan: @Jsyro renamed module to be cli_jobs, i agree it makes more sense than scheduled jobs

PR #869, opened 2019-07-16
Mds 1834

thegentlemanphysicist:
This deletes null entries in the tables address and party_address_xref.  Deletion of database fields is a newish tech stack for me, so feedback would be greatly appreciated.
I also altered the database constraint on mine manager apointments so a deleted mine manager no longer prevents a new mine manger from being appointed


thegentlemanphysicist:
Realized that creating new parties with blank addresses creates an empty table all over again.  There will need to be a backend fix to resolve this.

python-backend/app/api/parties/party/resources/party_resource.py
```diff
@@ -148,8 +147,7 @@ def delete(self, party_guid):
         if party is None:
             return self.create_error_payload(404, 'Party guid with "{party_guid}" not found.'), 404
 
-        mine_party_appts = MinePartyAppointment.find_all_by_mine_party_appt_guid(
-            party_guid)
+        mine_party_appts = MinePartyAppointment.find_by_party_guid(party_guid)
```
thegentlemanphysicist: This was a bug that wasn't marking mine party appts as deleted

Jsyro: Can you delete that method `find_all_by_mine_party_appt_guid` that filters by deleted_ind is misleading and it was only used here.

PR #870, opened 2019-07-16
Landing Page

mmumby:
- Added new 'home' page for users
- updated the styling in`SearchBar.js` to allow for other components to use the searchBox, the search will now inherit the width of its parent component vs being pre-set for the navBar.
-updated navigation responsiveness to switch to hamburger menu when links become too squished.
- added pdf to assets and updated webpack to allow for it.

NOTE- In the process of updating the unit tests to work with pdf, I ran through mineSpace tests and had to update the snapshots there.


frontend/src/actionCreators/reportingActionCreator.js
```diff
@@ -15,3 +13,21 @@ export const fetchCoreDashboard = () =>
...
+    .get(`${ENVIRONMENT.apiUrl + API.DASHBOARD_164}`, createRequestHeader())
+    .then((response) => {
+      const { dashboard_url } = response.data || {};
+      return dashboard_url;
+    })
+    .catch(() => error(Strings.ERROR));
+
+export const fetchDashboard165 = () =>
+  CustomAxios({ errorToastMessage: "Unable to fetch dashboard." })
+    .get(`${ENVIRONMENT.apiUrl + API.DASHBOARD_165}`, createRequestHeader())
```
NWCalvank: These don't dispatch any action-specific events. We could use a generic `fetchMetabaseDashboard` function instead, and provide an ID as an argument. The ID could be stored in a constant to maintain the readability.

Thoughts?

frontend/src/components/dashboard/HomePage.js
```diff
@@ -0,0 +1,144 @@
...
+  render() {
+    const iframeUrlOne = `${this.state.graph_urls[0]}#bordered=true&titled=false`;
+    const iframeUrlTwo = `${this.state.graph_urls[1]}#bordered=true&titled=false`;
+
+    return (
+      <div className="background" style={{ backgroundImage: `url(${BACKGROUND})` }}>
+        <div className="search-container">
+          <div className="center">
+            <h1>Welcome!</h1>
+            <p>To begin, please search or clicks the links below</p>
```
NWCalvank: `please search or click the links below`?

frontend/src/components/search/SearchBar.js
```diff
@@ -71,51 +72,44 @@ export class SearchBar extends Component {
...
-          onSearch={(searchTerm) => this.search(searchTerm)}
-          onChange={this.changeSearchTerm}
-          onClick={this.clearSearchBar}
-          onFocus={this.clearSearchBar}
-          onBlur={() =>
-            this.setState({
-              isSelected: false,
-            })
+  render() {
+    // The default behaviour will fix the dropdown to the parent (searchBox), except in the navBar where the user can scroll beyond the searchBox.
```
NWCalvank: This comment may be useful if I were more familiar with this component, but from what I can see here, I don't understand it

frontend/webpack.parts.js
```diff
@@ -164,6 +164,22 @@ exports.loadImages = ({
   },
 });
 
+exports.loadFiles = ({ include, exclude } = {}) => ({
+  module: {
+    rules: [
+      {
+        test: /\.(xls?m|pdf|PDF|doc?x)$/,
```
NWCalvank: I imagine since the `PDF` didn't fix the issue we were debugging, we can probably remove it

python-backend/app/api/reporting/resources/metabase.py
```diff
@@ -23,3 +23,32 @@ def get(self):
...
         return { 'dashboard_url': dashboard_url }
+
+class LandingPageGraphOneResource(Resource, UserMixin):
+    @requires_role_view_all
+    def get(self):
+        payload = {
+          "resource": {"dashboard": 164},
+          "params": {
+
+          }
```
NWCalvank: This can just be `"params": {}`. Could you update line 18 as well, please and thank you 😄

python-backend/app/api/reporting/resources/metabase.py
```diff
@@ -23,3 +23,32 @@ def get(self):
...
+    def get(self):
+        payload = {
+          "resource": {"dashboard": 165},
+          "params": {
+
+          }
+        }
+        token = jwt.encode(payload, current_app.config['METABASE_EMBEDDING_SECRET_KEY'], algorithm="HS256")
+
+        dashboard_url = current_app.config['METABASE_SITE_URL'] + "/embed/dashboard/" + token.decode("utf8")
```
NWCalvank: I think we can pull some of this out to avoid the duplication across every resource. Something like:

```python
def get(self):
    dashbord_url = _get_dashboard_url(165)
    return { 'dashboard_url': dashboard_url }


def _get_dashboard_url(self, id):
       payload = {
          "resource": {"dashboard": id},
          "params": {

           }
        }
        token = jwt.encode(payload, current_app.config['METABASE_EMBEDDING_SECRET_KEY'], algorithm="HS256")

        return current_app.config['METABASE_SITE_URL'] + "/embed/dashboard/" + token.decode("utf8")
```

frontend/src/actionCreators/reportingActionCreator.js
mmumby: I also had this thought, I'll update these functions.

frontend/src/components/dashboard/HomePage.js
mmumby: removing the s.

frontend/src/components/search/SearchBar.js
mmumby: ah, I was wondering how deep to go with this comment. Not deep enough apparently. I'll remove it

PR #872, opened 2019-07-18
Add inactive inspectors and display them seperately.

JustinMacaulay:
Allows inspectors to assign inactive inspectors to a variance or incident. Along with this an additional script will need to be run to add all the missing inactive inspectors to the system.


frontend/src/selectors/partiesSelectors.js
```diff
@@ -28,7 +29,19 @@ export const getPartyRelationshipTypeHash = createSelector(
...
+      .map((inspector) => ({ value: inspector.party_guid, label: inspector.name }));
+    const inactiveInspectors = parties
+      .filter((inspector) => moment(inspector.expiry_date) < moment())
+      .map((inspector) => ({ value: inspector.party_guid, label: inspector.name }));
+    return [
+      { groupName: "Active", opt: activeInspectors },
+      { groupName: "Inactive", opt: inactiveInspectors },
+    ];
+  }
+  // parties.map((party) => ({ label: party.expiry_date > CURRENT_DATE ? "Active" : "Inactive", opt: }))
```
NWCalvank: Remove comment

frontend/src/selectors/partiesSelectors.js
```diff
@@ -28,7 +29,19 @@ export const getPartyRelationshipTypeHash = createSelector(
...
 export const getDropdownInspectors = createSelector(
   [getInspectors],
-  (parties) => parties.map((party) => ({ value: party.party_guid, label: party.name }))
+  (parties) => {
+    const activeInspectors = parties
+      .filter((inspector) => moment(inspector.expiry_date) >= moment())
+      .map((inspector) => ({ value: inspector.party_guid, label: inspector.name }));
+    const inactiveInspectors = parties
+      .filter((inspector) => moment(inspector.expiry_date) < moment())
+      .map((inspector) => ({ value: inspector.party_guid, label: inspector.name }));
```
NWCalvank: What sort of numbers are we looking at for the number of inspectors being processed here? If it can get into the thousands, we may want to refactor this so that we only iterate twice instead of 4 times

frontend/src/components/common/RenderGroupedSelect.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  >
+    <Select
+      disabled={props.disabled}
+      getPopupContainer={props.doNotPinDropdown ? null : () => document.getElementById(props.id)}
+      showSearch
+      placeholder={props.placeholder}
+      optionFilterProp="children"
+      filterOption={(input, option) => {
+        if (typeof option.props.children !== "string") {
```
NWCalvank: How did we wind up needing a type check here?

Also, why is something called `children` a string? It seems like it should be a collection

frontend/src/components/common/RenderGroupedSelect.js
```diff
@@ -0,0 +1,83 @@
...
+          return false;
+        }
+        return option.props.children.toLowerCase().indexOf(input.toLowerCase()) >= 0;
```
NWCalvank: As a follow-up, if we need both conditions, we may as well do a single return since we have a conditional wrapping a boolean return.

```javascript
(typeof option.props.children === "string") && 
option.props.children.toLowerCase().includes(input.toLowerCase())
```

frontend/src/components/common/RenderGroupedSelect.js
```diff
@@ -0,0 +1,83 @@
...
+      }}
+      id={props.id}
+      {...props.input}
+      onSelect={props.onSelect}
+    >
+      {props.data.map((group) => (
```
NWCalvank: Again, how many elements are we looking at in this list? Is the performance okay? These nested maps are going to run on every single rerender. Only a concern if it's a performance cost. But it's something to be aware of

frontend/src/components/common/RenderGroupedSelect.js
```diff
@@ -0,0 +1,83 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Form, Select } from "antd";
+import CustomPropTypes from "@/customPropTypes";
+
+/**
+ * @constant RenderSelect - Ant Design `Select` component for redux-form - used for small data sets that (< 100);
+ * There is a bug when the data sets are large enough to cause the dropdown to scroll, and the field is in a modal.
+ * In the case where the modal cannot scroll, it is better to pass in the prop doNotPinDropdown.  It allows the
+ * dropdown to render properly
```
NWCalvank: This doesn't look like the right comment

frontend/src/components/common/RenderGroupedSelect.js
JustinMacaulay: With the antd select when using the filter function it returns the select options but also the option groups. I have to do a type check to remove the option groups from the set so that the filter can filter on the children, which are the options.

frontend/src/selectors/partiesSelectors.js
JustinMacaulay: I think thats a possibility but not for a very long time but I get what you're saying.

frontend/src/selectors/partiesSelectors.js
JustinMacaulay: Removed.

frontend/src/components/common/RenderGroupedSelect.js
JustinMacaulay: At the moment it is 28 inspectors. For now I think it is ok but we should keep an eye on it.

frontend/src/components/common/RenderGroupedSelect.js
JustinMacaulay: Changed a bit to reflect the grouped aspect of the select.

frontend/src/components/common/RenderGroupedSelect.js
JustinMacaulay: Changed and I like it a lot better, thanks!

PR #875, opened 2019-07-22
MDS-1281 Task/ Merge mine_location into mine

NWCalvank:


python-backend/app/api/mines/mine/resources/mine.py
```diff
@@ -146,20 +145,20 @@ def post(self):
             major_mine_ind=data.get('major_mine_ind'),
             mine_region=data.get('mine_region'),
             ohsc_ind=data.get('ohsc_ind'),
-            union_ind=data.get('union_ind'))
-
-        db.session.add(mine)
-
-        if lat and lon:
-            mine.mine_location = MineLocation(latitude=lat, longitude=lon)
+            union_ind=data.get('union_ind'),
```
itchison: Shouldn't the union_ind be above the lat/long if statement?

NWCalvank: The only if statement above this is on line 136, which has nothing to do with the `Mine` constructor. Not quite sure what the suggestion is here

itchison: I must've been reading the markup wrong - my mistake.

PR #879, opened 2019-07-24
Hackathon testcov

thegentlemanphysicist:
Added some tests to the frontend during the hackathon.


frontend/src/components/Forms/AdvancedContactSearchForm.js
```diff
@@ -32,12 +32,12 @@ export class AdvancedContactSearchForm extends Component {
...
 
   handleReset = () => {
+    this.setState({ contactType: "PER" });
     this.props.reset();
     this.props.handleSearch({}, true);
-    this.setState({ contactType: "PER" });
   };
 
-  handleContactTypeChange = (chars, value) => {
+  handleContactTypeChange = (value) => {
```
mmumby: what was this change for? Does this function still work correctly?

thegentlemanphysicist: if I remember right, the chars wasn't being used in the function

thegentlemanphysicist: good catch

PR #881, opened 2019-07-24
MDS-2153 Task/Finish incidents recommendations Part 1

NWCalvank:
This is a branch with some WIP comments. It's being pushed up at this point because it addresses all of the things that are currently known to be high-priority fixes.

There will be a follow-up PR to finish addressing any remaining issues.


python-backend/app/api/mines/mine_api_models.py
```diff
@@ -173,6 +173,7 @@ def format(self, value):
 MINE_INCIDENT_RECOMMENDATION_MODEL = api.model(
     'Mine Incident Recommendation', {
         'recommendation': fields.String,
+        'mine_incident_recommendation_guid': fields.String
```
thegentlemanphysicist: if we're passing in a dictionary, why isn't 'Mine Incident Recommendation' put in the dictionary?  Is it a particular patern?

python-backend/app/api/incidents/models/mine_incident.py
```diff
@@ -82,12 +84,16 @@ class MineIncident(AuditMixin, Base):
...
                                                   uselist=False)
 
-    recommendations = db.relationship('MineIncidentRecommendation', lazy='selectin')
+    _recommendations = db.relationship('MineIncidentRecommendation', lazy='dynamic')
     documents = db.relationship('MineIncidentDocumentXref', lazy='joined')
     incident_documents = db.relationship('MineDocument',
                                          lazy='joined',
                                          secondary='mine_incident_document_xref')
 
+    @hybrid_property
```
thegentlemanphysicist: Does this cause this property to be applied to all recomendations returned from queries on this model?

python-backend/app/api/incidents/models/mine_incident.py
```diff
@@ -82,12 +84,16 @@ class MineIncident(AuditMixin, Base):
                                                   lazy='joined',
                                                   uselist=False)
 
-    recommendations = db.relationship('MineIncidentRecommendation', lazy='selectin')
+    _recommendations = db.relationship('MineIncidentRecommendation', lazy='dynamic')
```
thegentlemanphysicist: what does 'dynamic' give us over 'selectin'

python-backend/app/api/incidents/models/mine_incident_recommendation.py
```diff
@@ -5,8 +5,10 @@
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
 from app.extensions import db
-from ....utils.models_mixins import AuditMixin, Base
 
+from app.api.utils.models_mixins import AuditMixin, Base
+
+INVALID_MINE_INCIDENT_RECOMMENDATION_GUID = 'Invalid mine_incident_recommendation_guid.'
```
thegentlemanphysicist: is there a constants file this could be moved to?

python-backend/app/api/incidents/models/mine_incident_recommendation.py
```diff
@@ -23,3 +32,12 @@ def create(cls, recommendation, mine_incident_id, add_to_session=True):
         if add_to_session:
             new_recommendation.save(commit=False)
         return new_recommendation
+
+
+    @classmethod
+    def find_by_mine_incident_recommendation_guid(cls, _id):
+        try:
+            uuid.UUID(str(_id), version=4)
+            return cls.query.filter_by(mine_incident_recommendation_guid=_id).filter_by(deleted_ind=False).first()
```
thegentlemanphysicist: Why does this use the '.first()'  mehtod, shouldn't the guid be unique?

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -295,8 +305,33 @@ def put(self, mine_guid, mine_incident_guid):
                     db.session.delete(doc)
                     db.session.commit()
 
-        try:
-            incident.save()
-        except Exception as e:
-            raise InternalServerError(f'Error when saving: {e}')
```
thegentlemanphysicist: Why is it safe to remove this error handling?

migrations/sql/V2019.07.23.16.20__add_guid_to_incident_recommendations.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE mine_incident_recommendation
+ADD COLUMN mine_incident_recommendation_guid uuid    DEFAULT gen_random_uuid() NOT NULL,
```
Jsyro: does this table need a guid to expose it directly publicly? seems like the integer id so it can FK to the incident seems like all this would need.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -295,8 +305,33 @@ def put(self, mine_guid, mine_incident_guid):
...
+    def _handle_recommendations(cls, incident, recommendations):
+        for payload in recommendations:
+            edited_rec_guid = payload.get('mine_incident_recommendation_guid')
+            rec_string = payload.get('recommendation')
+            if rec_string is None or rec_string == '':
+                if payload.get('mine_incident_recommendation_guid') is not None:
+                    # Remove existing empty recommendations
+                    rec = MineIncidentRecommendation.find_by_mine_incident_recommendation_guid(
+                        edited_rec_guid)
+                    rec.soft_delete()
```
Jsyro: not sure this needs a method.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -295,8 +305,33 @@ def put(self, mine_guid, mine_incident_guid):
...
-        try:
-            incident.save()
-        except Exception as e:
-            raise InternalServerError(f'Error when saving: {e}')
+        incident.save()
         return incident
+
+    @classmethod
+    def _handle_recommendations(cls, incident, recommendations):
+        for payload in recommendations:
```
nayr974: Minor naming quibble: the use of the word payload doesn't seem right. It's not as readable as 'recommendation' and not I haven't really seen it used as a generic iterator name vs 'i' or 'item'.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
NWCalvank: I can remove it if we don't like it. Just seemed like a nice complement to `.create()`

python-backend/app/api/incidents/models/mine_incident.py
```diff
@@ -5,12 +5,14 @@
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
 from app.extensions import db
-from ....utils.models_mixins import AuditMixin, Base
+
 from .mine_incident_followup_investigation_type import MineIncidentFollowupInvestigationType
-from app.api.mines.incidents.models.mine_incident_determination_type import MineIncidentDeterminationType
+
+from app.api.utils.models_mixins import AuditMixin, Base
+from app.api.incidents.models.mine_incident_determination_type import MineIncidentDeterminationType
```
nayr974: @Jsyro I remember you discussing that you thought incidents should be under mines, do you think this makes sense? We have some base level API methods so having 'incidients' at the top level makes sense to me, even if some routes are under /mines/.

migrations/sql/V2019.07.23.16.20__add_guid_to_incident_recommendations.sql
NWCalvank: Yeah. It gives the FE a unique identifier to pass along with an existing recommendation so that it's clear which record(s) need to be updated/soft deleted.

python-backend/app/api/incidents/models/mine_incident_document_type_code.py
```diff
@@ -4,14 +4,14 @@
 from app.api.utils.models_mixins import AuditMixin, Base
 
 
-class MineIncidentTypeCode(AuditMixin, Base):
+class MineIncidentDocumentTypeCode(AuditMixin, Base):
```
nayr974: Thanks. It bugged me that this was inconsistent with the table names.

python-backend/app/api/incidents/models/mine_incident_recommendation.py
nayr974: It doesn't seem useful to me to pull validation messages out as variables. What am I not seeing?

python-backend/app/api/mines/incidents/resources/mine_incidents.py
NWCalvank: Good point. I should have actually replaced `payload` before opening the PR. It was functioning as a placeholder when I first added it and then I became blind to it. I'll update

python-backend/app/api/incidents/models/mine_incident_recommendation.py
NWCalvank: I did this with Variances for the purpose of importing the constant(s) into the tests to assert the correct validation error.

There is a lot of missing test coverage, so I haven't added that in this PR.

There are potentially cons to having this split out into a constant, so I'm open to putting the string directly in the response if there are concerns. I think it's a net win, but that's only my perspective.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
nayr974: I get where you're coming from. Might be better served on the base model, or even as an override of delete which could set delete if the model has deleted_ind, else delete.

python-backend/app/api/incidents/models/mine_incident_recommendation.py
nayr974: Ok that makes sense. Asserting validation messages and the general overhead of string constants is a feel bad to me personally, but if the tests and API are using the same constants and we need to validate more granular than our response codes allow I do get it.

NWCalvank: I'm not sure that we _need_ to validate these strings. My reason for doing so was that we have had issues with useless errors and/or exception stack trace dumps in the responses historically, so I liked knowing that we had some coverage on that.

With the fix for global error handlers last month, this may no longer be an issue. At which point, we can remove the feel bad of validating error strings? 🙂

migrations/sql/V2019.07.23.16.20__add_guid_to_incident_recommendations.sql
Jsyro: Good points, sounds great.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: Yea I was going down Ryan's path, if we want a global soft_delete() then is should exist on the base model, but not all models have delete_ind (unless we want that to be a Mixin).

python-backend/app/api/incidents/models/mine_incident.py
Jsyro: The discussion about asking about ALL incidents (across all mines), ended with us saying that `/mines/<guid>/incidents` is a specific mine and `/incidents` is all mines (instead of `/mines/incidents`. Mostly this came true because that results in both `/mines/<guid>/incidents/<guid>` vs `/mines/incidents/<guid>` being supported and we didn't want to implement both of those paths, or have two paths map to the same code.

PR #882, opened 2019-07-24
Task/mds 1689 document resource refactor-part-1

Jsyro:
Lets do small PR's. Combine common doc service access methods.


python-backend/app/api/documents/mines/resources/mine_document_resource.py
```diff
@@ -2,24 +2,34 @@
...
 from ....utils.resources_mixins import UserMixin, ErrorMixin
 
+MINE_DOCUMENT_MODEL = api.model(
+    'MineDocument', {
+        'mine_document_guid': fields.String,
+        'mine_guid': fields.String,
+        'document_manager_guid': fields.String,
+        'document_name': fields.String,
+        'active_ind': fields.String,
+    })
```
NWCalvank: Is this intentionally left in this file instead of living in a file of api models?

python-backend/app/api/services/document_manager_service.py
```diff
@@ -0,0 +1,56 @@
...
+from flask import Response
+from app.config import Config
+
+ALLOWED_DOCUMENT_CATEGORIES = ['tailings', 'permits', 'variances', 'incidents']
+
+
+class DocumentManagerService():
+    document_manager_url = f'{Config.DOCUMENT_MANAGER_URL}/document-manager'
+
+    def initializeFileUploadWithDocumentManager(self, request, mine, document_category):
```
NWCalvank: Any reason for not making this a class method? It doesn't look like you're leveraging anything on the instance that couldn't be used as a constant/class method instead.

python-backend/app/api/documents/mines/resources/mine_document_resource.py
Jsyro: document models will be cleaned up and centralized, It's also possible part 2 doesn't come until reports are done, which will replace all of this code.

python-backend/app/api/services/document_manager_service.py
Jsyro: I think you are right, everything is static now.

PR #884, opened 2019-07-25
Feature/mds 2126

thegentlemanphysicist:
This pr allows anyone with permissions to edit a variance at any stage.
This pr also allows anyone to open the view variance modal. (This was not happening before)
There was an unused 'isOverdue' argument being passed into the open modal handler.  It seemed to be out of date.


thegentlemanphysicist:
> The `MineVarianceTable` is also a child of `VarianceTable`, Is there any clean up that needs to happen in that file or does this work as expected on that view?

I believe I looked at this, but I'll check it out to make certain

PR #896, opened 2019-07-27
Adjust resources

FreshGyan:
Reducing resource requests to match what the actual usage is like, left the limits still reasonable.

Also playing around with some pod numbers to see how that works.


pipeline/config-prod.groovy
```diff
@@ -341,33 +341,33 @@ environments {
...
-                    cpu_limit = "150m"
-                    memory_request = "256Mi"
-                    memory_limit = "512Mi"
-                    replica_min = 2
-                    replica_max = 4
+                    cpu_request = "10m"
+                    cpu_limit = "50m"
+                    memory_request = "96Mi"
+                    memory_limit = "160Mi"
+                    replica_min = 3
```
pdittaro: Any reason for increasing the nginx min replicas?

pipeline/config-prod.groovy
```diff
@@ -341,33 +341,33 @@ environments {
...
+                    memory_request = "512Mi"
+                    memory_limit = "2Gi"
                     uwsgi_threads = 2
                     uwsgi_processes = 4
                     replica_min = 2
                     replica_max = 4
                 }
                 python_lite {
-                    cpu_request = "100m"
+                    cpu_request = "10m"
```
pdittaro: This seems a little low compared to test, and doesn't have a cpu limit.

pipeline/config-prod.groovy
FreshGyan: After openshift 101, Jason A/S recommended we try with lower resources and more pods to see how things are working. 
Just so if a node gets evacuated or is faulty or something, we don't have downtime.
Im also curious to see how well this works and if it actually performs better somehow

pipeline/config-prod.groovy
FreshGyan: The limit is right underneath it 😅 
But yeah, the python_lite pod is actually not using more than 0.001 CPU in general, so I reduced the CPU to reflect that.

Good catch about test tho, updated test/dev to reflect the same values

PR #898, opened 2019-07-29
MDS-2181 add vFCBC and NROS schema

JustinMacaulay:
Added migrations to add a schema to hold the vFCBC and NROS data.


migrations/sql/V2019.07.24.11.55__add_tables_to_stage_NROS_and_VFCBC_data.sql
```diff
@@ -0,0 +1,494 @@
+CREATE SCHEMA IF NOT EXISTS now;
```
itchison: As discussed, I think we pivoted a little on our implementation. This schema won't hold NOW's but will hold NOW Submissions. Can we name the scheme to NOW_Submissions.

JustinMacaulay: Changed to now_submissions

migrations/sql/V2019.07.24.11.55__add_tables_to_stage_NROS_and_VFCBC_data.sql
```diff
@@ -0,0 +1,494 @@
...
+    MAILINGADDRESSLINE1 character varying(4000),
+    CONTACTTYPE character varying(4000),
+    CONTACTCERTIFICATIONTYPE character varying(200),
+    CONTACTCERTIFICATIONID character varying(4000),
+    MAILINGADDRESSLINE2 character varying(60),
+    MAILINGADDRESSCITY character varying(100),
+    MAILINGADDRESSPROVSTATE character varying(200),
+    MAILINGADDRESSCOUNTRY character varying(50),
+    MAILINGADDRESSPOSTALZIP character varying(10),
+
```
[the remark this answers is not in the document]
JustinMacaulay: Yes, added that in.

migrations/sql/V2019.07.24.11.55__add_tables_to_stage_NROS_and_VFCBC_data.sql
```diff
@@ -0,0 +1,494 @@
...
+CREATE TABLE now.document_nda (
+    ID serial PRIMARY KEY,
+    MESSAGEID integer,
+	DOCUMENTURL character varying(4000),
+	FILENAME character varying(4000),
+	DOCUMENTTYPE character varying(4000),
+	DESCRIPTION character varying(4000),
+
+    FOREIGN KEY (MESSAGEID) REFERENCES now.application(MESSAGEID) DEFERRABLE INITIALLY DEFERRED
```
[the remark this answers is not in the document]
JustinMacaulay: Yes, it should be to the application_nda table. Good catch and thanks.

migrations/sql/V2019.07.24.11.55__add_tables_to_stage_NROS_and_VFCBC_data.sql
```diff
@@ -0,0 +1,495 @@
+CREATE SCHEMA IF NOT EXISTS NOW_Submissions;
+
+CREATE TABLE NOW_Submissions.application_start_stop (
+    MESSAGEID integer PRIMARY KEY,
```
NWCalvank: So, I see a bunch of these don't have underscores, and others do. Are we leaving them as-is to simplify the ETL process? With the assumption that it doesn't matter if the naming patterns are different because it's in a different schema?

I'm creating the python models, and this stood out right away. Thanks!

JustinMacaulay: For this everything was meant to just be the same as it is in the databases we are pulling from. for the most part all of the columns are the same as what they use and we have added a few surrogate PK's on tables that had combo PK's. Half the tables had underscores in the name and the other half didn't so thats the one thing we normalized so they all have underscores now.

NWCalvank: Regarding underscores, I was referring to the column names. Our column names use underscores, so I would expect this to be called `message_id`, for example. Some of these column names have underscores, and some do not.

JustinMacaulay: Yeah, the SEQ_NO is a weird one that was added by NROS and for some reason doesn't match the rest of the naming conventions. They are a mor or less straight copy paste though.

PR #899, opened 2019-07-29
Feature/mds 2145

thegentlemanphysicist:
This is a short pr that fixes the sort on Variance and Compliance code collumns.  Incidents will be done ASANathan's done fixing the feature.  
This also sorts the drop down menus for codes so that 1.10 comes after 1.9. 
This has not been tested on Compliance because I could not get the tab working.


frontend/src/components/mine/Variances/MineVarianceTable.js
```diff
@@ -39,6 +39,20 @@ const propTypes = {
   }),
 };
 
+const codeSorter = (code_one, code_two) => {
```
mmumby: Can this be moved to 'utils' since it's the same in both files, also update the name to be a little more clear ie: `complinceCodeSorter`

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,25 @@ export const getFiscalYear = () => {
 };
 
 export const formatParamStringToArray = (param) => (param ? param.split(",").filter((x) => x) : []);
+
+// This method sorts codes of the for '#.#.# - Lorem Ipsum'
+// where the numbe of numbers is variable and the text is optional
```
mmumby: number*

frontend/src/components/mine/Compliance/ComplianceOrdersTable.js
```diff
@@ -19,6 +19,20 @@ const defaultProps = {
...
+  if (!code_one) {
+    return true;
+  }
+  if (!code_two) {
+    return false;
+  }
+  if (compareCodes(code_one, code_two) < 0) {
+    return true;
+  }
+  return false;
```
NWCalvank: These final returns are redundant due to the conditional. So, if the conditional passes (is true), you're returning true, otherwise you're returning false. In both cases, you return the value of the condition.

So, this will do the same thing:
```javascript
return compareCodes(code_one, code_two) < 0;
```

Also, I'd like to push for an explicit null check instead of the bang checks for the presence checks.

And I'm sure it was just a habit carried over from the python code, but JS favours camelCase, so we should stick to that when possible.

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,46 @@ export const getFiscalYear = () => {
...
+    bInt = parseInt(bCodes[i], 10);
+    if (aInt > bInt) return 1;
+    if (aInt < bInt) return -1;
+  }
+  // If the previous loop did not return anything and the codes are the same length then they are the same code.
+  if (aCodes.length === bCodes.length) {
+    return 0;
+  }
+  // The shorter of two codes with otherwise matching numbers is ordered first e.g 1.11 is before 1.11.12
+  return aCodes.length < bCodes.length ? -1 : 1;
```
NWCalvank: Nothing here actually leverages the regex very much, so one of those should be changed. Either don't use the regex at all or reference the named groups in the ordering logic.

The comments are an improvement in readability, but I'm always a bit leery about comments that describe this much implementation, since implementation can change without comments being changed, which makes things more confusing.

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,46 @@ export const getFiscalYear = () => {
...
+};
+
+export const codeSorter = (codeOne, codeTwo) => {
+  // if a code is null it set before one that is not in sort order.
```
NWCalvank: I think this is meant to say something like `if a code is null, set it before the other code`? As-written, I don't understand it.

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,46 @@ export const getFiscalYear = () => {
...
+  const bCodes = regexParse(b);
+  if (!aCodes) return -1;
+  if (!bCodes) return 1;
+  let aInt;
+  let bInt;
+  const k = Math.min(aCodes.length, bCodes.length);
+  // Compares then non-null parts of two strings of potentially different lengths e.g 1.11 and 1.4.12
+  for (let i = 1; i < k; i += 1) {
+    aInt = parseInt(aCodes[i], 10);
+    bInt = parseInt(bCodes[i], 10);
```
NWCalvank: Reason for using `parseInt` instead of `Number`?

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,46 @@ export const getFiscalYear = () => {
...
+  const regexParse = (input) =>
+    input.match(/(?<first>[0-9]+)\.(?<second>[0-9]+)\.(?<third>[0-9]+)/) ||
+    input.match(/(?<first>[0-9]+)\.(?<second>[0-9]+)/) ||
+    input.match(/(?<first>[0-9]+)/);
+  const aCodes = regexParse(a);
+  const bCodes = regexParse(b);
+  if (!aCodes) return -1;
+  if (!bCodes) return 1;
+  let aInt;
+  let bInt;
```
NWCalvank: These should be declared inside of the for loop block

frontend/src/utils/helpers.js
```diff
@@ -95,3 +95,46 @@ export const getFiscalYear = () => {
...
+export const compareCodes = (a, b) => {
+  // Returns the first match that is non-null.
+  const regexParse = (input) =>
+    input.match(/(?<first>[0-9]+)\.(?<second>[0-9]+)\.(?<third>[0-9]+)/) ||
+    input.match(/(?<first>[0-9]+)\.(?<second>[0-9]+)/) ||
+    input.match(/(?<first>[0-9]+)/);
+  const aCodes = regexParse(a);
+  const bCodes = regexParse(b);
+  if (!aCodes) return -1;
+  if (!bCodes) return 1;
```
NWCalvank: It's not common to omit curly braces in this codebase, so I'd recommend against it here, even though it adds lines.

frontend/src/utils/helpers.js
thegentlemanphysicist: It was the one I found, is there a reason we want Number?

frontend/src/utils/helpers.js
thegentlemanphysicist: done

frontend/src/utils/helpers.js
thegentlemanphysicist: done

frontend/src/utils/helpers.js
thegentlemanphysicist: I english gooder

frontend/src/utils/helpers.js
thegentlemanphysicist: I simplified the regex as much as possible.  I'm in favour of keeping the regex, mostly because generalizing to other code types is much easier to do.  i.e. 1.12.3.(3)

thegentlemanphysicist: I also rewrote the comments slightly, to be less verbose.

frontend/src/utils/helpers.js
```diff
@@ -96,6 +96,58 @@ export const getFiscalYear = () => {
...
+  // if both codes have the same length then they must be the same code.
+  if (aCodes.length === bCodes.length) {
+    return 0;
+  }
+  // The shorter of two codes with otherwise matching numbers is sorted before the longer one.
+  // since we sort null before non-null (e.g 1.11 is before 1.11.12)
+  return aCodes.length < bCodes.length ? -1 : 1;
+};
+
+export const codeSorter = (codeOne, codeTwo) => {
```
Jsyro: i don't think this method meets the spec for a compare method. Why not have this method return +1, 0, or -1?

Jsyro: also, if both codeOne and codeTwo are null, this will say they aren't equal... it's nonsense, but just a thought for completeness's sake.

thegentlemanphysicist: turns out you were right and sorter accepts '0',  if both are null this just boils down to the initial check where 1 was returned if a>b.

Jsyro: Looks great.

PR #900, opened 2019-07-29
MDS-2153 Task/Finish incidents recommendations Part 2

NWCalvank:
This updates the BE with requested changes from Part 1 and updates the FE to use constants instead of magic strings. I've also moved the dropdown field requests out of `Dashboard.js` and into `MineIncidents.js`, which will help with first load performance and leverages the static asset caching added during the second Hackathon.

Part 3 will include adding the missing regression tests.

~**Note**~
~At the time of opening this PR, I am unable to test the document upload behaviour. This should be unaffected by my changes, meaning that the code is still ready for review, but we should ensure that document uploading is working as-expected on this branch before merging.~


frontend/src/components/modalContent/AddIncidentModal.js
```diff
@@ -156,8 +158,9 @@ const StepForms = (
         incidentStatusCodeOptions={props.incidentStatusCodeOptions}
         hasFatalities={props.addIncidentFormValues.number_of_fatalities > 0 || false}
         hasFollowUp={props.addIncidentFormValues.followup_inspection || false}
+        determinationTypeCode={props.addIncidentFormValues.determination_type_code}
```
thegentlemanphysicist: Was this prop missing?

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -310,15 +310,16 @@ def put(self, mine_guid, mine_incident_guid):
...
+            rec_string = recommendation.get('recommendation')
             if rec_string is None or rec_string == '':
-                if payload.get('mine_incident_recommendation_guid') is not None:
+                if recommendation.get('mine_incident_recommendation_guid') is not None:
                     # Remove existing empty recommendations
                     rec = MineIncidentRecommendation.find_by_mine_incident_recommendation_guid(
                         edited_rec_guid)
-                    rec.soft_delete()
+                    rec.deleted_ind = True
+                    rec.save()
```
thegentlemanphysicist: Did this mean the delete wasn't being saved before?

NWCalvank: This is no change in behaviour. I just removed the `soft_delete` method, as per feedback on the Part 1 PR

frontend/src/components/modalContent/AddIncidentModal.js
NWCalvank: I added it in order to conditionally show a warning message about an uncommon use of the incident form, as per UX feedback.

frontend/src/components/Forms/incidents/AddIncidentFollowUpForm.js
```diff
@@ -124,7 +132,11 @@ export class AddIncidentFollowUpForm extends Component {
                   id="InitialIncidentFileUpload"
                   name="InitialIncidentFileUpload"
                   onFileLoad={(document_name, document_manager_guid) =>
-                    this.props.onFileLoad(document_name, document_manager_guid, "FIN")
+                    this.props.onFileLoad(
+                      document_name,
+                      document_manager_guid,
+                      Strings.INCIDENT_DOCUMENT_TYPES.final
```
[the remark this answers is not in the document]
NWCalvank: You can see in on line 127 from the original version of the file that the type was `"FIN"`, which is now referenced by `Strings.INCIDENT_DOCUMENT_TYPES.final`, so there should be no change in behaviour here

PR #905, opened 2019-07-30
Task/mds 1689 document resource refactor 2

Jsyro:
Part 2
Required-documents stay at highest level (no relation to mines) 
MineDocuments gets lots of updates because that will be staying around
Anything related to expected documents simply had their files move and flaskrestplus marshalling as this will become deprecated very soon. 

Lots of .json and create_error_payload removed, only a few places left.


python-backend/app/api/mines/documents/expected/resources/expected_documents.py
```diff
@@ -3,17 +3,38 @@
 from datetime import datetime
 
 from flask import request
-from flask_restplus import Resource, reqparse
+from flask_restplus import Resource, reqparse, fields
```
JustinMacaulay: missed unused import?

Jsyro:
clean-up

python-backend/app/api/mines/documents/expected/resources/expected_documents.py
Jsyro: fixed.

PR #908, opened 2019-07-31
MDS-2125 Feature/ Show variance number

NWCalvank:


migrations/sql/V2019.07.31.10.08__add_variance_no.sql
```diff
@@ -0,0 +1,29 @@
+ALTER TABLE variance
+ADD COLUMN variance_no integer;
```
nayr974: Since this is a net new column I'm curious why you didn't define this as SERIAL, which I think would have had the same outcome as the statements below?

NWCalvank: I don't think that would have worked for this particular situation.

We don't want to show a variance number on historical variances because MMS has already assigned variance numbers, but we don't have those (they need to be added to CORE manually). So, those records have `null` variance_no.

From what I can tell, SERIAL populates the column as it creates the column. So everything would have a variance_no.

If we do that, we'll probably need a separate column to indicate if a variance is historical (create_timestamp before/on June 26th), since we don't expose create_timestamp via the API. This second option seemed like a bit of overkill, as we're not currently sure what/if we want to do about historical variance numbers.

Open to thoughts on this.

nayr974: Makes sense; I missed the create_timestamp clause. 👍

PR #911, opened 2019-08-01
Code Required Reports Edit and Delete

nayr974:
This PR adds editing and removing of the MVP code required reports functionality.


python-backend/app/api/mines/reports/models/mine_report.py
```diff
@@ -56,7 +59,7 @@ def create(cls,
     def find_all_by_mine_guid(cls, _id):
```
Jsyro: Can you remove `all` from the prefix in this method, now that it filters on deleted_ind (as it should). thanks. Not a blocker

nayr974: No.

PR #912, opened 2019-08-01
Trim some data constants

Jsyro:
marshal mine region endpoint.


Jsyro:
cleanup

migrations/sql/V2019.08.01.13.17__remove_region_suffix_from_mine_region.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE mine_region_code set description = 'South West' where description = 'South West Region';
```
nayr974: It makes sense in this case because we're updating description specifically from one to another, but it's generally best to use a PK

python-backend/app/api/constants.py
```diff
@@ -205,22 +205,6 @@
     }
 }
 
-
-# Status Code for party_type_code
-PARTY_STATUS_CODE = {
```
nayr974: 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯

python-backend/app/api/constants.py
```diff
@@ -292,31 +276,6 @@
     }
 }
 
-
-#Region Code
-MINE_REGION_OPTIONS = [
```
nayr974: 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯

PR #916, opened 2019-08-02
Feature/mds 2127

thegentlemanphysicist:
I have moved the variance table from the custom dashboard.  
The frontend of 2205 is mostly implemented however the search bar is wrapped in IN_Development".  
2127 was completed incidentally in the process of making 2205.
Since local doc uploads are failing for me I need to test this in the open shift pod

Fixed some props validation console errors
Updated the pencil edit icon used so it is consistent now


frontend/src/components/Forms/variances/VarianceSearchForm.js
```diff
@@ -0,0 +1,162 @@
...
+                  data={this.props.complianceCodes}
+                />
+              </Col>
+            </Row>
+            <Row gutter={6}>
+              <Col md={12} xs={24}>
+                <Field
+                  id="issue_date_min"
+                  name="issue_date_min"
+                  placeholder="Select Min Issue Date"
```
NWCalvank: Is "min" the "earliest" date to be included in the range of valid dates? Such that anything before this date is filtered out?

If so, something like "earliest" might be more intuitive than "min," but I'd suggest confirming this with UX if you haven't already

frontend/src/components/Forms/variances/VarianceSearchForm.js
```diff
@@ -0,0 +1,162 @@
...
+                  // validate={}
+                />
+              </Col>
+              <Col md={12} xs={24}>
+                <Field
+                  id="expiry_date_max"
+                  name="expiry_date_max"
+                  placeholder="Select Max Expiry Date"
+                  component={renderConfig.DATE}
+                  // validate={}
```
NWCalvank: There are a few of these commented out that should be removed or updated to include validators

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -0,0 +1,352 @@
...
+      search,
+      ...remainingParams
+    } = queryString.parse(params);
+    const format = (param) => (param ? param.split(",").filter((x) => x) : []);
+    this.setState({
+      params: {
+        issue_date_min,
+        issue_date_max,
+        expiry_date_max,
+        expiry_date_min,
```
NWCalvank: Any reason why these are broken-out? It looks like they can be left inside of `remainingParams`, since they're not being transformed (like `region` and `compliance_codes`)

frontend/src/selectors/staticContentSelectors.js
```diff
@@ -215,6 +215,8 @@ export const getDangerousOccurrenceSubparagraphOptions = createSelector(
       })
 );
 
+// TODO:  this seems to double count compliance codes, particularly the dangerous occurences
```
NWCalvank: Small semantic thing, but this sounds more like a FIXME 🙂

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: Good catch, fixed. :)

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -100,27 +101,34 @@ export class CustomHomePage extends Component {
...
     if (params) {
       this.renderDataFromURL(params);
     } else {
       this.props.history.push(
         router.VARIANCE_DASHBOARD.dynamicRoute({
-          page: String.DEFAULT_PAGE,
-          per_page: String.DEFAULT_PER_PAGE,
+          page,
+          per_page,
+          type,
```
NWCalvank: Am I missing something? It looks like `page`, `per_page`, and `type` are pulled out of `parsedParams`, which is created by parsing `params`. So, if `params` doesn't exist (which could cause this block to run), then `page`, `per_page`, and `type` would always be the default values (`this.state.params.whatever`).

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -145,24 +153,12 @@ export class CustomHomePage extends Component {
...
-      expiry_date_min,
-      region,
-      compliance_code,
-      major,
-      search,
-      ...remainingParams
-    } = queryString.parse(params);
+    const { region, compliance_code, major, search, ...remainingParams } = queryString.parse(
+      params
+    );
```
NWCalvank: This formatting is weird. `parse(params);` looks like it should all be on one line

thegentlemanphysicist: auto formatted by vs code

NWCalvank: Okay. Well, it doesn't look like Prettier formatting 🤷‍♂I wonder if this is related to the dependency update...

Getting any errors like this? https://github.com/prettier/prettier-vscode/issues/870

PR #918, opened 2019-08-02
Task/run migrations for developers

Jsyro:
microservices kick their own migrations when run from docker-compose.


docker-compose.yml
```diff
@@ -122,6 +122,7 @@ services:
     depends_on:
       - flyway
       - redis
+    entrypoint: ./dev-entrypoint.sh
```
pdittaro: As an alternative to this, you could create a new service in the docker-compose file that has the same context as this one but has the different cmd of 'flask db upgrade' and  a failure restart policy. And it kind of mirrors our flyway convention. And would also save having an extra file.

PR #920, opened 2019-08-06
Feature / Add and remove files from mine reports.

nayr974:


python-backend/app/api/mines/reports/resources/mine_report_document.py
```diff
@@ -0,0 +1,37 @@
...
+from app.api.mines.mine.models.mine import Mine
+from app.api.documents.mines.models.mine_document import MineDocument
+from app.api.mines.reports.models.mine_report import MineReport
+
+from app.extensions import api, db
+from app.api.utils.custom_reqparser import CustomReqparser
+from ....utils.access_decorators import requires_role_edit_do
+from ....utils.resources_mixins import UserMixin, ErrorMixin
+from ....utils.url import get_document_manager_svc_url
+
```
nayr974: I should really prune these imports.

nayr974: I pruned the shit out of these imports.

frontend/src/components/Forms/reports/ReportFileUpload.js
```diff
@@ -0,0 +1,34 @@
...
+  allowMultiple: PropTypes.bool.isRequired,
+};
+
+export const ReportFileUpload = (props) => (
+  <Form.Item>
+    <Field
+      id="fileUpload"
+      name="fileUpload"
+      component={FileUpload}
+      uploadUrl={`${MINE_INCIDENT()}/documents?mine_guid=${props.mineGuid}`}
```
Jsyro: This url looks weird. should it be `mine/<guid>/reports/documents` which should become a constant in the API file.

JustinMacaulay: That looks weird because it shouldn't be here that component was going to be used but was never used. I have removed the whole component.

PR #923, opened 2019-08-08
Feature/mds 2092 crr default due date

Jsyro:
default_due_date is served with the report definitions as long as due_date in untouched on the form. it will set the due_date to the default_due_date.


frontend/src/components/Forms/reports/AddReportForm.js
```diff
@@ -214,6 +236,8 @@ export default compose(
     mineReportDefinitionOptions: getMineReportDefinitionOptions(state),
     selectedMineReportCategory: selector(state, "mine_report_category"),
     selectedMineReportDefinition: selector(state, "mine_report_definition_guid"),
+    currentDueDate: selector(state, "due_date"),
+    formMeta: state.form.ADD_REPORT,
```
Jsyro: Trying to find a better way of getting the form meta. but i never got this to work. 

https://redux-form.com/7.0.2/docs/api/selectors.md/#-code-getformmeta-formname-string-code-returns-code-state-gt-formmeta-object-code-

frontend/src/components/Forms/reports/AddReportForm.js
```diff
@@ -90,6 +92,25 @@ export class AddReportForm extends Component {
     }));
   };
 
+  updateDueDateWithDefaultDueDate = (selectedMineReportDefinition) => {
+    let formMeta = this.props.formMeta;
+    if (
+      !(
+        formMeta &&
+        formMeta.fields &&
```
pdittaro: You probably don't need to safe traverse .fields?

frontend/src/components/Forms/reports/AddReportForm.js
```diff
@@ -90,6 +92,25 @@ export class AddReportForm extends Component {
...
+        formMeta.fields.due_date &&
+        formMeta.fields.due_date.touched == true
+      )
+    ) {
+      this.props.change(
+        "due_date",
+        this.props.mineReportDefinitionOptions.find(
+          (x) => x.mine_report_definition_guid == selectedMineReportDefinition
```
NWCalvank: Please use strict equality (`===`). This should be causing a linting error. Are you not seeing errors?

Jsyro: i'm getting this `Failed to load plugin 'react' declared in 'frontend-public\.eslintrc.js': Cannot find module 'eslint-plugin-react'`. and the pre-hook is saying on .js files found.

NWCalvank: Oh. The same regression that @mmumby is having. I didn't realize that it was wide-spread. It's a result of the dependency upgrade. I'll push the priority up on this fix. Thanks for replying with the details

PR #933, opened 2019-08-14
Feature/mds 2245 migrate tailings to crr

Jsyro:
add command that runs petl script to convert tailings documents to core required_reports .


python-backend/app/api/mines/reports/resources/mine_reports.py
```diff
@@ -81,14 +82,11 @@ def post(self, mine_guid):
...
 
         submissions = data.get('mine_report_submissions')
-        if submissions is not None:
-            report_submission_guid = uuid.uuid4()
+        if submissions:
             submission = submissions[0]
             if len(submission.get('documents')) > 0:
-                report_submission = MineReportSubmission(
-                    mine_report_submission_guid=report_submission_guid,
-                    mine_report_submission_status_code='MIA',
```
pdittaro: Why do we need this status on MineReportSubmission; or - at least - why are we requiring it to be provided in this create?

python-backend/app/commands.py
```diff
@@ -75,3 +77,11 @@ def _create_data(num):
     def run_etl():
         from app.cli_jobs import ETL_jobs
         ETL_jobs.run_ETL()
+
+    @app.cli.command()
+    @click.option('-c', '--commit')
+    def run_append_tailings_reports_to_code_required_reports_then_destroy_tailings_data(commit):
```
pdittaro: This is the most monstrous command line argument I've ever read. The extended description for this function should be provided in a pydoc/click description.

Also, lets split this into two commands just to be safe. Migrate *then* truncate.

python-backend/app/api/mines/reports/resources/mine_reports.py
Jsyro: That's a good point, not sure the origin of that decision. @JustinMacaulay or @nayr974 ?

python-backend/app/commands.py
Jsyro: Given that we can nuke the CRR data if something goes wrong, that's a good idea, i didn't think there was a recoverable state inbetween the two steps.

python-backend/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,4 +1,6 @@
 import uuid
+from flask import current_app
```
FreshGyan: Unused import?

python-backend/app/api/mines/reports/models/mine_report.py
```diff
@@ -74,3 +73,15 @@ def find_by_mine_report_guid(cls, _id):
...
             return None
+
+    @classmethod
+    def find_by_mine_guid_with_category(cls, _id, category):
+        try:
+            uuid.UUID(_id, version=4)
+            reports = cls.query.filter_by(mine_guid=_id).all()
+            return [
+                r for r in reports if category.upper() in
+                [c.mine_report_category.upper() for c in r.mine_report_definition.categories]
```
FreshGyan: If i read this correctly, is this:
Return a list of reports whose category belongs to one or more of the categories this mine report belongs in in the list of mine report definition categories?

I may be wrong but the inner list may get formed and looped over for every single report and you can probably avoid that by defining the inner list outside of the return statement and save some iterative cycles.

python-backend/app/api/mines/reports/resources/mine_reports.py
```diff
@@ -47,7 +46,12 @@ class MineReportListResource(Resource, UserMixin):
     @api.doc(description='returns the reports for a given mine.')
     @requires_role_view_all
     def get(self, mine_guid):
-        mine_reports = MineReport.find_by_mine_guid(mine_guid)
+        mrd_category = request.args.get('mine_report_definition_category')
+        if mrd_category:
+            mine_reports = MineReport.find_by_mine_guid_with_category(mine_guid, mrd_category)
```
FreshGyan: Maybe nitpicky but is mine_guid_and_category more clearer than mine_guid_with_category? Just up for thought

python-backend/app/commands.py
```diff
@@ -75,3 +77,17 @@ def _create_data(num):
     def run_etl():
         from app.cli_jobs import ETL_jobs
         ETL_jobs.run_ETL()
+
+    @app.cli.command()
+    @click.option('-c', '--commit')
+    def run_tailings_reports_migration(commit):
```
FreshGyan: I'd love a technical demo on petl in the next sprint ceremomy if you can :D

python-backend/app/api/mines/reports/models/mine_report.py
Jsyro: good catch.

python-backend/app/api/mines/reports/resources/mine_reports.py
Jsyro: i like your name.

python-backend/app/api/mines/reports/models/mine_report.py
Jsyro: I think that's true, I took the first pass and got something that was easier to follow. Can be improved in the future.

pdittaro: No, the `for c in r.mine_report_definition.categories` is generating the list of categories of the selected report, not the list of categoreis to filter by. 

This is a pain to read, but I think its right. Its doing this:
```
for report in reports
   for report_category in report.categories
       return report_category == filter_category
```

PR #935, opened 2019-08-15
Feature/mds 2205

thegentlemanphysicist:
Created the new variances dashboard
Can search and filter variances
extended the endpoint to handle additional filtering
Created front and backend tests to handle it

Two existing issues:
~~If a filtered page is refreshed, clear does not reset the form (same issue is on the contacts search page)~~ FIXED
The variance tables do not sort, I would like to put this in a different PR since it isn't needed on this functionality
The status filter should be moved up to the be with the other filters.


frontend/src/components/Forms/variances/VarianceSearchForm.js
```diff
@@ -20,7 +20,7 @@ const defaultProps = {
   isAdvanceSearch: false,
 };
 
-const validate = (values) => {
+export const validate = (values) => {
   const errors = {};
```
mmumby: Let's avoid exporting methods in this way, If two files are using the same validate function, move it to a common resource and import it into both files.

frontend/src/components/dashboard/contactsHomePage/ContactSearch.js
```diff
@@ -51,6 +51,7 @@ export class ContactSearch extends Component {
                   ...this.props.partyRelationshipTypesList,
                 ]}
                 initialValues={this.props.initialValues}
+                enableReinitialize
```
mmumby: This should be added in the export like so: 
```
export default reduxForm({
  form: FORM.CONTACT_ADVANCED_SEARCH,
  touchOnBlur: false,
  enableReinitialize: true
})(AdvancedContactSearchForm);
```

Makes it easier to track what the form is enabling/disabling

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -68,67 +66,69 @@ const propTypes = {
...
-  ...remainingParams,
-});
-
+export const joinOrRemove = (param, key) => {
+  if (isEmpty(param)) {
+    return {};
+  }
+  return typeof param === "string" ? { [key]: param } : { [key]: param.join(",") };
+};
+export const removeEmptyStings = (param, key) => (isEmpty(param) ? {} : { [key]: param });
```
NWCalvank: What are all of these changes for?

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -68,67 +66,69 @@ const propTypes = {
...
     const parsedParams = queryString.parse(params);
-    const {
-      page = this.state.params.page,
-      per_page = this.state.params.per_page,
-      type = this.state.params.type,
-    } = parsedParams;
+    const { page = this.state.params.page, per_page = this.state.params.per_page } = parsedParams;
     if (params) {
-      this.renderDataFromURL(params);
+      this.renderDataFromURL();
```
NWCalvank: Reason for removing this? Did it fix a bug? I imagine it was there for a reason

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -145,44 +145,67 @@ export class VarianceHomePage extends Component {
...
     this.handleVarianceSearchDebounced.cancel();
     this.setState({
       params: {},
     });
   }
 
-  renderDataFromURL = (params) => {
+  renderDataFromURL = (nextProps) => {
+    const params = nextProps ? nextProps.location.search : this.props.location.search;
     const { region, compliance_code, major, search, ...remainingParams } = queryString.parse(
```
mmumby: the params names are confusing here - `nextProps` should only be used in `componentsWillReceiveProps`, as the data passed into this function is only params not all of the available props which this implies. 

You do a check for params in `componentWillReceiveProps`, would be better to pass them in at that point since `params` already  equal `this.props.location.search`

Then the function will already accept the proper params and this check will no longer be needed. 
`renderDataFromURL = (params) `

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -68,67 +66,69 @@ const propTypes = {
...
+    const formatedParams = {
+      compliance_code: formatParamStringToArray(this.params.compliance_code),
+      region: formatParamStringToArray(this.params.region),
+      major: this.params.major,
+      search: this.params.search,
+      issue_date_after: this.params.issue_date_after,
+      issue_date_before: this.params.issue_date_before,
+      expiry_date_before: this.params.expiry_date_before,
+      expiry_date_after: this.params.expiry_date_after,
+    };
```
NWCalvank: Is there a reason for pulling this out into a variable?

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -145,44 +145,67 @@ export class VarianceHomePage extends Component {
...
-        per_page: this.state.params.per_page ? this.state.params.per_page : String.DEFAULT_PER_PAGE,
-      })
+    this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(updatedParams));
+    this.setState(
+      {
+        params: updatedParams,
+      },
+      // Fetch parties once state has been updated
+      () => {
+        return this.renderDataFromURL();
```
mmumby: `this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(updatedParams));`
this will trigger `ComponentWillReceiveProps` to detect a change in the location, which will invoke `renderdataFromUrl`, so the last return is not necessary.

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -68,67 +66,69 @@ const propTypes = {
...
+}) => {
+  return {
+    ...joinOrRemove(region, "region"),
+    ...joinOrRemove(compliance_code, "compliance_code"),
+    ...removeEmptyStings(issue_date_after, "issue_date_after"),
+    ...removeEmptyStings(issue_date_before, "issue_date_before"),
+    ...removeEmptyStings(expiry_date_before, "expiry_date_before"),
+    ...removeEmptyStings(expiry_date_after, "expiry_date_after"),
+    ...removeEmptyStings(search, "search"),
+    ...removeEmptyStings(major, "major"),
```
NWCalvank: What makes these different from the `remainingParams` that causes them to need to be handled by `removeEmptyStrings`?

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -280,15 +303,14 @@ export class VarianceHomePage extends Component {
           <h1>Browse Variances</h1>
         </div>
         <div className="landing-page__content">
-          <AuthorizationWrapper permission={Permission.IN_TESTING}>
```
mmumby: Does this PR include the sorting? We shouldn't put this in prod until the table has sorting

frontend/src/components/dashboard/varianceHomePage/VarianceSearch.js
```diff
@@ -72,6 +64,7 @@ export class VarianceSearch extends Component {
                 initialValues={this.props.initialValues}
                 complianceCodes={this.props.complianceCodes}
                 mineRegionOptions={this.props.mineRegionOptions}
+                enableReinitialize
               />
```
mmumby: Same here, move this into the default export of the form

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -145,44 +145,67 @@ export class VarianceHomePage extends Component {
...
-    });
+      () => {
+        this.props.fetchVariances(this.state.params);
+      }
+    );
   };
 
   handleVarianceSearch = (searchParams, clear = false) => {
     const formattedSearchParams = formatParams(searchParams);
+    // const formattedSearchParams = clear ? {} : formatParams(searchParams);
```
NWCalvank: Unused code 🙂

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -145,44 +145,67 @@ export class VarianceHomePage extends Component {
...
+    const updatedParams = {
+      // Start from existing state
+      ...persistedParams,
+      // Overwrite prev params with any newly provided search params
+      ...formattedSearchParams,
+      // Reset page number
+      page: String.DEFAULT_PAGE,
+      // Retain per_page if present
+      // eslint-disable-next-line react/no-access-state-in-setstate
+      per_page: this.state.params.per_page ? this.state.params.per_page : String.DEFAULT_PER_PAGE,
```
NWCalvank: Please follow the linting rule instead of disabling it. I can chime in if you're not sure how to address it

frontend/src/components/dashboard/varianceHomePage/VarianceSearch.js
```diff
@@ -10,15 +10,7 @@ import CustomPropTypes from "@/customPropTypes";
...
-  initialValues: PropTypes.shape({
-    region: PropTypes.arrayOf(PropTypes.string),
-    compliance_code: PropTypes.arrayOf(PropTypes.string),
-    major: PropTypes.string,
-    issue_date_after: PropTypes.string,
-    issue_date_before: PropTypes.string,
-    expiry_date_before: PropTypes.string,
-    expiry_date_after: PropTypes.string,
-  }),
+  initialValues: CustomPropTypes.varianceSearchInitialValues,
```
NWCalvank: That's probably not a propType worth storing, imo. But I haven't given it a ton of thought

frontend/src/tests/components/Forms/variances/VarianceSearchForm.spec.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { shallow } from "enzyme";
+import { VarianceSearchForm, validate } from "@/components/Forms/variances/VarianceSearchForm";
+
+const dispatchProps = {};
+const props = {};
+// TODO FIX PROPS
```
NWCalvank: Still a TODO?

frontend/src/components/dashboard/contactsHomePage/ContactSearch.js
thegentlemanphysicist: done

frontend/src/components/dashboard/varianceHomePage/VarianceSearch.js
thegentlemanphysicist: done

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: removed

frontend/src/tests/components/Forms/variances/VarianceSearchForm.spec.js
thegentlemanphysicist: removed

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: I thought eslint was misfiring because I wasn't setting state their, having looked at it again I realized the set state was being invoked two lines below. X( Will fix now.

thegentlemanphysicist: Easy fix in the end, feel kind of silly now.

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: I wouldn't have thought of this, implemented.

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: I had put the 'params' in a previous pr because I had misread the logic. passing the params ended up being redundant.

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: done

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: done

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: There was some edge case where a url param 'expiry_date_after=' was getting pushed up.  This keeps the param from being pushed up

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: I think this was legacy of trying to fix the clear bug.  I have put it back

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: This dropdown selection list was throwing an error when a single item was selected.  It was passing in a string '50' instead of and arrays ['50','39'], the param.join was causing problems.  This checks for a string getting put in

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
...
+        regional_mine = MineFactory(major_mine_ind=False)
+
+        VarianceFactory(mine=major_mine)
+        VarianceFactory(mine=regional_mine)
+        get_resp = test_client.get(f'/variances?major=f', headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+        assert get_resp.status_code == 200
+        assert all(map(
+            lambda v: v['mine_name'] == regional_mine.mine_name,
+            get_data['records']))
```
NWCalvank: Same comment here about only returning 1 record

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
...
+
+    def test_get_variances_application_filter_by_compliance_code(self, test_client, db_session, auth_headers):
+        """Should filter variances by compliance code"""
+        compliance_codes = [54,66]
```
NWCalvank: You can just generate 2 random compliance codes, which is better than hardcoding them

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
...
+        VarianceFactory(approved=True,issue_date=date_plus_7_days)
+        dateFormat = "%Y-%m-%d"
+        get_resp = test_client.get(
+            f'/variances?issue_date_before={date_plus_1_day.strftime(dateFormat)}&issue_date_after={date_minus_1_day.strftime(dateFormat)}',
+            headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+        assert get_resp.status_code == 200
+        assert all(map(
+            lambda v: v['issue_date'] == date_today.strftime(dateFormat),
+            get_data['records']))
```
NWCalvank: Looks like another single record in the response?

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
...
+        VarianceFactory(approved=True,expiry_date=date_plus_7_days)
+
+        get_resp = test_client.get(
+            f'/variances?expiry_date_before={date_plus_1_day.strftime(dateFormat)}&expiry_date_after={date_minus_1_day.strftime(dateFormat)}',
+            headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+        assert get_resp.status_code == 200
+        assert all(map(
+            lambda v: v['expiry_date'] == date_today.strftime(dateFormat),
+            get_data['records']))
```
NWCalvank: Last one that appears to be returning a single record

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,107 @@ def get(self):
...
+                        'model': 'Mine',
+                        'field': 'mine_no',
+                        'op': 'ilike',
+                        'value': '%{}%'.format(search_term)
+                    },
+                )
+            conditions.append(
+                {
+                    'or': search_conditions
+                })
```
NWCalvank: Consider checking out how I handled the "or" condition for search terms in the NOW Applications. There's a `or_` function that works really well.

https://github.com/bcgov/mds/pull/936/files#diff-c3f49332dad6bb23875dae032ba7ea71R83

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,107 @@ def get(self):
...
+              'value': args["issue_date_after"]
+            })
+
+        if args["major"] is not None:
+            conditions.append({
+                'model': 'Mine',
+                'field': 'major_mine_ind',
+                'op': '==',
+                'value': args["major"]
+            })
```
NWCalvank: I think you can reduce the bloat on these in a couple different ways. You could either:
1. Refactor this to use the query object directly. Something like:
```python
if args["major"] is not None:
    conditions.append(Mine.major_mine_ind=args["major"])
```
or
2. Abstract this into a separate method (I'm calling it `_build_filter` here) that's invoked something like:
```python
if args["major"] is not None:
    conditions.append(cls._build_filter('Mine', 'major_mine_ind', '==', args))
```

I prefer option 1, personally. If you go with 2, you'll want to name the args the same as the columns, which is probably better for readability anyway. It would look something like:
```python
@classmethod 
def _build_filter(cls, model, field, op, args):
    return {
                'model': model,
                'field': field,
                'op': op,
                'value': args[field]
    }
```

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -23,14 +22,32 @@ class VarianceResource(Resource, UserMixin, ErrorMixin):
...
             'variance_application_status_code':
             'Comma-separated list of code statuses to include in results. Default: All status codes.',
-        })
+            'compliance_codes':'Comma-separated list of compliance codes to be filtered. Default: All compliance codes',
+            'major': 'boolean indicating if variance is from a major or regional mine',
+            'region': 'Comma-separated list of regions the mines associated with the variances are located in',
+            'issue_date_before': 'Latest possible issue date returned',
+            'issue_date_after': 'Earliest possible issue date returned',
+            'expiry_date_before': 'Latest possible expiry date returned',
+            'expiry_date_after': 'Earliest possible expiry date returned'})
```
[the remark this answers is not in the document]
thegentlemanphysicist: inclusive, done

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,107 @@ def get(self):
...
         status_filter_values = list(map(
             lambda x: x.variance_application_status_code,
             VarianceApplicationStatusCode.active()))
 
-        if application_status is not None:
-            status_filter_values = application_status.split(',')
+        if args["page_number"] is None:
+            args["page_number"]=PAGE_DEFAULT
+        if args["page_size"] is None:
+            args["page_size"] = PER_PAGE_DEFAULT
```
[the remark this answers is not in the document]
thegentlemanphysicist: deleted

python-backend/tests/factories.py
```diff
@@ -211,6 +211,7 @@ class Params:
 
     variance_guid = GUID
     compliance_article_id = factory.LazyFunction(RandomComplianceArticleId)
+
```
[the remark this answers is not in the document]
thegentlemanphysicist: removed

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
...
+        mineWith_region_NW = MineFactory(mine_region='NW')
+        mineWith_region_SW = MineFactory(mine_region="SW")
+        batch_size = 3
+        VarianceFactory.create_batch(size=batch_size)
+        VarianceFactory(mine=mineWith_region_NW)
+        VarianceFactory(mine=mineWith_region_SW)
+
+        get_resp = test_client.get(f'/variances?region={region_code}',headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+        print(get_data)
```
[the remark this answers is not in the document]
thegentlemanphysicist: removed

python-backend/tests/variances/resources/test_variance_resource.py
```diff
@@ -68,3 +69,93 @@ def test_get_variances_application_status_filter_list(self, test_client, db_sess
         assert all(map(
             lambda v: v['variance_application_status_code'] in status_codes,
             get_data['records']))
+
+    def test_get_variances_application_filter_by_mine_region(self, test_client, db_session, auth_headers):
+        """Should filter variances by mine region"""
+        region_code = "NW"
+        mineWith_region_NW = MineFactory(mine_region='NW')
+        mineWith_region_SW = MineFactory(mine_region="SW")
```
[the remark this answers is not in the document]
thegentlemanphysicist: snake ate a camel, removeing

python-backend/tests/variances/resources/test_variance_resource.py
thegentlemanphysicist: done

python-backend/tests/variances/resources/test_variance_resource.py
thegentlemanphysicist: set it to compare the single result instead

python-backend/tests/variances/resources/test_variance_resource.py
thegentlemanphysicist: set it to compare the single result instead

python-backend/app/api/variances/resources/variance_resource.py
thegentlemanphysicist: I've implemented and tested this idea

python-backend/app/api/variances/resources/variance_resource.py
thegentlemanphysicist: So I picked the convention I did because it matched the one that had already been implemented in this resource.  I can't see a way of implementing or_ without rewriting the entire pattern on this resource.

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -23,14 +22,32 @@ class VarianceResource(Resource, UserMixin, ErrorMixin):
...
+            'expiry_date_before': 'Latest possible expiry date returned (inclusive)',
+            'expiry_date_after': 'Earliest possible expiry date returned (inclusive)'})
     @requires_any_of([VIEW_ALL])
     @api.marshal_with(PAGINATED_VARIANCE_LIST, code=200)
     def get(self):
-        records, pagination_details = self._apply_filters_and_pagination(
-            page_number=request.args.get('page', PAGE_DEFAULT, type=int),
-            page_size=request.args.get('per_page', PER_PAGE_DEFAULT, type=int),
-            application_status=request.args.get('variance_application_status_code', type=str))
+        args={
```
FreshGyan: I'm not so sure the args dict needs to be there, it seems to me that you can just pass down request.args to apply_filters_pagination method directly and access values there.

Alternatively, if you're unpacking all the args beforehand, you can also pass them as an individual parameter to the function.

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,70 @@ def get(self):
             'total': pagination_details.total_results,
         }
 
+    @classmethod
```
FreshGyan: Should this be a staticmethod instead? Doesn't look like you're using the class instance at all.

Or even a regular internal method like the filter method

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,70 @@ def get(self):
...
+                search_conditions.append(
+                    self._build_filter('Mine', 'mine_no', 'ilike', '%{}%'.format(search_term)))
+
+            conditions.append({'or': search_conditions})
+
+        if args["region"] is not None:
+            region_list = args["region"].split(',')
+            conditions.append(self._build_filter('Mine', 'mine_region', 'in', region_list))
+
+        query = Variance.query.join(Mine)
```
FreshGyan: You can probably merge this and next line together, doesn't seem like query object is complicated enough to be its own thing

python-backend/app/api/variances/resources/variance_resource.py
```diff
@@ -43,24 +60,70 @@ def get(self):
...
         status_filter_values = list(map(
             lambda x: x.variance_application_status_code,
             VarianceApplicationStatusCode.active()))
 
-        if application_status is not None:
-            status_filter_values = application_status.split(',')
+        conditions = []
+        if args["application_status"] is not None:
+            status_filter_values = args["application_status"].split(',')
+            conditions.append(self._build_filter('Variance' , 'variance_application_status_code', 'in',  status_filter_values))
```
FreshGyan: FYI you can also do these filtering using native sql-alchemy provided functions directly on the model like in here, just to keep everything following the same pattern.
https://github.com/bcgov/mds/pull/936/files#diff-c3f49332dad6bb23875dae032ba7ea71R67

I do see the file already was using the filter helper method so i guess we dont have a hard set pattern for this kind of stuff yet so maybe that should be something to talk about 🤔

PR #936, opened 2019-08-15
MDS-2174 Feature/ Add filtering & sorting to NOW Applications API

NWCalvank:
Also updates the view to reference the data provided by the API, and updates the NOWApplicationFactory to seed this data.


python-backend/app/api/now_submissions/resources/application_list_resource.py
```diff
@@ -20,14 +20,24 @@ class ApplicationListResource(Resource, UserMixin, ErrorMixin):
         description='Get a list of applications. Order: receiveddate DESC',
         params={
             'page': f'The page number of paginated records to return. Default: {PAGE_DEFAULT}',
-            'per_page': f'The number of records to return per page. Default: {PER_PAGE_DEFAULT}'
+            'per_page': f'The number of records to return per page. Default: {PER_PAGE_DEFAULT}',
+            'status': 'Comma-separated list of statuses to include in results. Default: All statuses.',
+            'noticeofworktype': 'Substring to match with a NoW\s type',
```
thegentlemanphysicist: is there a reason we don't follow the convention notice_of_work_type like we do for other multy word filters?

python-backend/app/api/now_submissions/response_models.py
```diff
@@ -307,27 +307,12 @@
...
-        'bcexplosivespermitexpiry': fields.DateTime,
-        'sandgrvqryalrpermitnumber': fields.String,
-        'freeusepermit': fields.String,
-        'vfcbcapplicationurl': fields.String,
-        'messagecreateddate': fields.DateTime,
-        'processed': fields.String,
-        'processeddate': fields.DateTime,
-        'nrsosapplicationid': fields.String,
+        'noticeofworktype': fields.String,
+        'trackingnumber': fields.Integer,
```
thegentlemanphysicist: same as above comment

python-backend/app/api/now_submissions/resources/application_list_resource.py
NWCalvank: Because this is the column name. It was pulled over from NROS

thegentlemanphysicist: fair enough

PR #938, opened 2019-08-16
MDS-2174 Feature/Add NOW Applications list view

NWCalvank:
Adds the data (sans Region, which we don't know yet) and functionality to the NOW Application view, including sorting/filtering by columns & searching by mine name/number.

Small backend change to support filtering NOW Applications by mine name/number.


frontend/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
```diff
@@ -12,20 +16,99 @@ import { getNoticeOfWorkList, getNoticeOfWorkPageData } from "@/selectors/notice
...
+    }
   }
 
+  componentWillUnmount() {
+    this.setState({ params: {} });
+  }
+
+  renderDataFromURL = (nextProps) => {
+    const params = nextProps ? nextProps.location.search : this.props.location.search;
+    const parsedParams = queryString.parse(params);
```
mmumby: Wrote the same feedback on Jons PR, copying it here, assuming you both copied this logic from `ContactSearch`: 

the params names are confusing here - nextProps should only be used in componentsWillReceiveProps, as the data passed into this function only needs to be params not all of the available props which `nextProps` implies.

You do a check for params in `componentDidMount`, It would be better to pass params in at that point since params already equal this.props.location.search
Then the `renderDataFromURL` will already accept the proper params and this check will no longer be needed.

frontend/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
```diff
@@ -12,20 +16,99 @@ import { getNoticeOfWorkList, getNoticeOfWorkPageData } from "@/selectors/notice
...
+
+    this.props.history.push(router.NOTICE_OF_WORK_APPLICATIONS.dynamicRoute(updatedParams));
+    this.setState(
+      {
+        params: updatedParams,
+      },
+      // Fetch parties once state has been updated
+      () => this.renderDataFromURL()
+    );
+  };
```
mmumby: Same thing on Jon's PR: 

`this.props.history.push(router.NOTICE_OF_WORK_APPLICATIONS.dynamicRoute(updatedParams));`
Will trigger ComponentWillReceiveProps to detect a change in the location, which will invoke `renderdataFromUrl`, so the last return shouldnt be needed.

frontend/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
mmumby: chatted IRL

frontend/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
mmumby: Chatted IRL

PR #940, opened 2019-08-19
Feature/mds 2189 2

mmumby:
Putting what we currently have for NoW applications `inTesting` - Will update the View with more data in the following PR. 

What we be in test after this gets merged: 
- 'Browse Notice Of Work' view, with a paginated list of NoW applications - the ability to sort and filter (or search in appropriate fields)
- NoW application view


frontend/src/customPropTypes/noticeOfWork.js
```diff
@@ -7,8 +7,152 @@ export const nowApplication = shape({
   mine_guid: PropTypes.string,
   mine_name: PropTypes.string,
   minenumber: PropTypes.string,
-  noticeofworktype: PropTypes.string,
-  trackingnumber: PropTypes.number,
+  trackingnumber: PropTypes.string,
```
NWCalvank: The API is setup with `'trackingnumber': fields.Integer,`, so this should be a number

mmumby: oops, Didn't intend to delete that one

PR #941, opened 2019-08-19
Add region to Application List view

NWCalvank:
We've been told that the NOW Region is just the mine region. This adds that.


frontend/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkTable.js
```diff
@@ -60,6 +49,20 @@ const applySortIndicator = (_columns, field, dir) =>
   }));
 
 export class NoticeOfWorkTable extends Component {
+  transformRowData = (applications) =>
```
thegentlemanphysicist: Is there a reason this is moved inside the class?

NWCalvank: It uses `this.props`, and since it's already a class component, I just decided to move the function onto the class instead of passing the params around.

PR #949, opened 2019-08-21
MDS-2212 Feature/ Add GET mines/{mine_guid}/now-submissions/applications

NWCalvank:
Backend changes to support the FE Mine NOW Applications view, which is a table of NOW Applications (filtered by mine) on the Mine view


python-backend/tests/mines/now_submissions/resources/test_application_resource.py
```diff
@@ -0,0 +1,118 @@
...
+
+        get_resp = test_client.get(
+            f'mines/{mine.mine_guid}/now-submissions/applications?status=app,rej&noticeofworktype=dog&trackingnumber=1',
+            headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+
+        assert get_resp.status_code == 200
+        assert len(get_data['records']) == 1
+        assert all(
+            str(application.application_guid) in map(lambda x: x['application_guid'], get_data['records'])
```
thegentlemanphysicist: Is this necessary if we're getting a single result?

python-backend/app/api/mines/namespace/mines.py
```diff
@@ -1,27 +1,27 @@
...
-from ..mine.resources.mine_disturbance_code import MineDisturbanceCodeResource
-from ..mine.resources.mine_commodity_code import MineCommodityCodeResource
-from ..status.resources.status import MineStatusResource, MineStatusListResource
-from ..region.resources.region import MineRegionResource
-from ..tailings.resources.tailings import MineTailingsStorageFacilityListResource
-from ..documents.mines.resources.mine_document_resource import MineDocumentListResource
-from ..compliance.resources.compliance import MineComplianceSummaryResource
-from ..compliance.resources.compliance_article import ComplianceArticleResource
-from ..mine.resources.mine_basicinfo import MineBasicInfoResource
+from app.api.mines.mine.resources.mine import MineResource, MineListSearch, MineListResource
```
thegentlemanphysicist: Are we now following the pattern of importing from the root as much a s possible.

python-backend/app/api/mines/now_submissions/resources/application_resource.py
```diff
@@ -0,0 +1,72 @@
...
+        if noticeofworktype is not None:
+            filters.append(func.lower(Application.noticeofworktype).contains(func.lower(noticeofworktype)))
+        if trackingnumber is not None:
+            filters.append(Application.trackingnumber == trackingnumber)
+
+        status_filter_values = []
+        if status is not None:
+            status_filter_values = status.split(',')
+
+        if len(status_filter_values) > 0:
```
thegentlemanphysicist: Should this be indented one?

python-backend/app/api/mines/namespace/mines.py
NWCalvank: I just checked the change history. More recent changes were using the absolute path. So, I updated them all to be the same=

python-backend/app/api/mines/now_submissions/resources/application_resource.py
NWCalvank: Nope. `status_filter_values` is initialized as an empty list (in other words, no filters applied) at the same scope level as this line. It's modified on line 64 if status exists. And then this line checks if that modification occurred (ie. if any status filters need to be applied).

python-backend/tests/mines/now_submissions/resources/test_application_resource.py
NWCalvank: Not necessary, but it's consistent with all of the other tests in this suite, which I think makes it easier to read than it would be if one assertion was structured differently.

PR #952, opened 2019-08-22
added new code required reports migration.

JustinMacaulay:
Added 4 new code required reports to the definitions table.


pdittaro:
Very edifying PR regarding updating code tables.

PR #960, opened 2019-08-27
MDS-2131 Feature/ Split date and time inputs

NWCalvank:
- Splits the `reported_timestamp` and `incident_timestamp` values into Date and Time form fields
- Joins the Date and Time form fields into timestamp fields in the add/edit payload
- No default values are provided for those fields
- All fields are required


frontend/src/tests/components/common/RenderTime.spec.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import { shallow } from "enzyme";
+import moment from "moment";
+import RenderTime from "@/components/common/RenderTime";
+
+let props = {};
+
+const setupProps = () => {
```
thegentlemanphysicist: Is it possible to get a unit test on the new functions like parseFormDataIntoPayload?  They seem like a worthwhile place to add some coverage

PR #969, opened 2019-09-04
MDS-2277 Feature/ Add mine determination fields to Mine Incident

NWCalvank:


frontend/src/components/modalContent/ViewIncidentModal.js
```diff
@@ -135,12 +135,25 @@ export class ViewIncidentModal extends Component {
           </div>
         )}
         <div className="inline-flex padding-small">
-          <p className="field-title">Who made the determination?</p>
+          <p className="field-title">Inspector who made the determination?</p>
```
pdittaro: This isn't a question. E.g. Either "Which inspector made the determination?" or "Inspector who made the determination"

frontend/src/tests/components/modalContent/__snapshots__/ViewIncidentModal.spec.js.snap
```diff
@@ -201,7 +201,31 @@ exports[`ViewIncidentModal renders properly 1`] = `
         <p
           className="field-title"
         >
-          Who made the determination?
+          Inspector who made the determination?
```
pdittaro: Similarly here, not a question.

frontend/src/tests/components/modalContent/__snapshots__/ViewIncidentModal.spec.js.snap
```diff
@@ -201,7 +201,31 @@ exports[`ViewIncidentModal renders properly 1`] = `
...
+          N/A
+        </p>
+      </div>
+      <div
+        className="inline-flex padding-small"
+      >
+        <p
+          className="field-title"
+        >
+          Mine representative who made determination?
```
pdittaro: Similarly here, not a question.

python-backend/app/api/incidents/models/mine_incident.py
```diff
@@ -70,11 +70,16 @@ class MineIncident(AuditMixin, Base):
         db.ForeignKey(
             'mine_incident_followup_investigation_type.mine_incident_followup_investigation_type_code'
         ))
+    mine_determination_type_code = db.Column(
```
pdittaro: Should this be `incident_determination_type_code` rather than `mine_determination_type_code`?

frontend/src/components/modalContent/ViewIncidentModal.js
NWCalvank: I agree. I'll run it by UX and get these all updated

python-backend/app/api/incidents/models/mine_incident.py
NWCalvank: It's being used to indicate the determination that a representative at the mine made (in contrast to an inspector -- which is held in the column `determination_type_code`), so I felt that this name reflected that.

frontend/src/components/Forms/incidents/AddIncidentFollowUpForm.js
```diff
@@ -138,7 +138,7 @@ export class AddIncidentFollowUpForm extends Component {
                 <Field
                   id="status_code"
                   name="status_code"
-                  label="Incident status?*"
+                  label="Incident status*"
```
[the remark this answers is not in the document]
thegentlemanphysicist: looks like it's the convention, so approved

migrations/sql/V2019.09.04.11.35__incident_add_mine_determination_columns.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE mine_incident
```
FreshGyan: Is there already data in mine_incident? Just flagging it to test on a non-empty db as well. FK errors usually get missed till we hit test 😅

NWCalvank: There is already data in that table. I ran this locally on a pre-seeded database without any issue. The FK is being added to a nullable column that is being created in this migration. So, every record will be `null`, which does not violate the FK constraint.

I think this should be totally safe. Anything you think I may have missed?

FreshGyan: Nope, looks good 👍

frontend/src/components/Forms/incidents/AddIncidentFollowUpForm.js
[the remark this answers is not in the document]
NWCalvank: Thanks! Yeah, it's to indicate a required field.

python-backend/app/api/incidents/models/mine_incident.py
pdittaro: I see what you mean. Its almost like we need the possessive "mine's" in there. But I'm fine with this.

PR #972, opened 2019-09-05
MDS-2227 bug/Reduce minimum search character count to 2

NWCalvank:
Bug:
There is a record in the system that is 2-characters long. As a result of the 3-character minimum, this record cannot be found using the search bar.

Solution:
Drop the minimum character count per search term to 2.
There may be negative performance implications of this change that could warrant a more thoughtful solution.


NWCalvank:
As expected, there is a performance concern with this change. There will be a discussion with the business side to decide how we'd like to proceed.

nayr974:
> 
> 
> As expected, there is a performance concern with this change. There will be a discussion with the business side to decide how we'd like to proceed.

Delete QR.

PR #974, opened 2019-09-06
Removing command line of legacy scripts

pdittaro:
Removing tailings commands.
Also, moving scripts to 'bin' convention of python project folder structure.
Adding bak suffix to indicate obsolescence.


pdittaro:
Clopening to deal with deploy error.

PR #976, opened 2019-09-10
Feature/mds 2197 now attached documents

nayr974:
Streams file downloads from NROS and VFCBC through the Core API. Authentication and system specific workflow are provided by service classes.


python-backend/app/api/now_submissions/resources/application_document_resource.py
```diff
@@ -0,0 +1,27 @@
...
+from app.api.now_submissions.models.application import Application
+from app.api.now_submissions.response_models import APPLICATION
+from app.api.utils.access_decorators import requires_role_view_all
+from app.api.utils.resources_mixins import UserMixin, ErrorMixin
+
+from app.api.services.nros_download_service import NROSDownloadService
+from app.api.services.vfcbc_download_service import VFCBCDownloadService
+
+
+class ApplicationDocumentResource(Resource, UserMixin, ErrorMixin):
```
nayr974: This is test code at the moment.

microservices/now_etls/now_etls.py
```diff
@@ -24,7 +24,10 @@ def etl_now_submission_data():
         user=DB_USER, password=DB_PASS, dbname=DB_NAME)
 
     click.echo('Beginning ETL')
-    NOW_submissions_ETL(connection)
+    try:
+        NOW_submissions_ETL(connection)
+    finally:
+        connection.close()
```
Jsyro: I just learned you can do this using `with`, which guarantees that the connection is closed. 

Looks like it the psycopg2 connection class supports it, https://pynative.com/python-postgresql-transaction-management-using-commit-and-rollback/

pdittaro: I'm using with in the actual ETL function with the passed connection. From the documentation I read it only said that with committed transaction if it completed, not that it would close the connection if it failed. 

http://initd.org/psycopg/docs/usage.html#with-statement

There's actually a specific warning in the documentation saying that it *doesn't* close the connection.

PR #981, opened 2019-09-11
Commiting only the backend for 2263

thegentlemanphysicist:
This is the complete and tested new enpoint for the mine incidents dashboard page.
There is currently some ambiguity about incident_no vs incident_id, etc, so it's sorting and filtering has not been implemented.

### DIFF
--- a/python-backend/app/api/incidents/resources/incidents_resource.py
+++ b/python-backend/app/api/incidents/resources/incidents_resource.py
@@ -1,0 +1,142 @@
+from flask_restplus import Resource
+from flask import request
+from datetime import datetime
+from sqlalchemy_filters import apply_sort, apply_pagination, apply_filters
+from sqlalchemy import desc, cast, NUMERIC, extract
+from app.extensions import api
+from app.api.mines.mine.models.mine import Mine
+from ..models.mine_incident import MineIncident
+from app.api.incidents.models.mine_incident_do_subparagraph import MineIncidentDoSubparagraph
+from ..response_models import PAGINATED_INCIDENT_LIST
+from ...utils.access_decorators import requires_any_of, VIEW_ALL
+from ...utils.resources_mixins import UserMixin, ErrorMixin
+
+PAGE_DEFAULT = 1
+PER_PAGE_DEFAULT = 25
+
+
+class IncidentsResource(Resource, UserMixin, ErrorMixin):
+    @api.doc(
+        description='Get a list of incidents. Order: received_date DESC',
+        params={
+            'page': f'The page number of paginated records to return. Default: {PAGE_DEFAULT}',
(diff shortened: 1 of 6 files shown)


python-backend/app/api/incidents/resources/incidents_resource.py
```diff
@@ -0,0 +1,142 @@
...
+            # "incident_no": 'MineIncident',
+            "determination": 'MineIncident',
+            "status": 'MineIncident',
+            "mine_name": 'Mine',
+        }
+
+        sort_field = {
+            "date": 'incident_timestamp',
+            # TODO: Mine incident_no is not currently implemented
+            # "incident_no": 'mine_incident_no',
```
NWCalvank: Still have questions around these?

python-backend/app/api/incidents/response_models.py
```diff
@@ -1,6 +1,12 @@
 from app.extensions import api
 from flask_restplus import fields
 
+
+class DateTime(fields.Raw):
+    def format(self, value):
+        return value.strftime("%Y-%m-%d %H:%M") if value else None
```
NWCalvank: What's this for? Are we formatting timestamps like this elsewhere?

If we are, we need to be very sure that we accept data in the same format that we return it.

python-backend/tests/incidents/resources/test_incidents_resource.py
```diff
@@ -0,0 +1,204 @@
...
+import random
+
+from tests.status_code_gen import *
+from tests.factories import MineIncidentFactory, MineFactory
+
+class TestGetIncidents:
+    """GET /incidents"""
+    # get all incidents
+    # filter by each of the following fields: "status" "determination" "codes" "year" "major" "search_terms" "region"
+    # sort by each field: "date",  "incident_no", "determination", "status", "mine_name"
```
NWCalvank: Looks like this type of info should go in the docstrings

python-backend/tests/incidents/resources/test_incidents_resource.py
```diff
@@ -0,0 +1,204 @@
...
+        get_resp = test_client.get(f'/incidents?sort_field={sort_field}&sort_dir={sort_dir}', headers=auth_headers['full_auth_header'])
+        get_data = json.loads(get_resp.data.decode())
+        assert get_resp.status_code == 200
+        assert len(get_data['records']) == batch_size
+        assert all(get_data['records'][i]['incident_timestamp'] >= get_data['records'][i+1]['incident_timestamp']
+                   for i in range(len(get_data['records'])-1))
+
+    # TODO: When this incident_no is implemented a test on sorting by it will be needed
+    # def test_get_incidents_incident_no_sort(self, test_client, db_session, auth_headers):
+    #     """Should respect incidents incident_no sort"""
```
NWCalvank: Is this still relevant?

PR #983, opened 2019-09-11
MDS-304 Feature/ Mine Manager document - Upload & View / Download

NWCalvank:
Supports uploading a document while adding a Mine Manager appointment.

If there is a document on an existing Mine Manager appointment, you'll see a link to view/download the document in the contact card.


frontend/src/components/mine/ContactInfo/PartyRelationships/DefaultContact.js
```diff
@@ -79,7 +81,24 @@ export const DefaultContact = (props) => (
         <br />,
         <br />,
         <h6>{props.partyRelationshipTitle} Since</h6>,
-        <span>{formatDate(props.partyRelationship.start_date) || "Unknown"}</span>,
+        <span>
+          {formatDate(props.partyRelationship.start_date) || "Unknown"}
+          {props.partyRelationship.mine_party_appt_type_code === "MMG" &&
+            props.partyRelationship.documents.length > 0 && (
+              <span>
+                {" "}
```
mmumby: these white spaces can be deleted and just have `-`

NWCalvank: I tried that. With the `<LinkButton>` tag right there, it doesn't seem to work. Prettier adds those 🙂

thegentlemanphysicist:
Did some smoke testing, looks solid.  Just had a couple of backend questions.

PR #988, opened 2019-09-13
Task/2313 schemaspy all schemas

Jsyro:


Jsyro:
Remove schema specification and add `-all` flag to show all schemas.

PR #997, opened 2019-09-16
Removing originating system logic to troubleshoot NOW ETL

pdittaro:


pdittaro:
Squashing since it isn't touched by any of the tests.

PR #998, opened 2019-09-17
Feature/mds 2197 now attached documents

nayr974:
This PR reverses the order of the dictionary for file extensions and mimetypes.


frontend/src/constants/fileTypes.js
```diff
@@ -1,39 +1,40 @@
...
 export const PDF = {
-  "application/pdf": ".pdf",
+  ".pdf": "application/pdf",
 };
+
 export const DOC = {
-  "application/msword": ".doc",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document": ".docx",
-  "application/rtf": ".rtf",
-  "text/rtf": null,
```
pdittaro: As I would understand this, files without an extension would be treated as rich text files. I don't think we have a requirement to deal with this so am alright to remove it.

PR #1000, opened 2019-09-18
MDS 2318 - Adding mine_guid and originating system to NoW ETLs

pdittaro:
Need to add two fields to the ETL. One is a programmatic lookup of the mine_guid in the Core database.


pdittaro:
@Jsyro  are you able to re-review?

PR #1002, opened 2019-09-18
Doc/mds 2206

thegentlemanphysicist:
this moves a testing plan doc into the codebase


thegentlemanphysicist:
> Can you change the document name to not have spaces and use underscores or dashes instead. Thanks.

fixed

PR #1005, opened 2019-09-18
MDS-2327 Performance of now app view

JustinMacaulay:
Changed all the relationships on application in the Now Submissions schema to be select instead of joined.


JustinMacaulay:
As far as I can tell this has helped on local, and it is faster on dev but the dev instances are considerably slower than local.

PR #1006, opened 2019-09-19
Fixed tests

thegentlemanphysicist:
This should allow the incident python tests to run much faster and therefore not fail in the pipeline as often.


pdittaro:
Are the performance of the python unit tests an issue? I don't think theres any issue in scaling this down, but what sort of performance increases were you seeing?

thegentlemanphysicist:
> Are the performance of the python unit tests an issue? I don't think theres any issue in scaling this down, but what sort of performance increases were you seeing?

I noticed when I ran these locally that the tests I wrote were slower than others.   They were always passing locally, but I was curious if that delay may be why the tests in the pipeline were failing intermittently.

PR #1007, opened 2019-09-19
Feature/MDS-2178 Hide fields that will be overridden by the ETL when editing a permit amendment for a regional mine.

nayr974:
After talking with Aaron he only intends for Admins to upload documents at this time, so no additional roles or authorization changes. As we are still ETLing and overwriting regional mine permit information, I have hidden fields other than documents on the edit form just in case an Admin assumes they can modify these fields while uploading documents.


pdittaro:
I don't think that this PR description matches the changes contained. Why are you rendering the select party and description components on the PermitAmendmentForm for major mines?

frontend/src/tests/components/Forms/__snapshots__/PermitAmendmentForm.spec.js.snap
```diff
@@ -10,58 +10,6 @@ exports[`PermitAmendmentForm renders properly 1`] = `
...
   >
-    <Col
-      md={12}
-      sm={24}
-    >
-      <FormItem
-        hasFeedback={false}
-      >
-        <Connect(PartySelectField)
-          allowAddingParties={true}
```
Jsyro: Just pointing out that the snapshot test does not render the component with isMajorMine is true. Is that intended.

nayr974:
> 
> 
> I don't think that this PR description matches the changes contained - are you able to update what these changes encompass?

Renamed the title to better describe the outcome of the task.

pdittaro:
Very succinct title, thank you.

frontend/src/tests/components/Forms/__snapshots__/PermitAmendmentForm.spec.js.snap
nayr974: set isMajorMine=true in the test which now renders the full component

PR #1008, opened 2019-09-19
Set content-disposition by generating it from the file name instead of passing the header parameter from VFCBC

nayr974:


pdittaro:
This looks to match the solution for NROS.

PR #1011, opened 2019-09-20
Mds 2337

thegentlemanphysicist:
This fixes the page and per page url params and makes the link sharable


frontend/src/components/dashboard/incidentsHomePage/IncidentsHomePage.js
```diff
@@ -110,20 +111,31 @@ export class IncidentsHomePage extends Component {
...
+        per_page: Strings.DEFAULT_PER_PAGE,
         region: formatParamStringToArray(this.params.region),
         major: this.params.major,
         search: this.params.search,
         year: this.params.year,
         incident_status: this.params.incident_status,
         codes: this.params.codes,
         determination: this.params.determination,
+        ...this.params,
       },
```
mmumby: what is left in this spread? everything should be defined in the state

frontend/src/components/dashboard/incidentsHomePage/IncidentsHomePage.js
```diff
@@ -110,20 +111,31 @@ export class IncidentsHomePage extends Component {
...
     };
   }
 
   componentDidMount() {
     const params = this.props.location.search;
-    this.renderDataFromURL(params);
+    if (params) {
+      this.renderDataFromURL(params);
+    } else {
+      const defaultParams = {
```
mmumby: Option 1) 
if there are no 'params' in the URL - then 'page' and 'per_page' are already defined in the state with their default when the page loads, so this latter step is not necessary

Option 2) 
the URL contains the params - if 'ComponentWillReceiveProps' is handling the params correctly, it should be setting the new values for 'page' and 'per_page'

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -120,15 +123,22 @@ export class VarianceHomePage extends Component {
         issue_date_before: this.params.issue_date_before,
         expiry_date_before: this.params.expiry_date_before,
         expiry_date_after: this.params.expiry_date_after,
+        ...this.params,
       },
```
mmumby: same here, what is being spread? should be defined

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -207,25 +217,26 @@ export class VarianceHomePage extends Component {
...
+      this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(updatedParams));
+      return { params: updatedParams };
+    });
   };
 
   handleVariancePageChange = (page, per_page) => {
     this.setState({ variancesLoaded: false });
     const params = { ...this.state.params, page, per_page };
+    this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(params));
     return this.props.fetchVariances(params).then(() => {
```
mmumby: `this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(params))` will cause `CompopnentWillReceiveProps' to send the new params to 'renderDataFromUrl` which will `fetchVariances`, this path will set the params correctly within the state as well as passing the correct params to the API

so, `fetchVariances` here is not needed.

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -120,15 +123,22 @@ export class VarianceHomePage extends Component {
...
-
+    if (params) {
+      this.renderDataFromURL(params);
+    } else {
+      const defaultParams = {
+        page: Strings.DEFAULT_PAGE,
+        per_page: Strings.DEFAULT_PER_PAGE,
+      };
+      this.props.history.push(router.VARIANCE_DASHBOARD.dynamicRoute(defaultParams));
+    }
```
mmumby: same here ^

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: removed

frontend/src/components/dashboard/incidentsHomePage/IncidentsHomePage.js
thegentlemanphysicist: removed

frontend/src/components/dashboard/varianceHomePage/VarianceHomePage.js
thegentlemanphysicist: un removed

frontend/src/components/dashboard/incidentsHomePage/IncidentsHomePage.js
mmumby: @thegentlemanphysicist

PR #1018, opened 2019-09-24
Feature/mds 2302

mmumby:
Added a skeleton loading view to all tables
- reusable `TableLoadingWrapper.js` which accepts the column titles and creates a loading view specific to each table
- if the table is paginated, the loading view will have 25 rows for a smoother loading experience
- Minor styling updates throughout app


frontend/src/components/dashboard/customHomePage/CustomHomePage.js
```diff
@@ -41,9 +39,12 @@ const propTypes = {
...
 
 export class CustomHomePage extends Component {
+  state = { isLoaded: false };
+
   componentDidMount() {
-    this.props.fetchSubscribedMinesByUser();
-    this.props.fetchInspectors();
+    this.props.fetchSubscribedMinesByUser().then(() => {
+      this.setState({ isLoaded: true });
+    });
```
NWCalvank: We didn't need to call `this.props.fetchInspectors();` at all?

frontend/src/components/dashboard/minesHomePage/MineSearch.js
```diff
@@ -38,7 +38,7 @@ const defaultProps = {
 };
 
 const checkAdvancedSearch = ({ status, region, tenure, commodity, tsf, major }) =>
-  tsf || major || some([status, region, tenure, commodity], negate(isEmpty));
+  tsf || major || some([status, region, tenure, commodity, tsf], negate(isEmpty));
```
NWCalvank: I don't think this change is necessary with the incorrect `tsf: []` initialValues removed

frontend/src/components/mine/Incidents/MineIncidents.js
```diff
@@ -70,8 +70,12 @@ const defaultProps = {
 };
 
 export class MineIncidents extends Component {
+  state = { isLoaded: false };
+
   componentDidMount() {
-    this.props.fetchMineIncidents(this.props.mineGuid);
+    this.props.fetchMineIncidents(this.props.mineGuid).then(() => {
+      this.setState({ isLoaded: true });
+    });
```
NWCalvank: Just wondering. Would it be better for this to be an array of requests passed to Promise.all() so that `isLoaded` could be set to `true` once all dependencies were fetched?

This question applies to files with similar logic as well.

frontend/src/utils/helpers.js
```diff
@@ -169,3 +169,7 @@ export const formatComplianceCodeValueOrLabel = (code, showDescription) => {
 
   return `${section}${formattedSubSection}${formattedParagraph}${formattedSubParagraph}${formattedDescription}`;
 };
+
+export const getTableHeaders = (tableColumns) => {
+  return tableColumns.map((column) => column.title);
+};
```
NWCalvank: Totally fine with this implementation. If we'd like to leverage existing functions, lodash's `map` has this out-of-the-box: `_.map(tableColumns, 'title');`

https://lodash.com/docs/4.17.15#map

frontend/src/components/dashboard/customHomePage/CustomHomePage.js
mmumby: no, it should have been removed when variances was removed from this view.

frontend/src/utils/helpers.js
mmumby: good to know!

frontend/src/components/mine/Incidents/MineIncidents.js
mmumby: I was wondering this as well, through my testing it was never an issue, but local is much faster than other envs.

frontend/src/components/dashboard/minesHomePage/MineSearch.js
mmumby: removed

frontend/src/components/common/wrappers/TableLoadingWrapper.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { PropTypes } from "prop-types";
+import { Table } from "antd";
+
+/**
+ * @constant TableLoadingWrapper renders react children || or a skeleton loading view using the column headers
```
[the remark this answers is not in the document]
mmumby: removed

PR #1030, opened 2019-09-27
Task/ Remove old maps files

NWCalvank:


pdittaro:
It looks like the migration to leaflet was successful, but if possible lets keep this open for a while longer just in case we need to pivot quickly. I think the acceptance criteria for this task has been fulfilled.

Jsyro:
migration mismatch, close and reopen to clean openshift.

PR #1033, opened 2019-09-28
Leaflet map performance improvement

FreshGyan:
Loading clusters and Web map was too much work on the main thread. Causing browser slowing down and even page crashes. Couldn't catch it with happy PROD data or locally seeded one.
Needed TEST data for it to really push the limits 😅 

Anyways, I moved the order of loading map components around for no-jitter loading experience.

Basically load the lightweight leaflet map first and add the clusters while the external widget and esri webmap layer libraries are being downloaded by react-async-script-loader. 

Once the libraries are downloaded, asynchronously add the widgets, layers and other controls to the map.

### DIFF
--- a/frontend/src/components/maps/MineMapLeaflet.js
+++ b/frontend/src/components/maps/MineMapLeaflet.js
@@ -183,6 +183,8 @@ class MineMapLeaflet extends Component {
       center: [this.props.lat, this.props.long],
       zoom: this.props.zoom,
       worldCopyJump: true,
+      zoomAnimationThreshold: 8,
+      minZoom: 4,
     });
 
     // Add Topographic BaseMap by default
@@ -334,7 +336,7 @@ class MineMapLeaflet extends Component {
           // Stop checking
           clearInterval(resetZoomCheckID);
         }
-      }, 10);
+      }, 5);
     });
   }
 
--- a/frontend/src/components/maps/MineMapLeaflet.js
+++ b/frontend/src/components/maps/MineMapLeaflet.js
@@ -176,19 +176,34 @@ const tenureLayerStyles = {
 };
 
 class MineMapLeaflet extends Component {
-  state = {
-    mapCreated: false,
-  };
-
   componentDidMount() {
+    // Create the basic leaflet map
+    this.map = L.map("leaflet-map", {
+      attributionControl: false,
+      center: [this.props.lat, this.props.long],
+      zoom: this.props.zoom,
+      worldCopyJump: true,
+    });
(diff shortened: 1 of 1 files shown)


frontend/src/components/maps/MineMapLeaflet.js
```diff
@@ -176,19 +176,36 @@ const tenureLayerStyles = {
...
   componentDidMount() {
+    // Create the basic leaflet map
+    this.map = L.map("leaflet-map", {
+      attributionControl: false,
+      center: [this.props.lat, this.props.long],
+      zoom: this.props.zoom,
+      worldCopyJump: true,
+      zoomAnimationThreshold: 8,
+      minZoom: 4,
+    });
```
NWCalvank: Confirming that this doesn't require anything fetched by `scriptLoader`? I'd be worried about a race condition here

frontend/src/components/maps/MineMapLeaflet.js
```diff
@@ -213,6 +230,67 @@ class MineMapLeaflet extends Component {
...
+        color: "red",
+        fillColor: "#f03",
+        fillOpacity: 0.25,
+        radius: 500,
+        stroke: false,
+      }).addTo(this.map);
+    }
+  };
+
+  addMinePinClusters = async () => {
```
NWCalvank: Any intended reason why this is an async function? There's no `await` here and no returned value.

frontend/src/components/maps/MineMapLeaflet.js
```diff
@@ -213,6 +230,67 @@ class MineMapLeaflet extends Component {
...
+    // Add Clustered MinePins
+    this.markerClusterGroup = L.markerClusterGroup({ animate: false });
+    this.props.minesBasicInfo.map(this.createPin);
+    this.map.addLayer(this.markerClusterGroup);
+    this.addLatLongCircle();
+  };
+
+  addWidgets = async () => {
```
NWCalvank: Same question here. Perhaps it's a misunderstanding around `async` functions? They're still invoked synchronously. And if all of the statements in the function are also synchronous, there is no performance benefit provided.

Ex.
```javascript
const fn = async () => {
  console.log("async function");
}

fn();
console.log("sync");
```

Output:
```
async function
sync
```

frontend/src/components/maps/MineMapLeaflet.js
NWCalvank: Pulled the branch down and tested this. All is well

PR #1039, opened 2019-10-01
Bug - Fixed contact name on NoW view

mmumby:
- updated contact to include first or last name
- added org name
- added "n/a' if the above 3 do not exist


frontend/src/components/noticeOfWork/NOWGeneralInfo.js
```diff
@@ -130,9 +130,12 @@ export class NOWGeneralInfo extends Component {
                     >
                       <div>
                         <h3>
-                          {contact.ind_firstname && contact.ind_lastname
-                            ? `${contact.ind_firstname} - ${contact.ind_firstname}`
-                            : Strings.EMPTY_FIELD}
+                          {contact.ind_firstname && <span>{contact.ind_firstname}</span>}
+                          {contact.ind_lastname && <span> {contact.ind_lastname}</span>}
+                          {contact.org_legalname && <span>{contact.org_legalname}</span>}
```
NWCalvank: Do you want to check that there is not a firstname & lastname & org_legalname all at the same time?

mmumby: yeah, from what I'm gathering - the firstName & lastName only exist if type = individual, and org_legalname only exists if type = org. 

I don't think it will make it to the last situation ever, but just in case it's better to have "na" than nothing

PR #1040, opened 2019-10-01
Task/remove error mixin

Jsyro:
don't like it, get rid of it.


Jsyro:
Also fixed url registration bug from namespace file rename.

PR #1041, opened 2019-10-02
Task/remove party xref

Jsyro:
Remove party_address_xref table and add party_guid xref on address table. 

Need to ensure all addresses have parties in existing prod databases with this query `select COUNT(*) from address a left join party_address_xref pax on a.address_id = pax.address_id where pax.party_guid is null;`


pdittaro:
The provided SQL was tested in prod and returned 0 results. Good to go.

PR #1043, opened 2019-10-03
add party mms pk.

Jsyro:
etl testing


Jsyro:
closing for now, pivoting to different approach .

PR #1045, opened 2019-10-09
Feature/sql join addresses to mms

Jsyro:
First attempt at sql join to import addresses from their source location.


tools/digdag/etl/tasks/mms_data_import.py
```diff
@@ -1,6 +1,6 @@
 from models.pod import POD
 
 def run_job():
-    pod = POD(pod_name='digdag-mms-job', env_pod='mds-python-backend', command=["flask","run-etl"])
+    pod = POD(pod_name='digdag-mms-job', env_pod='mds-python-backend', command=["flask","run-etl && flask run-address-etl"])
```
pdittaro: Lets leave this out for now and we can test manually in Test

PR #1061, opened 2019-10-30
version and share project vscode settings

Jsyro:
We should share this file, it should stay light, but there are somethings that will be easier to manage here than anywhere else.


pdittaro:
I approved, but I think we should get more consensus from the other developers before making the decision to include this.

PR #1063, opened 2019-11-07
Feature/mds 2339 incident type category

nayr974:
This PR adds an "Incident type" field when adding editing, or viewing an incident.


frontend/src/components/modalContent/AddIncidentModal.js
```diff
@@ -157,7 +162,6 @@ const StepForms = (
         initialValues={props.initialValues}
         incidentDeterminationOptions={props.incidentDeterminationOptions}
         followupActionOptions={props.followupActionOptions}
-        incidentStatusCodeOptions={props.incidentStatusCodeOptions}
         hasFatalities={props.addIncidentFormValues.number_of_fatalities > 0 || false}
```
mmumby: was this an intentional deletion?

python-backend/app/api/incidents/models/mine_incident.py
```diff
@@ -21,12 +21,10 @@ class MineIncident(AuditMixin, Base):
     __tablename__ = 'mine_incident'
 
     mine_incident_id = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
-    mine_incident_id_year = db.Column(db.Integer,
-                                      nullable=False,
-                                      default=datetime.datetime.now().year)
-    mine_incident_guid = db.Column(UUID(as_uuid=True),
-                                   nullable=False,
-                                   server_default=FetchedValue())
+    mine_incident_id_year = db.Column(
```
pdittaro: Looks like we might have some conflicting python formatters by different developers. We should probably standardize on one.

python-backend/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -99,6 +104,7 @@ def post(self, mine_guid):
             followup_investigation_type_code=data['followup_investigation_type_code'],
             reported_timestamp=data['reported_timestamp'],
             reported_by_name=data['reported_by_name'],
+            mine_incident_category_code=data['mine_incident_category_code'],
```
pdittaro: Re: formatting - This looks like the only change in this file.

frontend/src/components/modalContent/AddIncidentModal.js
pdittaro: Doesn't look like it, the Form still has this as a parameter.

nayr974: I did not, I must have selected an extra line when removing the unnecessary passing of categories that I added then removed here. Thank you for noticing! Added back.

python-backend/app/api/incidents/models/mine_incident.py
nayr974: Yes please

PR #1066, opened 2019-11-08
Fix creation of duplicate permits, merge duplicate permit records

nayr974:
This PR modifies the ETL to handle cases when:
- A permit was created and amended in the same day, causing two records to be created
- A bug when a new amendment is created for an existing permit, where the ETL_Permit table would not have the existing permit_guid id

It also merges amendments then deletes duplicate permit records.


pdittaro:
How are we going to test this? Is running it in test sufficient? @nayr974

pdittaro:
Talked with Ryan and am alright with approving this. We'll need to keep an eye on what happens in Test.

PR #1067, opened 2019-11-12
Feature/MDS-2432

mmumby:
- Edit/view for core schema NoW applications.


frontend/package.json
```diff
@@ -54,6 +54,7 @@
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.13.3",
     "reselect": "^4.0.0",
+    "save": "^2.4.0",
     "string-similarity": "^3.0.0",
```
mmumby: @Jsyro did you install this? wondering if it was it an accidental install on my part...

frontend/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -25,24 +32,42 @@ const propTypes = {
...
     isLoaded: false,
     associatedMineGuid: "",
+    isViewMode: true,
+    fixedTop: false,
   };
 
   componentDidMount() {
-    const params = queryString.parse(this.props.location.search);
-    const { id } = params;
+    window.scrollTo(0, 0);
```
nayr974: I'm not _sure_ this works as expected in all browsers. I have a vague memory of some complexity when I looked into this issue in the past, although it may not longer be true.

mmumby: I think you are correct, I will remove it from here (in the follow up PR) and we can address it in MDS-2449 (add scroll on every route change)

frontend/package.json
nayr974: 100% confidant that Jason didn't add this unless also by accident, so probably good to remove.

PR #1069, opened 2019-11-15
Upgrade NoW view

mmumby:
- Added a table for reclamation summary, creating a selector that pulls reclamation data out of the nested activities, 
- created endpoint to get NoticeOfWorkActivityTypeCodes, created staticContentActionCreator, staticContentSelector etc for NOWActivityTypeDropdownOptions
- Updated data on water_supply activity table
- Added a loading state around content of steps"
- added tests for ststicContent actionCreator/Reducer/Selector
- added `blasting_operation` to payload


migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -544,7 +544,7 @@ VALUES
     ('mechanical_trenching', 'Mechanical Trenching / Test Pits', 'system-mds', 'system-mds'),
     ('surface_bulk_sample', 'Surface Bulk Sample', 'system-mds', 'system-mds'),
     ('blasting_operation', 'Blasting Operations', 'system-mds', 'system-mds'),
-    ('placer', 'Placer Opertations', 'system-mds', 'system-mds')
+    ('placer_operation', 'Placer Opertations', 'system-mds', 'system-mds')
 ON CONFLICT DO NOTHING;
```
mmumby: Not sure if this can be done, updated the seed data so `placer` matched the format of the other activities.

frontend/package.json
```diff
@@ -54,7 +54,6 @@
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.13.3",
     "reselect": "^4.0.0",
-    "save": "^2.4.0",
```
pdittaro: If a dependency is being removed the package-lock.json file needs to be updated as well. Why is this being removed?

mmumby: This was accidentally added, Likely as a result of forgetting `--` when saving other packages. 

I attempted to remove it in the previous sprint, but with merge conflicts, it kept getting added back in. 

Will remove from the package-lock.json, although I was under the impression that once you ran 'npm ci', the `package-lock.json` would be updated according to the current `package.jon`, is this not the case?

mmumby:
I agree in regards to `Proptypes.any`... this was a lazy attempt of not creating `customProps` for every activity of the NoW. If we're against using any, I will update all the new `PropTypes.any` introduced in this PR before we merge it.

frontend/src/selectors/noticeOfWorkSelectors.js
```diff
@@ -1,8 +1,36 @@
...
   getOriginalNoticeOfWork,
 } = noticeOfWorkReducer;
+
+export const getNOWReclamationSummary = createSelector(
+  [getNoticeOfWork, getDropdownNoticeOfWorkActivityTypeOptions],
+  (noticeOfWork, options) => {
+    const reclamationList = [];
+    if (options.length > 0) {
+      options
+        .filter(({ value }) => value !== "blasting_operation" && value !== "water_supply")
```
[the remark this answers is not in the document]
mmumby: updated!

PR #1076, opened 2019-11-26
Feature/now application edit

Jsyro:
Now application has an edit endpoint that accepts a modified version of the get payload and editing values, or adding objects to nested lists. this will not be the final implementation, but it does satisfy the edit, 

Models now have a _edit_key, the group in which they edit in, and an _edit_groups, which says what edit they are allowed to be edited by.


python-backend/app/api/mines/permits/permit/resources/permit.py
```diff
@@ -47,7 +47,6 @@ class PermitListResource(Resource, UserMixin):
     parser.add_argument('uploadedFiles', type=list, location='json', store_missing=False)
 
     @api.doc(params={'mine_guid': 'mine_guid to filter on'})
-    @requires_role_view_all
```
nayr974: We still want to keep this role I believe.

python-backend/app/api/now_applications/resources/now_application_resource.py
```diff
@@ -29,17 +30,33 @@ class NOWApplicationResource(Resource, UserMixin):
...
+    )
+    @requires_role_edit_permit
+    @api.marshal_with(NOW_APPLICATION_MODEL, code=200)
+    def put(self, application_guid):
+        now_application_identity = NOWApplicationIdentity.find_by_guid(application_guid)
+        if not now_application_identity:
+            raise NotFound('No identity record for this application guid.')
+
+        if now_application_identity.now_application_id is None:
+            raise NotImplemented('this would cause transmogrify and save, not yet')
```
nayr974: This message should is public, please revise this without referencing transmogrify eg "This application has not been imported. Please import an application before making changes."

python-backend/app/api/now_applications/resources/now_application_resource.py
```diff
@@ -29,17 +30,33 @@ class NOWApplicationResource(Resource, UserMixin):
...
+        data = request.json
+        now_application_identity.now_application.deep_update_from_dict(data)
```
nayr974: Should this just be named update? Does this class need to know implementation details of the other class?

python-backend/app/api/mines/permits/permit/resources/permit.py
Jsyro: fixed

python-backend/app/api/mms_now_submissions/models/application.py
```diff
@@ -121,7 +119,7 @@ def __repr__(self):
     @classmethod
     def find_by_application_guid(cls, guid):
         cls.validate_guid(guid)
-        mms_cid = NOWApplicationIdentity.filter_by(application_guid=guid).first().mms_cid
```
Jsyro: reminder: this needs to be fixed

python-backend/app/api/now_applications/resources/now_application_resource.py
Jsyro: i think 'deep' is an important prefix, because it can be unclear if it's updating the object, or related objects as well. could remove the 'from' dict suffix, but without that the caller doesn't know what the input type should be.

python-backend/app/api/mms_now_submissions/models/application.py
Jsyro: why fix it when you can remove it .

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -13,10 +13,14 @@
 
 from . import permit_amendment_status_code, permit_amendment_type_code
 from app.api.utils.models_mixins import AuditMixin, Base
+from app.api.constants import *
 
 
 class PermitAmendment(AuditMixin, Base):
     __tablename__ = 'permit_amendment'
+    _edit_groups = [PERMIT_EDIT_GROUP, PERMIT_AMENDMENT_EDIT_GROUP, MINE_EDIT_GROUP]
+    _edit_key = PERMIT_AMENDMENT_EDIT_GROUP
```
Jsyro: Starting at permit amendments, you can only edit related entities with PERMIT_AMENDMENT_EDIT_GROUP in _edit_groups, in this example you cannot go up to edit a Permit from the amendment.

services/core-api/app/api/mines/permits/permit/models/permit.py
```diff
@@ -12,10 +12,14 @@
 from app.api.parties.party_appt.models.mine_party_appt import MinePartyAppointment
 
 from app.api.utils.models_mixins import AuditMixin, Base
+from app.api.constants import *
 
 
 class Permit(AuditMixin, Base):
     __tablename__ = 'permit'
+    _edit_groups = [PERMIT_EDIT_GROUP, MINE_EDIT_GROUP]
+    _edit_key = PERMIT_EDIT_GROUP
```
Jsyro: since Permit Amendment has PERMIT_EDIT_GROUP as an edit group, you can edit next permit_amendments from the permit.

services/core-api/app/api/utils/setup_marshmallow.py
```diff
@@ -47,17 +75,39 @@ class Meta(object):
...
+                    raise e
+
     return setup_schema_fn
 
 
 # TODO: finish this and resolve errors now_application/activity_detail_base.activity_type_code to all for programatic generation of schema
 # TODO: add call to model method to execute post_generation of schema.
-# event.listen(mapper, "after_configured", setup_schema(Base, db.session))
+
+#event.listen(mapper, "after_configured", setup_schema(BaseModel, db.session))
```
Jsyro: if this line was uncommented, this would run, for now. it's just the start of future code.

services/core-api/app/api/mines/permits/permit/models/permit.py
```diff
@@ -12,10 +12,14 @@
 from app.api.parties.party_appt.models.mine_party_appt import MinePartyAppointment
 
 from app.api.utils.models_mixins import AuditMixin, Base
+from app.api.constants import *
 
 
 class Permit(AuditMixin, Base):
     __tablename__ = 'permit'
+    _edit_groups = [PERMIT_EDIT_GROUP, MINE_EDIT_GROUP]
```
nayr974: I don't think we want permits to be in the mine edit group.

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -13,10 +13,14 @@
 
 from . import permit_amendment_status_code, permit_amendment_type_code
 from app.api.utils.models_mixins import AuditMixin, Base
+from app.api.constants import *
 
 
 class PermitAmendment(AuditMixin, Base):
     __tablename__ = 'permit_amendment'
+    _edit_groups = [PERMIT_EDIT_GROUP, PERMIT_AMENDMENT_EDIT_GROUP, MINE_EDIT_GROUP]
```
nayr974: Same comment about the mine_edit_group and permits.

services/core-api/app/api/utils/models_mixins.py
```diff
@@ -82,26 +86,47 @@ def deep_update_from_dict(self, data_dict, depth=0):
...
 
             if isinstance(v, list):
+                rel = getattr(self.__class__, k) #SA.relationship definition
                 obj_list = getattr(self, k)
+                obj_list_class = rel.property.entity.class_
+
+                if _edit_key not in obj_list_class._edit_groups:
+                    #not part of same edit group, skip
+                    current_app.logger.debug(
+                        f'COMBOBREAKER!!! {_edit_key} not in {obj_list_class} edit groups {obj_list_class._edit_groups}'
```
nayr974: Do we want to keep these debug statements in?

services/core-api/app/api/now_applications/resources/now_application_resource.py
```diff
@@ -29,17 +30,35 @@ class NOWApplicationResource(Resource, UserMixin):
...
+    )
+    @requires_role_edit_permit
+    @api.marshal_with(NOW_APPLICATION_MODEL, code=200)
+    def put(self, application_guid):
+        now_application_identity = NOWApplicationIdentity.find_by_guid(application_guid)
+        if not now_application_identity:
+            raise NotFound('No identity record for this application guid.')
+
+        if now_application_identity.now_application_id is None:
+            raise NotImplemented(
```
pdittaro: I don't think this is a correct invocation of the NotImplemented exception - it should be used if a specific function invocation isn't supported.

services/core-api/app/api/now_submissions/models/application.py
```diff
@@ -225,7 +227,11 @@ def __repr__(self):
     @classmethod
     def find_by_now_application_guid(cls, now_application_guid):
         cls.validate_guid(now_application_guid)
-        now_identity = NOWApplicationIdentity.query.filter_by(now_application_guid=now_application_guid).first()
+
+        from app.api.now_applications.models.now_application_identity import NOWApplicationIdentity
```
pdittaro: Imports should be at the top of the file

services/core-api/app/api/mines/permits/permit/models/permit.py
Jsyro: changed

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
Jsyro: changed

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
pdittaro: Do we need to disable this?

services/core-api/app/api/utils/models_mixins.py
Jsyro: Yes, but could use be tweaked, debug is the highest detail, we should use err,warn,info,debug logging levels better.

services/core-api/app/api/now_submissions/models/application.py
Jsyro: it causes a circular import at the top of the file, this works.... i can do some more digging if needed.

services/core-api/app/api/now_applications/resources/now_application_resource.py
Jsyro: this is http not implemented, not python.

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
Jsyro: apparently not.

services/core-api/app/api/now_submissions/models/application.py
pdittaro: I guess that's why its a should instead of shall, this seems like a valid reason.

services/core-api/app/api/now_applications/resources/now_application_resource.py
pdittaro: Ah, okay thats fine.

PR #1087, opened 2019-11-28
Change to only run functional tests on the 292 branch

pdittaro:
This change will cause the functional tests only to be run on the develop branch PR, meaning that the PRs to develop do not require this step. Any deployment to Test and Prod will still require the functional tests to pass develop.


nayr974:
I'm in favor of this change!

PR #1092, opened 2019-12-06
Fix/audit now application data

Jsyro:
Verify fields that we can retrieve are being transmogrified and displayed, only fields that we cannot initialize from NROS/vFCBC have ** indicator.


Jsyro:
closing to clean db

PR #1099, opened 2019-12-13
Feature/deep update delete and detail-append

Jsyro:
special flag caller can add to json blob that tells the backend to delete nested object.


services/core-api/app/api/constants.py
```diff
@@ -83,6 +83,9 @@ def DOWNLOAD_TOKEN(token_guid):
     return f'document-manager:download-token:{token_guid}'
 
 
+#Deep Update Special Flag
+DELETE_ITEM_FROM_LIST_JSON_KEY = "DELETE_ON_PUT"
```
nayr974: After discussion let's make this a more generic state name that includes delete.

pdittaro:
Jason and I had a discussion about behaviour differences between deleting records and disassociating records. E.g. remove a contact from a mine, but don't delete that contact. This will be something that we can expand into later with the 'state_modified' attribute.

Otherwise, this looks good.

services/core-api/app/api/utils/setup_marshmallow.py
```diff
@@ -85,29 +85,29 @@ class ModelSchema2(ModelSchema):
                 except Exception as e:
                     raise e
 
-        for class_ in Base._decl_class_registry.values():
-            if hasattr(class_, "_schema"):
-                pprint(class_.__name__)
-                try:
-                    mapper = inspect(class_)
+        # for class_ in Base._decl_class_registry.values():
```
pdittaro: Are we going to be using this later?

pdittaro: Talked with Jason, these are fine to keep for now as this is WIP.

PR #1101, opened 2019-12-13
Feature/mds 2511

mmumby:
- disabled all fields with "**"
- added a disclaimer to explain what "**" means at this stage


services/core-web/src/tests/components/noticeOfWork/applications/review/__snapshots__/ReviewNOWApplication.spec.js.snap
```diff
@@ -46,7 +61,7 @@ exports[`ReviewNOWApplication renders properly 1`] = `
               <div
                 className="field-title"
               >
-                Permit Status**
+                Permit Status
```
pdittaro: Are we getting the Permit status now?

PR #1103, opened 2019-12-16
Feature/mds 2510 cancel now edits

JustinMacaulay:
Styling fixes for the NoW application view and added functionality to cancel edits in progress and revert the form to its initial state.


services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -156,7 +158,11 @@ export class NoticeOfWorkApplication extends Component {
   };
 
   toggleEditMode = () => {
-    this.setState((prevState) => ({ isViewMode: !prevState.isViewMode }));
+    this.setState((prevState) => ({
+      isViewMode: !prevState.isViewMode,
+      isEditing: !prevState.isEditing,
+      menuVisible: false,
+    }));
```
mmumby: What is the difference between `isViewMode` and `isEditing` ? They are the same but opposite, can we use one?

JustinMacaulay: Changed to just use isViewMode.

PR #1111, opened 2019-12-19
Feature/MDS-2505 Provide back button from NoW view

exzizt:
Mostly just formatting left and a decision to be made on what to do if the user visits via a direct URL and not from another page.


services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -112,6 +117,22 @@ export class NoticeOfWorkApplication extends Component {
       });
     });
     this.props.fetchOriginalNoticeOfWorkApplication(id);
+    this.setState({
+      fromTitle:
+        this.props &&
```
nayr974: oof, but necessary :)

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -337,8 +358,13 @@ export class NoticeOfWorkApplication extends Component {
           <div className="inline-flex between">
             <div>
               <h1>NoW Number: {this.props.noticeOfWork.now_number || Strings.EMPTY_FIELD}</h1>
+              {this.state && this.state.fromRoute && (
```
nayr974: this.state will always exist on a component, so probably not necessary to add safety around it (but doesn't hurt either).

services/core-web/src/tests/components/dashboard/noticeOfWorkHomePage/__snapshots__/NoticeOfWorkTable.spec.js.snap
```diff
@@ -1,121 +1,7 @@
...
-          type="no-results"
-        />,
-      }
-    }
-    onChange={[Function]}
-    pagination={false}
-    rowClassName="fade-in"
-  />
-</TableLoadingWrapper>
+<ContextConsumer>
```
nayr974: It looks like there's a bug causing the snapshot to render as an empty component, probably something undefined that should be either added to the snapshot definition, or checked in code.

Check setupProps in MineNoticeOfWorkTable.spec.js, I'm thinking you should add a definition for location there.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
exzizt: I didn't know that it would, thanks!

services/core-web/src/tests/components/dashboard/noticeOfWorkHomePage/__snapshots__/NoticeOfWorkTable.spec.js.snap
exzizt: This was generated/changed when I ran the updateAll command. Not too sure about what much of this means yet to be honest.

exzizt:
* Added "back" buttons to the view/edit NoW pages to either return you to the Browse Notices of Work page or a specific mine's NoW page, depending on where you came from.
* Scrolling through the NoW sections no longer updates the URL; only clicking on a NoW heading in the side menu does (fixes the undesired back-button behavior).
* Changed "Browse Notice of Work" to "Browse Notices of Work" in various locations.

exzizt:
@mmumby 
I wasn't able to use the method of providing default props to circumvent the issue of having to check for nulls down the property chain, but I tidied up my initial approach a bit. This is because the props I need are stored in the history/location state, which gets wiped out after the default props are set.

exzizt:
Another thing I've noticed that we might want to look into is that if the user refreshes the URL "http://localhost:3000/dashboard/notice-of-work/application/7eea63a1-e88a-490e-99f4-ec82f20d7d43/view#state-of-land", for example, upon loading the page, we could automatically scroll them down to the #state-of-land section. Currently, any route is removed upon loading the page and they always start from the beginning.

services/core-web/src/components/common/wrappers/ScrollContentWrapper.js
```diff
@@ -10,30 +10,40 @@ import { PropTypes } from "prop-types";
...
-export const ScrollContentWrapper = (props) => {
-  const isActive = () =>
-    includes(props.location.hash, props.id) ? "circle purple" : "circle grey";
+class ScrollContentWrapper extends Component {
+  isActive = () => {
+    const activeRoute =
+      this.props.history && this.props.history.location && this.props.history.location.state
+        ? this.props.history.location.state.activeRoute
+        : undefined;
+    const isActiveRoute = includes(activeRoute, this.props.id);
```
mmumby: In this spot, if the 'this.props.history.location.state.activeRoute' doesn't exist( in the case where the user bookmarks the tab, and the '/#state-of-land' exist in the URL) You should be able to check the URL for that search term and use it as the second option if it exists instead of using `undefined`

exzizt: Good idea. I tried to do that earlier (on page load and if URL fragment present, go to corresponding NoW section), but I noticed an issue (see comment above) where URL fragments are being removed. It would be good to set that up!

exzizt:
* Page will now scroll to current NoW section found in URL fragment

services/core-web/src/components/common/wrappers/ScrollContentWrapper.js
```diff
@@ -10,30 +10,40 @@ import { PropTypes } from "prop-types";
...
+  history: PropTypes.shape({
+    location: PropTypes.shape({ state: PropTypes.shape({ currentActiveLink: PropTypes.string }) }),
+  }).isRequired,
 };
 
-export const ScrollContentWrapper = (props) => {
-  const isActive = () =>
-    includes(props.location.hash, props.id) ? "circle purple" : "circle grey";
+class ScrollContentWrapper extends Component {
+  isActive = () => {
```
mmumby: Looks like this doesn't use state, can you convert back to a stateless component.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -72,20 +78,25 @@ const defaultProps = {
...
-    fixedTop: false,
-    menuVisible: false,
-    isDecision: false,
-    buttonValue: "REV",
-    buttonLabel: "Technical Review",
-  };
+  static noticeOfWorkPageFromRoute = {};
+
+  constructor(props) {
+    super(props);
```
mmumby: This change isn't necessary unless you are using the constructor for something else, if not, initializing as 'state = {}` is consistent throughout the app.

services/core-web/src/styles/components/NoticeOfWork.scss
```diff
@@ -144,7 +144,7 @@
     position: relative;
     padding-right: 20px;
     left: 250px;
-    width: calc(100vw - 250px);
+    width: calc(100vw - 270px);
```
mmumby: What was this change addressing?

services/core-web/src/components/noticeOfWork/applications/ViewNoticeOfWorkApplication.js
```diff
@@ -43,21 +49,29 @@ const propTypes = {
...
-      );
       this.setState({ isLoaded: true });
     });
     this.props.fetchOriginalNoticeOfWorkApplication(id);
+    this.noticeOfWorkPageFromRoute =
+      this.props.location &&
+      this.props.location.state &&
+      this.props.location.state.noticeOfWorkPageFromRoute
+        ? this.props.location.state.noticeOfWorkPageFromRoute
+        : this.noticeOfWorkPageFromRoute;
```
mmumby: So I don't think this variable `noticeOfWorkPageFromRoute` is needed, you should be able to use`this.props.location.state.noticeOfWorkPageFromRoute` as your check

services/core-web/src/components/noticeOfWork/applications/ViewNoticeOfWorkApplication.js
```diff
@@ -43,21 +49,29 @@ const propTypes = {
...
 
 export class ViewNoticeOfWorkApplication extends Component {
-  state = {
-    isLoaded: false,
-    showOriginalValues: false,
-  };
+  static noticeOfWorkPageFromRoute = {};
+
+  constructor(props) {
+    super(props);
```
mmumby: same here, the conversion is not needed.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -113,6 +124,12 @@ export class NoticeOfWorkApplication extends Component {
       });
     });
     this.props.fetchOriginalNoticeOfWorkApplication(id);
+    this.noticeOfWorkPageFromRoute =
```
mmumby: same note here regarding `noticeOfWorkPageFromRoute`

mmumby:
https://minesdigitalservices-test.pathfinder.gov.bc.ca/dashboard/notice-of-work/application/bbbdec68-89d1-46ff-b92b-6408d4a696f3/#application-info


some things I noticed when testing: 
1) Active styling is not updated if you use the browser back button, using componentWillReceiveprops should address that issue
2) Fixed sideNav/header is no longer working on scroll, it works when the link is clicked, but the styling is off.

services/core-web/src/components/common/wrappers/ScrollContentWrapper.js
exzizt: Done!

services/core-web/src/styles/components/NoticeOfWork.scss
exzizt: If we keep it at -250px, there is an unnecessary horizontal scrollbar that is created on the page. We need to account not only for left's 250px, but also the padding-right's 20px to remove this unnecessary scrollbar.

services/core-web/src/components/noticeOfWork/applications/ViewNoticeOfWorkApplication.js
exzizt: I created the variable `noticeOfWorkPageFromRoute` because I need to access the value of `this.props.location.state.noticeOfWorkPageFromRoute` in a few other places, so instead of doing the long null check in those other places as well, I stored the result of it once.
I have also switched it from being a static variable to a state variable now.

services/core-web/src/components/common/wrappers/ScrollContentWrapper.js
```diff
@@ -10,13 +10,21 @@ import { PropTypes } from "prop-types";
...
 };
 
 export const ScrollContentWrapper = (props) => {
-  const isActive = () =>
-    includes(props.location.hash, props.id) ? "circle purple" : "circle grey";
+  const isActive = () => {
+    const currentActiveLink =
+      props.history && props.history.location && props.history.location.state
+        ? props.history.location.state.currentActiveLink
+        : undefined;
```
exzizt: Prevents error `TypeError: Cannot read property 'currentActiveLink' of undefined`.

services/core-web/src/components/noticeOfWork/applications/NOWSideMenu.js
```diff
@@ -10,51 +10,117 @@ const { Link } = Anchor;
...
+      !this.props.location.hash.startsWith("#state")
+        ? this.props.location.hash
+        : undefined;
+
+    if (!link) {
+      return;
+    }
+
+    // Extracts "#blasting" from "#blasting&state=bd74ea1c-09e5-4d7e-810f-d...", for example.
+    link = link.substr(0, link.indexOf("&"));
```
exzizt: I looked into using the "query-string" module but it doesn't provide a cleaner method of extracting the NoW URL fragment (if present).
E.g., for the URL "http://localhost:3000/dashboard/notice-of-work/application/7e6e23cf-0d98-45e2-a059-739d345aeed8/view#underground-exploration", the following code returns the below object (since Keycloak authorization redirects and appends various params to the URL).

`const params = queryString.parse(this.props.location.hash);  
console.log("params:", params);`

`{  
code: "777ef559-1cc2-4dbb-a646-488450d77a2d.e2cef18a-5b94…6b1096fbf63c.ba58208e-e55c-417b-b845-bd472b79c6da",  
session_state: "e2cef18a-5b94-4b69-8142-6b1096fbf63c",  
state: "fedeef5e-6487-4f7b-80df-310db80e69b3",  
underground-exploration: null  
}`

PR #1114, opened 2019-12-20
MDS 2491 Core NOW Documents

nayr974:
This PR adds document uploads to notices of work.


migrations/sql/V2019.12.13.14.08__update_now_documents.sql
```diff
@@ -0,0 +1,23 @@
...
+	('BLP', 'Blasting Procedure', true, 'system-mds', 'system-mds'),
+	('EMS', 'Explosives Magazine Storage and Use Permit Application', true, 'system-mds', 'system-mds'),
+	('LAL', 'Landowner Authorization Letter', true, 'system-mds', 'system-mds'),
+	('MRP', 'Mine Emergency Response Plan', true, 'system-mds', 'system-mds'),
+	('OTH', 'Other', true, 'system-mds', 'system-mds'),
+	('RFE', 'Record of First Nations Engagement', true, 'system-mds', 'system-mds'),
+	('TAL', 'Tenure Authorization Letter', true, 'system-mds', 'system-mds'),
+	('TMP', 'Tenure Map / Property Map', true, 'system-mds', 'system-mds'),
+	('MPW', 'Map of Proposed Work', true, 'system-mds', 'system-mds')
+on conflict do nothing;
```
mmumby: Should this be removed now that its in the seed file?

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+  openModal: PropTypes.func.isRequired,
+  closeModal: PropTypes.func.isRequired,
+  now_application_guid: PropTypes.string.isRequired,
+  mine_guid: PropTypes.string.isRequired,
+  documents: PropTypes.arrayOf(PropTypes.any).isRequired,
+  noticeOfWorkApplicationDocumentTypeOptionsHash: PropTypes.objectOf(PropTypes.any).isRequired,
+  arrayPush: PropTypes.func.isRequired,
+  isViewMode: PropTypes.bool.isRequired,
+};
+
```
mmumby: Unimportant change, but for clarity, can you move this to the top of the file.

migrations/sql/V2019.12.13.14.08__update_now_documents.sql
nayr974: I prefer to include a migration script with inserts at the same time, so it better represents what happens in production. If this data gets inserted into production, removing or changing a value in the seed data script won't remove or update that value in production - but would in our dev environment otherwise.

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
nayr974: I thought that typically we have been putting function definitions that are not part of the component (stateless functions) above PropTypes in other files, but I see it's inconstant in different places. I'll move it :)

PR #1116, opened 2019-12-24
Fix/MDS-2463 Users can multi-select incident type

exzizt:
Users can now assign multiple incident types to a mine incident report.


migrations/sql/V2019.12.21.23.39__create_mine_incident_category_codes_table.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE mine_incident DROP CONSTRAINT IF EXISTS mine_incident_mine_incident_category_code_fkey;
+ALTER TABLE mine_incident DROP COLUMN IF EXISTS mine_incident_category_code;
+
+ALTER TABLE mine_incident_category ALTER COLUMN mine_incident_category_code TYPE varchar(3);
```
exzizt: I changed mine_incident_category_code across the board to restrict its size to 3 characters.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -74,33 +75,33 @@ class MineIncident(AuditMixin, Base):
 
     determination_type = db.relationship(
         'MineIncidentDeterminationType',
-        backref='mine_incidents',
+        backref='mine_incident',
```
exzizt: Pretty sure these were typos

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -74,33 +75,33 @@ class MineIncident(AuditMixin, Base):
...
-    mine_incident_category_code = db.Column(
-        db.String, db.ForeignKey('mine_incident_category.mine_incident_category_code'))
-    mine_incident_category = db.relationship('MineIncidentCategory', lazy='joined')
 
     recommendations = db.relationship(
         'MineIncidentRecommendation',
         primaryjoin=
         "and_(MineIncidentRecommendation.mine_incident_id == MineIncident.mine_incident_id, MineIncidentRecommendation.deleted_ind==False)",
-        lazy='selectin')
+        lazy='select')
```
exzizt: And this was a typo as well

services/core-api/app/api/incidents/models/mine_incident_category.py
```diff
@@ -9,10 +9,10 @@
...
     __tablename__ = 'mine_incident_category'
-    mine_incident_category_code = db.Column(db.String, primary_key=True)
+    mine_incident_category_code = db.Column(db.String(3), primary_key=True)
     description = db.Column(db.String, nullable=False)
     active_ind = db.Column(db.Boolean, nullable=False, server_default=FetchedValue())
 
     @classmethod
     def active(cls):
-        return cls.query.filter_by(active_ind=True).all()
+        return cls.query.filter_by(active_ind=True).order_by(cls.description).all()
```
exzizt: Sorting it alphabetically will make it easier for users to find a code in the dropdown, especially as more codes are added. This would be redundant if we decide to sort all of the records in the mine_incident_category table by default as we insert new codes.

services/core-api/app/api/incidents/models/mine_incident_category_xref.py
```diff
@@ -0,0 +1,43 @@
...
+    @classmethod
+    def create_batch(cls, mine_incident_id, mine_incident_category_codes, add_to_session=True):
+        if mine_incident_category_codes is not None:
+            mine_incident_category_codes.sort()
+            for mine_incident_category_code in mine_incident_category_codes:
+                if mine_incident_category_code is None:
+                    continue
+                new_mine_incident_category_code = MineIncidentCategoryXref.create(
+                    mine_incident_id, mine_incident_category_code, add_to_session)
+                new_mine_incident_category_code.save()
```
exzizt: Everything regarding the add_to_session, .save(), .save(commit=False) seems weird to me, as I think we just need to save() it, but I followed a previous pattern someone did when creating the code for creating recommendations.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -178,6 +179,10 @@ def post(self, mine_guid):
                     rec_string, mine_incident_id=incident.mine_incident_id)
                 new_recommendation.save()
 
+        mine_incident_category_codes = data.get('mine_incident_category_xref')
+        MineIncidentCategoryXref.create_batch(incident.mine_incident_id,
+                                              mine_incident_category_codes)
+
```
exzizt: Code to create the incident types in the incident record (POST).

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -348,3 +361,8 @@ def _handle_recommendations(cls, incident, recommendations):
             new_recommendation = MineIncidentRecommendation.create(
                 recommendation['recommendation'], mine_incident_id=incident.mine_incident_id)
             new_recommendation.save()
+
+    @classmethod
+    def _handle_mine_incident_category_codes(cls, mine_incident_id, mine_incident_category_codes):
+        MineIncidentCategoryXref.delete(mine_incident_id)
+        MineIncidentCategoryXref.create_batch(mine_incident_id, mine_incident_category_codes)
```
exzizt: Code to update the incident types for an incident record (PUT).

services/core-api/app/api/mines/response_models.py
```diff
@@ -8,15 +8,17 @@ class DateTime(fields.Raw):
...
-BASIC_MINE_LOCATION_MODEL = api.model('BasicMineLocation', {
-    'latitude': fields.String,
-    'longitude': fields.String,
-    'utm_easting': fields.String,
-    'utm_northing': fields.String,
-    'utm_zone_number': fields.String,
-    'utm_zone_letter': fields.String,
-    'mine_location_description': fields.String,
-})
+
```
exzizt: Lots of auto-format stuff.

services/core-web/src/components/Forms/incidents/AddIncidentReportingForm.js
```diff
@@ -20,29 +28,31 @@ export const AddIncidentReportingForm = (props) => (
       <Row gutter={48}>
         <Col>
           {props.initialValues.mine_incident_id_year && (
-            <h4>{`Ministry Incident No. :  ${props.initialValues.mine_incident_report_no}`}</h4>
+            <h4>{`Ministry Incident No. ${props.initialValues.mine_incident_report_no}`}</h4>
```
exzizt: Should this say "Mine Incident No."?

services/core-web/src/components/mine/Incidents/MineIncidents.js
```diff
@@ -156,6 +156,9 @@ export class MineIncidents extends Component {
           dangerous_occurrence_subparagraph_ids: existingIncident.dangerous_occurrence_subparagraph_ids.map(
             String
           ),
+          mine_incident_category_xref: existingIncident.mine_incident_category_xref
+            ? existingIncident.mine_incident_category_xref.map((c) => c.mine_incident_category_code)
+            : [],
```
exzizt: Initial values for the incident types field

services/core-web/src/utils/Validate.js
```diff
@@ -50,6 +50,9 @@ export const Validate = new Validator();
 
 export const required = (value) => (value ? undefined : "This is a required field");
 
+export const requiredList = (value) =>
+  value && value.length > 0 ? undefined : "This is a required field";
+
```
exzizt: Created to validate fields with list data types

migrations/sql/V2019.12.21.23.39__create_mine_incident_category_codes_table.sql
```diff
@@ -0,0 +1,25 @@
...
+
+ALTER TABLE mine_incident_category_xref
+    DROP CONSTRAINT IF EXISTS mine_incident_category_xref_mine_incident_fkey CASCADE;
+ALTER TABLE mine_incident_category_xref
+    ADD CONSTRAINT mine_incident_category_xref_mine_incident_fkey FOREIGN KEY (mine_incident_id) REFERENCES mine_incident(mine_incident_id) ON UPDATE CASCADE ON DELETE CASCADE;
+
+ALTER TABLE mine_incident_category_xref
+    DROP CONSTRAINT IF EXISTS mine_incident_category_xref_mine_incident_category_fkey CASCADE;
+ALTER TABLE mine_incident_category_xref
+    ADD CONSTRAINT mine_incident_category_xref_mine_incident_category_fkey FOREIGN KEY (mine_incident_category_code) REFERENCES mine_incident_category(mine_incident_category_code) ON UPDATE CASCADE ON DELETE CASCADE;
```
Jsyro: This looks great and should run no issue. FYI Postgres has a short hand for this. 

from V2019.04.08.09.32__create_incident_reports.sql.

```
CREATE TABLE table_name( 
<column>
<column>
....
    CONSTRAINT mine_guid_fkey FOREIGN KEY (mine_guid) REFERENCES mine(mine_guid) DEFERRABLE INITIALLY DEFERRED,
    CONSTRAINT mine_incident_followup_fkey FOREIGN KEY (followup_type_code) REFERENCES mine_incident_followup_type(mine_incident_followup_type_code) DEFERRABLE INITIALLY DEFERRED
);```

File: services/core-api/app/api/incidents/models/mine_incident.py (hunk 2 above)
Comment (Jsyro) [reply services/core-api/app/api/incidents/models/mine_incident.py]: These are not typos. this defines the reverse relationship from the target object to this object. 

for example, we could go determination_type.mine_incidents and get all the incidents that use that code, so it is plural. Backref is weird and I think long term we are going to phase it out because it's tough to track what relationships exist. for now just revert this change.

File: services/core-api/app/api/incidents/models/mine_incident.py (hunk 3 above)
Comment (Jsyro) [reply services/core-api/app/api/incidents/models/mine_incident.py]: these are enumerated options. see here, selectin was used deliberately i believe. 

https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html

File: services/core-api/app/api/incidents/models/mine_incident.py [hunk 13]
```diff
@@ -74,33 +75,33 @@ class MineIncident(AuditMixin, Base):
...
     documents = db.relationship('MineIncidentDocumentXref', lazy='joined')
     mine_documents = db.relationship(
         'MineDocument', lazy='joined', secondary='mine_incident_document_xref')
 
+    mine_incident_category_xref = db.relationship(
+        'MineIncidentCategoryXref', backref='mine_incident', lazy='select')
+
```
Jsyro: see above with the use of the xref table as a secondary, can this do the same. 

```mine_incident_categories = db.relationship('MineIncidentCategory', secondary='mine_incident_category_xref', backref='mine_incidents', lazy='select')```

File: services/core-api/app/api/incidents/models/mine_incident_category.py (hunk 4 above)
Comment (Jsyro) [reply services/core-api/app/api/incidents/models/mine_incident_category.py]: This also appears to be inconsistent in our app, but we also have `display_order` as column to do just this. assuming it's stable even without your change, i'd rather not have a 3rd style introduced. let me know your thoughts.

File: services/core-api/app/api/incidents/models/mine_incident.py (hunk 3 above)
Comment (exzizt) [reply services/core-api/app/api/incidents/models/mine_incident.py]: Good catch, I didn't see that in the docs.

File: services/core-api/app/api/incidents/models/mine_incident_category_xref.py (hunk 5 above)
Comment (Jsyro) [reply services/core-api/app/api/incidents/models/mine_incident_category_xref.py]: this seems unnecessary because a simple for loop in the resource can do this. We also want to avoid creating these relationships by explicitly referring to id's. 

eg. what if the mine_incident is brand new and hasn't gotten gotten an ID yet.

the consumer of this function should just loop like this eg from permit_amendment.py resource file. 

`        for newFile in uploadedFiles:
            new_pa_doc = PermitAmendmentDocument(
                document_name=newFile['fileName'],
                document_manager_guid=newFile['document_manager_guid'],
                mine_guid=permit.mine_guid,
            )
            new_pa.related_documents.append(new_pa_doc)
`
using the relationship with the secondary flag will let you refer to it as a set (and just append) then the ORM deals with the PK/FK's as it gets saved.

File: services/core-api/app/api/incidents/response_models.py [hunk 14]
```diff
@@ -49,12 +47,18 @@
     }
 )
 
+MINE_INCIDENT_CATEGORY_XREF_MODEL = api.model(
```
Jsyro: this will also get hidden when you use the relationship with the secondary flag to get many-to-many.

services/core-api/app/api/incidents/response_models.py
```diff
@@ -83,7 +87,9 @@
         'mine_incident_no': fields.String,
         'documents': fields.List(fields.Nested(MINE_INCIDENT_DOCUMENT_MODEL)),
         'recommendations': fields.List(fields.Nested(
-            MINE_INCIDENT_RECOMMENDATION_MODEL))
+            MINE_INCIDENT_RECOMMENDATION_MODEL)),
+        'mine_incident_category_xref': fields.List(fields.Nested(
```
Jsyro: this just refers directly to the many-to-many relationship and reuse the existing MINE_INCIDENT_DOCUMENT_MODEL.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: this should almost identically mimic how `dangerous_occurrence_subparagraph_ids` which has identical behaviour and shouldn't have many edge cases

services/core-api/app/api/incidents/models/mine_incident.py
exzizt: I see. Are we even making use of these backrefs in our project?

services/core-api/app/api/incidents/models/mine_incident_category.py
exzizt: Are you suggesting I add a display_order (integer) column to the mine_incident_category table and then do order_by(cls.display_order) instead? If so, I could do that no problem.

migrations/sql/V2019.12.21.23.39__create_mine_incident_category_codes_table.sql
exzizt: Thanks. I thought about doing it shorthand too but I was just following how I saw it done in another file. The DDL that DBeaver generates would be good too.

services/core-api/app/api/incidents/models/mine_incident.py
exzizt: Like this?
`    mine_incident_category_xref = db.relationship('MineIncidentCategoryXref', lazy='joined')
    mine_incident_categories = db.relationship(
        'MineIncidentCategory', secondary='mine_incident_category_xref', backref='mine_incidents', lazy='select')`

Would this change actually affect anything (as everything is working now) and require more changes, or is it just to merely better represent the relationships/model?

services/core-api/app/api/incidents/response_models.py
exzizt: So this MINE_INCIDENT_CATEGORY_XREF_MODEL is not required if I make those changes you mentioned in your [other comment](https://github.com/bcgov/mds/pull/1116#discussion_r361201547)?

services/core-api/app/api/incidents/response_models.py
exzizt: I'm not sure what you mean exactly.

services/core-web/src/utils/Validate.js
mmumby: Does the `required` validation not work in this situation?

services/core-api/app/api/incidents/models/mine_incident_category_xref.py
exzizt: I've tested creating brand new mine incidents and this doesn't seem to be a problem, but I'll check it out.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
exzizt: Ok, I'll see how it compares to that. I was using 'recommendations' stuff as a reference a lot. I believe it is also structured similarily?

services/core-api/app/api/incidents/models/mine_incident.py
Jsyro: you could leave both, but you've had to write alot of code further down that doesn't match existing patterns, so i would like to see this cleaned up before we merge.

services/core-api/app/api/incidents/models/mine_incident_category.py
Jsyro: Yea, that would be best version of the code table we have so far.

services/core-web/src/utils/Validate.js
exzizt: I tried that one first and realized it wasn't working because an empty array is considered "truthy" in JS.

services/core-api/app/api/incidents/models/mine_incident_category_xref.py
Jsyro: hmm i see very similar assignments in the same POST, i'm surprised they work as they do. because until the object has been saved, the serial ID should still be null, maybe it's generated when it's attached to the session,,,

services/core-api/app/api/incidents/models/mine_incident_category.py
exzizt: I will make that change.

services/core-api/app/api/incidents/response_models.py
Jsyro: correct, as far the caller of our API is concerned, a mine_incident has a set of categories, that's how it should look. having a set of xrefs would not be ideal.

services/core-api/app/api/incidents/response_models.py
Jsyro: change line 91/92 to 

`'<relationship_name>' : fields.List(fields.Nested( MINE_INCIDENT_DOCUMENT_MODEL))`

services/core-api/app/api/incidents/models/mine_incident.py
exzizt: Ok. I was looking at the "recommendations" code as a reference for patterns actually, but now I see that looking at the "documents xref" table is a better candidate, as the relationship it has to mine_incident is the same as the new "category xref" table. (?)

services/core-web/src/utils/Validate.js
exzizt: I could probably change "required" to 
`export const required = (value) => (value && value.length > 0 ? undefined : "This is a required field");` and it could work for both string and array/object fields.

services/core-api/app/api/incidents/response_models.py
exzizt: Document?

Jsyro: Category, sorry.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -88,19 +89,21 @@ class MineIncident(AuditMixin, Base):
...
         lazy='selectin')
+
     documents = db.relationship('MineIncidentDocumentXref', lazy='joined')
     mine_documents = db.relationship(
         'MineDocument', lazy='joined', secondary='mine_incident_document_xref')
 
+    categories = db.relationship('MineIncidentCategoryXref', lazy='joined')
+    mine_categories = db.relationship(
+        'MineIncidentCategory', lazy='joined', secondary='mine_incident_category_xref')
+
```
exzizt: This now follows the same pattern used as the MineIncidentDocumentXref table as seen above.

services/core-api/app/api/incidents/models/mine_incident_category.py
```diff
@@ -9,10 +9,11 @@
...
-    mine_incident_category_code = db.Column(db.String, primary_key=True)
+    mine_incident_category_code = db.Column(db.String(3), primary_key=True)
     description = db.Column(db.String, nullable=False)
+    display_order = db.Column(db.Integer, nullable=False)
     active_ind = db.Column(db.Boolean, nullable=False, server_default=FetchedValue())
 
     @classmethod
     def active(cls):
-        return cls.query.filter_by(active_ind=True).all()
+        return cls.query.filter_by(active_ind=True).order_by(cls.display_order).all()
```
exzizt: Not sure if we want to just return them sorted (what we're doing here) or just use the display_order attribute on the client-side to handle the sorting.

services/core-api/app/api/incidents/models/mine_incident_category_xref.py
```diff
@@ -0,0 +1,29 @@
...
+    mine_incident_category_code = db.Column(
+        db.String(3),
+        db.ForeignKey('mine_incident_category.mine_incident_category_code'),
+        primary_key=True,
+        nullable=False)
+
+    mine_incident_category = db.relationship('MineIncidentCategory', lazy='joined')
+
+    description = association_proxy('mine_incident_category', 'description')
+    display_order = association_proxy('mine_incident_category', 'display_order')
```
exzizt: I had to set up these association proxies to implement the M:M relationship.

services/core-api/app/api/incidents/models/mine_incident_category_xref.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @classmethod
+    def delete(cls, mine_incident_id):
+        cls.query.filter_by(mine_incident_id=mine_incident_id).delete()
+        db.session.commit()
```
exzizt: This delete method is required (or at least something like it is) for our current PUT implementation.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -178,6 +173,21 @@ def post(self, mine_guid):
...
+                    continue
+                new_category = MineIncidentCategoryXref(
+                    mine_incident_id=incident.mine_incident_id,
+                    mine_incident_category_code=category)
+                incident.categories.append(new_category)
+
+        try:
+            incident.save()
+        except Exception as e:
+            raise InternalServerError(f'Error when saving: {e}')
```
exzizt: I moved this to the bottom of the POST function because it makes more sense there.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -178,6 +173,21 @@ def post(self, mine_guid):
...
 
+        categories = data.get('categories')
+        if categories is not None:
+            for category in categories:
+                if category is None:
+                    continue
+                new_category = MineIncidentCategoryXref(
+                    mine_incident_id=incident.mine_incident_id,
+                    mine_incident_category_code=category)
+                incident.categories.append(new_category)
```
exzizt: Follows the pattern used in "mine_incident_doc" above.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -274,6 +283,17 @@ def put(self, mine_guid, mine_incident_guid):
...
+        MineIncidentCategoryXref.delete(incident.mine_incident_id)
+        if categories is not None:
+            for category in categories:
+                if category is None:
+                    continue
+                new_category = MineIncidentCategoryXref(
+                    mine_incident_id=incident.mine_incident_id,
+                    mine_incident_category_code=category)
+                incident.categories.append(new_category)
+
```
exzizt: Incident PUT: Follows the pattern used in "mine_incident_doc" below.

services/core-api/app/api/mines/response_models.py
exzizt: Auto-format incoming...

services/core-web/src/components/modalContent/ViewIncidentModal.js
```diff
@@ -38,16 +38,20 @@ export class ViewIncidentModal extends Component {
...
             <p>
-              {this.props.incidentCategoryCodeHash[
-                this.props.incident.mine_incident_category_code
-              ] || Strings.EMPTY_FIELD}
+              {this.props.incident.categories && this.props.incident.categories.length > 0
+                ? this.props.incident.categories
+                    .sort((a, b) => (a.display_order > b.display_order ? 1 : -1))
+                    .map((c) => this.props.incidentCategoryCodeHash[c.mine_incident_category_code])
+                    .join(", ")
+                : Strings.EMPTY_FIELD}
```
exzizt: Display the incident categories sorted in ascending order by their display_order attribute.

services/core-web/src/components/Forms/incidents/AddIncidentReportingForm.js
```diff
@@ -20,29 +28,37 @@ export const AddIncidentReportingForm = (props) => (
...
+              placeholder="Select the incident type(s)"
+              component={renderConfig.MULTI_SELECT}
+              validate={[requiredList]}
+              data={
+                props.incidentCategoryCodeOptions
+                  ? props.incidentCategoryCodeOptions.sort((a, b) =>
+                      a.display_order > b.display_order ? 1 : -1
+                    )
+                  : []
+              }
```
exzizt: Display the incident category options sorted in ascending order by their display_order attribute.

migrations/sql/V2019.12.21.23.39__support_multiple_incident_types.sql
```diff
@@ -0,0 +1,46 @@
...
+UPDATE mine_incident_category
+SET display_order = 1
+WHERE mine_incident_category_code = 'GTC';
+
+UPDATE mine_incident_category
+SET display_order = 2
+WHERE mine_incident_category_code = 'H&S';
+
+UPDATE mine_incident_category
+SET display_order = 3
```
Jsyro: can we change the display orders to be 10, 20, 30, 40 so that we can add new records between in the future.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
Jsyro: we shouldn't have to create xref records in this instance. Look at lines 138-144 of this file.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -88,19 +89,21 @@ class MineIncident(AuditMixin, Base):
...
         primaryjoin=
         "and_(MineIncidentRecommendation.mine_incident_id == MineIncident.mine_incident_id, MineIncidentRecommendation.deleted_ind==False)",
         lazy='selectin')
+
     documents = db.relationship('MineIncidentDocumentXref', lazy='joined')
     mine_documents = db.relationship(
         'MineDocument', lazy='joined', secondary='mine_incident_document_xref')
 
+    categories = db.relationship('MineIncidentCategoryXref', lazy='joined')
+    mine_categories = db.relationship(
```
Jsyro: mine_categories is a bit misleading, it should be incident_categories, but i don't think we need the first relationship as the xref is just to FK's.

services/core-api/app/api/incidents/response_models.py
```diff
@@ -4,7 +4,9 @@
 MINE_INCIDENT_CATEGORY_MODEL = api.model(
     'Mine Incident Category', {
         'mine_incident_category_code': fields.String,
-        'description': fields.String
+        'description': fields.String,
+        'display_order': fields.Integer,
+        'active_ind': fields.Boolean
```
pdittaro: Do we use active_ind on other response models?

services/core-api/app/api/mines/response_models.py
```diff
@@ -81,29 +80,31 @@ def format(self, value):
...
+        'permit_amendment_id': fields.Integer,
         'permit_amendment_guid': fields.String,
         'permit_amendment_status_code': fields.String,
         'permit_amendment_type_code': fields.String,
         'received_date': fields.DateTime(dt_format='iso8601'),
         'issue_date': fields.DateTime(dt_format='iso8601'),
         'authorization_end_date': fields.DateTime(dt_format='iso8601'),
-        #'permit_amendment_status_description': fields.String,
-        #'permit_amendment_type_description': fields.String,
+                                                                                         #'permit_amendment_status_description': fields.String,
```
pdittaro: This indenting seems way off for some reason.

exzizt: I agree. It was the auto-formatter. Maybe there is a setting we can add to prevent this.

services/core-api/app/api/incidents/response_models.py
exzizt: I'm not sure but Jason requested I put it in for future use.

PR #1124, opened 2020-01-06
Pipelines PVC provisioning fix

pdittaro:


pdittaro:
This change to have seperate template files for the test and prod environments is not ideal, but a stopgap solution while we investigate how to deal with the netapp PVC provisioning.

PR #1127, opened 2020-01-08
MDS-2523: Environment-based feature flagging for MineSpace

exzizt:
Need feedback everything. I've basically copied over the required components/files from what I can tell. Some variables/lines in these files are probably Core-specific and may not be required in MineSpace. Let me know!

### DIFF
--- a/services/minespace-web/server.js
+++ b/services/minespace-web/server.js
@@ -4,6 +4,7 @@ const dotenv = require("dotenv").config({ path: `${__dirname}/.env` });
 
 let BASE_PATH = process.env.BASE_PATH;
 let BUILD_DIR = process.env.BUILD_DIR || "build";
+const VENDOR_DIR = process.env.VENDOR_DIR || "vendor";
 let PORT = process.env.PORT || 3020;
 if (dotenv.parsed) {
   BASE_PATH = dotenv.parsed.BASE_PATH || BASE_PATH;
@@ -29,16 +30,23 @@ const staticServe = express.static(`${__dirname}/${BUILD_DIR}`, {
   maxAge: "1y",
 });
 
+const vendorServe = express.static(`${__dirname}/${VENDOR_DIR}`, {
+  immutable: true,
+  maxAge: "1y",
+});
+
 app.get(`${BASE_PATH}/env`, (req, res) => {
   res.json({
     backend: "mds-python-backend",
     apiUrl: process.env.API_URL,
     docManUrl: process.env.DOCUMENT_MANAGER_URL,
+    firstNationsLayerUrl: process.env.FN_LAYER_URL,
     keycloak_resource: process.env.KEYCLOAK_RESOURCE,
     keycloak_clientId: process.env.KEYCLOAK_CLIENT_ID,
-    keycloak_idpHint: process.env.KEYCLOAK_IDP_HINT,
     keycloak_url: process.env.KEYCLOAK_URL,
+    keycloak_idpHint: process.env.KEYCLOAK_IDP_HINT,
     siteminder_url: process.env.SITEMINDER_URL,
+    environment: process.env.NODE_ENV,
   });
 });
 
@@ -49,8 +57,10 @@ app.get(`/health`, (req, res) => {
 });
 
 app.use(`${BASE_PATH}/`, staticServe);
+app.use(`${BASE_PATH}/vendor/`, vendorServe);
 app.use(`${BASE_PATH}*`, staticServe);
 app.use(`/`, staticServe);
+app.use(`/vendor/`, staticServe);
 app.use(`*`, staticServe);
 
 app.listen(PORT, "0.0.0.0", () => console.log("Server running"));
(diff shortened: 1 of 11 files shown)


services/minespace-web/src/constants/permissions.js
```diff
@@ -0,0 +1,13 @@
...
+export const EDIT_MINES = "role_edit_mines";
+export const EDIT_PERMITS = "role_edit_permits";
+export const EDIT_REPORTS = "role_edit_reports";
+export const EDIT_DO = "role_edit_do";
+export const EDIT_VARIANCES = "role_edit_variances";
+export const CLOSE_PERMITS = "role_close_permits";
+export const DEVELOPER = "role_developer";
+export const IN_DEVELOPMENT = "inDevelopment";
+export const IN_TESTING = "inTesting";
+export const EXECUTIVE = "role_executive_view";
```
exzizt: We probably don't need most of these

services/minespace-web/server.js
```diff
@@ -29,16 +30,23 @@ const staticServe = express.static(`${__dirname}/${BUILD_DIR}`, {
...
     docManUrl: process.env.DOCUMENT_MANAGER_URL,
+    firstNationsLayerUrl: process.env.FN_LAYER_URL,
     keycloak_resource: process.env.KEYCLOAK_RESOURCE,
     keycloak_clientId: process.env.KEYCLOAK_CLIENT_ID,
-    keycloak_idpHint: process.env.KEYCLOAK_IDP_HINT,
     keycloak_url: process.env.KEYCLOAK_URL,
+    keycloak_idpHint: process.env.KEYCLOAK_IDP_HINT,
     siteminder_url: process.env.SITEMINDER_URL,
+    environment: process.env.NODE_ENV,
   });
```
mmumby: This is the only thing that is needed right now

services/minespace-web/src/components/common/WarningBanner.js
```diff
@@ -58,7 +58,6 @@ const WarningBanner = (props) => (
     )}
     {props.type === "mobile" && (
       <Alert
-        style={{ backgroundColor: "#F3CD65" }}
```
exzizt: This banner had the same colour attribute as well so I removed it.

services/minespace-web/src/HOC/AuthorizationGuard.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @constant AuthorizationGuard - Higher Order Component that provides "feature flagging", in order to hide routes that
+ * are not ready to be in PROD, pass in `inDevelopment` or 'inTesting` param to keep the content environment specific.
+ */
+
+export const AuthorizationGuard = (permission) => (WrappedComponent) => {
+  const authorizationGuard = (props) => {
+    if (
+      (permission === "inDevelopment" && detectDevelopmentEnvironment()) ||
+      (permission === "inTesting" && !detectProdEnvironment())
```
exzizt: environmentUtils.js defines a `export const detectTestEnvironment = () => ENVIRONMENT.environment === "test";` yet we just use !detectProdEnvironment here?

services/minespace-web/src/components/common/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,54 @@
...
+  inDevelopment: false,
+  inTesting: false,
+};
+
+export const AuthorizationWrapper = (props) => (
+  <span>
+    {props.inDevelopment && detectDevelopmentEnvironment() && <span>{props.children}</span>}
+    {props.inTesting && !detectProdEnvironment() && <span>{props.children}</span>}
+  </span>
+);
```
exzizt: Do we have to wrap everything in a span? It seems like this could be an unwanted side-effect. That is, you just want to conditionally render your component, but this AuthorizationWrapper component will render your component(s) wrapped inside of a span element.

services/minespace-web/src/constants/permissions.js
```diff
@@ -0,0 +1,2 @@
+export const IN_DEVELOPMENT = "inDevelopment";
+export const IN_TESTING = "inTesting";
```
exzizt: These constants in this file and in Core's version don't seem to be used anywhere. Perhaps they're meant to be used in the `const authorizationGuard = ...` in AuthorizationGuard.js?

exzizt: Made it so they're used, modified Core to use them as well.

services/minespace-web/src/HOC/AuthorizationGuard.js
```diff
@@ -0,0 +1,28 @@
...
+export const AuthorizationGuard = (permission) => (WrappedComponent) => {
+  const authorizationGuard = (props) => {
+    if (
+      (permission === Permission.IN_DEVELOPMENT && detectDevelopmentEnvironment()) ||
+      (permission === Permission.IN_TESTING && !detectProdEnvironment())
+    ) {
+      return <WrappedComponent {...props} />;
+    }
+    return <NullScreen type="unauthorized-page" />;
+  };
```
mmumby: type="unauthorized"

PR #1128, opened 2020-01-09
MDS-2461: Form does not submit if you tab through time/date component

exzizt:
* Doing some more work in this PR as well, so WIP
* The cause of this issue is a bug in Ant Design. They are changing the value of the field in the onBlur event (to a string from a Moment), which is a side-effect and unintended and unexpected behaviour; just because a form field loses focus does not mean that its value should change. The solution (for now) is to just provide an onBlur event to prevent this from happening. I may open up a ticket in Ant Design. Other people have been posting similar issues to this one as well.


services/core-web/src/components/common/RenderTime.js
```diff
@@ -44,6 +44,7 @@ const RenderDate = (props) => (
       {...props.input}
       placeholder={props.placeholder}
       onChange={props.input.onChange}
+      onBlur={() => {}}
```
exzizt: If you'd rather have onBlur be a prop and the default value be `() => {}` and have `onBlur={props.onBlur}` let me know 🙂

services/core-web/src/styles/settings/theme.scss
```diff
@@ -15,8 +15,9 @@ $ok-green: #45a766;
 $primary-color: $violet;
 $menu-item-color: $white;
 
-$error-color: $alert-red;
-$success-color: $ok-green;
+// $error-color: $alert-red;
```
nayr974: Curious why these are commented out? If they won't be needed, please remove them entirely.

services/core-web/src/utils/helpers.js
```diff
@@ -86,6 +87,41 @@ export const normalizePhone = (value, previousValue) => {
...
+  // Extract the parts (name and extension) of the filename
+  const parts = /(^.+)\.([^.]+)$/.exec(filename);
+
+  // If the filename has no extension (e.g., the filename is "foo", not "foo.txt")
+  if (!parts) {
+    return `${filename.substring(0, max)}${trunc}`;
+  }
+
+  // Get the name of the filename (e.g., "foo.txt" will give "foo")
+  let name = parts[1];
```
nayr974: super super nitpicky but this could probably be done in one operation and maybe read a bit cleaner?

const name = parts[1].length > max ? `${parts[1].substring(0, max)}${trunc}` : parts[1];

services/core-web/src/utils/helpers.js
```diff
@@ -86,6 +87,41 @@ export const normalizePhone = (value, previousValue) => {
...
+  if (name.length > max) {
+    name = `${name.substring(0, max)}${trunc}`;
+  }
+
+  // Get the extension of the filename (e.g., "foo.txt" will give "txt")
+  let ext = parts[2];
```
nayr974: same comment here

pdittaro:
You mentioned about existing issues on the ant design library. In these cases are you able to add a link to the issue on the other issue board? Makes it easier for looking for resolutions later.

services/core-web/src/utils/helpers.js
exzizt: Sure, no problem!

services/core-web/src/constants/styles.js
```diff
@@ -10,3 +9,28 @@ export const COLOR = {
   dangerButtonRed: "#BC2929",
   backgroundWhite: "#ffffff",
 };
+
+// Styling types for various document/application status codes
```
pdittaro: I feel like this is the wrong location for these types. The purpose of this file, IMO, is to have a location to change styling attributes in a single location.

services/core-web/src/styles/settings/theme.scss
exzizt: Discussed in person.

services/core-web/src/constants/styles.js
pdittaro: Discussed with Luke, the issue I had with this is that there shouldn't be business domain logic in the styles file.

exzizt: Chatted and resolved in person.

PR #1129, opened 2020-01-09
Refactor Loading of Static Content

nayr974:
This PR moves all static content loading to the top level component. All action creators in staticContentActionCreators are executed automatically, so no hand maintained definition needs to be maintained besides the file itself.


services/core-web/src/components/Home.js
```diff
@@ -65,6 +71,15 @@ export class Home extends Component {
     this.setState((prevState) => ({ isMenuOpen: !prevState.isMenuOpen }));
   };
 
+  loadStaticContent = () => {
```
nayr974: This is the primary method that loads all static content.

services/core-web/src/components/Home.js
```diff
@@ -109,4 +124,4 @@ export class Home extends Component {
 
 Home.propTypes = propTypes;
 
-export default AuthenticationGuard(Home);
+export default connect(null)(AuthenticationGuard(Home));
```
mmumby: is the connect needed here?

nayr974: It is. Connect gives me access to the dispatch function, which then lets me dispatch each action without having to bind each action creator by name.

nayr974:
> 
> 
> Looks good, Ryan! Thank you for doing this.
> 
> Something to think about: what happens if one of the fetches fails inside the loop?

Yes, I'd like to handle this. I don't consider this PR complete yet - I wanted to throw it up and get feedback to see if anybody disagreed with this approach before putting more effort into it.

pdittaro:
> > Looks good, Ryan! Thank you for doing this.
> > Something to think about: what happens if one of the fetches fails inside the loop?
> 
> Yes, I'd like to handle this. I don't consider this PR complete yet - I wanted to throw it up and get feedback to see if anybody disagreed with this approach before putting more effort into it.

If we're consolidating our type code fetching in the frontend, does it make sense to do the same for the backend?

nayr974:
> 
> 
> > > Looks good, Ryan! Thank you for doing this.
> > > Something to think about: what happens if one of the fetches fails inside the loop?
> > 
> > 
> > Yes, I'd like to handle this. I don't consider this PR complete yet - I wanted to throw it up and get feedback to see if anybody disagreed with this approach before putting more effort into it.
> 
> If we're consolidating our type code fetching in the frontend, does it make sense to do the same for the backend?

I think so. Package it up in one call and have the API cache it as a next step.

services/core-web/src/components/dashboard/varianceHomePage/VarianceHomePage.js
```diff
@@ -328,31 +305,33 @@ export class VarianceHomePage extends Component {
...
-              handleNameFieldReset={this.handleNameFieldReset}
-              initialValues={this.state.params}
-              fetchVariances={this.props.fetchVariances}
-              handleVarianceSearch={this.handleVarianceSearchDebounced}
-              mineRegionOptions={this.props.mineRegionOptions}
-              complianceCodes={this.props.getDropdownHSRCMComplianceCodes}
-              filterVarianceStatusOptions={this.props.filterVarianceStatusOptions}
+          <div className="page__content">
+            <AuthorizationWrapper inTesting>
+              <VarianceSearch
```
mmumby: I know you didn't add this.. but should `VarianceSearch` be in prod??

PR #1130, opened 2020-01-09
Feature/mds 2542 - Adding Unit Codes, updating underground activity type

mmumby:
Opening the initial PR:
- added unit codes, updated some positioning so the new fields fit within the form. 
- Minor styling change to the Admin view when viewing a list of mines for proponents
Screen Shot 2020-01-07 at 2 19 07 PM
- added a new field input that has a built in drop down for situations like unit codes
- Will remove the `staticContent` fetches once Ryans PR is merged.


services/core-web/src/components/admin/MinespaceUserList.js
```diff
@@ -33,7 +33,13 @@ const columns = [
...
     render: (text) => (
       <div title="Mines">
-        {text && text.map(({ mine_guid, mine_name }) => <span key={mine_guid}> {mine_name} </span>)}
+        {text &&
+          text.map(({ mine_guid, mine_name }) => (
+            <span key={mine_guid}>
+              {mine_name}
+              <br />
+            </span>
+          ))}
```
exzizt: Do we want to display a message if they have no mines? Or is this dealt with already?

mmumby: chatted IRL - there should never be a situation where there isn't at least 1 mine in this list.

migrations/sql/V2020.01.09.9.30__update_sand_gravel_operation.sql
```diff
@@ -0,0 +1,22 @@
...
+ADD COLUMN average_overburden_depth_unit_type_code character varying(3),
+ADD COLUMN average_top_soil_depth_unit_type_code character varying(3);
+
+
+ALTER TABLE sand_gravel_quarry_operation
+ADD FOREIGN KEY (average_overburden_depth_unit_type_code) REFERENCES unit_type(unit_type_code) DEFERRABLE INITIALLY DEFERRED,
+ADD FOREIGN KEY (average_top_soil_depth_unit_type_code) REFERENCES unit_type(unit_type_code) DEFERRABLE INITIALLY DEFERRED;
+
+
+UPDATE unit_type set unit = 't' where unit_type_code = 'MTN';
```
nayr974: Ideally I think this column is named "unit" as you've used in the response model, or "short_description" or "display_value" or something along that line as "type" has a meaning in your database that might cause confusion. I'm fine leaving this as is if time constraints prevent a change.

migrations/sql/V2020.01.07.01.26__update_underground_exploration.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE underground_exploration
```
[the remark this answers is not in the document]
mmumby: It was on the wrong table, needed to be on the on underground_exploration_detail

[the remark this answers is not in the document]
mmumby: Which it already existed on.

migrations/sql/V2020.01.09.9.30__update_sand_gravel_operation.sql
mmumby: I can update this!

PR #1134, opened 2020-01-11
MDS-2477: Expose variances in MineSpace for proponents

exzizt:
# Main
* Variances are now exposed and can be viewed.
* Opening the view/edit variance modal shows the document type for each document.
# Other
* Fixed a bug that prevented mines from being displayed.
* Updated variance styling in the table (specifically for overdue variances). Note: will change in a future ticket to match the new mockup.
* Changed various text throughout the app. Note: will change later in a new ticket to match new mockups.
* Minimized some code.
* Sorted the mines list alphabetically in ascending order (at least there's some organization there for now even though that page will change a lot).
* Other minor changes.


services/minespace-web/src/components/common/QuestionsSidebar.js
```diff
@@ -1,17 +1,19 @@
...
-        We encourage your feedback and would like to hear from you. If you would like to get in
-        touch with us, send us an email to{" "}
-        <a className="underline" href="mailto:MDS@gov.bc.ca">
-          mds@gov.bc.ca
+        We encourage your feedback and would like to hear from you. If you have any questions,
+        feedback, or other concerns, please send us an email at&nbsp;
+        <a className="underline" href={`mailto:${Strings.MDS_EMAIL}`}>
+          {Strings.MDS_EMAIL}
         </a>
+        .
```
exzizt: I thought this looked/sounded nicer, but no worries if you want me to revert it to the previous text.

services/minespace-web/src/selectors/varianceSelectors.js
```diff
@@ -79,3 +80,17 @@ export const getVarianceStatusOptionsHash = createSelector(
...
+    options.map((option) => {
+      const composedLabel = `${option.description} Document`;
+      return { value: option.variance_document_category_code, label: composedLabel };
+    })
+);
+
+export const getVarianceDocumentCategoryOptionsHash = createSelector(
+  [getDropdownVarianceDocumentCategoryOptions],
+  createLabelHash
+);
```
exzizt: I have some big concerns about code duplication and maintainability issues. Will chat in person.

exzizt: Discussed in person during standup.

services/minespace-web/src/components/common/QuestionsSidebar.js
exzizt: I know have access to the MineSpace mockups which has the updated text to use.

exzizt: A lot has changed for this component/the home page in general so we can address this text again in a future ticket.

services/core-web/src/reducers/staticContentReducer.js
```diff
@@ -20,7 +20,7 @@ const initialState = {
   incidentStatusCodeOptions: [],
   incidentCategoryCodeOptions: [],
   varianceStatusOptions: [],
-  varianceDocumentCategoryOptions: [],
+  documentCategoryOptions: [],
```
Jsyro: are these document categories generic, why not keep the `variance` prefix?

exzizt: Not sure why I did that, I'll put that back in.

exzizt: Looks like that is done throughout the app in some Core files as well.

exzizt: I didn't mean to change it for that Core file so I put it back.

services/minespace-web/src/constants/strings.js
```diff
@@ -9,8 +9,11 @@ export const RETURN_PAGE_TYPE = {
   LOGOUT: "logout",
 };
 
-export const EMPTY_FIELD = "-";
+export const EMPTY_FIELD = "N/A";
```
Jsyro: CORE has more options.... do we want to change all the - to N/A, or add a new one and rename things. 

```export const EMPTY_FIELD = "N/A";
export const EMPTY = "";
export const ZERO = "0.00";
export const UNASSIGNED = "Unassigned";```

exzizt: Not sure. I thought I would update it to match what we're using in Core and I think with the new mockups the way to indicate an empty field might change. I can revert it if you want?

exzizt: Well, now that I think about it again, it doesn't make sense. And that makes me think that it doesn't make sense for where we use it in Core either. In most cases it IS an applicable field, we just don't have a value for it--that doesn't mean it's not applicable, just empty.

PR #1139, opened 2020-01-16
Feature/MDS-2559 - Conditionally render now Activities

mmumby:
Initial Implementation of conditionally rendering Activities based on NoW Type. We had gone back and forth between hiding the sections vs hiding the content. This PR hides the content. 
- Added a `<nullScreen />` to the sections that do not exist on the NoW. (see photo below)
Screen Shot 2020-01-16 at 8 36 38 AM


services/core-web/src/components/noticeOfWork/applications/review/ReviewNOWApplication.js
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable */
+
```
mmumby: disabled the entire file, the props are not being read correctly when passed into the "Field" component

PR #1142, opened 2020-01-16
Feature/mds 2500 update csv for gis team

JustinMacaulay:
- Adds columns to the view to accommodate additional fields
- removes trailing commas and columns that would contain commas
- fixes the bug where only one record would be returned for each mine.


services/core-api/app/api/exports/mines/resources/mine_summary_csv_resource.py
```diff
@@ -24,8 +24,10 @@ def get(self):
 
             csv_string = "\"" + '","'.join([c.name or "" for c in model.columns]) + "\"\n"
 
-            rows = MineSummaryCSVView.query.distinct(MineSummaryCSVView.mine_name,
-                                                     MineSummaryCSVView.permit_no).all()
+            rows = MineSummaryCSVView.query.all()
+
+            current_app.logger.debug(rows)
```
mmumby: remove logger

PR #1146, opened 2020-01-18
Feature/🌪 

nayr974:
This PR adds a whirlpool toggle to the file uploader. 

Additionally: 
- It updates the main Core logo.
- Optimizes the loading and size of some static content elements so they are not included in our bundle.
- Cleans up some linting errors.


pdittaro:
I think the toggle might be slightly too obvious - should be more of a hidden feature, but otherwise I'm all for it.

exzizt:
> I think the toggle might be slightly too obvious - should be more of a hidden feature, but otherwise I'm all for it.

Interesting idea. The toggle's off state could be styled to be completely transparent, i.e., you have to know where to click to activate it.

PR #1147, opened 2020-01-19
MDS-2589: Create MineSpace component library

exzizt:
# Notes
* The point of this PR is to set us up with **Ant Design components** whose styles have been overridden to suit the style of MineSpace. This way we can use these components to build MineSpace and have their styles automatically be correct, with having to resort to building custom components or using basic HTML elements/CSS **as little as possible**. This way we can use Ant Design as fully as possible and in the end, MineSpace will be easier to build and maintain.

* Ant Design styling is achieved using LESS, and we are using SASS, so they suggest setting up something like this (which I noticed we already have!): https://intoli.com/blog/antd-scss-theme-plugin/
Once that is set up, you use the variables found in Ant Design's _default.less_ file to override Ant Design's theme: https://github.com/ant-design/ant-design/blob/master/components/style/themes/default.less
We should do as much Ant Design overriding in **this file first** (as it usually performs broad strokes and helps maintain theme consistency)! This removes the need for style overriding in multiple individual components/scss files later on. But, Ant Design doesn't provide overrides for everything in this file, so I made a file named _antd-overrides.scss_ for us to do additional overrides on components as needed.

* Styling will evolve and change as time goes on, so not much time was spent fussing about certain things (table skeleton loader, loading bars in modals, hover effects on elements, etc).


services/minespace-web/src/components/common/RenderSelect.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
Jsyro: Why do we need this?

services/minespace-web/src/components/dashboard/mine/variances/VariancesTable.js
```diff
@@ -54,6 +56,11 @@ export class VarianceTable extends Component {
     }
   };
 
+  // NOTE: Luke/Brian: The expired "all red" styling is to be REMOVED.
```
Jsyro: is this a back and forth conversation? can this either be removed or summarized into a single direction for the next developer?

exzizt: It's meant as a note, from Luke and Brian, for the developer.

services/minespace-web/src/components/common/RenderSelect.js
exzizt: When I started making changes in this file, it made the linter starting checking it again, which caused some "no prop type defined" for a few of the props which I hadn't bothered to fix yet.

services/minespace-web/src/components/dashboard/mine/reports/ReportsTable.js
```diff
@@ -0,0 +1,150 @@
...
+  {
+    title: "Submitted By",
+    dataIndex: "submitted_by",
+    key: "submitted_by",
+    // sorter: (a, b) => (moment(a.received_date) > moment(b.received_date) ? -1 : 1),
+    render: (text, record) => (
+      <div title="Submitted By" className={record.isOverdue ? "color-error" : ""}></div>
+    ),
+  },
+  // NOTE: This is a newly requested column. Brian: "Update table to include this for both
```
nayr974: We should make sure to remove these notes during this sprint.

services/minespace-web/src/components/dashboard/mine/variances/VarianceDetails.js
```diff
@@ -22,57 +22,42 @@ const defaultProps = {
 };
 
 export const VarianceDetails = (props) => {
-  const isApproved =
-    props.variance.variance_application_status_code === Strings.VARIANCE_APPROVED_CODE;
+  // const isApproved =
```
nayr974: Is this still needed?

services/minespace-web/src/components/dashboard/mine/variances/VariancesTable.js
nayr974: IMO I think it's fine for this PR for this sprint, but these should all be removed by the end of this sprint.

services/minespace-web/src/components/dashboard/mine/variances/VariancesTable.js
```diff
@@ -25,15 +26,16 @@ const defaultProps = {
   openViewVarianceModal: () => {},
 };
 
-export class VarianceTable extends Component {
-  getOverdueStyle = (isOverdue) => (isOverdue ? { color: errorRed } : {});
+export class VariancesTable extends Component {
+  getOverdueClassName = (isOverdue) => (isOverdue ? "color-error" : "");
```
nayr974: Having this as a function seems really odd to me. Maybe define "color-error" as a local constant and put a ternary where it's consumed? No big deal, but I don't think this is done anywhere else

services/minespace-web/src/components/modalContent/config.js
```diff
@@ -3,11 +3,11 @@ import AddVarianceModal from "@/components/modalContent/variances/AddVarianceMod
 import ViewVarianceModal from "@/components/modalContent/variances/ViewVarianceModal";
 import EditVarianceModal from "@/components/modalContent/variances/EditVarianceModal";
 
-// This file is anticipated to have multiple exports
-// eslint-disable-next-line import/prefer-default-export
 export const modalConfig = {
   ADD_REPORT: AddReportModal,
-  ADD_VARIANCE: AddVarianceModal,
+  CREATE_VARIANCE: AddVarianceModal,
```
nayr974: We're inconsistent on nomenclature here. 
  **ADD**_REPORT: **Add**ReportModal,
  **CREATE**_VARIANCE: **Add**VarianceModal,

services/minespace-web/src/constants/strings.js
```diff
@@ -10,10 +10,11 @@ export const RETURN_PAGE_TYPE = {
 };
 
 export const EMPTY_FIELD = "-";
+export const NOT_APPLICABLE = "N/A";
```
nayr974: yay

services/minespace-web/src/components/dashboard/mine/variances/VariancesTable.js
exzizt: That was my intention :)

services/minespace-web/src/components/dashboard/mine/reports/ReportsTable.js
exzizt: That's what I was thinking too :)

services/minespace-web/src/components/modalContent/config.js
exzizt: This was a mishap on my part for forgetting to rename the other part too. I like to stick with CRUD action verbs for things like this. "Add" reads weird in my brain for "Create" actions. For example, do we "Subtract" something when we "Delete" it? Just a minor, very, very, important detail, obviously ;)

services/minespace-web/src/components/dashboard/mine/variances/VarianceDetails.js
exzizt: I'm thinking that it might be, come this Sprint. I had to comment it out to get the linter to not complain about an unused variable at the time.

services/minespace-web/src/components/dashboard/mine/overview/Overview.js
```diff
@@ -0,0 +1,152 @@
...
+      <Text strong style={{ textTransform: "uppercase" }}>
+        {props.label}
+      </Text>
+      <br />
+      <Text>{props.value}</Text>
+    </Col>
+  </Row>
+);
+
+const ContactCard = (props) => (
```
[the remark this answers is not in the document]
exzizt: I forgot to do that when I did it for the TableSummaryCard component. I'll do that in another PR or in this one tomorrow morning! Can't merge it in because it's failing to build.

exzizt:
Closing and reopening under a new PR to try and bypass build failure on pipeline

PR #1149, opened 2020-01-21
Feature/Fix Whirlpool and revert API retrys

nayr974:
This PR fixes the whirlpool and adds a binary file type definition for .gif files. It also reverts the API retry logic with a hunch that it might be interfering with the metabase fetch.


services/core-web/src/customAxios.js
```diff
@@ -3,30 +3,20 @@ import { notification } from "antd";
 import { isEmpty } from "lodash";
```
mmumby: Was this change intentional? All the deletions to the customAxios?

nayr974: Yes, I just updated the description to explain why I updated this file as well :)

PR #1155, opened 2020-01-27
Feature/mds 2561 major mine permit application

Jsyro:
Major Mines now see 'Notice of Work Applications' as 'Permit Application'
Major mines see an 'Add Permit Application' button, which take them to the Permit Application Initialization stage, 
Permits on Major mines have an action to create a permit amendment application, which takes them to the same form as above, with the permit pre-selected. 
After creation, users need to select a lead inspector. 
Stages 2-4 look identical to regional NoW applications for now. 

Now Application components were restructured, could use some renames as well.


services/core-api/app/api/mines/mine/models/mine.py
```diff
@@ -134,29 +134,36 @@ def find_by_mine_no(cls, _id):
...
             mines_q = Mine.query.filter(name_filter).filter_by(deleted_ind=False)
-            mines = mines_q.limit(MINE_LIST_RESULT_LIMIT).all()
         else:
-            mines = Mine.query.limit(MINE_LIST_RESULT_LIMIT).all()
-        return mines
+            mines_q = Mine.query
+
+        if major is not None:
+            mines_q = mines_q.filter_by(major_mine_ind=major)
+            current_app.logger.debug(f'FILTER BY MAJOR = {major}')
```
pdittaro: I don't mind having debug statements, but I think this one probably isn't too necessary.

services/core-api/app/api/mines/mine/models/mine.py
```diff
@@ -134,29 +134,36 @@ def find_by_mine_no(cls, _id):
...
+    def find_by_name_no_permit(cls, term=None, major=None):
         MINE_LIST_RESULT_LIMIT = 50
         if term:
             name_filter = Mine.mine_name.ilike('%{}%'.format(term))
             number_filter = Mine.mine_no.ilike('%{}%'.format(term))
             permit_filter = Permit.permit_no.ilike('%{}%'.format(term))
             mines_q = Mine.query.filter(name_filter | number_filter).filter_by(deleted_ind=False)
             permit_q = Mine.query.join(Permit).filter(permit_filter)
-            mines = mines_q.union(permit_q).limit(MINE_LIST_RESULT_LIMIT).all()
+            mines_q = mines_q.union(permit_q)
```
pdittaro: If you put the `permit_q` declaration in front of the initial `mines_q` declaration you can assign this in one line.

services/core-api/app/api/mines/mine/models/mine.py
```diff
@@ -134,29 +134,36 @@ def find_by_mine_no(cls, _id):
...
         else:
-            mines = Mine.query.limit(MINE_LIST_RESULT_LIMIT).all()
-        return mines
+            mines_q = Mine.query
+
+        if major is not None:
```
pdittaro: Should this be a `not None` rather than true? I'd rather treat the parameter as a boolean on the function.

services/core-api/tests/now_application_factories.py
```diff
@@ -363,7 +363,7 @@ class Params:
 
     now_application_id = factory.SelfAttribute('now_application.now_application_id')
     #application_progress_id = factory.Sequence(lambda n: n)
-    application_progress_status_code = factory.LazyFunction(RandomNOWProgressStatusCode)
+    application_progress_status_code = 'REV'
```
pdittaro: Is this a limitation of the status code workflow checks that we've put in?

services/core-api/app/api/mines/mine/models/mine.py
Jsyro: if major is undefined, ignore it ,but if it if it's true, or false, filter by that defined value.

services/core-api/tests/now_application_factories.py
Jsyro: there should only be one of each, and they should be ordered. so the first one is REV.. I think we should look at better ways of sequencing this.

services/core-web/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -158,7 +159,26 @@ const columns = [
               />
               Edit permit status
             </button>
-          </Menu.Item>
+          </Menu.Item>{" "}
+          <AuthorizationWrapper isMajorMine={text.major_mine_ind}>
+            <div className="custom-menu-item" key="3">
```
mmumby: Should this also have the "permit_create" permissions?

services/core-web/src/components/mine/NoticeOfWork/MineNOWApplications.js
```diff
@@ -102,13 +104,28 @@ export class MineNOWApplications extends Component {
...
   render() {
+    const isMajorMine = this.props.mines[this.props.mineGuid].major_mine_ind;
+    const title = isMajorMine ? "Permit Applications" : "Notice of Work Applications";
     return (
       <div className="tab__content">
         <div>
-          <h2>Notice of Work Applications</h2>
-          <Divider />
+          <h2>{title}</h2>
+          {isMajorMine && (
```
mmumby: update to use AuthWrapper and Should this also have the "permit_create" permissions?

services/core-api/app/api/mines/mine/models/mine.py
pdittaro: Right, okay.

services/core-web/src/components/mine/Permit/MinePermitTable.js
Jsyro: I've actually removed it as it was unnecessary as the parent as the same auth requirements.

services/core-web/src/components/mine/NoticeOfWork/MineNOWApplications.js
Jsyro: yup, done.

services/core-api/app/api/mines/mine/models/mine.py
Jsyro: removed.

Jsyro:
Review Reminders:
- [ ] test_expected_auth tests (if new endpoints)
- [ ] snapshot test (if new react components)

services/core-api/tests/now_application_factories.py
nayr974: I agree with setting it to rev. Having it be random caused mismatched state permutations that adding more complex logic to define doesn't seem useful.

PR #1156, opened 2020-01-28
Reconciling PVC Pipeline Creation

pdittaro:
The migration of all relevant PVCs has been completed in the Test and Prod namespaces. An exception to this are the nginx logs which for some reason had errors during the test migration. I decided to just start from scratch on those with a new pvc in the format of 'mds-access-logs-prod'.

So basically, this PR should allow us to do the full pipeline deployment again.


openshift/pipeline/config-prod.groovy
```diff
@@ -332,7 +332,7 @@ environments {
     'prod' {
         vars {
             DB_PVC_SIZE = '50Gi'
-            DOCUMENT_PVC_SIZE = '390Gi'
+            DOCUMENT_PVC_SIZE = '80Gi'
```
nayr974: Is this based on a legitimate value of projected use until we switch over to s3?

pdittaro: No, but one nice thing is that the new netapp pvcs can scale to larger sizes if we need to. This size was created because of the quota limit with the two PVCs existing in parallel. 

At present we're only using 15gb of document storage.

PR #1161, opened 2020-01-29
Feature/mds 2587 create reporting db

nayr974:
This PR adds a new database pod for reporting purposes, as well as a new digdag job to restore a database backup to this databbase.


openshift/templates/dbbackup.dc.json
```diff
@@ -58,7 +76,23 @@
...
+                  "bash",
+                  "-c",
+                  "$FLYWAY_HOME/scripts/run_migrations.sh"
+                ]
+              }
+            },
+            "maxSurge": "50%",
+            "maxUnavailable": 0,
+            "timeoutSeconds": 1200
+          }
```
Jsyro: the dbbackup pod is long-living, meaning it has no need to redeploy on sql changes, not sure this will keep the datastore db up to date without intervention.

openshift/templates/dbbackup.dc.json
```diff
@@ -74,6 +108,17 @@
...
+          {
+            "type": "ImageChange",
+            "imageChangeParams": {
+              "automatic": true,
+              "containerNames": ["flyway-migration"],
+              "from": {
+                "kind": "ImageStreamTag",
+                "name": "${FLYWAY_NAME}:${VERSION}"
+              }
+            }
```
Jsyro: This will mean any SQL change will cause dbbackup to redeploy. 

@pdittaro What if we merged in a PR while this pod was in the middle of backing up and shipping date to the store? Something noteworthy. can we disable PR's from midnight-1am?

openshift/templates/dbbackup.dc.json
nayr974: (discussed offline) :)

PR #1169, opened 2020-01-31
Upgrade/mds 2400

mmumby:
- Updated Verification Step to include lat/lon validations
- Added dropdown in actions to update lat/lon after verification is complete
- Update map component to handle an additionalPin


services/core-web/src/components/common/wrappers/AddPartyComponentWrapper.js
```diff
@@ -99,7 +99,7 @@ export class AddPartyComponentWrapper extends Component {
           <Radio.Group
             defaultValue
             size="large"
-            value={this.props.isPerson}
+            value={this.state.isPerson}
             onChange={this.togglePartyChange}
```
mmumby: During the merge conflict, this came up as a linting error, "this.props.isPerson"

PR #1178, opened 2020-02-05
MDS-2607: View Permits on MineSpace (Major Mines Only)

exzizt:
# Main
* Major mines can now view their permit information on MineSpace via the Permits tab
# Other
* Updates Core home page styling/structure and link card design


services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.js
```diff
@@ -1,33 +1,157 @@
...
-  <Table
-    size="small"
-    pagination={false}
-    loading={!props.isLoaded}
-    columns={columns}
-    expandedRowRender={(record) => <p>{record.description}</p>}
-    dataSource={data}
-    locale={{ emptyText: "This mine has no permit data." }}
-  />
+const renderDocumentLink = (file, text) => (
```
nayr974: Does this being a function provide value and it seems like this is only used once? Can't expandedColumn include a <LinkButton directly?

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.js
```diff
@@ -1,33 +1,157 @@
...
+          <Text>
+            {renderDocumentLink(file, file.document_name)}
+            <br />
+          </Text>
+        ))) ||
+      Strings.NONE,
+  },
+];
+
+const transformRowData = (permit, permitStatusOptions) => {
```
nayr974: Don't refactor for the time being, but going forward we should avoid transform functions for tables as briefly discussed.

PR #1179, opened 2020-02-06
MDS-2516 Incidents in MineSpace

nayr974:
This PR adds mine Incidents to Minespace, and fixes a few Incident related bugs and content.


services/core-web/src/components/dashboard/customHomePage/VarianceTable.js
```diff
@@ -38,7 +38,7 @@ const defaultProps = {
 };
 
 export const VarianceTable = (props) => (
-  <div className="tab__content">
+  <React.Fragment>
     <MineVarianceTable
```
mmumby: Since removing the class, Does this still have the same margins as the other table views?

nayr974: Yes - the outer component is already adding this class so it was wrapped in two tab__contents. Now it is the same as the NoW "HomePage".

mmumby: Oh perfect.

PR #1182, opened 2020-02-07
MDS-2617: MineSpace Documents Table Component

exzizt:
# Main
* This ticket has been deferred, so please just see **Other**
# Other
* Updates SearchBar styling
* Updates NavBar styling
* Adds a class we can use to cap the height of columns (see its usage for how to use, particularly for columns containing lists of documents/files), and applied it in various places
* Use "truncate filename" function in more places, and apply "title" attribute so the full filename can be seen
* Modifies the LinkButton component to pass down props


services/core-web/src/components/search/SearchBarDropdown.js
```diff
@@ -12,78 +12,86 @@ const propTypes = {
...
-  ({
-    mine: route.MINE_GENERAL.dynamicRoute(item.result.id),
-    party: route.PARTY_PROFILE.dynamicRoute(item.result.id),
-    permit: route.MINE_PERMITS.dynamicRoute(item.result.id),
-  }[item.type]);
+  const URLFor = (item) =>
+    ({
+      mine: route.MINE_GENERAL.dynamicRoute(item.result.id),
+      party: route.PARTY_PROFILE.dynamicRoute(item.result.id),
+      permit: route.MINE_PERMITS.dynamicRoute(item.result.id),
```
exzizt: Do we actually search through permits? If so, I'll make a section for those to go with Mines and Contacts.

PR #1191, opened 2020-02-10
Task/crr dates table

Jsyro:
Sorting dates treats null like a date very far in the future. Ok'ed by Brian. 

I think it would be cool to have a priority sort (show me the oldest due date without any documents) and possibly force fufilied reports and/or, reports without due dates, to the bottom, but that's for another time. 

This is consistent and reasonable. 

e.g.
(ascending)
Feb 1,
Feb 20,
<no date>

(descending)
<no date>,
Feb 20,
Feb 1


services/minespace-web/src/components/dashboard/mine/reports/ReportsTable.js
```diff
@@ -46,7 +46,11 @@ const columns = [
     dataIndex: "due_date",
     key: "due_date",
     defaultSortOrder: "ascend",
-    sorter: (a, b) => (moment(a.due_date) > moment(b.due_date) ? -1 : 1),
+    sorter: (a, b) => {
+      const a_date = a.due_date == null ? moment().add(200, "y") : moment(a.due_date);
+      const b_date = b.due_date == null ? moment().add(200, "y") : moment(b.due_date);
+      return a_date - b_date;
```
mmumby: Should we update all date sorters to handle null?

Jsyro: I turned this into a util function that can be applied wherever we need.

PR #1197, opened 2020-02-12
MDS-2538: Permit Application Letters - Template Downloads (docx)

exzizt:
# Main
* Inserts three NoW document type codes to the table `now_application_document_type`.
* Adds a submenu under the NoW Action dropdown menu for generating documents. Clicking on an option downloads its respective DOCX file.
* Adds backend code to house "document generation" endpoints, and implements the NoW document generation endpoint (whose temporary solution is to serve the DOCX file associated with the received document code). Downloading the document templates requires a download token.
# Other
* Re-structured components on the NoW Action dropdown menu.
* Fixes some spelling/grammar/inconsistency issues with various code tables.


services/core-web/src/actionCreators/noticeOfWorkActionCreator.js
```diff
@@ -0,0 +1,48 @@
...
+      createRequestHeader()
+    )
+    .then((response) => {
+      const token = { token: response.data.token_guid };
+      window.open(
+        `${ENVIRONMENT.apiUrl + API.NOTICE_OF_WORK_APPLICATION_DOCUMENT(documentTypeCode, token)}`,
+        "_blank"
+      );
+    });
+};
```
exzizt: This isn't actually an "action creator" but I placed it in this file to follow used patterns.

services/core-web/src/actionCreators/noticeOfWorkActionCreator.js
```diff
@@ -0,0 +1,48 @@
...
+
+export const generateNoticeOfWorkApplicationDocument = (documentTypeCode, payload) => (
```
exzizt: Current functionality doesn't use this yet but eventually, it will.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -289,8 +296,7 @@ export class NoticeOfWorkApplication extends Component {
       .then(() => finalAction());
   };
 
-  openUpdateLeadInspectorModal = (event) => {
-    event.preventDefault();
+  openUpdateLeadInspectorModal = () => {
```
exzizt: I restructured things so this isn't needed here anymore and for similar use-cases.

exzizt:
I wasn't entirely sure what names/folder structures to use so please let me know if those need changes.

services/core-api/app/api/constants.py
```diff
@@ -67,6 +67,12 @@
     }
 }
 
+NOW_DOCUMENT_TEMPLATE_TYPES = {
```
nayr974: In the end we will probably want to move this to the database, along with a JSON definition of the input form. We have an interesting problem later because we will have a type, form template, and file. We could end up storing the template file in the database as well, but we'll talk about that later.

exzizt: Definitely. I figured this was just temporary.

services/core-api/app/api/document_generation/resources/now_document.py
```diff
@@ -0,0 +1,79 @@
...
+        description='Generates the specified document for the NoW using the provided template data.',
+        params={'document_type_code': 'The code indicating the type of document to generate.'})
+    @requires_role_edit_permit
+    def post(self, document_type_code):
+        data = self.parser.parse_args()
+
+        if document_type_code not in NOW_DOCUMENT_TEMPLATE_TYPES:
+            raise NotFound('Document type code not found')
+
+        raise NotImplemented('This endpoint is under construction')
```
Jsyro: This is the endpoint that should return a token to the file.

services/core-api/app/api/document_generation/resources/now_document.py
```diff
@@ -0,0 +1,79 @@
...
+    def get(self, document_type_code):
+        if document_type_code not in NOW_DOCUMENT_TEMPLATE_TYPES:
+            raise NotFound('Document type code not found')
+
+        token_guid = uuid.uuid4()
+        cache.set(
+            NOW_DOCUMENT_DOWNLOAD_TOKEN(token_guid), {'document_type_code': document_type_code},
+            TIMEOUT_5_MINUTES)
+
+        return {'token_guid': token_guid}
```
Jsyro: The endpoint that issues the token is the 'generate file' endpoint  I think we only need two.

services/core-api/app/api/constants.py
```diff
@@ -67,6 +67,12 @@
...
+    'CAL': 'Client Acknowledgment Letter Template (NoW).docx',
+    'RJL': 'Rejection Letter Template (NoW).docx',
+    'WDL': 'Withdrawl Letter Template (NoW).docx'
+}
+
```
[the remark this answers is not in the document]
exzizt: Definitely!

services/core-api/app/api/document_generation/namespace.py
```diff
@@ -0,0 +1,8 @@
+from flask_restplus import Namespace
+from app.api.document_generation.resources.now_document import NoticeOfWorkDocumentResource, NoticeOfWorkDocumentTokenResource, NoticeOfWorkDocumentGenerationResource
+
+api = Namespace('now-document-types', description='Generate Notice of Work documents')
+
+api.add_resource(NoticeOfWorkDocumentResource, '/<string:document_type_code>')
+api.add_resource(NoticeOfWorkDocumentGenerationResource, '/<string:document_type_code>/generate')
+api.add_resource(NoticeOfWorkDocumentTokenResource, '/<string:document_type_code>/token')
```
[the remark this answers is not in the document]
Jsyro: I'm not sure we can do this properly right now, it might have to hard code the context.

services/core-api/app/api/document_generation/resources/now_document.py
exzizt: I don't think we should modify this POST endpoint's functionality to behave in a way that will be entirely different from how it will behave in the future. I think for the temporary functionality we have now of downloading a docx we need a temporary endpoint, which is what I've created as the GET endpoint.

services/core-api/app/api/document_generation/resources/now_document.py
```diff
@@ -0,0 +1,79 @@
...
+        data = self.parser.parse_args()
+
+        if document_type_code not in NOW_DOCUMENT_TEMPLATE_TYPES:
+            raise NotFound('Document type code not found')
+
+        raise NotImplemented('This endpoint is under construction')
+
+
+class NoticeOfWorkDocumentResource(Resource, UserMixin):
+    @api.doc(
```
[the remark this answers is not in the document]
exzizt: That is what we want for now, that's the purpose of the ticket. It's a temporary solution. See my other comment, too.

services/core-api/app/api/document_generation/resources/now_document.py
Jsyro: I think that is the POST's behaviour, 
REQUEST: Create me a document 
RESPONSE: Here is a token to get the document you asked for.

exzizt:
> Some changes, the URLs and design can be set up for the final version. Let me know if you have any questions.

I agree with your comments mostly but I think the changes can be done as work as part of the next step. I think this implementation satisfies the ticket and even more potential changes will probably be realized as we get closer to the next step and figure out how to generate the PDFs etc.

services/core-api/app/api/document_generation/resources/now_document.py
exzizt: I see what you mean now. I guess I could pass over the NoW application GUID and an empty template_data object even though it's not being used at the moment.

Jsyro:
> > Some changes, the URLs and design can be set up for the final version. Let me know if you have any questions.
> 
> I agree with your comments mostly but I think the changes can be done as work as part of the next step. I think this implementation satisfies the ticket and even more potential changes will probably be realized as we get closer to the next step and figure out how to generate the PDFs etc. If you think it's blocking then I can make some of the changes next week.

It's not blocking per se, but the tickets were divided in this way so that the frontend considerations could be done ahead of time and not redone later, this is kinda putting us in a spot where we do have to redo them later, so I'd rather see this done now then have to context switch and remember next week the changes we want to do.

exzizt:
> > > Some changes, the URLs and design can be set up for the final version. Let me know if you have any questions.
> > 
> > 
> > I agree with your comments mostly but I think the changes can be done as work as part of the next step. I think this implementation satisfies the ticket and even more potential changes will probably be realized as we get closer to the next step and figure out how to generate the PDFs etc. If you think it's blocking then I can make some of the changes next week.
> 
> It's not blocking per se, but the tickets were divided in this way so that the frontend considerations could be done ahead of time and not redone later, this is kinda putting us in a spot where we do have to redo them later, so I'd rather see this done now then have to context switch and remember next week the changes we want to do.

Sounds good! I'll put the PR back to WIP.

services/core-api/app/api/now_applications/transmogrify_now.py
```diff
@@ -10,9 +10,9 @@
```
exzizt: All auto-formatting below. I just changed these "ton" values to match the fixed code values in the DB.

services/core-web/src/actionCreators/noticeOfWorkActionCreator.js
exzizt: The front-end uses this now.

services/core-api/app/api/document_generation/namespace.py
```diff
@@ -0,0 +1,7 @@
+from flask_restplus import Namespace
+from app.api.document_generation.resources.now_document import NoticeOfWorkDocumentResource, NoticeOfWorkDocumentGenerationResource
+
+api = Namespace('document-generation', description='Generate documents for business objects.')
+
+api.add_resource(NoticeOfWorkDocumentResource, '/now')
+api.add_resource(NoticeOfWorkDocumentGenerationResource, '/now/<string:document_type_code>')
```
Jsyro: Sorry i didn't notice this on the first time. I think we want generate documents to be an action on that document type. 

`/api/now-applications/application-document-types/<code>/generate`

or something like that. Let me know if that's not trivial.

Jsyro: We will tackle this when the templates go into the db.

exzizt:
> Thanks for making those changes. Great work.

Thanks!

PR #1202, opened 2020-02-13
MDS-2584 new table component that includes the loading skeleton.

JustinMacaulay:
Changes the table loading wrapper into a CoreTable component that either loads the skeleton table or takes the props passed in and spreads them over the standard antd table.


services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkTable.js
```diff
@@ -272,24 +271,22 @@ export class NoticeOfWorkTable extends Component {
...
-      </TableLoadingWrapper>
+        columns={applySortIndicator(
+          this.columns(this.props),
+          this.props.sortField,
+          this.props.sortDir
+        )}
+        dataSource={this.transformRowData(this.props.noticeOfWorkApplications)}
+        tableProps={{
+          rowClassName: "fade-in",
+          align: "left",
```
mmumby: This can be a default class on CoreTable as it goes with the transition of the loading skeleton and every table should inherit

PR #1206, opened 2020-02-19
stop nuking mds_test everyrun.

Jsyro:
Don't reconstruct mds_test db everytime. Let's see if that's necessary.


migrations/sql/afterMigrate__97__create_test_db.sql
```diff
@@ -5,12 +5,6 @@ BEGIN
   CREATE EXTENSION IF NOT EXISTS dblink;
   IF (EXISTS (SELECT 1 FROM pg_database WHERE datname = 'mds_test')) THEN
     RAISE NOTICE 'Database already exists';
-    IF (EXISTS (SELECT * FROM current_catalog WHERE current_database='mds')) THEN
```
pdittaro: The only thing I can think of would be that we should set the owner of mds_test to be mds, but I really don't think its a big deal. Lets see if this works.

PR #1209, opened 2020-02-21
Test Instance created for NROS/vFCBC integration

nayr974:


Jsyro:
This instance's swagger is being used as reference for other systems.

PR #1216, opened 2020-02-24
fixed some auth tests after a 403 return was added

JustinMacaulay:
Fixes some unit tests to allow for a 403 return instead of a 401.


exzizt:
Thanks again for doing this, Justin. It'd be nice to figure out why these tests aren't triggering for my dev environment.

PR #1217, opened 2020-02-25
MDS-2351: SPIKE - Filtering Mine Pins on Map

exzizt:
While this was just a SPIKE and focussed on research I did make a few changes to the code:

* Reduced the structure/size of the JSON object used to draw the mine pins on the map (when a pin is selected, additional information is properly retrieved and displayed)
* Reduced a prop/index-out-of-range error on the map page which prevented the page from loading
* Markers that are clicked/selected get the "selected" icon and have it unapplied when a new marker is selected
* Fixed a contact information issue for mines in MineSpace

**Note:** This PR doesn't fix the "search by name" bug, that existed before work was done in this branch.


services/core-web/src/components/maps/MineMapLeaflet.js
```diff
@@ -1,3 +1,6 @@
+// TODO: Remove this line when file is more properly implemented.
```
nayr974: What do you mean "more properly implemented"? We shouldn't have introduced a bunch of linting issues with your changes. If there's one or two issues, disable the rule on a per line basis.

exzizt: There were a couple issues disabled per line and I figured they could be addressed in the upcoming future work. I can put them back in.

exzizt: Addressed comments in person. Thanks Ryan

PR #1227, opened 2020-03-02
MDS-2654: Update "submissions_only" flag logic/NoW View

exzizt:
# Main
* Fixes a bug where NoWs created in Core were not retrieved when "submissions_only" query param was set to true
* Major Mines Permit Applications page now has "submissions_only" query param set to true

Demonstration of working changes (note: the code has changed):
image

# Other
* Fixes an issue with the AutoComplete component where moving through the options with the arrow keys would place the value in the field


exzizt:
> In a sea of code
> this PR is only a drop.
> The water is warm.

Lovely

PR #1230, opened 2020-03-03
pod specs for docgen

Jsyro:


pdittaro:
The team is agreeing to force merge this PR because it is only affecting pod quotas.

PR #1238, opened 2020-03-05
fix/ mine_report_definition-duplicate fix. 

Jsyro:


migrations/sql/V2020.03.05.12.47__mine_report_definition_unique_report_name_clean_up.sql
```diff
@@ -0,0 +1,2 @@
+--One time FIX from seed data bug inserting duplicate records. is fixed now.
+DELETE FROM mine_report_definition where mine_report_definition_id > 61;
```
exzizt: Would it be safer to delete "where not unique/distinct?" than "id > 61"? Unless you're certain that this case covers removing all the duplicated values.

migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -620,88 +620,10 @@ VALUES
...
-(report_name, description, due_date_period_months, mine_report_due_date_type, active_ind, create_user, create_timestamp, update_user, update_timestamp)
-select report_name, '', due_date_period, due_date_type, 'true', 'system-mds', now(), 'system-mds', now() from tmp_report_definition_compliance;
--- hide appeal to CIM until notifications are built into CORE/MineSpace
-UPDATE mine_report_definition SET active_ind='false' where report_name='Appeal to CIM Report';
-
-
+-- Mine Report Definitions have been managed by sequential migrations. 
+-- V2019.07.05.15.01
+-- V2019.07.09.16.01
+-- V2019.09.28.14.16
```
exzizt: These files didn't need to be changed?

PR #1241, opened 2020-03-09
MDS-2638: Provincial CRR Dashboard in CORE!

exzizt:
# Main
* Adds a CRR dashboard into Core for provincial analysis. The user is able to filter and sort the results by various data points and columns. Adds/modifies various action creators/selectors/reducers and models/resources to achieve this.
Filters on Provincial Reports Dashboard:
image
Table on Provincial Reports Dashboard (Note: columns are not set in stone yet. Showing the type might be too much, could show compliance year instead, etc.):
image
# Other
* Identified a solution to fix the issue where select inputs would not display their placeholder values, multiple-select inputs would populate an "empty" value, tabbing through date inputs would set the value to "empty string" value, etc. The solution is to provide a "format={null}" prop to Field components. See more details [here](https://redux-form.com/8.2.2/docs/api/field.md/) on the docs for Redux Form.
* Fixes various issues with and adds additional sorting and filtering to the mine-specific CRR dashboard (e.g., received date, requested by, submission status), and also allows for sorting on the mine-specific tailings report view, despite there being no filtering there.
* Removes a duplicate record from the mine_report_category_xref table.
* Fixes an issue in a few resource files where BadRequest was not imported (but it was being used).
* Adds a resource to neatly get all of the mine report category codes. Before, this was being done in a bit of a janky way by using the mine report definition resource, then processing that result to extract the code values.
* Creates a function to compare nullable string values in our tables that use front-end sorting.
* Modifies the function to compare nullable date values in our tables that use front-end sorting.
* Adds additional properties to the mine report model and tidied its related code (e.g., removed a duplicate property, set/unset nullable parameter in column definitions). Some additional properties are hybrid and are required for sorting/filtering in the provincial reports dashboard.
* Fixes various issues with the contact page and tidies its code, e.g., where the "contact type" select input was not populated to the default "Person" value, and the render() event was triggering more than necessary.
* Adjusts various text throughout the app (e.g., placeholders, labels) to try and make them more consistent with their sibling fields. We should come up with a convention to be used throughout the app.
* Tidies/slightly modifies the view for the "add report" modal/form (e.g., placeholders, only showing report code requirements field once category and definition are selected, allowing for initial values to be populated correctly).
* Fixes a prop type issue with Core's authorization wrapper component.
* Edits the "incidents dashboard" route to be more consistent with the other dashboard routes (i.e., it was missing the "/reporting/" that the others have).
* Removes the table-level pagination from the provincial variances table:
image
* Fix issue where CoreTable didn't pass down other necessary props required to render the skeleton table correctly
* Other minor changes to text throughout the app.


services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
+    report_name = association_proxy('mine_report_definition', 'report_name')
+    mine_report_category_xref = db.relationship(
+        'MineReportCategoryXref',
+        lazy='joined',
+        primaryjoin=
+        'MineReport.mine_report_definition_id==MineReportCategoryXref.mine_report_definition_id',
+        foreign_keys=mine_report_definition_id,
+        uselist=True,
+        order_by='MineReportCategoryXref.mine_report_category')
+    mine_report_category = association_proxy('mine_report_category_xref', 'mine_report_category')
```
Jsyro: mine_report_categories should be a relationship using `secondary='mine_report_category_xref', don't need to traverse the xref table.

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
+
+    mine_guid = db.Column(UUID(as_uuid=True), db.ForeignKey('mine.mine_guid'), nullable=False)
+    mine_table = db.relationship('Mine', lazy='joined')
```
Jsyro: this is relating to a specific mine, not the table, can you rename this to `mine = `

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
 
-    mine_report_definition_guid = association_proxy('mine_report_definition',
-                                                    'mine_report_definition_guid')
-    report_name = association_proxy('mine_report_definition', 'report_name')
-    report_name = association_proxy('mine_report_definition', 'report_name')
-    permit_guid = association_proxy('permit', 'permit_guid')
+    @hybrid_property
+    def mine_report_submission_status_code(self):
+        if self.mine_report_submissions and len(self.mine_report_submissions) > 0:
+            return self.mine_report_submissions[len(self.mine_report_submissions) -
```
Jsyro: use array slicing to access last element of the array, `self.mine_report_submissions[-1].mine_report_submission_status_code` i'm wondering if there is a way to define a relationship so that it's eagerly loaded.

Also, you are using the status of the submission with the highest id to represent the status of the report. can you rename the attribute to match -> `mine_report_status_code`

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
+    @mine_report_submission_status_code.expression
+    def mine_report_submission_status_code(cls):
+        return select([
+            MineReportSubmission.mine_report_submission_status_code
+        ]).where(MineReportSubmission.mine_report_id == cls.mine_report_id).order_by(
+            desc(MineReportSubmission.mine_report_submission_id)).limit(1).as_scalar()
+
+    @hybrid_property
+    def mine_report_submission_status_code_description(self):
+        if self.mine_report_submissions and len(self.mine_report_submissions) > 0:
```
Jsyro: lists of length 0 are falsy in python, don't need a length check, just a truth check.

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
+            return None
+
+    @mine_report_submission_status_code_description.expression
+    def mine_report_submission_status_code_description(cls):
+        return select([MineReportSubmissionStatusCode.description]).where(
+            and_(
+                MineReportSubmission.mine_report_id == cls.mine_report_id,
+                MineReportSubmissionStatusCode.mine_report_submission_status_code ==
+                MineReportSubmission.mine_report_submission_status_code)).order_by(
+                    desc(MineReportSubmission.mine_report_submission_id)).limit(1).as_scalar()
```
Jsyro: same naming concern. Please look into defining a relationship for the most recent mine_report_submission, that might be better performance and easier to follow than this set of functions.

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,43 +1,97 @@
...
+    mine_region = association_proxy('mine_table', 'mine_region')
+    major_mine_ind = association_proxy('mine_table', 'major_mine_ind')
+
     permit_id = db.Column(db.Integer, db.ForeignKey('permit.permit_id'))
     permit = db.relationship('Permit', lazy='selectin')
+    permit_guid = association_proxy('permit', 'permit_guid')
+
     received_date = db.Column(db.DateTime)
-    due_date = db.Column(db.DateTime, nullable=False)
+    due_date = db.Column(db.DateTime)
```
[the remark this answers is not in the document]
exzizt: I removed it because the table in the database doesn't have a not-null constraint.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: I tried to do that and I couldn't find a way to create a property that just gets the last element, so I opted to create a hybrid property instead.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: It doesn't really need to be eagerly loaded since this is only used in filtering/sorting?

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: Sure, I just copied the naming convention from another file(s). Should I rename it there as well?

services/core-api/app/api/incidents/resources/incidents_resource.py
```diff
@@ -3,6 +3,7 @@
 from datetime import datetime
 from sqlalchemy import desc, cast, NUMERIC, extract, asc
 from sqlalchemy_filters import apply_sort, apply_pagination, apply_filters
+from werkzeug.exceptions import BadRequest
```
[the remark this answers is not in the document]
exzizt: Yeah, BadRequest was being used in a couple of files but never imported.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: Will talk to you in person.

services/core-api/app/api/variances/resources/variance_resource.py
```diff
@@ -2,6 +2,7 @@
 from flask_restplus import Resource
 from sqlalchemy_filters import apply_sort, apply_pagination, apply_filters
 from sqlalchemy import desc, cast, NUMERIC, func, or_
+from werkzeug.exceptions import BadRequest
```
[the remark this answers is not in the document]
exzizt: It was being referenced in the file but it was never imported.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: I will talk to you in person.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: It doesn't really need to be eagerly loaded since its only being used for sorting/filtering?

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: Changing it to "mine" throws the following error:
```Error creating backref 'mine' on relationship 'Mine.mine_reports': property of that name exists on mapper 'mapped class MineReport->mine_report'```

I guess this is why those other files called it "mine_table" as well.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: Resolved in person.

services/core-api/app/api/mines/reports/models/mine_report.py
exzizt: Resolved in person!

services/core-api/app/api/mines/reports/models/mine_report.py
Jsyro: Maybe a comment on why these exists, After explanation i understand. We a sqlalchemy to sort/filter on these members, but they don't exist in the db, so we have these expressions to tell the ORM how to execute operations on these members.

I have no idea around the performance implications.

PR #1242, opened 2020-03-10
Spike/mms-data-validation

Jsyro:


Jsyro:
This also revealed that is test and prod, no records will exist that only have MMS data, they may be in mms but are not being loaded into our mms_now_submissions schema.

PR #1244, opened 2020-03-13
MDS-2526: Provide NoW Securities Info

exzizt:
# Main
* There is a Securities section in the NoW Technical Review page between the Application Info and Contacts section. This Securities section has a dollar input field for the security total and a date field for the security received date. Validation is done on the dollar field to ensure it is a dollar amount. These fields are only editable in edit mode. There is a new NoW document type for “Security Calculation Documents” so users can attach these types of documents under the Additional Documents section.
image
image
* Adds securities fields to permit amendment model
# Other
* Fixed a bug with the "deep update dictionary" function in the API where, for example, setting a date or decimal field in the NoW Technical Review section back to an empty value would cause an error and not appropriately reset the value/set it to null.
* Creates tests for testing new logic in the "deep update dictionary" function.
* Adds a package to Core/MineSpace for Redux Forms to create easy input masks on Field components (which I used on the security total field).
* Fixes incorrect casting of a string value (tenure_number) in our NoW Application factory.
* Fixed some typos in code.


migrations/sql/V2020.03.12.11.35__add_ddl_for_securities.sql
```diff
@@ -0,0 +1,13 @@
+
+-- Add a new NoW document type for the security calculation document
+INSERT INTO now_application_document_type
+(now_application_document_type_code, description, active_ind, create_user, update_user)
+VALUES ('SCD', 'Security Calculation Document', true, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
```
Jsyro: add this to seed data as well.

exzizt: Done

services/core-api/app/api/utils/models_mixins.py
```diff
@@ -198,32 +206,45 @@ def _deep_update_from_dict(self, data_dict, depth=0, _edit_key=None):
...
                     if py_type == decimal.Decimal:
-                        #if Decimal column, cast whatever you get to Decimal
-                        dec = decimal.Decimal(v)
-                        #don't care about anything more precise, procection if incoming data is float
-                        setattr(self, k, dec.quantize(decimal.Decimal('.0000001')))
+                        #values that are considered null
+                        if v in [None, '']:
+                            setattr(self, k, None)
+                        else:
+                            #if Decimal column, cast whatever you get to Decimal
```
Jsyro: Looks good, thanks for reworking this as discussed.

exzizt: Of course!

Jsyro: Can you add to the deep_update testing suite this exact behaviour. (one happy like you have, and one nonsense like we talked about `description=[]` or something like that).

exzizt: Sure, I'll take a look.

PR #1245, opened 2020-03-13
Provisioning additional database resources and adding new readiness check

Jsyro:
4gb on a production database is not enough.


Jsyro:
> Will this affect both postgres instances? Seems like a contradiction to the PR name.
> 
> Should we split the metabase and postgres resources?

Yes it will, i've pulled back to 12gb max ram for now, could probably split, but i'm thinking they will use what they need. minimum is only 4.

PR #1246, opened 2020-03-13
Force Validations on NoW edit

mmumby:
- When editing a NoW application, if there is an error on any of the fields the user touched, it will prevent the form from submitting until the user fixes the errors.
- Created some new validation fun

- Styling/verbiage has been run past Brian
Screen Shot 2020-03-13 at 11 52 57 AM


services/core-web/common/utils/helpers.js
```diff
@@ -227,3 +228,28 @@ export const formatComplianceCodeValueOrLabel = (code, showDescription) => {
...
   return `${section}${formattedSubSection}${formattedParagraph}${formattedSubParagraph}${formattedDescription}`;
 };
+
+// function to flatten an object for nested items in redux form
+// eslint-disable-snippets
+export const flattenObject = (ob) => {
+  const toReturn = {};
+  let flatObject;
+  for (const i in ob) {
+    if (!ob.hasOwnProperty(i)) {
```
JustinMacaulay: shouldn't this always return false since your iterating through the values on the object?

mmumby: yes you are correct, I'll remove this block on both files.

mmumby: this block has been removed

PR #1247, opened 2020-03-13
Upgrade/static content endpoint

Jsyro:
One bulk static content loading. with the json cached.


services/core-api/app/api/exports/response_models.py
```diff
@@ -0,0 +1,89 @@
...
+        fields.List(
+            fields.Nested(APPLICATION_PROGRESS_STATUS_CODES),
+            attribute='NOWApplicationProgressStatus'),
+        'noticeOfWorkApplicationPermitTypeOptions':
+        fields.List(
+            fields.Nested(NOW_APPLICATION_PERMIT_TYPES), attribute='NOWApplicationPermitType'),
+        'noticeOfWorkApplicationReviewOptions':
+        fields.List(
+            fields.Nested(NOW_APPLICATION_REVIEW_TYPES), attribute='NOWApplicationReviewType')
+    })
```
exzizt: Looks good. Maybe get the bond static content in here before we pull it in?

services/core-api/app/api/exports/static_content/resources/core_static_content_resource.py
```diff
@@ -0,0 +1,77 @@
...
+    for model in MODELS_GET_ACTIVE:
+        static_content[model.__name__] = model.get_active()
+
+    return static_content
+
+
+class StaticContentResource(Resource):
+    @api.doc(
+        description=
+        'Returns a subset of mine data in a CSV. Column headers: mine_guid, mine_name, mine_no, mine_region, major_mine_ind, operating_status, operating_status_code, effective_date, tenure, tenure_code, commodity, commodity_code, disturbance, disturbance_code, permit_no, permittee_party_name'
```
exzizt: Incorrect comment?

services/core-api/app/api/compliance/models/compliance_article.py
```diff
@@ -27,4 +26,8 @@ def __repr__(self):
 
     @classmethod
     def find_by_compliance_article_id(cls, id):
-        return cls.query.filter_by(compliance_article_id=id).first()
\ No newline at end of file
+        return cls.query.filter_by(compliance_article_id=id).first()
+
+    @classmethod
+    def get_active(cls):
+        return cls.query.all()
```
JustinMacaulay: Is there a reason this get active is different than all the others?

services/core-api/app/api/parties/party/models/sub_division_code.py
```diff
@@ -14,19 +14,6 @@ class SubDivisionCode(AuditMixin, Base):
     def __repr__(self):
         return '<SubDivisionCode %r>' % self.sub_division_code
 
-
     @classmethod
-    def all_options(cls):
```
exzizt: Why isn't this function needed anymore?

services/core-web/common/actionCreators/staticContentActionCreator.js
```diff
@@ -25,6 +25,18 @@ import {
   getVarianceDocumentCategoryOptions,
 } from "../selectors/staticContentSelectors";
 
+export const loadBulkStaticContent = () => (dispatch) => {
+  console.log("LOAD BULK STATIC CONTENT");
```
exzizt: Remove console.log

services/core-web/common/reducers/staticContentReducer.js
```diff
@@ -225,11 +227,11 @@ export const getNoticeOfWorkApplicationPermitTypeOptions = (state) =>
...
-const isStaticContentLoaded = (state) =>
-  Object.keys(state)
-    // eslint-disable-next-line no-prototype-builtins
-    .filter((p) => state.hasOwnProperty(p) && Array.isArray(state[p]))
-    .every((p) => state[p].length > 0);
+const isStaticContentLoaded = (state) => true;
+// Object.keys(state)
+//   // eslint-disable-next-line no-prototype-builtins
+//   .filter((p) => state.hasOwnProperty(p) && Array.isArray(state[p]))
+//   .every((p) => state[p].length > 0);
```
exzizt: Is this commented-out code needed?

services/core-web/src/components/Home.js
```diff
@@ -80,10 +83,15 @@ export class Home extends Component {
 
   loadStaticContent = () => {
     this.props.dispatch(showLoading());
-    const staticContentActionCreators = Object.getOwnPropertyNames(staticContent).filter(
-      (property) => typeof staticContent[property] === "function"
-    );
-    staticContentActionCreators.forEach((action) => this.props.dispatch(staticContent[action]()));
+    console.log("HELPS");
```
exzizt: Remove console.log, and is commented-out code below needed?

services/core-api/app/api/compliance/models/compliance_article.py
Jsyro: i need all classes to have a get_active member to work with my global static content fetcher (i do now realize how similar this is to your static_data loading for marshmallow/swagger, but not sure they can be combined easier). This model didn't have a fetch all helper at all was being queried directly. since there isn't one all of them are active, we can add an active indicator and filter and this still works as intended.

services/core-api/app/api/exports/static_content/resources/core_static_content_resource.py
Jsyro: updated.

services/core-web/common/actionCreators/staticContentActionCreator.js
Jsyro: removed.

services/core-web/common/reducers/staticContentReducer.js
Jsyro: this is more to demonstrate the change, it would need to be revisited if we want to fuly adopt this change.

services/core-web/src/components/Home.js
Jsyro: removed.

services/core-web/common/actionCreators/staticContentActionCreator.js
```diff
@@ -25,6 +25,18 @@ import {
...
+  dispatch(request(reducerTypes.LOAD_ALL_STATIC_CONTENT));
+  return CustomAxios()
+    .get(ENVIRONMENT.apiUrl + API.CORE_STATIC_CONTENT, createRequestHeader())
+    .then((response) => {
+      dispatch(success(reducerTypes.LOAD_ALL_STATIC_CONTENT));
+      dispatch(staticContentActions.storeBulkStaticContent(response.data));
+    })
+    .catch(() => dispatch(error(reducerTypes.LOAD_ALL_STATIC_CONTENT)));
+};
+
```
Jsyro: note that all of static content creators in this file only exist for frontend tests, we could remove all of that is we prefer.

services/core-web/common/reducers/staticContentReducer.js
Jsyro: Actually, i think this check is still the correct check, (make sure all keys have values of arrays with length greater than 0... let me check.

services/core-api/app/api/parties/party/models/sub_division_code.py
Jsyro: This was formatting the dictionary for serialisation. it should just be querying objects and flask-restplus models handle the serialisation.

services/core-web/src/tests/components/dashboard/varianceHomePage/__snapshots__/VarianceHomePage.spec.js.snap
```diff
@@ -69,28 +69,6 @@ exports[`VarianceHomePage renders properly 1`] = `
               ],
             }
           }
-          filterVarianceStatusOptions={
```
mmumby: should this have been removed?

services/core-web/src/components/Home.js
```diff
@@ -80,10 +83,14 @@ export class Home extends Component {
 
   loadStaticContent = () => {
     this.props.dispatch(showLoading());
-    const staticContentActionCreators = Object.getOwnPropertyNames(staticContent).filter(
-      (property) => typeof staticContent[property] === "function"
-    );
-    staticContentActionCreators.forEach((action) => this.props.dispatch(staticContent[action]()));
+    this.props.dispatch(loadBulkStaticContent());
+    this.props.dispatch(fetchInspectors());
+    //  this.props.dispatch(fetchMineDisturbanceOptions());
```
mmumby: Why are we fetching Inspectors here? 
also comments can be deleted

Jsyro: inspectors is a query the front end treats as static content, but the backend doesn't think is static content, so it's special.

services/core-web/src/tests/components/dashboard/varianceHomePage/__snapshots__/VarianceHomePage.spec.js.snap
Jsyro: fixed. that constant doesn't exist anymore, used the new one.

mmumby:
https://minesdigitalservices-dev.pathfinder.gov.bc.ca/1247/dashboard/mines/?page=1&per_page=25

I just seeded dev with some data, it looks like the API gives a 200, but a lot of the dropdowns are empty.

PR #1248, opened 2020-03-13
Feature/mds 2665 create bond data structure

JustinMacaulay:
- Adds the base data structure to handle bonds.
- Also creates marsmallow schema for the bond structure.
- Adds the bond relationship to Permits.


migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE bond_status_code (
```
Jsyro: naming consistency i'm now keenly aware of, the table should be bond_status
and the pk should be bond_status_code

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
```diff
@@ -0,0 +1,56 @@
...
+    create_timestamp                 timestamp with time zone DEFAULT now() NOT NULL            ,
+    update_user                      varchar                                NOT NULL            ,
+    update_timestamp                 timestamp with time zone DEFAULT now() NOT NULL
+);
+
+ALTER TABLE bond_status_code OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS bond (
+    bond_id                                                          SERIAL PRIMARY KEY,
+    bond_guid                        uuid DEFAULT gen_random_uuid()            NOT NULL,
```
Jsyro: unique constraint for safety.

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
```diff
@@ -0,0 +1,56 @@
...
+    FOREIGN KEY (bond_status_code) REFERENCES bond_status_code(bond_status_code) DEFERRABLE INITIALLY DEFERRED,
+    FOREIGN KEY (bond_type_code) REFERENCES bond_type_code(bond_type_code) DEFERRABLE INITIALLY DEFERRED
+
+);
+
+ALTER TABLE bond OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS bond_permit_xref
+(
+    bond_permit_xref_guid uuid    DEFAULT gen_random_uuid() NOT NULL PRIMARY KEY,
```
exzizt: The XREF GUID isn't needed if we use a primary key of (bond_id, permit_id). I thought we wanted to ditch serial keys on XREF tables but maybe it doesn't matter so much anymore.

services/core-api/app/api/utils/static_data.py
```diff
@@ -45,3 +45,5 @@ def setup_static_data(Base):
 
             except Exception as e:
                 raise e
+
+    current_app.logger.debug(STATIC_DATA)
```
exzizt: Is this wanted debug?

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
JustinMacaulay: Im not saying one way or the other is right but we have a lot of type codes and status codes in our system named "x_status_code"  going forward are we aiming to change everything to the "bond_status" style?

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
JustinMacaulay: added

services/core-api/app/api/utils/static_data.py
JustinMacaulay: removed.

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
JustinMacaulay: I don't remember that conversation or choice but am more than happy to.

migrations/sql/V2020.03.12.08.33__create_bond_structure.sql
Jsyro: The code should refer to the identifier, each row in this table are a 'bond_status' the code is the identifier, i just looked at all our code tables (as you did too), and this inconsistency stood out.

PR #1249, opened 2020-03-16
Create a reusable component for the NoW "field origin" tooltip

exzizt:
* Creates and implements a variant (`<NOWFieldOriginTooltip />`) of a new component, `CoreTooltip.js` , a file which exports various tooltips that can be reused and all adhere to the same style
* Changes **VFCBC** to **vFCBC** in a couple places
* Fixes a typo in a tooltip's text: "to to" to "to"


pdittaro:
Great thought. The Tooltips were super well received on our regional office trip.

exzizt:
> So, if possible, can we make a more generic CoreTooltip component that we can work off of?
> 
> I'm something like thinking `<CoreTooltip class='noOrigin' />` to produce the message but that could also handle `<CoreTooltip>Tooltip content</CoreTooltip>` or other classes to produce different messages. That way we can control all the Tooltip formatting in our app in the same location.

That's a good idea. Maybe something to implement down the road if the need for more tooltips arises?

pdittaro:
I think the need is here presently, we need to get more info points on all our fields.

Still, if you want to defer we can just get this in for now.

exzizt:
> I think the need is here presently, we need to get more info points on all our fields.
> 
> Still, if you want to defer we can just get this in for now.

I can do it quickly. No worries!

services/core-web/src/components/common/CoreTooltip.js
```diff
@@ -0,0 +1,35 @@
...
+
+const types = {
+  "now-field-origin":
+    "This field is not being sent by NROS or vFCBC. Open the original PDF to see the data.",
+};
+
+const propTypes = {
+  type: PropTypes.oneOf(Object.keys(types)),
+  title: PropTypes.string,
+};
```
mmumby: title looks like its derived from 'props.type' and not a prop that's passed in.
Are there situations where a title prop is passed instead of a type? Can we standardize that?

exzizt: The component will use props.type if one is provided, otherwise it will display props.title. I think it might actually be nicer to restructure this component so that it exports multiple components that all adhere to the same style. E.g., <NowFieldOriginTooltip />, <SomethingElseTooltip />, <FooTooltip />. I think it's nicer to import and use the tooltips you want explicitly rather than having to toss a magic string around (e.g., type="now-field-origin"), which is prone to typos, etc., especially if you're using it a lot across multiple files. I don't expect this minor change to take very long so I'll set it back for ready for review when it's done.

PR #1257, opened 2020-03-18
Mds 2665 post and get for bonds

JustinMacaulay:
This contains the post, put and get for bonds as well as all the relevant tests for them. Also contains some changes to the bond data structure.


migrations/sql/V2020.03.17.16.19__update_bond_structure.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE bond ADD COLUMN institution_name varchar,
+ADD COLUMN institution_street varchar,
+ADD COLUMN institution_city varchar,
+ADD COLUMN institution_province varchar,
+ADD COLUMN institution_postal_code varchar,
```
Jsyro: Given that these are similar fields to the attributes on a party, can institution be a party of type company?

mmumby: We had a larger discussion about this - you must have missed it. We had two options here 1) create a special type of party for institutions or 2) add this information on the bond object. 

Given that institutions aren't always the same - the chances of them getting re-used sounded slim - so we opted for option 2.

Reasoning, If I have 10 bonds saved to the institution BMO, all with different branches and addresses, in our parties table we would have 10 BMO party names, with 10 different addresses. 

How we surface parties doesn't allow users to see the difference between the 10 BMO options, and it didn't sound like the reuse was there, so we'd be cluttering our parties table with little benefit. 
The lower cost option was to add this data on the bond object.

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -0,0 +1,88 @@
...
+
+        mine_guid = request.args.get('mine_guid', None)
+
+        if mine_guid is None:
+            raise BadRequest('Please provide a mine_guid.')
+
+        mine = Mine.find_by_mine_guid(mine_guid)
+
+        if mine is None:
+            raise BadRequest('No mine found with the provided mine_guid.')
```
exzizt: raise NotFound instead

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -0,0 +1,88 @@
...
+
+        try:
+            bond = Bond._schema().load(request.json['bond'])
+        except MarshmallowError as e:
+            raise BadRequest(e)
+
+        permit = Permit.find_by_permit_guid(request.json['permit_guid'])
+
+        if permit is None:
+            raise BadRequest('No Permit found with the guid provided.')
```
exzizt: raise NotFound instead

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -0,0 +1,88 @@
...
+
+class BondResource(Resource, UserMixin):
+    @requires_role_view_all
+    @api.marshal_with(BOND, code=200)
+    def get(self, bond_guid):
+
+        bond = Bond.find_by_bond_guid(bond_guid)
+
+        if bond is None:
+            raise BadRequest('No bond was found with the guid provided.')
```
exzizt: raise NotFound instead

services/core-api/tests/factories.py
```diff
@@ -577,7 +577,7 @@ def bonds(obj, create, extracted, **kwargs):
             return
 
         if not isinstance(extracted, int):
-            extracted = random.randint(0, 3)
+            extracted = random.randint(1, 3)
```
exzizt: I believe we want to create some permits that don't yet have bonds attached to them. It's good to have test data for that case. I did have a discussion with @Jsyro about this when I added this to the factories.

exzizt:
Great work!

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -0,0 +1,88 @@
+from flask_restplus import Resource, reqparse
+from datetime import datetime
+from flask import current_app, request
+from werkzeug.exceptions import BadRequest, NotFound, InternalServerError
```
exzizt: InternalServerError is an unused import, can you remove it from this file and the others it's imported into but not being used?

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: Changed

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: Changed

services/core-api/app/api/securities/resources/bond.py
Jsyro: I actually think this should return empty set. given it's a querystring parameter.

services/core-api/app/api/securities/resources/bond.py
Jsyro: I actually think this should return empty set. given it's a querystring parameter.

services/core-api/tests/factories.py
JustinMacaulay: I changed this specifically because when I was trying to create a permit to test bonds I would sometimes get a permit created with now bonds. That might not be the best way to do it but I wasn't super sure on how to pass in the argument needed to guarantee a bond is created.

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: removed

PR #1258, opened 2020-03-19
MDS-2566: Attach Generated NoW Documents to NoW

exzizt:
# Main
* Generated NoW documents are now attached to the NoW.

## Important:
* Modify the following line in `mds\services\document-manager\backend\.env`:
``DOCUMENT_MANAGER_URL=http://localhost:5001`` changes to ``DOCUMENT_MANAGER_URL=http://document_manager_backend:5001``


services/document-manager/backend/app/docman/resources/document.py
```diff
@@ -16,20 +16,21 @@
 class DocumentListResource(Resource):
```
exzizt: Since tuspy can't add params to the body of a POST, I had to send them in the header (see below). It might be nicer to clean this up so that params **must** come from the header (and not the body or the header).

services/core-api/app/api/now_applications/resources/now_application_document_type_resource.py
```diff
@@ -73,7 +73,9 @@ def post(self, document_type_code):
             NOW_DOCUMENT_DOWNLOAD_TOKEN(token), {
                 'document_type_code': document_type_code,
                 'now_application_guid': data['now_application_guid'],
-                'template_data': template_data
+                'template_data': template_data,
+                'username': User().get_user_username(),
+                'authorization_header': request.headers['Authorization']
```
exzizt: For now, we don't have a "proper" means of authorizing communication between our microservices, so this temporary solution has been put in place to authorize with the document manager (pass the authorization headers into the token and re-use them later). A ticket to set something else up as been created [here](https://bcmines.atlassian.net/browse/MDS-2744?atlOrigin=eyJpIjoiM2NiYTk2N2YyNTkwNDFiMGE2ODYwNWVlOTZiNDI4YTAiLCJwIjoiaiJ9).

services/core-api/requirements.txt
```diff
@@ -25,6 +25,7 @@ requests==2.22.0
 six==1.12.0
 SQLAlchemy==1.3.5
 sqlalchemy-filters==0.10.0
+tuspy==0.2.4
```
Jsyro: FunFact: This library is supported by the same group that supports our frontend tus client. 

https://github.com/tus/tus-py-client

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -183,10 +183,11 @@ export class NoticeOfWorkApplication extends Component {
 
   loadNoticeOfWork = (id) => {
     this.props.fetchImportedNoticeOfWorkApplication(id).then(({ data }) => {
-      this.setState({ isLoaded: true });
       this.loadMineInfo(data.mine_guid);
       this.handleProgressButtonLabels(data.application_progress);
-      this.props.fetchOriginalNoticeOfWorkApplication(id);
+      this.props
+        .fetchOriginalNoticeOfWorkApplication(id)
+        .then(() => this.setState({ isLoaded: true }));
```
Jsyro: The state isloaded should be probably happen after all of those promises resolve, is is that possible?

services/document-manager/backend/app/services/nris_download_service.py
```diff
@@ -1,63 +0,0 @@
-import requests
```
Jsyro: It looks like this was accidentally deleted.

services/document-manager/backend/app/utils/resources_mixins.py
```diff
@@ -1,8 +1,2 @@
 from flask_restplus import abort
 from .include.user_info import User
-
-
-class UserMixin(object):
-    def get_user_info(self):
-        user = User()
```
Jsyro: if this isn't being used, please delete the file completely.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
exzizt: Done. And I made it so the two calls that just need the NoW ID happen concurrently as well, rather than sequentially.

services/core-api/requirements.txt
nayr974: nice

services/core-api/app/api/document_generation/resources/now_document.py
```diff
@@ -1,35 +1,79 @@
...
 class NoticeOfWorkDocumentResource(Resource, UserMixin):
     @api.doc(
-        description='Returns the generated document associated with the received token.',
+        description=
+        'Returns the generated document associated with the received token and pushes it to the \
+        Document Manager and associates it with the Notice of Work.',
         params={'token': 'uuid4 token to execute the planned file generation for that token'})
     def get(self):
+
+        # Ensure that the token is valid
```
pdittaro: We should also make a comment in the code here about our requirement to have a valid user's token to generate the document. And describe the required order of operations to satisfy saving the generated now: first hit the generate document and then immediately hit download resource.

services/core-api/app/api/now_applications/resources/now_application_document_type_resource.py
pdittaro: Are you able to put this comment into the code as well in the form of a note or `TODO`?

pdittaro: Also, this resource will be where we are enabling the 'preview' functionality as well, right? We'll probably want to implement the service account ticket before doing the immediate preview.

services/document-manager/backend/.env-example
```diff
@@ -5,7 +5,7 @@ DB_PORT=5432
 DB_NAME=mds
 DB_NAME_TEST=mds_test
 
-DOCUMENT_MANAGER_URL=http://localhost:5001
+DOCUMENT_MANAGER_URL=http://document_manager_backend:5001
```
pdittaro: This will only work with Docker. Because of the way bridge networking works, with docker it should be able to work with both `localhost` and `document_manager_backend`, but if someone decides to run the document manager locally - say for debugging - this will no longer work.

Was this changed because it no longer worked as intended?

services/document-manager/backend/app/config.py
```diff
@@ -43,7 +43,8 @@ class Config(object):
     CACHE_REDIS_PASS = os.environ.get('CACHE_REDIS_PASS', 'redis-password')
     CACHE_REDIS_URL = 'redis://:{0}@{1}:{2}'.format(CACHE_REDIS_PASS, CACHE_REDIS_HOST,
                                                     CACHE_REDIS_PORT)
-    DOCUMENT_MANAGER_URL = os.environ.get('DOCUMENT_MANAGER_URL', 'http://localhost:5001')
+    DOCUMENT_MANAGER_URL = os.environ.get('DOCUMENT_MANAGER_URL',
+                                          'http://document_manager_backend:5001')
```
pdittaro: Similar here, this will only work on Docker

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -33,6 +35,22 @@ def initializeFileUploadWithDocumentManager(cls, request, mine, document_categor
...
+    def pushFileToDocumentManager(cls, file_content, filename, mine, document_category,
+                                  authorization_header):
+        folder, pretty_folder = cls._parse_upload_folders(mine, document_category)
+        data = {
+            'folder': folder,
+            'pretty_folder': pretty_folder,
+            'filename': filename,
+            'authorization': authorization_header
+        }
+        my_client = client.TusClient(cls.document_manager_url, headers=data)
```
pdittaro: This is where you are unable to send parameters in the body of the request?

exzizt: Correct.

services/document-manager/backend/.env-example
exzizt: I had a note in my PR comments about needing to change an .env file for it to work locally.
We had to change it from localhost to use the explicit hostname because of resource resolution issues. The docman will say "come get the PDF at localhost" which the API will fail to do, because, to itself, it is localhost. Here's the error that bubbles up without the modifications to the .env file:

```{"status":500,"message":"HTTPConnectionPool(host='localhost', port=5001): Max retries exceeded with url:  (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5d49465a20>: Failed to establish a new connection: [Errno 111] Connection refused',))"}```

services/document-manager/backend/app/config.py
exzizt: See above comment

services/core-api/app/api/now_applications/resources/now_application_document_type_resource.py
exzizt: Added comment

services/core-api/app/api/document_generation/resources/now_document.py
exzizt: Added comment

services/document-manager/backend/.env-example
pdittaro: Might be able to get around it with the loopback address 127.0.0.1 or the default 0.0.0.0 rather than localhost.

exzizt: That gives the same error because, to the API, it is also 127.0.0.1/0.0.0.0.

Jsyro: Read Line 85 of the document.py resource in the document manager. It sets the response header and tells the client where they can continue uploading the file. So the backend pod tries to send the the data to the document manager via 'localhost' but localhost:5001 doesn't exist on the backend pod, so we need to use the service name. this will not be an issue in openshift because it's always uses the service name.

pdittaro: Yeah, but the document_manager should be able to bind to the loopback address across all the containers, which is why I'm confused.

But @exzizt walked me through the implementation and it does seem like blocking behaviour, so I'm alright with it.

PR #1263, opened 2020-03-23
Health and Readiness checks for remaining services

pdittaro:
This PR creates health and readiness checks for our remaining deployments. Hopefully this will help us address issues where we have been unaware that a service is no longer responding - especially in the case of node evacuations.

Created checks for the following services:
- Document generation 
- Nginx
- Digdag
- Metabase
- Schemaspy
- Logstash


pdittaro:
I've done all of them except logstash, since it was a pain and we're going to remove it soon.

pdittaro:
Was unable to resolve the metabase liveness and readiness check issues in a timely manner, saving for another PR. The rest are ready for review.

PR #1265, opened 2020-03-24
wrong model to marshal with.

Jsyro:
oooops.

### DIFF
--- a/services/core-api/tests/utils/test_swagger.py
+++ b/services/core-api/tests/utils/test_swagger.py
@@ -1,13 +1,9 @@
-
 import json
-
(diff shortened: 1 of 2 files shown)


services/core-api/tests/utils/test_swagger.py
```diff
@@ -0,0 +1,13 @@
+
+import json
+
+from tests.factories import BondFactory, MineFactory, PermitFactory, PartyFactory
+from app.api.now_applications.resources.now_application_list_resource import PAGE_DEFAULT, PER_PAGE_DEFAULT
+
+
+class TestSwagger:
```
nayr974: Fantastic idea.

PR #1270, opened 2020-03-26
MDS-2758: View/Add/Edit Documents on Bonds

exzizt:
# Main
* Users can now view/add/edit documents on bonds

# Other
* Creates/fixes reusable DocumentTable component and uses it for Variances and Bonds
* Minor changes to some text throughout the app
* Fixed bug in formatMoney helper function


migrations/sql/V2020.03.18.10.40__create_bond_documents_type.sql
```diff
@@ -0,0 +1,12 @@
...
+    description                      varchar                                NOT NULL            ,
+    active_ind                       boolean                  DEFAULT true  NOT NULL            ,
+    create_user                      varchar                                NOT NULL            ,
+    create_timestamp                 timestamp with time zone DEFAULT now() NOT NULL            ,
+    update_user                      varchar                                NOT NULL            ,
+    update_timestamp                 timestamp with time zone DEFAULT now() NOT NULL
+);
+
+ALTER TABLE bond_document_type OWNER TO mds;
+COMMENT ON TABLE bond_document_type IS 'Contains a list of documents related to bonds used used by the Ministry. Bond document types include; Scan of Reclamation Security Bond, Release of Security Form (signed by inspector), Release of Security Letter (my letter to the bank or proponent), Confiscation of Security Form (signed by inspector), Confiscation of Security Letter (my letter to bank or proponent), Reminder Letter, Acknowledgement Letter.';
```
Jsyro: This comment will go out of date, can we remove the content, as that is in the rows. Show the intent and related tables.

Jsyro: also `used used` on the first line.

exzizt: "The types of documents that can be uploaded with a bond."

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -73,16 +76,19 @@ def get(self, bond_guid):
         return bond
 
     @api.doc(description='Update a bond')
-    @requires_role_edit_bonds
+    @requires_role_edit_securities
     @api.expect(BOND)
     @api.marshal_with(BOND, code=200)
     def put(self, bond_guid):
-
+        current_app.logger.debug("loading bond")
```
Jsyro: debug statement.

services/core-web/src/components/mine/Securities/BondDocumentsTable.js
```diff
@@ -0,0 +1,106 @@
+import React from "react";
```
mmumby: Instead of creating this new table, can we use 'DocumentTable.js' in common, It might need some tweaks to make is fully reusable, but should be used for this exact situation.

services/core-web/src/components/modalContent/AddBondModal.js
```diff
@@ -14,20 +15,24 @@ const propTypes = {
...
 
 const defaultProps = {
+  bond: {},
   editBond: false,
 };
 
 export const AddBondModal = (props) => {
-  const handleAddBond = (values) =>
-    props.editBond
+  const handleAddBond = (values) => {
```
mmumby: the change shouldn't be needed as it was an implicit return, but it doesn't matter.

services/core-web/src/components/modalContent/AddBondModal.js
```diff
@@ -14,20 +15,24 @@ const propTypes = {
...
+    return props.editBond
```
nayr974: I prefer the implicit return and we have done that in most places. Is there a reason you added return?

exzizt: Hmm... Can't remember. I put it back to normal.

services/core-web/src/components/modalContent/AddBondModal.js
exzizt: I put it back to normal, can't remember why I changed it.

services/core-web/src/components/mine/Securities/BondDocumentsTable.js
exzizt: Addressed in-person. Conclusion: Will make DocumentTable.js reusable (as it's currently Variance-specific).

migrations/sql/V2020.03.18.10.40__create_bond_documents_type.sql
klefler: Is it possible to leave a couple examples of a bond type in the description, it helps provide context to someone that may not have access to the database. Maybe just update the comment to say something like "Examples of bond types may include... and then have 2 or 3"?

JustinMacaulay: I have changed this based on the suggestions to be a may include with 3 examples

services/core-api/app/api/securities/models/bond.py
```diff
@@ -37,7 +38,8 @@ class _ModelSchema(Base._ModelSchema):
     project_id = db.Column(db.String)
 
     payer = db.relationship('Party', lazy='joined')
-    permits = db.relationship('Permit', uselist=False, lazy='select', secondary='bond_permit_xref')
+    permits = db.relationship('Permit', uselist=False, lazy='joined', secondary='bond_permit_xref')
```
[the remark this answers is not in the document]
JustinMacaulay: changed

PR #1272, opened 2020-03-27
MDS-2745: New Reports Tables Filter for Only Received Reports

exzizt:
# Main
* Creates a new filter for only showing reports that have a set received date (enabled by default) on the provincial reports table and the mine-specific reports table

## Provincial Reports
image

## Mine-Specific Reports
image

# Other
* Adjusts structure of Variance and NoW Provincial pages so they match the other Provincial pages
* Updates Provincial NoW page disclaimer text


services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
```diff
@@ -148,25 +148,27 @@ export class NoticeOfWorkHomePage extends Component {
...
-                applicationStatusOptions={this.props.applicationStatusOptions}
-                applicationTypeOptions={this.props.applicationTypeOptions}
-              />
-              <div className="center">
-                <ResponsivePagination
-                  onPageChange={this.onPageChange}
-                  currentPage={Number(this.state.params.page)}
-                  pageTotal={Number(this.props.pageData.total)}
-                  itemsPerPage={Number(this.state.params.per_page)}
+              <div className="tab__content">
```
mmumby: I see you added this class to Variance and Reports (likely to match the other tables on the dashboard view). This is just personal preference but I don't think these views need that much padding. Thoughts?  

I was just looking at all the tables on Test and I prefer the ones without this class. (This is in no way blocking to getting your PR in)

exzizt: I agree that it doesn't need that much padding, but I figured I would add it to the Provincial pages that didn't have it, rather than remove it from the ones that did. Maybe we can look at it in the future because it's nice to have as much width as possible for the tables.

PR #1275, opened 2020-03-31
Mds 2743 Exemption Fee Status

nayr974:
This PR adds Exemption Fee Status to a mine.


services/core-api/app/api/mines/mine/models/excemption_fee_status.py
```diff
@@ -0,0 +1,22 @@
+from sqlalchemy.schema import FetchedValue
+from app.extensions import db
+
+from app.api.utils.models_mixins import AuditMixin, Base
+
+class ExemptionFeeStatus(AuditMixin, Base):
+    __tablename__ = 'exemption_fee_status'
+    exemption_fee_status_code = db.Column(db.String, nullable=False, primary_key=True)
+    description = db.Column(db.String(100), nullable=False)
```
exzizt, line 9: This size constraint isn't enforced on the DB level, maybe it should be?

services/core-api/tests/status_code_gen.py
```diff
@@ -35,6 +36,8 @@ def RandomMineRegionCode():
 def RandomPermitStatusCode():
     return random.choice([x.permit_status_code for x in db.session.query(PermitStatusCode).all()])
 
+def RandomExemptionFeeStatusCode():
+    return random.choice([x.exemption_fee_status_code for x in db.session.query(ExemptionFeeStatus).all()])
```
exzizt, line 40: Do we want to generate records with inactive exemption fee statuses? If not, maybe use ExemptionFeeStatus.get_active().

services/core-web/common/reducers/staticContentReducer.js
```diff
@@ -37,6 +37,7 @@ const initialState = {
   bondStatusOptions: [],
   bondTypeOptions: [],
   bondDocumentTypeOptions: [],
+  exemptionFeeSatusOptions: [],
```
exzizt, line 40: Satus --> Status :D
(In a few locations)

services/core-web/src/components/mine/MineHeader.js
```diff
@@ -202,13 +203,7 @@ export class MineHeader extends Component {
...
                 )}
                 {this.props.mine.mine_status[0] && (
-                  <img
-                    alt="info"
-                    className="dashboard__header--card__content--status__img"
-                    src={INFO_CIRCLE}
-                    style={{ marginLeft: 5 }}
-                    title={this.props.mine.mine_status[0].status_description}
-                  />
+                  <CoreTooltip title={this.props.mine.mine_status[0].status_description} />
```
exzizt, line 206: ❤️

services/core-web/src/components/mine/MineHeader.js
```diff
@@ -284,20 +279,23 @@ export class MineHeader extends Component {
...
                 </Popover>
               ) : (
                 <p>{String.EMPTY_FIELD}</p>
               )}
             </div>
           </div>
+          <div className="inline-flex padding-small wrap">
+            <p className="field-title">Exemption Status</p>
+            <div>
+              {(null && this.props.mine.exemption_fee_status.description) || String.EMPTY_FIELD}
```
exzizt, line 293: What's the desired output here? This always displayed `String.EMPTY_FIELD` because of the `null &&`.

services/core-api/app/api/mines/mine/models/excemption_fee_status.py
nayr974: I actually removed it, we don't want to enforce string lengths at the data level, and this definition is not actually enforced by SQLAlchemy in any way so we have been meaning to just use db.String.

services/core-api/tests/status_code_gen.py
nayr974: Good observation, changed

services/core-web/common/reducers/staticContentReducer.js
nayr974: Did I really do that and then copy and paste that EVERYWHERE? Yes. Fixed!

services/core-web/src/components/mine/MineHeader.js
nayr974: 💯

services/core-web/src/components/mine/MineHeader.js
nayr974: This was test code I forgot to undo. Updated

PR #1276, opened 2020-03-31
MDS-2772: Securities - Reclamation Invoices (Backend)

exzizt:
# Note
* This ticket is a subtask of MDS-2701. This ticket captures the backend work while the frontend work is being completed on MDS-2771.

# Main
* Implements the API and database changes required to provide reclamation invoice management
* Creates backend tests for reclamation invoices

# Other
* Restructures the URL for our API's endpoints for general "securities" objects, which are currently bonds and reclamation invoices
* Adjusts permissions on various bond endpoints
* Fixes a semantic error for a bond test (should test for 400 error code instead of 404)


services/core-api/app/api/securities/resources/bond.py
```diff
@@ -26,7 +26,7 @@ def get(self):
         mine = Mine.find_by_mine_guid(mine_guid)
 
         if mine is None:
-            raise NotFound('No mine found with the provided mine_guid.')
+            raise NotFound('No mine was found with the guid provided.')
```
Jsyro: platonically, this should just return `[]`. it's a querystring param.

services/core-api/app/api/securities/resources/reclamation_invoice.py
```diff
@@ -0,0 +1,101 @@
...
+
+        mine_guid = request.args.get('mine_guid', None)
+
+        if mine_guid is None:
+            raise BadRequest('Please provide a mine_guid.')
+
+        mine = Mine.find_by_mine_guid(mine_guid)
+
+        if mine is None:
+            raise NotFound('No mine was found with the guid provided.')
```
Jsyro: same here, just return `[]`

services/core-api/app/api/securities/resources/bond.py
exzizt: `raise NotFound()`?

exzizt: Nevermind, I see what you mean now... Are you sure that would be good behaviour? The user is trying to get all bonds on a specific mine, if they provide an invalid mine GUID, I think we should raise not found. I think returning an empty [] for bonds is only appropriate if they search for a valid mine that happens to have no bonds.

services/core-api/app/api/securities/resources/reclamation_invoice.py
```diff
@@ -0,0 +1,101 @@
...
+        try:
+            reclamation_invoice = ReclamationInvoice._schema().load(
+                request.json['reclamation_invoice'])
+        except MarshmallowError as e:
+            raise BadRequest(e)
+
+        permit = Permit.find_by_permit_guid(request.json['permit_guid'])
+
+        if permit is None:
+            raise NotFound('No permit was found with the guid provided.')
```
Jsyro: that should be BadRequest, 404 is `resource(uri) not found` or `object not found.`

services/core-api/app/api/securities/resources/bond.py
Jsyro: querystring param should be treated like a `WHERE` clause, we are enforcing that mine_guid is provided because we refuse to return all securities across all mines. 

ivory tower stuff, not the end of the world.

exzizt: I understand more what you mean now. I'm happy to change it to return an empty result set then.

services/core-api/app/api/securities/resources/reclamation_invoice.py
exzizt: I'll make the same change to the similar bond endpoint.

PR #1278, opened 2020-04-01
Loki and grafana deployment

pdittaro:


docker-compose.yml
```diff
@@ -1,26 +1,24 @@
...
-      - ./settings/keycloak/keycloak-local-realm.json:/tmp/keycloak-local-realm.json
-      - ./settings/keycloak/keycloak-local-user.sh:/tmp/keycloak-local-user.sh
-    ports:
-      - "8080:8080"
-      - "8443:8443"
-    networks:
-      - mds
-    # logging:
-    #     driver: none
+  # keycloak:
```
exzizt, line 5: Any reason why this is commented out? Should it be removed?

docker-compose.yml
```diff
@@ -220,14 +197,123 @@ services:
...
+    ports:
+      - 3010:3000
+    volumes:
+      - grafana_storage:/var/lib/grafana
+      - ./monitoring/grafana/provisioning/:/etc/grafana/provisioning/
+    env_file:
+      - ./monitoring/grafana/env
+    restart: always
+
+  # jaeger:
```
exzizt, line 251: No jaegar?

pdittaro: Not yet, next PR, didn't feel like removing the WIP though

docker-compose.yml
pdittaro: This was commented out as I felt like our docker-compose was getting pretty full. But seriously, we might need to consider a different approach for how we manage local services.

PR #1279, opened 2020-04-01
MDS-2776: Edit NoW Application Status

exzizt:
# Main
* Users can now edit a NoW's status via the NoW page's Action dropdown menu
* Creates new NoW status codes and deletes the old ones
* Creates a display order column for NoW statuses (and assigns this value so they are alphabetical in ascending order by description) 
* Modifies our factories to only use active NoW statuses when generating NoWs

### Visuals
image
A temporary spot to display the status. Roxanne is working on a mockup that will determine where it will go.

image
Option in Actions dropdown

image
Change Application Status Modal

# Other
* Fixes an issue where some columns in the NoW View were not being determined properly

## Note
* Our `flask create-data` command needs to re-run the NoW View creation SQL after each run, otherwise, the view will contain outdated information. Please talk to me about this.
* As far as the frontend implementation goes, I followed the same pattern that updating the lead inspector uses.


migrations/sql/V2020.04.01.10.33__add_now_statuses.sql
```diff
@@ -0,0 +1,23 @@
+-- Create a new column for defining the desired display order of the statuses
+ALTER TABLE now_application_status ADD COLUMN display_order integer;
+
+-- Inactivate all of the existing Core statuses and set their display order
+UPDATE now_application_status SET active_ind = false, display_order = 5 WHERE now_application_status_code = 'ACC';
```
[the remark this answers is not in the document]
exzizt: I agree. If they're not in production, I guess we don't need an ETL.
Also, I already adjusted our factories to only generate test data with the active statuses.
Looks like the default status is one of the old ones, `UNR`:
image
Maybe it should be null (if it's a nullable field), otherwise... What's a new good default given the new statuses?

migrations/sql/V2020.04.01.10.33__add_now_statuses.sql
```diff
@@ -0,0 +1,23 @@
+-- Create a new column for defining the desired display order of the statuses
+ALTER TABLE now_application_status ADD COLUMN display_order integer;
+
+-- Delete all of the old Core statuses (Accepted, Under Review, and Withdrawn)
+DELETE FROM now_application_status WHERE now_application_status_code = 'ACC';
```
nayr974, line 5: I think this will fail when deployed to test. In test, these statuses do exist and are set. So I think the appropriate order of operations here is:

1. Insert new statuses
2. Set any applications to the default new status (Active/Issued/Approved)
3. Delete the old statuses

exzizt: Good point, fixed!

PR #1293, opened 2020-04-09
Fix fee exemption ETL

nayr974:
This PR fixes a bug causing the fee exemption status to not update for mines that already exist.


migrations/sql/R__stored_procedure_for_mine_information.sql
```diff
@@ -153,7 +153,8 @@ CREATE OR REPLACE FUNCTION transfer_mine_information() RETURNS void AS $$
             WHERE ETL_MINE.mine_guid = mine.mine_guid
             AND (ETL_MINE.mine_name != mine.mine_name
                 OR ETL_MINE.mine_region != mine.mine_region
-                OR ETL_MINE.major_mine_ind != mine.major_mine_ind);
+                OR ETL_MINE.major_mine_ind != mine.major_mine_ind
+                OR ETL_MINE.exemption_fee_status_code != mine.exemption_fee_status_code););
```
pdittaro, line 157: Is this right? Can we get some indentation to suggest where the second `);` is coming from?

nayr974: It's not right :) Fixed

PR #1294, opened 2020-04-09
Encrypt NRIS database connection

pdittaro:


pdittaro:
I'm testing in Test to see if this is all we need to do. I'm hoping that the `makedsn` function is flexible enough to handle this change.

pdittaro:
image

PR #1297, opened 2020-04-09
MDS-2722: Document Generation Improvements

exzizt:
# Main
* Contact info in the three letters' footers is automatically populated depending on the mine's region
* Creates regional contact types and regional contacts models to store data to be used in auto-populating letter footers for some generated documents (and wherever else in the app in the future)
* Changes "Client Acknowledgement Letter" to "Acknowledgement Letter"
* Combines the multiple paragraph form inputs for the Client Acknowledgement Letter into just one (for the letter body)
* Addresses multiple formatting/layout/structure inconsistencies in the three generated letters

# Other
* Fixes linting errors from a previously-merged PR ;)
* Fixes a typo for some generated letters (propnent --> proponent)


migrations/sql/V2020.04.08.15.06__add_new_fields_to_now_document_templates.sql
```diff
@@ -0,0 +1,352 @@
+UPDATE document_template SET form_spec_json = '[
```
exzizt, line 1: Is this entire file necessary? Or are the equivalent updates in seed_data enough?..

nayr974: The inserts in seed data typically have "ON CONFLICT DO NOTHING", which for our intents means if the unique key already exists do nothing. It doesn't handle updating any fields if they have already been inserted. IMO the seed data file shouldn't exist, but it can be nice to have to review at times.

services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
```diff
@@ -75,8 +67,10 @@ export class NOWPermitGeneration extends Component {
 
   handlePremitGenSubmit = (values) => {
     if (this.props.isAmendment) {
+      // eslint-disable-next-line no-param-reassign
```
nayr974, line 70: Can you please add a comment as to why disabling these linting rules is needed?

exzizt: `no-param-reassign` is the linting rule being prevented from being triggered. Is that not enough?
Justin didn't have his linter running yesterday, so when I merged from develop today I had to quickly fix some issues.

migrations/sql/V2020.04.08.15.06__add_new_fields_to_now_document_templates.sql
exzizt: I think because seed data runs last, this file might not be needed, as seed data will insert these values.

services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
exzizt: I've pushed up some code to fix the linting issue.

migrations/sql/V2020.04.08.15.06__add_new_fields_to_now_document_templates.sql
nayr974: If these are net new, the files is not needed. If they existed and you are updating them, they are needed.

exzizt: Thanks for clarifying. 😄

PR #1298, opened 2020-04-09
Fix/mds 2799 multi post on importing no w submissions

JustinMacaulay:
small fix to prevent a double click of the import now submission button, added some additional checking on the backend as well.


services/core-api/app/api/now_applications/resources/now_application_import_resource.py
```diff
@@ -58,6 +58,9 @@ def post(self, application_guid):
         application.latitude = latitude
         application.longitude = longitude
         application.now_application_guid = application_guid
+        db.session.refresh(now_application_identity)
+        if now_application_identity.now_application_id is not None:
+            raise BadRequest('This redcord has already been imported.')
         application.save()
```
Jsyro, line 64: can we leave a note that this is not perfect due to the request processing time and multiple requests fired at once. 

Great first cut to avoid issues.

exzizt: redcord --> record

services/core-api/app/api/now_applications/resources/now_application_import_resource.py
```diff
@@ -58,6 +58,12 @@ def post(self, application_guid):
         application.latitude = latitude
         application.longitude = longitude
         application.now_application_guid = application_guid
+
+        # This is a first pass but by no means exhaustive solution to preventing the now application from being saved more than once.
+        # In the event of multiple requests being fired simultaneously this can still sometimes fail.
+        db.session.refresh(now_application_identity)
+        if now_application_identity.now_application_id is not None:
+            raise BadRequest('This redcord has already been imported.')
```
exzizt, line 66: typo (redcord)

PR #1299, opened 2020-04-09
Fix incorrect phone number value for SW Regional Office contact

exzizt:
* Fixes incorrect phone number value for SW Regional Office contact


pdittaro:
A migration is kind of a strange way to fix a data record.

exzizt:
> A migration is kind of a strange way to fix a data record. This pattern can introduce a lot of side effects.

I thought this was the only way it should be done. Any changes we make to existing data/structures should be done in the flyway SQL scripts. Does this not fall in that category?

PR #1303, opened 2020-04-15
MDS-2780: CRR Fixes and Improvements (Core and MineSpace)

exzizt:
# Main
## MDS-2780
* [**MineSpace**] Hide the summary cards on the Reports tab pane
* [**MineSpace**] Update the info text on the Reports tab pane
* [**MineSpace**] Update the styling for required fields in the Add Report modal
* [**MineSpace**] Disallow setting the compliance year to a date in the future in the Add Report modal
* [**MineSpace**] Update various info text and labels for fields in the Add Report modal
* [**MineSpace**] Adds a "Code Section" column to the Reports table
* [**Core**] Adds a "Code Section" column to the Reports table (visible in provincial and mine-specific views)
* [**Core**] When adding or editing a report, received date is now a required field when at least one file is uploaded to the report

## MDS-2781
* [**MineSpace**] File Pond's functionality to remove uploaded files is now used in the add/edit report modals

# Other
* [**MineSpace**] When editing and adding files to a report that previously had no attached files, set the report's received date to today
* [**MineSpace**] When adding a report in MineSpace, set the received date to today if at least one file is attached (otherwise, don't set the received date)
* [**MineSpace**] Adds "accepted file type" validation to the files upload for reports

# Screenshots
## MineSpace
image
image
image

## Core
image
image


services/core-web/common/selectors/staticContentSelectors.js
```diff
@@ -321,6 +321,18 @@ export const getDropdownMineReportDefinitionOptions = createSelector(
...
+  (options) =>
+    options.reduce(
+      (map, mine_report_definition) => ({
+        [mine_report_definition.mine_report_definition_guid]: mine_report_definition,
+        ...map,
+      }),
+      {}
+    )
+);
+
```
mmumby, line 335:
```
export const getMineReportDefinitionHash = createSelector(
  [getDropdownMineReportCategoryOptions],
  createLabelHash
);
```

there is a function you can use to make the hash from the dropdown options.

mmumby: There should also be a test for this selector

services/minespace-web/src/components/Forms/reports/ReportSubmissions.js
```diff
@@ -55,28 +63,46 @@ export const ReportSubmissions = (props) => {
...
+                  >
+                    {file.document_name}
+                  </LinkButton>
+                </li>
+              ))}
+            </ul>
+          )) || <div>This report has no uploaded files.</div>}{" "}
+        </Form.Item>
+      )}
+      <Form.Item label="Upload Files">
```
mmumby, line 88: This is great!

services/core-web/common/selectors/staticContentSelectors.js
exzizt: I looked at `createLabelHash` but it doesn't create the shape of object I need. It creates "string:string" when I needed "string:object".

services/minespace-web/src/components/common/RenderYear.js
```diff
@@ -79,6 +81,7 @@ export class RenderDate extends Component {
         onOpenChange={this.handleOpenChange}
         onPanelChange={this.handlePanelChange}
         onChange={this.clearValue}
+        disabledDate={this.props.disabledDate}
         disabled={this.props.disabled}
```
mmumby, line 85: Can the disabled flag be used here instead of creating a new disabled prop?

services/minespace-web/src/utils/Validate.js
```diff
@@ -83,3 +85,6 @@ export const validateStartDate = memoize((previousStartDate) => (value) =>
     ? "New manager's start date cannot be on or before the previous manager's start date."
     : undefined
 );
+
+export const yearNotInFuture = (value) =>
+  value && value > moment().year() ? "Year cannot be in future" : undefined;
```
mmumby, line 90: I thought utils were in the common folders?

services/minespace-web/src/components/common/RenderYear.js
exzizt: `disabledDate` is a prop for `DatePicker`: https://ant.design/components/date-picker/

services/minespace-web/src/utils/Validate.js
mmumby: Correction, it is. Can you remove this and update the file to use yearNotInFuture form the common folder?

exzizt: MineSpace still has its own Validate file in src. I could move this function into the common folder if you like

services/core-web/common/selectors/staticContentSelectors.js
mmumby: ohhh, okay I didn't notice that was an object. Okay.

services/minespace-web/src/components/common/RenderYear.js
mmumby: Okay!

services/minespace-web/src/utils/Validate.js
mmumby: It should already exist in the common folder!

exzizt: Nice, I looked but I didn't see it the first time. Using it now.

services/core-web/common/selectors/staticContentSelectors.js
exzizt: > There should also be a test for this selector

I made the test for this new selector :)

mmumby: 👍

exzizt:
> Good work Luke!

Thanks!

PR #1305, opened 2020-04-17
MDS-2800/MDS-2785: Handle Currency Felds in Document Generation/Bond Edit Form Payer Field Bug

exzizt:
# Main
## MDS-2800:
* Formats the bond amount as currency in the Acknowledgement Letter document template
image
* The bond amount input field behaves like a currency input field
image

## MDS-2785:
* Fixed a bug when editing a bond, the "Payer" field was not populating properly

# Other
* ~~As a solution for this ticket, created the ability to define "context props" for generated form input fields in our form specs, which allows us to pass multiple props of any kind to our Field components. This was used to provide the currency mask and currency-related validation props to a FIELD component. And, for example, could be used in the future to provide the dataSource (e.g., inspectors list) for a SELECT component for inspectors.~~


migrations/sql/V2020.04.17.12.48__add_context_props_to_document_template.sql
```diff
@@ -0,0 +1,132 @@
...
+      "placeholder": "Enter the exploration type",
+      "required": true
+    },
+    {
+      "id": "bond_inc_amt",
+      "label": "Bond Amount",
+      "type": "FIELD",
+      "placeholder": "Enter the bond amount",
+      "required": true,
+      "context-props": ["currency"]
```
exzizt, line 69: This is the only change.

services/core-web/.eslintrc.js
```diff
@@ -40,5 +40,6 @@ module.exports = {
     "jsx-a11y/label-has-for": 0, // deprecated rule
     "import/no-cycle": 0,
     "no-param-reassign": 0,
+    "no-return-assign": 0,
```
exzizt, line 43: I feel like we can do without this one, personally. Line 12 in documentGeneration.js was the cause in this case.

migrations/sql/V2020.04.17.12.48__add_context_props_to_document_template.sql
Jsyro: I updated Seed data, there are update statements that will run to set these values, this can just be changed there. my guess is that would actually overwrite this given it runs after.

migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -933,7 +933,8 @@ UPDATE document_template SET form_spec_json = '[
       "label": "Bond Amount",
       "type": "FIELD",
       "placeholder": "Enter the bond amount",
-      "required": true
+      "required": true,
+      "context-props": ["currency"]
```
Jsyro, line 937: adding another member seems a bit unnessecary, would it be easier to make the "type": "CURRENCY' and make a fieldtype of currency that extends the base field?

exzizt: I tried that at first, but if you look at how we get the currency functionality we don't pass props at that level, e.g.: 
```
            <Form.Item>
              <Field
                id="amount"
                name="amount"
                label="Bond Amount*"
                component={RenderField}
                {...currencyMask}
                validate={[required, number, currency]}
              />
            </Form.Item>
```

migrations/sql/V2020.04.17.12.48__add_context_props_to_document_template.sql
exzizt: True, I was skeptical about the need for this file. I can delete it then?

services/core-web/src/components/Forms/GenerateDocumentForm.js
```diff
@@ -31,6 +32,7 @@ const createFields = (fields) => (
                   placeholder={field.placeholder}
                   component={renderConfig[field.type]}
                   validate={field.required ? [required] : null}
+                  {...getContextProps(field["context-props"])}
```
exzizt, line 35: To provide a rough example, essentially what this is doing is this:

```
<Form.Item>
    <Field
    id="amount"
    name="amount"
    label="Bond Amount*"
    component={RenderField}
    format: ƒ (e)
    normalize: ƒ (e,t)
    onChange: ƒ (e)
    onFocus: ƒ (e)
    autoComplete: "off"
    validate: (3) [ƒ, ƒ, ƒ]
    />
</Form.Item>
```

migrations/sql/afterMigrate__1__seed_data.sql
nayr974: It's extensible, but doesn't seem as maintainable given that the intent of what's being doing is not intuitive. I also don't like how it's blending concerns of very react-specific front-end component constructs verses the back-end defining the expectations of the form data. I agree with Jason and I would expect that whatever is processing the form definition acts on this type versus the back-end trying to tell the front-end how to work on this level.

exzizt: No problem. I have an alternate solution that I will implement then.

exzizt:
> I like this approach a lot better, thanks for making the changes Luke!

I do too! Thank you for suggesting them!

PR #1307, opened 2020-04-17
Added DocGen component

garywong-bc:
Just a doc change:
* added the stateless DocGen component
* added an .svg version of the same exported PDF


garywong-bc:
Overrode review rule and just merged in as it's only *.md changes

PR #1310, opened 2020-04-21
NROS and VFCBC get status by NoW Number

JustinMacaulay:
This changes the way Notice of Work statuses are queried and returned to NROS and VFCBC


services/core-api/app/api/now_submissions/resources/application_status_resource.py
```diff
@@ -42,9 +42,8 @@ def get(self):
         updated_status_records = []
         for now_application in now_applications:
             updated_status_record = {
-                "messageid":
-                None if not now_application.now_application_identity else
-                now_application.now_application_identity.messageid,
+                "now_number":
+                now_application.now_application_identity.now_number,
```
exzizt, line 46: We sure the null check isn't needed anymore?

PR #1315, opened 2020-04-22
MDS-2728: Add a "Source" Column to NoW Table Views

exzizt:
# Main
* Adds a "source" column to indicate the originating system for NoW applications in the provincial and mine-specific views (and allows for sorting/filtering by this column)
image
image

* Removes the 'submissions_only=true' flag from relevant URLs

# Other
* Conditional text for "Add a Permit Application"/"Add a Notice of Work Application" button depending on if a major or regional mine in mine dashboard
* Fixes the sorting for the mine-specific NoW table, and disables the buggy filtering (it doesn't work for all columns and it should be changed to be frontend filtering, as are all of our other mine-specific tables. This should be addressed in a future task that gives NoW the same filtering as all of the other views)
* Fixes factories to generate better "originating system" data for NoW submissions


services/core-web/src/components/mine/NoticeOfWork/MineNOWApplications.js
```diff
@@ -93,35 +87,33 @@ export class MineNOWApplications extends Component {
...
 
   render() {
     const isMajorMine = this.props.mines[this.props.mineGuid].major_mine_ind;
-    const title = isMajorMine ? "Permit Applications" : "Notice of Work Applications";
+    const type = isMajorMine ? "Permit Application" : "Notice of Work Application";
     return (
       <div className="tab__content">
         <div>
-          <h2>{title}</h2>
+          <h2>{type}s</h2>
```
mmumby, line 103: {types}

exzizt: That's intentional 😄 If you look at where else it is used

mmumby: haha wait whaaat, how is that correct syntax? 🤔

PR #1321, opened 2020-04-29
mms now don't have status, give them SUB

Jsyro:


services/core-api/app/api/now_applications/transmogrify_now.py
```diff
@@ -9,7 +9,7 @@
 
 from flask import current_app
 
-status_code_mapping = { "Accepted": "AIA", "Withdrawn": "WDN", "Under Review": "SUB" }
+status_code_mapping = {"Accepted": "AIA", "Withdrawn": "WDN", "Under Review": "SUB", None: "SUB"}
```
exzizt, line 12: None --> "None" ?

Jsyro: It is lookup None, not the string 'None' both are valid in python though. Good Question.

PR #1322, opened 2020-04-30
Fix bugs in Provincial NoW searching, sorting, and filtering

exzizt:
* Fixes a variety of bugs regarding searching, sorting, and filtering on the Provincial NoW page
_Looking forward to when we have this page match the others (with the filtering form, etc.!)_


services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.js
```diff
@@ -35,58 +33,62 @@ const propTypes = {
...
-  }
+  renderDataFromURL = (params) => {
+    const parsedParams = queryString.parse(params);
+    this.props.fetchNoticeOfWorkApplications(parsedParams).then(() => {
+      this.setState({ isLoaded: true });
+    });
+  };
 
   renderDataFromURL = (queryParams) => {
     const params = queryParams || this.props.location.search;
```
mmumby, line 87: was this removed? Seems the same function was added above

exzizt: This was the one push I meant to make before you reviewed, I forgot to delete the old version of the function (but everything still worked as intended since the new one was defined above the old one).

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkTable.js
```diff
@@ -108,15 +118,15 @@ export class NoticeOfWorkTable extends Component {
             value={selectedKeys[0] || this.props.searchParams[field]}
             onChange={(e) => setSelectedKeys(e.target.value ? [e.target.value] : [])}
```
mmumby, line 119: [field] is different than the other patterns, is this due to the filter within the table?

exzizt: Yup!

exzizt:
> no BE changes were required to update this following the other formats?

No, sort of oddly enough. I actually think this file got reverted to an old version at some point in the past but the backend changes remained intact. Maybe a bad merge conflict resolution, but I'm not sure. I was pretty sure I removed the join/split query list params thing a lot time ago from these files. My memory Is too foggy to remember

PR #1334, opened 2020-05-08
Transfer a bond to a different permit

mmumby:
- Ability to release a current bond and transfer it to another permit on the same mine
- on The UI showing only the permit dropdown when transferring, and the note field as requested from Eva/Em
Screen Shot 2020-05-08 at 11 55 25 AM
Screen Shot 2020-05-11 at 11 22 46 AM


services/core-web/src/components/mine/Securities/MineSecurityInfo.js
```diff
@@ -139,6 +139,34 @@ export class MineSecurityInfo extends Component {
...
+        mineGuid: this.props.mineGuid,
+        permits: this.props.permits,
+      },
+      width: "50vw",
+      content: modalConfig.TRANSFER_BOND_MODAL,
+    });
+  };
+
+  transferBond = (previousBondPayload, previousBondGuid, newBondPayload) => {
+    this.props.updateBond(previousBondPayload, previousBondGuid).then(() => {
```
pdittaro, line 160: I'm not sure if I agree with this implementation - I think we should create a specific 'transfer' endpoint action to deal with this rather than orchestrating it on the front-end. 

With this we don't have a way of capturing whether someone has explicitly transferred a bond, rather than just release and recreate. Plus, there's some error handling concerns as well.

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -97,4 +97,55 @@ def put(self, bond_guid):
...
+        # Get the permit to transfer the bond to and validate it
+        permit_guid = request.json.get('permit_guid', None)
+        if not permit_guid:
+            raise BadRequest('permit_guid is required.')
+        permit = Permit.find_by_permit_guid(permit_guid)
+        if not permit:
+            raise BadRequest('No permit was found with the permit_guid provided.')
+        if permit.permit_guid == bond.permit.permit_guid:
+            raise BadRequest('This bond is already associated with this permit.')
+
```
exzizt, line 124: Should validate that the permit_guid belongs to the same mine as the current permit.

JustinMacaulay:
Im not sure if I'm missing something in the ticket for this feature but I see a specific section on the user needs to be able to transfer this bond from one mine to another. Is there a reason that functionality was removed?

PR #1335, opened 2020-05-08
MDS-2811 NoW status change push to NROS

JustinMacaulay:
Whenever a NoW status changes in core the change will be pushed to NROS.


services/core-api/app/api/now_applications/resources/now_application_import_resource.py
```diff
@@ -65,5 +67,12 @@ def post(self, application_guid):
         if now_application_identity.now_application_id is not None:
             raise BadRequest('This record has already been imported.')
         application.save()
+        db.session.refresh(now_application_identity)
+        status = NOWApplicationStatus.find_by_now_application_status_code(
+            now_application_identity.now_application.now_application_status_code)
```
Jsyro, line 72: now_application.now_application_status should return this. if it doesn't exist, can you add that relationship and use it here?

services/core-api/app/api/services/nros_now_status_service.py
```diff
@@ -0,0 +1,56 @@
...
+        resp = requests.post(
+            url=cls.nros_now_url,
+            headers={
+                'Authorization': f'Bearer {token}',
+                'Content-type': 'application/json',
+                'lobName': 'MMS'
+            },
+            data=json.dumps(data))
+        current_app.logger.info(f'Status for NoW Number{now_number} pushed to NROS')
+        return
```
Jsyro, line 45: We should return something to the caller (even if they ignore it) and maybe log to the warning stream if the post responses.status_code is not 204, or something like that. If this needs to go live ASAP, we can defer this.

services/core-api/app/api/now_applications/resources/now_application_import_resource.py
JustinMacaulay: Added

services/core-api/app/api/services/nros_now_status_service.py
JustinMacaulay: Added more to the logger just to have this at least a bit more useful.

services/core-api/app/api/services/nros_now_status_service.py
```diff
@@ -0,0 +1,57 @@
...
+                'Content-type': 'application/json',
+                'lobName': 'MMS'
+            },
+            data=json.dumps(data))
+        current_app.logger.info(
+            f'Status for NoW Number{now_number} pushed to NROS with response: {resp.__dict__}')
+        return
+
+    @classmethod
+    def get_nros_now_token(cls):
```
exzizt, line 49: I think this function could use some error handling, e.g., to ensure that the status code received is 200 and if not throw an appropriate error message.

services/core-api/app/api/services/nros_now_status_service.py
```diff
@@ -41,17 +43,28 @@ def nros_now_status_update(cls, now_number, status, status_updated_date):
...
+            resp.raise_for_status()
+            resp_body = json.loads(resp.text)
 
-        resp_body = json.loads(resp.text)
-
-        token = resp_body["access_token"]
+            token = resp_body["access_token"]
+        except:
+            current_app.logger.info(f'Failed to get a token from NROS. Response: {resp.__dict__}')
+            pass
```
exzizt, line 68: I think the `pass` can be removed.

migrations/sql/V2020.05.08.12.51__update_now_statuses.sql
```diff
@@ -0,0 +1,29 @@
...
+VALUES
+	('REC', 'Received', 90, true, 'system-mds', 'system-mds'),
+    ('PAP', 'Pending Approval', 50, true, 'system-mds', 'system-mds'),
+	('REI', 'Rejected-Initial', 100, true, 'system-mds', 'system-mds'),
+	('PCO', 'Permit Closed', 40, true, 'system-mds', 'system-mds'),
+	('NPR', 'No Permit Required', 110, true, 'system-mds', 'system-mds'),
+	('RCO', 'Referral Complete', 120, true, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
+-- Set all of the current NoW statuses to the new default status (Submitted)
```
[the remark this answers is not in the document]
JustinMacaulay: removed

services/core-api/app/api/services/nros_now_status_service.py
```diff
@@ -0,0 +1,57 @@
...
+    nros_now_token_pass = Config.NROS_NOW_CLIENT_SECRET
+    nros_now_token_user = Config.NROS_NOW_CLIENT_ID
+    nros_now_token_url = Config.NROS_NOW_TOKEN_URL
+
+    @classmethod
+    def nros_now_status_update(cls, now_number, status, status_updated_date):
+
+        token = cache.get(NROS_NOW_TOKEN)
+        if not token:
+            token = cls.get_nros_now_token()
```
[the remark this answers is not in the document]
JustinMacaulay: Added more logging around around failing calls.

services/core-api/app/api/services/nros_now_status_service.py
JustinMacaulay: I dont think its needed but I put it in just to show it was a deliberate choice to ignore the error.

PR #1336, opened 2020-05-08
Connect MineSpace users to mines with BCeID username

mmumby:
- added ability to use either the email or username was connecting a mineSpace user to a mine
- Also fixed the bug where a white screen appears when navigating to the inspections tab if there are no orders.


services/core-web/src/components/Forms/AddMinespaceUser.js
```diff
@@ -30,21 +30,21 @@ export const AddMinespaceUser = (props) => (
         <Col>
           <Form.Item>
             <Field
-              id="user_bceid_email"
-              name="user_bceid_email"
-              label="Email*"
+              id="email"
+              name="email"
+              label="Email or BCeID username*"
```
pdittaro, line 35: We should indicate a preference here, email first BCeID second.

services/core-api/app/auth.py
```diff
@@ -42,8 +43,9 @@ def get_current_user():
     rv = getattr(g, 'current_user', None)
     if rv == None:
         email = get_user_email()
-        rv = MinespaceUser.query.unbound_unsafe().filter_by(email=email).filter_by(
-            deleted_ind=False).first()
+        username = get_user_username()
+        rv = MinespaceUser.query.unbound_unsafe().filter(
```
pdittaro, line 47: This is actually really clever. Though I think we should rename the 'email' column in the database to avoid confusion about the two types of values contained.

mmumby: I was thinking of renaming it, something generic like 'identifier' ?

PR #1338, opened 2020-05-26
MDS-2820: Bond Info on MineSpace

exzizt:
# Main
* Provides a Bonds tab on MineSpace for proponents to view their bond information (hidden from proponents for the time being)
image


services/core-api/app/api/securities/response_models.py
```diff
@@ -45,9 +45,26 @@
         'payer': fields.Nested(BOND_PARTY),
         'project_id': fields.String,
         'permit_guid': fields.String(attribute='permit.permit_guid'),
+        'permit_no': fields.String(attribute='permit.permit_no'),
         'documents': fields.List(fields.Nested(BOND_DOCUMENT_MODEL))
     })
 
+BOND_MINESPACE = api.model(
```
mmumby, line 52: Why does the MineSpace bond need a different model? How does it differ from the current bond?

exzizt: There was a concern about sending certain bond info to proponents or users with the "MineSpace proponent" role, particularly info about the bonds' institution data, notes, and documents. So I made the API use an entirely different model to strip this "sensitive" info from the bonds payload for proponents.

mmumby: ah, okay that's what I was thinking it was for. 👍

PR #1343, opened 2020-05-30
Hackathon follow up - Admin permission mode

mmumby:
- Added a ToolTip to all `AuthenticationWrapper`
- Tooltip is visible to admin user when hovering over any buttons wrapped in the wrapper (visible after 1 second of hovering) 
- Tooltip displays the permissions, if its visible for regional mines, and if it's visible only in Test
Screen Shot 2020-05-29 at 6 22 52 PM
Screen Shot 2020-05-29 at 6 22 17 PM

Note: Was looking for a way to pass props into child component, can do so with `React.createElement`, but this only passes props to the span and not the children - so not helpful, but it did solve the issue of `{props.children}` not rendering tooltip on all actions, for example this menu. (Open to suggestions)
Screen Shot 2020-05-29 at 6 24 16 PM


services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
```diff
@@ -19,42 +19,46 @@ class MinePartyApptResource(Resource, UserMixin):
     parser = CustomReqparser()
```
mmumby, line 19: This entire file is formatting... no changes were made

PR #1344, opened 2020-06-01
MDS 2854 add incident type to tables

JustinMacaulay:


services/core-web/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -168,6 +176,16 @@ export class MineIncidentTable extends Component {
           </div>
         ),
       },
+      {
+        title: "Incident Type(s)",
+        key: "incident_types",
+        dataIndex: "incident_types",
+        render: (text) => (
+          <div title="Commodity">
```
exzizt, line 184: Update div `title` to match column title.

JustinMacaulay: changed

PR #1346, opened 2020-06-01
MDS-2857: Core CRR Page on Mines Crashes (reported by Silvertip)

exzizt:
# Main
* MDS-2857: Fixes an issue where inactive CRR definitions weren't being sent to the FE but the FE was trying to render data for inactive CRR definitions (this issue caused a WSOD on Silvertip CRR page, as they had a report with definition ID 17 which is now considered inactive)
* MDS-2849: Adds new bond document type code: Change of Name Certificate
* MDS-2860: Adds multiple new bond document types, renames some, inactivates one, and creates/uses a "display order" column.

**Note**: We need to have a discussion at some point about how we only send over static content whose objects are considered "active" when the FE generally needs inactive and active in order to correctly render information.

# Other
* Updates the mine response model to return an empty object for verified_status and mine_location if they are null instead of an object with empty properties.


services/core-web/src/components/Forms/reports/AddReportForm.js
```diff
@@ -72,10 +72,12 @@ export class AddReportForm extends Component {
...
 
   updateMineReportDefinitionOptions = (mineReportDefinitionOptions, selectedMineReportCategory) => {
-    let mineReportDefinitionOptionsFiltered = mineReportDefinitionOptions;
+    let mineReportDefinitionOptionsFiltered = mineReportDefinitionOptions.filter(
+      (option) => option.active_ind
+    );
 
     if (selectedMineReportCategory) {
-      mineReportDefinitionOptionsFiltered = mineReportDefinitionOptions.filter(
+      mineReportDefinitionOptionsFiltered = mineReportDefinitionOptionsFiltered.filter(
```
JustinMacaulay, line 80: If your removing all of the inactive codes in the selectors will it not have the same issue as it did before?

exzizt: Not using this variable here was a logic error, as it was intended to be used here. A couple of lines above `mineReportDefinitionOptionsFiltered` removes all the inactive codes, which is wanted, then the options are filtered further by the selected category. We want to remove inactive codes in this case because we don't want them to be able to select them.
The issue that caused the WSOD was on `MineReportTable.js` but didn't require any code changes:
```
      render: (record) => (
        <div title="Code Section">
          {formatComplianceCodeValueOrLabel(
            props.mineReportDefinitionHash[record.mine_report_definition_guid]
              .compliance_articles[0],
            false
          )}
        </div>
```

PR #1349, opened 2020-06-05
MDS-2847 bond audit history

JustinMacaulay:
Adds a table and functions to store a bond record every time information on it is edited.


services/core-api/app/api/securities/resources/bond.py
```diff
@@ -89,6 +89,7 @@ def get(self, bond_guid):
     def put(self, bond_guid):
         #remove the amount from the request if it exists as it should not be editable.
         temp_bond = Bond.find_by_bond_guid(bond_guid)
+        temp_bond.save_bond_history()
```
exzizt, line 92: What if the below operations fail? We'll have incorrectly saved bond history for a failed edit operation. I think this should happen at the end after the bond has been successfully modified.

JustinMacaulay: If I want to save the history the way that I am with a class method the bond history save has to happen before the bond is modified

exzizt: I guess an option is to then delete this `temp_bond` object in the `catch` block.

JustinMacaulay: Temp_bond is just the bond before we modify it. So I dont want to delete it, the function lives on the bond model so it takes self and marshals it then loads it into the history table.

exzizt: I understand. Do we want to create a bond history record if the PUT operation fails?

services/core-api/app/api/securities/resources/bond.py
```diff
@@ -98,8 +99,11 @@ def put(self, bond_guid):
```
exzizt, line 99: The catch of this block should delete the bond history object too

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: I have changed this and it should be good now if you can take another look

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: it does

services/core-api/app/api/securities/models/bond.py
```diff
@@ -44,6 +48,17 @@ class _ModelSchema(Base._ModelSchema):
     def __repr__(self):
         return '<Bond %r>' % self.bond_guid
 
+    def save_bond_history(self):
+        bond_json = marshal(self, BOND)
```
Jsyro, line 52: can this use BondHistory._schema().dumps, this is currently using two different librarys for dumping and loading. Which is how the api works, but can you see if that's an easy change?

services/core-api/app/api/securities/resources/bond.py
exzizt: image
We need to do `history.delete()` between lines 97 and 98 too.

exzizt: I would also raise a specific error on 106, maybe InternalServerError with the exception message?

services/core-api/app/api/securities/models/bond.py
JustinMacaulay: I will try that

services/core-api/app/api/securities/resources/bond.py
JustinMacaulay: The second exception block around the save already raises a SQLAlchemy exception to I am just catching it again so I can delete the history and then re-raising the initial exception.

services/core-api/app/api/securities/models/bond_history.py
```diff
@@ -0,0 +1,43 @@
...
+    __tablename__ = "bond_history"
+
+    bond_history_id = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
+    bond_id = db.Column(db.Integer, db.ForeignKey('bond.bond_id'), nullable=False)
+    amount = db.Column(db.Numeric(14, 2), nullable=False)
+    bond_type_code = db.Column(db.String)
+    issue_date = db.Column(db.DateTime)
+    permit_no = db.Column(db.String)
+    permit_guid = db.Column(UUID(as_uuid=True))
+    payer = db.Column(db.String)
```
Jsyro, line 23: can this be payer_name, payer on the bond model is used and is different.

PR #1350, opened 2020-06-10
Set deleted mines as deleted based off 'REG_CD'

nayr974:
This small change will set 'deleted_ind' to true based off mmsmin.reg_cd, which is set to X when a mine is deleted in MMS.


nayr974:
> 
> 
> `like %delete% or like %deleted%` should be alright with just the first clause, right?

lol totally. removed the second

PR #1351, opened 2020-06-10
MDS-2533: Configure Document Manager to Use S3 Object Store

exzizt:
# Main
- Now our Document Manager can upload files to our object store (S3 bucket) as well as its filesystem (configurable via an environment variable). When files are requested it can correctly identify where the file lives (filesystem or object store) and retrieve it.
- Adds a TUSD container which this functionality depends on.

**Notes**:
- Might want to look into this for the future (for dealing with large S3 objects, zips, compression, etc.): https://github.com/RaRe-Technologies/smart_open
- This was a good thread and useful in figuring out how to stream objects from S3: https://github.com/boto/boto3/issues/426#issuecomment-184889230
  - We might want to do this in Dormant as well? There was a TODO that mentioned figuring out how to stream objects instead of downloading and sending the entire object in one go.
- After reading that our TUSD S3 service internally caches PATCH request chunks until it meets AWS multi-part upload requirements of 5MB (talked about here: https://godoc.org/github.com/tus/tusd/pkg/s3store#hdr-Considerations) I don't think my changes to the file upload chunk size (1MB --> 5MB) are necessary anymore, but I also feel like 1MB was too small.


services/core-web/src/components/common/FileUpload.js
```diff
@@ -30,8 +28,8 @@ const propTypes = {
 const defaultProps = {
   maxFileSize: "750MB",
   acceptedFileTypesMap: {},
-  onFileLoad: doNothing,
-  onRemoveFile: doNothing,
+  onFileLoad: () => {},
```
nayr974, line 31: how dare u

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -0,0 +1,41 @@
...
+        resp = Response(
+            generate(s3_response),
+            mimetype='application/pdf' if '.pdf' in display_name.lower() else 'application/zip',
+            headers={
+                'Content-Disposition':
+                ('attachment; ' if as_attachment else '') + ('filename=' + display_name)
+            })
+        return resp
+
+    # NOTE: Leaving this old method in here for reference and in case we want to use it in the future.
```
nayr974, line 33: we should probably just remove this, we have history if needed

PR #1360, opened 2020-06-18
Show `Edited` on NoW Field with original value

mmumby:
- Added functionality to compare current value with the previous value. If different show that the field has been edited with a tooltip containing the original value. 
- Handle situations where the value is a code, show the label to the user
- Handle situations where fields are undefined due to just being added to the NoW (Ie Activities), only show edited if the current value is also `!null`
- Handle situations where the previous value is null and have been changed, show "N/A" as the previous value in the tooltip
Screen Shot 2020-06-18 at 2 04 07 PM
Screen Shot 2020-06-18 at 2 04 25 PM


services/core-api/app/api/mines/mine/models/mine.py
```diff
@@ -23,44 +23,60 @@ class Mine(AuditMixin, Base):
     __tablename__ = 'mine'
```
mmumby, line 23: The majority of the changes on the BE files are formatting.

migrations/sql/V2020.06.08.13.30__add_updated_date_to_now_tables.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE now_application ADD COLUMN last_updated_by character varying(60);
+ALTER TABLE now_application ADD COLUMN IF NOT EXISTS last_updated_date timestamp with time zone DEFAULT NOW() NOT NULL;
```
exzizt, line 2: I don't think there should be a default value for this column and it should be nullable.

services/core-api/app/api/mines/mine/models/mine.py
exzizt: I wonder why. We should try to get the team using a consistent formatted :(
In VS Code you can do "Ctrl + K, Ctrl + S" to save without applying formatting.

services/core-api/app/api/now_applications/models/now_application.py
```diff
@@ -21,30 +23,46 @@ class NOWApplication(Base, AuditMixin):
...
     now_application_status_code = db.Column(
         db.String,
         db.ForeignKey('now_application_status.now_application_status_code'),
         nullable=False)
-    status_updated_date = db.Column(db.Date, nullable=False, server_default=FetchedValue())
+    status_updated_date = db.Column(db.Date,
+                                    nullable=False,
+                                    server_default=FetchedValue())
+    last_updated_date = db.Column(db.Date,
+                                  default=datetime.utcnow,
```
exzizt, line 61: I don't think `last_updated_date` and `last_updated_by` should have default values and they should be nullable because NoW's can exist without having been updated.

services/core-api/app/api/now_applications/models/now_application.py
```diff
@@ -139,11 +187,17 @@ def __repr__(self):
...
+            now_application_id=now_application_id).first()
 
     @classmethod
     def validate_guid(cls, guid, msg='Invalid guid.'):
         try:
             uuid.UUID(str(guid), version=4)
         except ValueError:
             raise AssertionError(msg)
+
+    def save(self, commit=True):
```
exzizt, line 200: Is this function necessary? I think the `onupdate=...` you defined in the model should do this for you already.

migrations/sql/V2020.06.08.13.30__add_updated_date_to_now_tables.sql
mmumby: I'll remove the default! I agree

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -180,6 +182,30 @@ export class NoticeOfWorkApplication extends Component {
     });
   };
 
+  renderOriginalValues = (path) => {
+    const prevValue = get(this.props.originalNoticeOfWork, path);
+    const currentValue = get(this.props.noticeOfWork, path);
+    // only show edited if user has updated the field
+    const isNewValue = isUndefined(prevValue) && !isNull(currentValue);
```
exzizt, line 189: Isn't `isNewValue = prevValue !== currentValue` enough and more correct? Why does the previous value have to be undefined and the current value can't be null? Data can be changed to null and that is considered a new value.

services/core-api/app/api/mines/mine/models/mine.py
mmumby: I thought we did have consistent formatting in our settings.JSON? Im wondering if mine got removed for some reason

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
mmumby: ah, great question Luke! 

When a new activity is added, the fields are not registered in the initialValues, so by default they are undefined, and without that check - all new fields added to the form automatically say "edited" although they are brand new to the form. 

These are the situations I have identified: 
undefined - because they do not exist in the original application (isNewValue = prevValue !== currentValue works for this case)
undefined - because they are net new fields - the above check doesn't work in this situation as previous the check would be undefined !== null, which is true. So an additional check was added, does current === a value

mmumby: also, a good point that the current value can be changed back to null. 

The issue with that, is we do not capture any history. So if the original was 'null' || 'undefined', and the value updated to "directions to site", the FE would show "edited", but if the field was changed to null, we would be comparing the null to very first value of null/undefined which would show that the field wasn't edited. 

Following? There are so many edge cases with this one, but without history we have no way to capture if it was filled in and then changed back. 

Now if the original had a value, and is now null.... I dont handle that correctly. I'll add an additional case for that situation

services/core-web/src/components/noticeOfWork/applications/review/ReviewNOWApplication.js
```diff
@@ -618,6 +745,14 @@ export const ReviewNOWApplication = (props) => {
       <Form layout="vertical">
         <div className="side-menu--content">
           <h2>General Information</h2>
+          {props.noticeOfWork.last_updated_date && (
+            <p className="violet">
+              Last Updated: {formatDate(props.noticeOfWork.last_updated_date)}
```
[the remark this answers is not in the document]
mmumby: There is an additional check if null do not show the date.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
exzizt: Sounds good. That sounds wrong that we don't capture any history if the value is set back to null.

mmumby: It does capture the audit history of the most current date/person. But it doesn't track every individual field, we only have access to original values and current values and the last time the entire form was edited.

PR #1361, opened 2020-06-18
Update Bond View

exzizt:
* I need the bond type and status code descriptions in the view


pdittaro:
Good point to note on this one is that it looks like you can't use CREATE OR REPLACE with views in Postgres if you're going to be removing columns. Makes it not a great candidate for repeatable migrations.

@nayr974

PR #1368, opened 2020-06-23
Docman conditional object store per environment + fixes

pdittaro:


exzizt:
I think we should update the docman env-example to have OBJECT_STORE_ENABLED set to 0 as well for local development.

PR #1371, opened 2020-06-25
MDS-2585: Encrypt Connection to NRIS Database

mhuseinov:


pdittaro:
If we can we should load the certificate as a configmap from openshift

pdittaro:
Closing to deal with repo issues.

PR #1372, opened 2020-06-26
VSCode pylint and virtualenv setting

pdittaro:


pdittaro:
Unnessecary with Workspace settings

PR #1373, opened 2020-06-29
Update NoW nagivation

mmumby:
- Removed progress tracking from component files - redux still exists to be reused later
- removed `Steps` and added `Tabs`
- Verfication Tab is in place until removed in a follow up ticket, at which point the default tab will be "technical review"
- Added 'Administrative' Tab, follow up ticket will populate that view with more data.
- Updated Fixed header logic, the Tabs function differently than steps - added a different header that stays fixed on scroll for "technical review" tab, the same logic can later be applied to the "draft permit" tab
Screen Shot 2020-06-29 at 9 07 14 AM
Screen Shot 2020-06-29 at 9 08 43 AM

- Added a null screen to "draft permit" tab if there is no permittee on the NoW - previously this caused a white screen 
Screen Shot 2020-06-29 at 9 09 03 AM

Paths to get to this view: 
- NoW edit, 
- NoW view, 
- create MM permit application


pdittaro:
Some things I've noticed:
- If you're on a tab and you click on the tab again it shows the loading spinner indefinitely
- We shouldn't be using H1 headers in different locations around the page - just at the top. Advise that the tab header be de-emphasized.
- In facing the issue of no assigned permittee, there isn't a clear way to add a permittee from the draft permit page

mmumby:
I'll take a look at the bug you found with tabbing, and update the headers. 

In regards to the permittee, We currently have no way to add a permittee, they should be getting carried over from the NoW application when we import but we have yet to build that functionality. 
Adding the null screen was a quick fix to the current bug on test when there is no permittee, if you click the "draft permit" button the user will get a white screen. It is not a long term solution, it just allows this view to be present within the tab navigation structure without causing a white screen.

pdittaro:
Thanks for the fixes, @mmumby.

PR #1374, opened 2020-06-30
Tracking specific page views for Matomo

pdittaro:


mmumby:
Is this supposed to be added to every route on the app?

pdittaro:
> Is this supposed to be added to every route on the app?

I'd like to come up with a better way of capturing it based on the router, but for now this is just to start getting information in.

PR #1383, opened 2020-07-15
MDS-2929: Implement Bond Closed Date/Notes (Modify Bond Confiscate/Release Flow)

exzizt:
# Main
* Modifies the bond release/confiscation flow by allowing users to set the release/confiscation date as well as provide release/confiscation notes. **Note**: Adding release/confiscation documents is supported in the Edit Bond modal after the bond has been closed. This was decided because it would add a lot more scope if we allowed the user to select/upload multiple release/confiscation document types in the modal.
* Adds the columns `closed_date` and `closed_note` to the tables `bond` and `bond_history` and updates the `bond_view` to account for these new columns. **Note**: Securities reporting will be addressed in another ticket to account for these new columns.
* Updates the View Bond Modal by displaying the closed date/notes if the bond is closed.
* Updates the Edit Bond Modal by allowing the editing of the closed date/notes if the bond is closed. **Note**: If the bond is confiscated, the bond type cannot be edited, as changing the bond type to cash when the bond was confiscated was an enforced business rule.

**Note**: The language for the closed date/notes changes dynamically depending on whether the bond was released or confiscated.
**Note**: Existing bond and bond history entries, if they are released or confiscated, have their closed date set to the update timestamp.

## Screenshots
image
image
image
image

# Other
* Moves the Notes field in the Edit Bond Modal up to the other general information area rather than the Institution area

### DIFF
--- a/migrations/sql/V2020.07.15.08.33__update_bond_view_with_closed_date_and_notes.sql
+++ b/migrations/sql/V2020.07.15.08.33__update_bond_view_with_closed_date_and_notes.sql
@@ -1,0 +1,79 @@
+DROP VIEW IF EXISTS bond_view;
+CREATE VIEW bond_view AS
+SELECT
+    *
+FROM
+    (
+    SELECT
+        bond.bond_id,
+        amount,
+        bond.bond_type_code,
+        bond_type.description as "bond_type_description",
+        bond.bond_status_code,
+        bond_status.description as "bond_status_description",
+        issue_date,
+        note,
+        closed_date,
+        closed_note,
+        payer_party_guid,
+        party_name AS "payer_name",
+        institution_city,
+        institution_name,
+        institution_postal_code,
+        institution_province,
+        institution_street,        
+        permit_guid,
+        permit_no,
+        project_id,
+        reference_number,
+        bond.update_timestamp,
+        bond.update_user,
+        'Yes' AS "is_current_record"
+    FROM
+        bond,
+        party,
+        permit,
+        bond_permit_xref,
+        bond_type,
+        bond_status
+    WHERE
+        bond.payer_party_guid = party.party_guid
+        AND permit.permit_id = bond_permit_xref.permit_id
+        AND bond.bond_id = bond_permit_xref.bond_id
+        AND bond.bond_type_code = bond_type.bond_type_code
+        AND bond.bond_status_code = bond_status.bond_status_code        
+    UNION ALL 
+    SELECT
+        bond_id,
+        amount,
+        bond_history.bond_type_code,
+        bond_type.description as "bond_type_description",
+        bond_history.bond_status_code,
+        bond_status.description as "bond_status_description",
+        issue_date,
+        note,
+        closed_date,
+        closed_note,        
+        payer_party_guid,
+        payer_name,
+        institution_city,
+        institution_name,
+        institution_postal_code,
+        institution_province,
+        institution_street,        
+        permit_guid,
+        permit_no,
+        project_id,
+        reference_number,
+        bond_history.update_timestamp,
+        bond_history.update_user,
+        'No' AS "is_current_record"
+    FROM
+        bond_history,
+        bond_type,
+        bond_status
+    WHERE
+        bond_history.bond_type_code = bond_type.bond_type_code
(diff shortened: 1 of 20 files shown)


exzizt:
Do we want to add the scope of adding a "Closed Date" column to the bond tables? It will be N/A for all active bonds and show the closed date for confiscated/released bonds.

migrations/sql/V2020.07.15.08.33__update_bond_view_with_closed_date_and_notes.sql
```diff
@@ -0,0 +1,79 @@
+DROP VIEW IF EXISTS bond_view;
+CREATE VIEW bond_view AS
```
nayr974, line 2: What is bond_view used for? In general we have chosen to avoid views unless there is a constraint or performance concern that makes them a significantly better choice than working in our ORM

services/core-web/src/components/Forms/Securities/BondForm.js
```diff
@@ -29,12 +29,14 @@ const propTypes = {
   closeModal: PropTypes.func.isRequired,
   title: PropTypes.string.isRequired,
   submitting: PropTypes.bool.isRequired,
+  pristine: PropTypes.bool.isRequired,
```
nayr974, line 32: I think we've used the flag is_dirty or 'dirty' to indicate if an object has been changed, can you take a look? If so, we should keep this consistent. If not, pristine is fine.

services/core-web/src/tests/components/Forms/Securities/BondForm.spec.js
```diff
@@ -17,6 +17,8 @@ const setupProps = () => {
   props.submitting = false;
   props.provinceOptions = MOCK.DROPDOWN_PROVINCE_OPTIONS;
   props.bondTypeOptions = [];
+  props.bondStatusOptionsHash = MOCK.BULK_STATIC_CONTENT_RESPONSE.bondStatusOptions;
+  // eslint-disable-next-line prefer-destructuring
```
nayr974, line 21: Is there a reason to not destructure here? If you disable a linting rule, please put a comment as to why.

migrations/sql/V2020.07.15.08.33__update_bond_view_with_closed_date_and_notes.sql
exzizt: The bond view was created to support securities reporting, to almagate the bond and bond_history tables for easier querying. If you take a look at the securities dashboard reports you can see how it's used.
https://github.com/bcgov/mds/pull/1359

services/core-web/src/components/Forms/Securities/BondForm.js
exzizt: Pristine is the opposite of Dirty, its redundant to have two but that's what Redux Forms does. I've noticed we've used Pristine throughout the app already and don't use Dirty.

services/core-web/src/tests/components/Forms/Securities/BondForm.spec.js
exzizt: I googled it briefly to figure out what it was complaining about but I couldn't figure it out exactly. I didn't even add the line it's complaining about so I'm not entirely why it started to complain about it.

nayr974: Can you please add a comment as such?

migrations/sql/V2020.07.15.08.33__update_bond_view_with_closed_date_and_notes.sql
nayr974: sounds good

services/core-web/src/components/Forms/Securities/BondForm.js
nayr974: sounds good

services/core-web/src/tests/components/Forms/Securities/BondForm.spec.js
exzizt: Apparently this is the fix for it. So silly
`
[props.bond] = MOCK.BONDS.records[0];
`

PR #1386, opened 2020-07-18
Fix Loading Bar Issues

exzizt:
I have a [PR](https://github.com/mironov/react-redux-loading-bar/pull/104) for the react-redux-loading-bar repo but in the meantime, we can import the source for the LoadingBar and related classes into our project. When and if the PR gets in we can resort to using the package for the modal wrapper's loading bar (the navbar still uses the package). I resolved the issue by using an alternate animation approach (using width as opposed to translate3d). I also fixed an issue where the width being set in the navbar's loading bar which was preventing it from functioning.

## Old
image

## New 0 (current)
_3 px with background_
image

## New 1
_8 px with background_
image

## New 2
_8 px with no background (close to original style)_
image

## Scrap it all together and use Ant Design's button loading effect (I personally lean towards this)
mds_core_modal_button_submitting


exzizt:
> OK

⭕ 🇰🇿

PR #1399, opened 2020-07-29
MDS-2967 nris data fix

JustinMacaulay:
Issue with the defined lengths on strings in the NRIS API. This PR just changes all of the string lengths to an arbitrarily large number due to the fact that alembic cannot detect changes to columns when going from a string length constraint to no string length constraint.


services/nris-api/backend/app/extensions.py
```diff
@@ -10,7 +9,7 @@
 
 apm = ElasticAPM()
 db = SQLAlchemy()
-migrate = Migrate()
+migrate = Migrate(compare_type=True)
```
nayr974, line 12: interesting that this is needed. should be default IMO?

services/nris-api/backend/app/nris/models/attendee_type.py
```diff
@@ -5,10 +5,12 @@
...
     __tablename__ = "attendee_type"
     __table_args__ = {
-        'comment': 'Lookup table that provides types of Attendees on an Inspection; i.e. Inspector, Management, FirstNations.'}
+        'comment':
+        'Lookup table that provides types of Attendees on an Inspection; i.e. Inspector, Management, FirstNations.'
+    }
 
     attendee_type_id = db.Column(db.Integer, primary_key=True)
-    attendee_type = db.Column(db.String(256))
+    attendee_type = db.Column(db.String(10485760))
```
nayr974, line 13: coo coo

services/nris-api/backend/app/extensions.py
JustinMacaulay: As far as Jason and I could find yesterday it is not.

PR #1406, opened 2020-08-08
Feature/conditions

mmumby:
- Adding template conditions for Sand and Gravel permits to the creation of New Permits.


migrations/sql/R__stored_procedure_for_mine_status_information.sql
```diff
@@ -140,4 +140,4 @@ CREATE OR REPLACE FUNCTION transfer_mine_status_information() RETURNS void AS $$
             RAISE NOTICE 'Finish updating mine status in MDS';
         END;
     END;
-$$LANGUAGE plpgsql;
+$$LANGUAGE plpgsql;
```
mmumby, line 143: Had merge conflicts in this file for some reason, so I copied directly from dev, not sure why there is a dif here.

migrations/sql/V2020.08.04.15.09__add_template_permit_conditions.sql
```diff
@@ -0,0 +1,104 @@
+DROP SEQUENCE standard_permit_conditions_standard_permit_condition_id_seq cascade;
+
```
mmumby, line 2: Removing the Serial from PK, this appears to work fine, but please advise on a better approach if there is one.

mmumby:
This PR is actually quite annoying as the majority of my "changes" are formatting, questioning what happened to my formatter. Will look into it on Monday.

PR #1407, opened 2020-08-10
Add Template conditions to "New Permits"

mmumby:
- Seeded template conditions for "Sand and Gravel"
- Ability to automatically add  template conditions to draft permits when created
- Only get added to "New Permits", templates are not added to "Amendments"
- Template conditions are treated as normal conditions on the UI, they can be deleted, etc.
Screen Shot 2020-08-10 at 10 42 51 AM


services/core-api/app/api/mines/permits/permit/resources/permit.py
```diff
@@ -111,6 +128,24 @@ def post(self, mine_guid):
         db.session.add(permit)
         db.session.add(amendment)
 
+        now_application_guid = data.get('now_application_guid')
```
exzizt, line 131: I think we should raise a Bad Request error if `now_application_guid` is not provided.

mmumby: This endpoint works without a now_guid - should we still throw a 'Bad Request error` given that the majority of the POSTS do not have a now_guid?

exzizt: I didn't realize that. If it's an optional parameter then it's all good!

PR #1410, opened 2020-08-13
MDS 2979 add query param to mines for verified only in return

JustinMacaulay:
Added a new query param to the GET /api/mines that allows for filtering by verified mines.


JustinMacaulay:
image

PR #1415, opened 2020-08-17
Feature/MDS 2893 - Permit Condition Updates

nayr974:
This PR adds step's (A, B, C, D) to the permit condition categories, as well as shows the number of conditions under each section in the UI.


migrations/sql/V2020.08.11.09.05__add_letters_to_condition_sections.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE permit_condition_category ADD COLUMN step varchar;
```
exzizt, line 1: Should we add a not-null constraint?

nayr974: For sure, I'll add a not null

nayr974: Updated

PR #1419, opened 2020-08-19
Re-ordering conditions once they are deleted

mmumby:
- Updated BE to re-order conditions after they are deleted.


services/core-api/app/api/mines/permits/permit_conditions/resources/permit_conditions_resource.py
```diff
@@ -96,4 +96,22 @@ def delete(self, mine_guid, permit_guid, permit_amendment_guid, permit_condition
 
         permit_condition.save()
 
+        if permit_condition.parent_permit_condition_id is not None:
+            parent = PermitConditions.find_by_permit_condition_id(
+                permit_condition.parent_permit_condition_id)
```
Jsyro, line 101: Why don't we just traverse 'permit_condition.parent_condition'. Shouldn't need a whole lookup.

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_conditions_resource.py
```diff
@@ -96,4 +96,22 @@ def delete(self, mine_guid, permit_guid, permit_amendment_guid, permit_condition
...
+            sorted_conditions = sorted(parent.sub_conditions, key=lambda x: x.display_order)
+            for i, condition in enumerate(sorted_conditions):
+                condition.display_order = i + 1
+            parent.save()
+        else:
+            conditions = PermitConditions.find_all_by_permit_amendment_id(
+                permit_condition.permit_amendment_id)
```
Jsyro, line 108: Might want to load permit_amendment then get conditions member. but i suppose they are equivalent.

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_conditions_resource.py
nayr974: .parent didn't seem accessible outside of the scope of the model itself. Maybe a get_parent method on the model might work

Jsyro: PermitCondition should get a relationship to it's parent (just like relationship to it's set of children). Also, that model should be singular.

PR #1420, opened 2020-08-20
Update "Total Assessed" To sum all permit_amendment security totals

mmumby:
-Update "Total Assessed" value on the Bond table To sum all permit_amendment security totals vs pulling from the most recent
- "Security Total" was not being saved on permit_amendment POST, updated API to save that value
Screen Shot 2020-08-20 at 1 08 53 PM

Permit P-4365246 in the image has two security totals of "2, 000, 000" and "1, 000, 000"
while M-1535797 has one security_total of "2, 000, 000"


services/core-api/app/api/mines/permits/permit_conditions/models/permit_conditions.py
```diff
@@ -22,30 +22,22 @@ class _ModelSchema(Base._ModelSchema):
         permit_condition_guid = fields.UUID(dump_only=True)
```
mmumby, line 22: Nothing was changed in this file, this was just formatting.

services/core-api/app/api/mines/permits/permit_amendment/resources/permit_amendment.py
```diff
@@ -132,6 +133,7 @@ def post(self, mine_guid, permit_guid, permit_amendment_guid=None):
             permit_amendment_type_code=permit_amendment_type_code
             if permit_amendment_type_code else 'AMD',
             description=data.get('description'),
+            security_total=data.get('security_total'),
```
Jsyro, line 136: given this field is actually the delta of the security on the permit, we should change it's name, or at least leave a comment

mmumby: Added a comment to the modal

PR #1425, opened 2020-08-25
Misc permit bugs

mmumby:
- Merge conflicts passed the wrong function to the delete condition button, updated to pass the correct function 
- Filtered out permits with status_code == 'D' from get permit by mine guid
- Updated Permit table to use permitStatusOptionsHash to follow other patterns for showing code/labels on the UI
- Added draft status to permit status options table


services/core-api/app/api/mines/permits/permit/resources/permit.py
```diff
@@ -23,55 +23,42 @@ class PermitListResource(Resource, UserMixin):
     parser.add_argument(
```
mmumby, line 23: This file is all formatting changes.. 

my settings.json is the same as the mds/settings.json. Not sure why this is happening

```
  "python.formatting.yapfArgs": [
    "--style={column_limit:100, spaces_before_comment:'15,30,50', SPLIT_BEFORE_EXPRESSION_AFTER_OPENING_PAREN:TRUE}"
  ],
```

migrations/sql/V2020.08.25.13.45__add_draft_status_to_permit_status.sql
```diff
@@ -0,0 +1,11 @@
...
+    (
+    permit_status_code,
+    description,
+    display_order,
+    create_user,
+    update_user
+    )
+VALUES
+    ('D', 'Draft', 20, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
```
nayr974, line 11: This was already added in the 2020.07.09 draft_permit sql script

mmumby: Okay, I didn't notice that. 
I think it was being overwritten by the script in the seed file, I wasn't able to see this status in my db thats why I added it. 

I will remove this file

PR #1427, opened 2020-08-26
Feature/permit condition reordering

nayr974:
This PR adds up and down arrows to re-order permit conditions.


services/core-web/src/components/Forms/permits/conditions/SubCondition.js
```diff
@@ -76,12 +78,41 @@ const SubCondition = (props) => {
...
+                  type="primary"
+                  onClick={() => {
+                    props.reorderConditions(props.condition, true);
+                  }}
+                  disabled={props.editingConditionFlag}
+                >
+                  <Icon type="up" theme="outlined" />
+                </Button>
+              </AuthorizationWrapper>
+              <AuthorizationWrapper permission={Permission.ADMIN}>
```
mmumby, line 98: This isn't specific to this PR, but curious what permissions should be on these actions. 

I think edit_permits would be better than admin. I'll check with Brian/Aaron

PR #1429, opened 2020-08-27
Feature/mds 3025

mmumby:
- Updated disclaimers on security UI components. 
- Adding screenshots to all disclaimers here, and showing to Brian/Aaron for any final changes to make sure everything is consistent and clear (including components that were not updated in this PR)

**NoW flow:** 
Screen Shot 2020-08-27 at 11 05 45 AM
Screen Shot 2020-08-27 at 11 05 59 AM

**Bond table on Mine Profile**
Screen Shot 2020-08-27 at 11 12 07 AM

Screen Shot 2020-08-27 at 11 07 54 AM
Screen Shot 2020-08-27 at 11 08 06 AM
Screen Shot 2020-08-27 at 11 08 25 AM

**Permit Amendment Modal**
- Also added security_received date field for historical permits (as per discussion in sprint planning)
Screen Shot 2020-08-31 at 3 24 51 PM

**Ant design table bug**
When adding a tooltip to the table headers, there is a bug that appends a column to the beginning of the table. (This column 'Total Assessed' is not duplicated in the code)
1 solution to address this is to change the MineBondTable to use an ant design `Table` vs the `CoreTable` - I did not like this approach since it removes some consistency among our tables, 
Another solution I found was to add a class that removes the extra column - I went with this approach for now.
Pictures below show the bug before and after the class 'additional-table-header' was added.

Screen Shot 2020-08-27 at 11 35 41 AM
Screen Shot 2020-08-27 at 11 36 49 AM

**Negative currency**
added flag to allow negative numbers on currencyMask, you must enter a number first and then include the negative
Screen Shot 2020-09-01 at 9 03 49 AM


pdittaro:
Were you also going to change it so that the Security Assessment value for permit amendments could be negative?

mmumby:
Uodated wording: "Changes to this value in CORE will not be updated in MMS." on Security Assessment field

mmumby:
> Were you also going to change it so that the Security Assessment value for permit amendments could be negative?

Added the ability to allow negatives!

PR #1430, opened 2020-08-28
MDS 3016 Fix securities project ID

JustinMacaulay:
Moved the Project Id on the bond records to permit.
- When changing the project id on one bond the id is changed on all and the permit.
- The project id now lives at the top level of the bond reclamation invoice and bonds tables.
- When changing the project id a warning is displayed to make sure you know you are changing all of them.
- The project ID is pre-populated on bonds and reclamation invoices now.


Jsyro (changes requested):
Looks good, can you write a test or two, specifically around editing the permit.project_no through the bond PUT endpoint?

Jsyro (approved):
Thanks for adding that test. looks great.

PR #1438, opened 2020-09-03
Feature/Fix-Permits-Not-Being-ETL'd

nayr974:


mmumby (approved):

pdittaro (approved):
Thanks for the fix

pdittaro (changes requested):
Actually, this will do a weird side effect if we build out a new environment from scratch. I think its alright if we just get rid of that migration and then add the column as part of the stored procedure.

migrations/sql/V2020.09.03.14.10__etl_permit_security_adjustment.sql
```diff
@@ -1,2 +0,0 @@
-ALTER TABLE ETL_PERMIT
-ADD COLUMN IF NOT EXISTS security_adjustment numeric;
```
Jsyro, line 2: Migration got removed?

pdittaro: Yep

pdittaro: It hasn't run in Dev/Test/Prod

pdittaro (approved):

PR #1454, opened 2020-09-17
Linting errors from antd 4 update

mmumby:
Hopefully, I have removed all linting issues that were merged into develop

I saved every file that was updated in the antd PR, and let the commit hooks do their thing, addressed any concerns that were raised.


nayr974 (approved):
image

PR #1455, opened 2020-09-17
Fix bond document_date bug; Fix party dropdown not loading intital value

nayr974:


services/core-web/src/components/common/PartySelectField.js
```diff
@@ -84,6 +84,7 @@ export class PartySelectField extends Component {
 
   componentDidMount() {
     if (this.props.initialValue) {
+      alert(JSON.stringify(this.props.initialValue));
       this.handleSearch(this.props.initialValue.label);
```
mmumby, line 88: Should this be removed?

nayr974: Sure as heck should

mmumby (dismissed):
🚀

PR #1457, opened 2020-09-21
Update search bar behavior; Rename top level menu items; Fix bug assigning inspectors

nayr974:


services/core-web/src/components/common/RenderGroupedSelect.js
```diff
@@ -50,10 +51,6 @@ const RenderGroupedSelect = (props) => (
       showSearch
       placeholder={props.placeholder}
       optionFilterProp="children"
-      filterOption={(input, option) =>
```
JustinMacaulay, line 53: Are these the changes that fixed the issue with the inspector selection? Also is there no filtering needed on this common component for other sections of the application where it's used?

JustinMacaulay (commented):
I just had the one comment but everything else looks good to me.

services/core-web/src/components/common/RenderGroupedSelect.js
nayr974: Yes. The filtering happens natively so I don't know if this is a workaround that was needed but everything works as expected now.

JustinMacaulay: Sounds good!

JustinMacaulay (approved):
All looks good to me!

Jsyro:
Looks awesome, Ryan. 

There is an existing bug that i wanted to raise. if you type in the quick search bar, then try and click to move the cursor, the field gets cleared, doesn't have to be addressed here, but it might be a quick fix.

Jsyro (approved):
Thanks for fixing the onclick.

PR #1459, opened 2020-09-21
MDS-3072 Change 'assessed value' ETL to be the sum of the held bonds on permit 

pdittaro:
### Assessed Value Totals
This PR addresses an issue where the Total Assessed liability value wasn't being totalled correctly because of the relationship between Permit, Permit Amendments, Mines, and Bonds. Previously we were calculating the assessed liability as the sum of all the permit amendment security adjustments in the front-end; this is is constrained by only retrieving permit amendments for a specific mine number. Instead the assessed value total is now being calculated by the backend across all permit-amendments, regardless of mine number. Similarly moved the total held bond amount and confiscated bond amount as properties in this fashion.
image
image

### Initializing Assessed Liability Values
This PR removes the first attempt to normalize assessed values by reading the 'bond_inc_amt' from the 'mms_streamnow' table. The data in this table didn't accurately reflect the state of these permits.

As part of this PR there is also a stored procedure which sets the permit assessed value total to be the sum of held and confiscated bonds. It achieves this by setting the adjustment amount of the most recent permit amendment on that permit to be equal to the value of said bonds. This was created as a utility to help baseline these amounts for the administrators. All values are editable and can be corrected after the fact if needed. It should only be run once before we go live with securities, and then should no longer be needed.


services/core-web/src/components/mine/Securities/MineSecurityInfo.js
```diff
@@ -94,23 +94,26 @@ export class MineSecurityInfo extends Component {
...
     ).length;
 
-  getSum = (status, permit) =>
-    this.props.bonds
-      .filter(
-        ({ bond_status_code, permit_guid }) =>
-          bond_status_code === status && permit_guid === permit.permit_guid
-      )
-      .reduce((sum, bond) => +sum + +bond.amount, 0);
+  // getSum = (status, permit) =>
```
mmumby, line 97: remove commented code

services/core-web/src/components/mine/Securities/MineReclamationInvoiceTable.js
```diff
@@ -207,7 +207,7 @@ export const MineReclamationInvoiceTable = (props) => {
     permits.map((permit) => {
       return {
         key: permit.permit_guid,
-        amount_confiscated: props.getSum("CON", permit),
+        amount_confiscated: props.getTotalConfiscatedSum(permit),
         amount_spent: props.getAmountSum(permit),
```
mmumby, line 211: if `permit.confiscated_bond_total` is on the permit, you can just reference that here 

`amount_confiscated: permit.confiscated_bond_total`

services/core-web/src/components/mine/Securities/MineBondTable.js
```diff
@@ -287,8 +284,8 @@ export const MineBondTable = (props) => {
       return {
         key: permit.permit_guid,
         total_bonds: props.activeBondCount(permit),
-        amount_confiscated: props.getSum("CON", permit),
-        amount_held: props.getSum("ACT", permit),
+        amount_confiscated: props.getTotalConfiscatedSum(permit),
+        amount_held: props.getTotalHeldSum(permit),
         total_assessed: props.getTotalAssessedSum(permit),
```
mmumby, line 289: same here: `amount_held: permit.active_bond_total`

services/core-web/src/components/mine/Securities/MineReclamationInvoiceTable.js
pdittaro: Thats a fair point, I just wanted to abstract the implementation in case we go with something different in the future - but that's easy enough to change down the road.

docker-compose.yml
```diff
@@ -149,8 +149,10 @@ services:
     container_name: tusd
     build:
       context: services/tusd
-    ports:
-      - 1080:1080
+    # ports:
```
nayr974, line 152: Remove?

nayr974 (commented):
image

services/core-web/src/components/mine/Securities/MineSecurityInfo.js
pdittaro: Removed

services/core-web/src/components/mine/Securities/MineReclamationInvoiceTable.js
pdittaro: Updated

services/core-web/src/components/mine/Securities/MineBondTable.js
pdittaro: Updated

docker-compose.yml
pdittaro: Removed

mmumby (approved):
Great work :)

PR #1463, opened 2020-09-24
Fix Reclamation Invoice Document Bug

nayr974:
This PR fixes a bug saving reclamation documents, and also disable "virtual scrolling" in the Ant design select component, which does not show a scroll bar and confuses users.


services/core-web/src/components/common/RenderLargeSelect.js
```diff
@@ -28,7 +28,7 @@ const propTypes = {
   handleSelect: PropTypes.func,
 };
 
-const doNothing = () => { };
+const doNothing = () => {};
```
exzizt, line 31: 😠

exzizt (approved):
💥

PR #1467, opened 2020-09-29
MDS-2719 - Graceful pre-kill for Postgres instances

nayr974:


pdittaro (approved):
Tested by shutting down a db pod manually and observed shutdown behaviour as expected. Good work.

PR #1468, opened 2020-09-30
more to rerunnable and add ses records with easy conversions.

Jsyro:
Moved this to a new re-runnable file.


exzizt (approved):
Looks good

PR #1471, opened 2020-10-02
Bond/add ses to view

Jsyro:
Flyway doesn't have access to foreign schema until after first run.
instead of direct access, staging table was built and is used in union on bond_view (used in aggregation in Metabase). 
refresh job is run nightly. 
Table is created in a sequential migration for view creation.


pdittaro:
Does this solve the issue posited in your description?

exzizt (dismissed):
✔️ ✔️ ✔️

Jsyro:
@pdittaro Now it does, needed to add a sequential migration as well.

pdittaro (approved):
unnamed

PR #1474, opened 2020-10-02
MDS-3075 Bond ETL changes to fix metabase reporting.

JustinMacaulay:
Adds a new section to the bond ETL to insert records into the bond history table so that the Metabase reports can function correctly. Also fixed the prod Metabase reports to function properly and run off the correct variables.


exzizt (approved):
Looks good, Justin! Nice work.

exzizt (approved):

exzizt (approved):

PR #1475, opened 2020-10-05
Remove NoW view only view 

mmumby:
-Remove NoW view only view 
- Update all actions inside the now view to have "edit_permits" permissions
- Add application pdf to "notice_of_work_view" to display the application in the table if it exists
- Remove the pencil button and replace with "Open"
Screen Shot 2020-10-05 at 10 51 31 AM


exzizt (approved):
Looks good. Nice work, Meredith!

PR #1481, opened 2020-10-07
MDS-3090 duplicate permitee creation

JustinMacaulay:
Turns out this was a much smaller and easier problem than I thought. Every night we purge the PMT from our mine_party_appt table but this was done using the mine_guid on the mine_party_appt to ensure it isn't for a major mine. Now the delete joins the mine_party_appt to permit to get the mine_guid so it can function as expected.


exzizt (approved):
Well, I'll be danged. Nice job finding the fix ✔️

Jsyro (approved):
Thanks Justin.

PR #1483, opened 2020-10-08
MDS-fix for sand and gravel template section conditionla rendering

mhuseinov:


exzizt (approved):
Looked at the docx as well, looks good.

mmumby (approved):

PR #1484, opened 2020-10-09
MDS-2930: add historical amendments to the permit

mhuseinov:
image
mds


services/core-web/src/components/Forms/PermitAmendmentForm.js
```diff
@@ -16,6 +16,13 @@ import PermitAmendmentFileUpload from "@/components/mine/Permit/PermitAmendmentF
 const originalPermit = "OGP";
 const amalgamtedPermit = "ALG";
 
+const ROLES_ALLOWED_TO_CREATE_HISTORICAL_AMENDMENTS = [
```
mmumby, line 19: Can you move this to the permissions.js file to keep all roles in one spot?

mmumby: Also, should this be handled through a group on keycloak? @pdittaro thoughts?

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -136,6 +136,11 @@ def find_by_permit_id(cls, _id):
     def find_by_now_application_guid(cls, _id):
         return cls.query.filter_by(now_application_guid=_id).first()
 
+    @classmethod
+    def find_original_permit_amendment_by_permit_guid(cls, _guid):
+        return cls.query.filter_by(permit_guid=_guid).filter_by(
+            permit_amendment_type_code='OGP').first()
+
```
Jsyro, line 143: This may return multiple results in situations where a permit is used by multiple mines (each mine,permit pair will have an OGP).

services/core-web/src/components/Forms/PermitAmendmentForm.js
pdittaro: Correct, we should never use the 'mds' prefixed roles within the application, they are just for management in Keycloak. This should be a new permission if that's the case, i.e. prefixed with "core_"

services/core-api/app/api/utils/access_decorators.py
```diff
@@ -15,6 +15,9 @@
 EDIT_SECURITIES = "core_edit_securities"
 MINESPACE_PROPONENT = "mds_minespace_proponents"
 EDIT_SUBMISSIONS = "core_edit_submissions"
+DATA_CLEANUP = "mds_data_cleanup"
```
pdittaro, line 18: These are "user groups" and should not be modelled in our codebase. This should instead reflect the permission or capability necessary in the system. User group role assignment is achieved in Keycloak.

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -160,14 +165,12 @@ def validate_received_date(self, key, received_date):
...
     def validate_issue_date(self, key, issue_date):
-        if self.permit_amendment_type_code != 'OGP':
-            original_permit_amendment = self.query.filter_by(permit_id=self.permit_id).filter_by(
-                permit_amendment_type_code='OGP').first()
-            if original_permit_amendment and original_permit_amendment.issue_date:
-                if issue_date and original_permit_amendment.issue_date > issue_date.date():
-                    raise AssertionError(
-                        'Permit amendment issue date cannot be before the permits First Issued date.'
-                    )
+        # TODO DO NOT REMOVE NEXT LINE. If this validation removed then exception will be thrown on permit creation/editing:
```
pdittaro, line 168: Is this comment actually driving some sort of validation?

mhuseinov: For some reason, if you remove db querying in this method it causes sqlalchemy to randomly fail with strange errors

services/core-web/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -161,6 +162,22 @@ const columns = [
               </button>
             </Menu.Item>
           )}
+          <AuthorizationWrapper permission={Permission.ADMIN}>
+            <div className="custom-menu-item">
```
mmumby, line 166: Can you update this permission to be the historical_permit group?

services/core-web/src/components/Forms/PermitAmendmentForm.js
```diff
@@ -40,10 +43,23 @@ const validateBusinessRules = (values) => {
       (x) => x.permit_amendment_type_code === originalPermit
     )[0];
     const mostRecentAmendment = values.amendments[0];
-    if (originalPermitAmendment && values.issue_date < originalPermitAmendment.issue_date) {
+    const isHistoricalAmendmentsAllowed =
+      values.is_historical_amendment &&
+      values.userRoles &&
+      values.userRoles.some((role) => USER_ROLES.role_edit_historical_amendments === role);
+    if (
```
mmumby, line 50: Can you use includes? `values.userRoles.includes(USER_ROLES.role_edit_historical_amendments)` ?

mmumby (approved):
🔥

PR #1485, opened 2020-10-09
MDS-2376/deletable mixin

Jsyro:
Reorganized delete code to central mixin. Tests functionality. 

Should I write a test ensure the assignment of the mixin to a set of models?
I will play around with inspecting models to find children to cascade soft deletes.


mhuseinov (approved):
💯

PR #1487, opened 2020-10-14
Search/dropdown bug

mmumby:
- Added the dropdown back to the searchBar component, but still allows the Search Icon to search not clear.


nayr974 (approved):
image

PR #1488, opened 2020-10-16
Bugfix/mds 3036 switch to redux field array in now access road

mhuseinov:
POC of NOW Review Form:
switch from HTML inputs to redux FieldArray for Access Road
mds


exzizt (approved):
Good job with the conversion!

PR #1492, opened 2020-10-16
VFCBC new fields fix

nayr974:


services/core-api/app/api/now_applications/models/now_application.py
```diff
@@ -86,7 +86,7 @@ class NOWApplication(Base, AuditMixin):
 
     permit_status = db.Column(db.String)
     term_of_application = db.Column(db.Numeric(14, 0))
-    is_applicant_individual_or_company = db.Column(db.Boolean, nullable=True)
+    is_applicant_individual_or_company = db.Column(db.String)
     relationship_to_applicant = db.Column(db.String)
```
mmumby, line 90: Why is this a string now?

nayr974: It's actually a string from VFCBC, it's not a Yes or No boolean.

mmumby (approved):

PR #1494, opened 2020-10-21
Bug/access roads

mmumby:
- Added null check to field array component
- Renamed FE term_of_application field as it was throwing errors when editing the NoW and does not need to be saved.


nayr974 (approved):
image

PR #1496, opened 2020-10-22
MDS-3108 Auto Generate new mine numbers on mine creation

JustinMacaulay:
This adds a new column to the mine table for a mine number sequence as well as a trigger to copy the mine number sequence into the mine number if that column is null.

Had to leave the factories auto generating mine numbers as many tests would fail if that was removed.

Also modified the mine ETL to set the mine sequence number to null if adding a new mine.

Removed old code and functions to generate prefixed mine numbers.


nayr974 (commented):
I don't see anything out of place. Happy to approve if you're done with the work.

nayr974 (approved):
image

PR #1497, opened 2020-10-22
Remove link to NoW from mineTable

mmumby:
- Removed the link to a NoW table from the mine record, users have been using this to access NoW in prod 
- The "Open" button is still available to access NoW records in test
Screen Shot 2020-10-22 at 4 07 01 PM


nayr974 (approved):
image

mhuseinov (approved):

PR #1499, opened 2020-10-22
MDS 2399 Contact Verification - Fixes and Updates

nayr974:


services/core-web/src/components/Forms/noticeOfWork/EditNoWContacts.js
```diff
@@ -83,11 +93,19 @@ const renderContacts = ({
...
+                          if (fields.get(index)) {
+                            // add state_modified and set to "delete" for backend
+                            contacts[index].state_modified = "delete";
+
+                            changeArray(FORM.EDIT_NOTICE_OF_WORK, "contacts", contacts);
 
-                          arrayPushReduxForms(FORM.EDIT_NOTICE_OF_WORK, "contacts", updatedContact);
-                          arrayRemoveReduxForms(FORM.EDIT_NOTICE_OF_WORK, "contacts", index);
+                            // move updated object, this will cause rerendering of the react component, setTimeout is required to bypass react optimization
+                            setTimeout(() => {
```
mhuseinov, line 103: This definitely a hack but it is the only way to rerender those FieldArray :fire:

mhuseinov (approved):
:100:

services/core-web/src/components/Forms/noticeOfWork/EditNoWContacts.js
nayr974: I never want to use field array again. Next time I'll create a custom component that deals with arrayPush, arrayChange, or change directly... seems like it will be easier. :/

PR #1502, opened 2020-10-26
MDS-3031 Quick Fix for Permit ETL to use appr_dt if iss_dt is NULL

JustinMacaulay:
added a coalesce to use appr_dt if the iss_dt on mmspmt is null.


Jsyro (approved):
I like this. Is there corresponding work from Neil Wells to help this data issue

PR #1505, opened 2020-10-28
Bug - update to delete NoW contacts 

mmumby:
- Updated logic to removing current NoW contacts, and newly added contacts
- Update roles dropdown - if mine manager is deleted, MM gets added back into the select options

- RenderRadio buttons were not showing error validations, also `requiredRadioButton` was not validating correctly - updated component to force validation on required fields instead of submitting "false"


mhuseinov (approved):
:fire:

PR #1506, opened 2020-10-29
MDS-3109 Update bond_view to include Active record of un-etl-able REL/CON bonds

nayr974:


migrations/sql/R__view_bonds.sql
```diff
@@ -101,6 +101,35 @@ FROM
                   null as "update_user", 
                   'Yes' AS "is_current_record", 
                   'MMS_FDW' as "source"
-     FROM SES_BOND_STAGING) AS bond_data
+     FROM SES_BOND_STAGING
```
exzizt, line 104: On line 87 of this file we should be setting the closed date.

exzizt (changes requested):
This solution looks correct, nice work. But I noticed that we should be setting the closed_date for the closed records on line 87.

migrations/sql/R__view_bonds.sql
nayr974: Added logic to set closed date :)

exzizt (approved):
🍏 📗 🟢 💚 🥗 ✔️ ✅

PR #1509, opened 2020-10-29
MDS-2798: Import Notice of Work Submission Documents

mhuseinov:
# Main
- When a Notice of Work is verified a request is sent to the Document Manager to start a job (a _Celery_ task) to import its submission documents.
  - "Importing" a document means downloading/streaming it from its source (either NROS or vFCBC, depending on the NoW's origin) and saving it as a file on our object storage. A Document Manager document (_docman.document_) record is created and becomes associated with a created "NoW document identity" (_public.now_application_document_identity_) record.
- Imported submission documents can now be added to the NoW's "final application package" of documents.
- Imported submission documents can now be downloaded as part of a "referral package".
- Import jobs that fail are retried following a custom "exponential" back-off set of delays. E.g., if the job fails, try again in 1 minute, then if that fails again, in 5 minutes, then in 30 minutes, etc...

### A freshly-started import job
image

### An import job that ran into some errors
image

### Adding some submission documents to the final application package
image

### Viewing a final application package with some submission documents
image

### Imported documents can also be selected in the Referral Package modal
image

# Other
- Minor text changes


services/core-web/src/components/noticeOfWork/applications/administrative/NOWApplicationAdministrative.js
```diff
@@ -17,18 +17,24 @@ const propTypes = {
   inspectors: CustomPropTypes.groupOptions.isRequired,
   setLeadInspectorPartyGuid: PropTypes.func.isRequired,
   handleUpdateLeadInspector: PropTypes.func.isRequired,
+  importNowSubmissionDocumentsJob: PropTypes.objectOf(PropTypes.any),
+  handleSaveNOWEdit: PropTypes.func.isRequired,
```
mmumby, line 21: Is this being used?

services/core-web/common/actionCreators/noticeOfWorkActionCreator.js
```diff
@@ -26,6 +26,25 @@ export const fetchNoticeOfWorkApplication = (applicationGuid) => (dispatch) => {
     .finally(() => dispatch(hideLoading()));
 };
 
+export const fetchImportNoticeOfWorkSubmissionDocumentsJob = (applicationGuid) => (dispatch) => {
```
mmumby, line 29: would love to see a test for all the new redux actionCreators/reducer states/selectors

services/core-web/src/components/Forms/permits/GeneratePermitForm.js
```diff
@@ -220,6 +220,7 @@ export const GeneratePermitForm = (props) => (
         <FinalPermitDocuments
           mineGuid={props.noticeOfWork.mine_guid}
           noticeOfWork={props.noticeOfWork}
+          importNowSubmissionDocumentsJob={props.importNowSubmissionDocumentsJob}
         />
```
mmumby, line 224: Looks like FinalPermitDocuments is a connected component, can you get these props through the store vs passing to this component?

services/core-web/src/components/noticeOfWork/applications/NOWSubmissionDocuments.js
```diff
@@ -64,33 +105,167 @@ export const NOWSubmissionDocuments = (props) => {
     },
   ];
 
-  const columns = [fileNameColum, ...otherColumns];
+  if (!props.hideImportStatusColumn) {
```
mmumby, line 108: Is it possible to add an additional column to this table to show its the document in the final application package or not? Similar to the other document tables. (Not a requirement for merging this or, more of curiosity if we have that data available here)

mmumby (commented):
FE looks good, no huge concerns. 

Where is the trigger for this job?  does it happen on the verification/Import?

services/core-web/src/components/noticeOfWork/applications/administrative/NOWApplicationAdministrative.js
mhuseinov: importNowSubmissionDocumentsJob - is used in modal to display the import job status.
handleSaveNOWEdit - nope, it seems like something old and unused. I will delete it

mhuseinov:
> FE looks good, no huge concerns.
> 
> Where is the trigger for this job? does it happen on the verification/Import?

It is on the backend side in now_application_resources.py in PUT endpoint. It is triggered by lead inspector assignment

mmumby:
> > FE looks good, no huge concerns.
> > Where is the trigger for this job? does it happen on the verification/Import?
> 
> It is on the backend side in now_application_resources.py in PUT endpoint. It is triggered by lead inspector assignment
> 
> And just in case we can manually trigger it by using the POST endpoint

AH, I see. 

It is possible to put that on the import resource? I don't see a reason why we couldn't start at that point vs waiting until they assign an inspector

mhuseinov:
> > > FE looks good, no huge concerns.
> > > Where is the trigger for this job? does it happen on the verification/Import?
> > 
> > 
> > It is on the backend side in now_application_resources.py in PUT endpoint. It is triggered by lead inspector assignment
> > And just in case we can manually trigger it by using the POST endpoint
> 
> AH, I see.
> 
> It is possible to put that on the import resource? I don't see a reason why we couldn't start at that point vs waiting until they assign an inspector

👍  Sure we can do that, but I would rather discuss these changes first with the team and have them in a separate pr

services/core-web/src/components/noticeOfWork/applications/NOWSubmissionDocuments.js
mhuseinov: Hm, interesting, but then the table may look too overloaded and be a duplication because we already have a separate Final Application Package on the Administrative tab with that information

services/core-api/app/api/now_applications/models/now_application_document_identity_xref.py
```diff
@@ -0,0 +1,63 @@
...
+               document_url, file_name, document_type, description):
+
+        new_document_identity_xref = cls(
+            messageid=message_id,
+            documenturl=document_url,
+            documenttype=document_type,
+            description=description,
+            filename=file_name,
+            now_application_id=now_application_id,
+            mine_document=MineDocument(
```
JustinMacaulay, line 57: I am not sure how I feel about the Xref table crating its entry while simultaneously creating the mine document it references...

mhuseinov: changed

mmumby (dismissed):

mmumby (approved):

PR #1516, opened 2020-11-04
misc bugs

mmumby:
- When uploading multiple documents and 1 fails, deleting it disabled the submit button. Fixed
- SelectValidation was forcing input if null. Fixed
- Unable to remove contacts at the initial import stage. Fixed


nayr974 (approved):
Please let this be it 👍

PR #1517, opened 2020-11-04
Remove- required fields on activity detail tables

mmumby:
- Removed all "required" validations from activity_details according to what is nullable in the db.


exzizt (approved):
🏁 ☑️

PR #1520, opened 2020-11-05
Disable no w application

mmumby:
`<NOWActionWrapper />` handles all the logic for disabling NoW actions (Controls still need to be added), It currently accepts the same props as the `<AuthorizationWrapper />`


services/core-web/src/components/Forms/permits/PreDraftPermitForm.js
```diff
@@ -9,7 +9,7 @@ import { resetForm, createDropDownList } from "@common/utils/helpers";
 import * as FORM from "@/constants/forms";
 import { renderConfig } from "@/components/common/config";
 import CustomPropTypes from "@/customPropTypes";
-import AuthorizationWrapper from "@/components/common/wrappers/AuthorizationWrapper";
+import NOWActionWrapper from "@/components/noticeOfWork/NOWActionWrapper";
```
nayr974, line 12: ACTION! NOW!

services/core-web/src/components/noticeOfWork/NOWActionWrapper.js
```diff
@@ -0,0 +1,57 @@
...
+    PropTypes.element.isRequired,
+  ]).isRequired,
+};
+
+const defaultProps = {};
+// eslint-disable-next-line react/prefer-stateless-function
+export class NOWActionWrapper extends Component {
+  // state = { currentTab: "" };
+
+  // componentDidMount() {
```
nayr974, line 28: I assume that later, we conditionally set disabled here based on the combined logic of delayed + processed status?

services/core-web/src/components/noticeOfWork/NOWActionWrapper.js
```diff
@@ -0,0 +1,57 @@
...
+      <div />
+    );
+  }
+}
+
+NOWActionWrapper.propTypes = propTypes;
+NOWActionWrapper.defaultProps = defaultProps;
+
+const mapStateToProps = (state) => ({
+  // can disable all actions based off application status === rejected || withdrawn
```
nayr974, line 51: Or, are you thinking we map the conditional here based on multiple gets?

nayr974 (approved):
I like this approach. A couple questions based on how we might want to implement our logic later, but nothing that affects what's here. Thanks MD!

services/core-web/src/components/noticeOfWork/NOWActionWrapper.js
mmumby: I was planning to use this logic to know if a specific tab should be disabled - ie if consultation has not started yet, then any button on the consultation tab should be hidden until that stage is in progress. There will likely be additional logic added for the progress related disabled-buttons.

services/core-web/src/components/noticeOfWork/NOWActionWrapper.js
mmumby: Yeah, I assumed that all logic would be stored in redux, depending on where the "complete" states are saved, it might not be the noticeofWork object - but I tested the logic with the current now_status and worked well!

PR #1523, opened 2020-11-06
MDS 3122 Migrate Keycloak URL

JustinMacaulay:
On Nov 12th we are required to change from using sso.pathfinder.gov.bc.ca to oidc.gov.bc.ca for our keycloak auth. This Pr changes on the URLS, however when this is merged in all of our clients will need to change the URL they are using to get tokens to the respective "test.oidc.gov.bc.ca" or "oidc.gov.bc.ca". All developers will also need to ensure that their .env files are updated to match the new URL's.


pdittaro:
To be merged at 1pm.

pdittaro (approved):

PR #1524, opened 2020-11-07
Separated Referral, Consultation, Public Comment into their own tabs

mmumby:
- Separated Referral, Consultation, Public Comment into their own tabs
- All functionality remains the same, This move will allow us to start tracking each stage with current sprints progress work
Screen Shot 2020-11-06 at 4 24 41 PM
Screen Shot 2020-11-06 at 4 25 01 PM
Screen Shot 2020-11-06 at 4 25 11 PM

- Updated the logic of the NOWStatusBanner to include other types of icons since they are all driving off the same logic
-Banner
-Badges
- Timeline (Didn't add this yet, but thinking it could live in this component as well due to the similar logic/colouring)
- Also updated styling, added color  references to the Alert Banner (Still hidden for now)
Screen Shot 2020-11-06 at 4 24 16 PM

- Added Badges to the tab headers (Hidden for now until conditions are implemented)
Screen Shot 2020-11-06 at 5 18 48 PM


services/core-web/src/components/noticeOfWork/NOWStatusIndicator.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable */
```
exzizt, line 1: Should we remove this and address any linting issues?

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -861,35 +870,94 @@ export class NoticeOfWorkApplication extends Component {
             </Tabs.TabPane>
 
             <Tabs.TabPane
-              tab="Referral/Consultation"
-              key="referral-consultation"
+              tab={
+                <span>
```
exzizt, line 874: Can we pull this out into a function, as this is being used in the other tabs and they should follow the same format.
```
function renderTabTitle = (title) =>
    (
                <>
                  <NOWStatusIndicator type="badge" />
                  {title}
                </>
    );
```
for example.

exzizt (changes requested):
Looks really good! Just some minor comments. 👍

services/core-web/src/components/noticeOfWork/NOWStatusIndicator.js
mmumby: Yes, was planning to remove once this component is live/has all the logic - This is just a starting point to have this component available for Ryan and myself.

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
mmumby: Yes, I can make this change.

mmumby: Updated!

exzizt (approved):
✔️

PR #1525, opened 2020-11-09
MDS-3014: Render Permit Conditions in Generated Permit Document

exzizt:
# Main
* Permit conditions are now rendered in our generated permit documents

### New margins
image

### Old margins
image

# Other
* Combines our "new permit" and "permit amendment" document templates into one
* Improves some formatting and styling in our permit template


mhuseinov (approved):
💯

PR #1528, opened 2020-11-12
MDS-3170: Fix Error When Viewing MMS Notice of Works

exzizt:
# Main
* Fixes an error when viewing a Notice of Work from MMS which was caused by a null reference exception when trying to read variable `agent` which doesn't exist for MMS submissions


nayr974 (approved):
image

PR #1530, opened 2020-11-13
MDS-3140: Create script for deleting mine_permit_xref records and their associated objects

exzizt:
# Main
* Creates a script that creates, uses, and then drops a function that accepts a permit number and mine number and deletes it and all of its associated records (permit amendments, reports, and Notice of Work data)


JustinMacaulay (commented):
I am assuming you have tested it on dev but other than that it all looks good to me. I would like another set of eyes on it since it will be deleting records from prod.

Jsyro (approved):
This code does what it seems to do. So i've approved it, however should be there be extra consideration for deleting records that have been entered by users in core (and not just ETL'ed in). I'd be tempted to say this function should not delete things that have been input by users through core. but it's also not setup for reuse.... maybe just a comment warning the next developer who uses this as reference?

exzizt:
> This code does what it seems to do. So i've approved it, however should be there be extra consideration for deleting records that have been entered by users in core (and not just ETL'ed in). I'd be tempted to say this function should not delete things that have been input by users through core. but it's also not setup for reuse.... maybe just a comment warning the next developer who uses this as reference?

Sure, I can add a comment. Setting it up for reuse via CLI for example seemed like it was going down a bit of a rabbit hole so I held off on doing it. Thanks for reviewing!

Jsyro (approved):
Looks good.

nayr974 (approved):
image

PR #1532, opened 2020-11-13
Remove referral/consultation/public comment progress tracking

mmumby:
- Wanted to keep this PR independent of the Progress PR
- When  referral/consultation/public comment was built originally, we added in simple progress tracking flow - this is to be replaced with the newer progress functionality. 
- Removed all progress tracking on the UI and fields in the db


exzizt (dismissed):
☑️

exzizt (approved):

PR #1537, opened 2020-11-16
MDS-3074: Accept shape files for NoW file uploads

exzizt:
# Main
* Users can now upload spatial files to a NoW

# Other
* Fixes an issue where TUSD wouldn't return proper "file type" metadata for obscure file types, such as spatial files
* Tidies some code

# Screenshots
image

# Notes
- It's not possible to nicely display all accepted file extensions for files that share the same MIME type (in our case, "application/octet-stream"--many of these files have no MIME type). https://github.com/pqina/filepond-plugin-file-validate-type/issues/13


services/core-web/src/components/common/FileUpload.js
```diff
@@ -53,7 +53,7 @@ class FileUpload extends React.Component {
           chunkSize: this.props.chunkSize,
           metadata: {
             filename: file.name,
-            filetype: file.type,
+            filetype: file.type || "application/octet-stream",
```
mhuseinov, line 56: May be "application/octet-stream" can be moved to constants. What do you think?

mhuseinov (approved):

services/core-web/src/components/common/FileUpload.js
exzizt: Not sure what the heck happened but it's resorted back to not working--so I'll have to take another look. :/

services/core-web/package.json
```diff
@@ -44,7 +44,7 @@
     "react": "^16.13.1",
     "react-async-script-loader": "^0.3.0",
     "react-dom": "^16.13.1",
-    "react-filepond": "^7.0.1",
+    "react-filepond": "^7.1.0",
```
mhuseinov, line 47: Have you verified if filepond works correctly after this change? Especially the inspector signature upload?

exzizt: Yep, it works!

mhuseinov (approved):
💯

PR #1538, opened 2020-11-17
Fix UID issue for Oracle encryption running as any PID

pdittaro:
Refer to this POC for details: https://github.com/bcgov/nr-alm-ops/tree/master/oracle-instant-client-with-tls


pdittaro:
This fix may still be needed. For a couple runs it was working without it, but now its consistently failling on the 'digdag task' but we're still able to manually validate the ssl connection in the nris_api pod.

PR #1541, opened 2020-11-19
Bug - remove permit_no character limit

mmumby:
- Draft permits are created with a draft number - MX-DRAFT-1621396-2014-01 - which exceeds the character limit set in the model
- Removed limit
- "starting" progress does not succeed if there is an error when creating the draft permit
- If the Now_no doesn't exist, use the mine_no for the draft permit no creation - there was a non-unique error when trying to use the number 'P-DRAFT-None'
- Added a check for party address so the draft can be created without one
- There is a bug with previewing the draft if there is no permittee, so disabled button until a permittee is assigned


exzizt (approved):
☑️

PR #1545, opened 2020-11-23
MDS-3113: Allow users to delete documents uploaded to a NoW

mhuseinov:
Allows users to delete documents from NOW applications which are not included in the Final Package.
image
image
image

Fixed export of 'Edited Application'.


services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -161,6 +174,49 @@ export const NOWDocuments = (props) => {
...
     }
 
+    const deleteButtonColumn = {
+      title: "",
+      dataIndex: "isDeletionAllowed",
+      key: "isDeletionAllowed",
+      render: (isDeletionAllowed, record) => {
+        if (isDeletionAllowed) {
+          return (
+            <Popconfirm
```
mmumby, line 184: Can you wrap all the delete actions into a <NOWActionWrapper  /> with the correct permissions and tabs - this will allow them to be disabled based off the status of the tab/application

exzizt (commented):
Looks good to me, once Meredith's comment is addressed. :)

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
mhuseinov: Thanks! it is done.

mmumby (approved):

mmumby (approved):

PR #1546, opened 2020-11-24
BUG - fixed hide/show buttons based off application status

mmumby:
- Any button with conditional logic was not being rendered correctly if a stage was started or stopped
- Updated logic surrounding buttons - they display correctly now

ALSO
- Purposely hid the Admin tooltip on the navigation "verified mines" actions as this gets triggered via the scroll and was driving me crazy.


exzizt (approved):
🐈

PR #1547, opened 2020-11-24
Feature/mds 2977 display who edited conditions

mhuseinov:
Displays who and when last edited Conditions on a permit.
image


services/core-api/app/api/mines/permits/permit_conditions/models/permit_conditions.py
```diff
@@ -46,6 +47,15 @@ class _ModelSchema(Base._ModelSchema):
         order_by='asc(PermitConditions.display_order)',
         backref=backref('parent', remote_side=[permit_condition_id]))
 
+    update_user = db.Column(
```
mmumby, line 50: I don't believe we want to expose the Audit columns to the FE. 

Can you do something similar to `last_updated_by` and `last_updated_date` on `now_application` ?

mmumby: @nayr974 @JustinMacaulay I might be wrong about the audit columns, can you guys confirm?

exzizt: > @nayr974 @JustinMacaulay I might be wrong about the audit columns, can you guys confirm?

That's correct. `update_user`, `update_timestamp`, `create_user`, and `create_timestamp` fields are to never be exposed

exzizt (approved):
Thank you for making those changes. Looks correct! 🥇

PR #1549, opened 2020-11-25
MDS-3167: Update "Total Confiscated" Column On Bond Table

exzizt:
# Main
* Changes the "Total Confiscated" column to be a "Confiscated Cash on Hand" column in the Securities tab. This is being calculated the same way as "Balance" in the Reclamation Invoices tab.

image


mmumby (approved):
🎉 Looks good

PR #1551, opened 2020-11-26
Reject/Withdrawal application with letter

mmumby:
- Ability to reject an application with the "Rejection Letter"
- Ability to reject an application with the "Withdrawal Letter"
- Capturing and displaying the reason for a rejection
- Admin can "Undo rejection" - Reused the old status update modal for this, kept the dropdown options to select from 
- Removed ability to update status via the admin tab
- Removed the ability to generate anything other than the "Acknowledgment Letter" from the admin tab
- If the issuing inspector does not have a signature, they cannot generate the document (This is to prevent the BE error after the user goes through all the work of updating the letter) 
- Added this alert to both the generate document workflows
- Reminder to return the security if it has been received.

PICTURES
Screen Shot 2020-11-26 at 11 34 57 AM
Screen Shot 2020-11-26 at 11 35 12 AM
Screen Shot 2020-11-26 at 11 36 33 AM
Screen Shot 2020-11-26 at 11 37 18 AM
Screen Shot 2020-11-26 at 11 38 25 AM
Screen Shot 2020-11-26 at 11 39 31 AM
Screen Shot 2020-11-26 at 11 39 58 AM
Screen Shot 2020-11-26 at 11 40 12 AM


services/core-web/src/components/Forms/noticeOfWork/RejectApplicationForm.js
```diff
@@ -1,23 +1,54 @@
...
     <Row>
-      <Col span={24} />
+      <Col span={24}>
+        <Form.Item>
+          <Field
+            id="status_reason"
+            name="status_reason"
+            label={props.type === "REJ" ? "Reason for Rejection" : "Reason for Withdrawal"}
+            component={renderConfig.AUTO_SIZE_FIELD}
+            validate={[maxLength(280)]}
```
exzizt, line 48: The max length here is 280 but it is really 4000. Should we update this to 4000?

services/core-web/src/components/modalContent/RejectApplicationModal.js
```diff
@@ -1,33 +1,100 @@
...
+            draftAmendment={this.props.draftAmendment}
+          />
+        ),
+      },
+    ];
+    return (
+      <div>
+        <Alert
+          message="This action is final"
+          description="No changes or additions can be made to this application after it has been rejected."
```
exzizt, line 69: Let's say "after it has been **withdrawn**." if it is withdrawn and not rejected?

services/core-web/src/components/modalContent/RejectApplicationModal.js
```diff
@@ -1,33 +1,100 @@
...
+          type="warning"
+          showIcon
+          style={{ textAlign: "left" }}
+        />
+        <br />
+        {!this.props.signature && (
+          <>
+            <Alert
+              message="Signature needed."
```
exzizt, line 78: Let's remove the "."? :) 

"Signature needed"

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
```diff
@@ -220,10 +278,10 @@ export class ProcessPermit extends Component {
       <Menu.Item key="issue-permit" onClick={this.openIssuePermitModal} disabled={validationErrors}>
         Issue permit
       </Menu.Item>
-      <Menu.Item key="reject-application" onClick={this.openRejectApplicationModal}>
+      <Menu.Item key="reject-application" onClick={() => this.openRejectApplicationModal("REJ")}>
         Reject application
       </Menu.Item>
-      <Menu.Item key="withdraw-application" onClick={this.openWithdrawApplicationModal}>
+      <Menu.Item key="withdraw-application" onClick={() => this.openRejectApplicationModal("WDN")}>
```
exzizt, line 284: The logic in the called function will still work, but it's using "WDL" as a withdrawn status, and not "WDN".

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
```diff
@@ -155,24 +191,46 @@ export class ProcessPermit extends Component {
...
+        newValues[field.id] = formatDate(newValues[field.id]);
+      });
+    const payload = {
+      now_application_guid: this.props.noticeOfWork.now_application_guid,
+      template_data: newValues,
+    };
+    this.props.generateNoticeOfWorkApplicationDocument(
+      documentTypeCode,
+      payload,
+      "Successfully Created Document and Attached it to this Notice of Work"
```
exzizt, line 230: I fixed the casing in this text to match our other toast messages, if you merge in from develop you should resolve the changes.

"Successfully created document and attached it to Notice of Work"

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
```diff
@@ -248,11 +305,22 @@ export class ProcessPermit extends Component {
             </h2>
             <NOWProgressActions tab="PRO" />
             {!isProcessed && (
-              <Dropdown overlay={this.menu(validationErrors)} placement="bottomLeft">
-                <Button type="secondary" className="full-mobile">
-                  Process <DownOutlined />
+              <>
```
exzizt, line 308: I'm not sure if this fragment is necessary as there is already one child "AuthorizationWrapper"?

services/core-web/src/constants/NOWConditions.js
```diff
@@ -66,12 +66,12 @@ export const draftPermitMenu = [
     title: "Conditions",
     alwaysVisible: true,
     children: [
-      { href: "GEC", title: "A. General" },
-      { href: "HSC", title: "B. Healthy and Safety" },
-      { href: "GOC", title: "C. Geotechnical" },
-      { href: "ELC", title: "D. Environmental Land" },
-      { href: "RCC", title: "E. Reclamation and Closure" },
-      { href: "ADC", title: "F. Additional Conditions" },
+      { href: "GEC", title: "> General", light: true },
```
exzizt, line 69: Why don't we want the lettering anymore. Do we still want these lettered in the generated permit documents?

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
```diff
@@ -696,25 +653,14 @@ export class NoticeOfWorkApplication extends Component {
...
         )}
         {!isReview && Object.values(this.props.generatableApplicationDocuments).length > 0 && (
           <Menu.SubMenu key="generate-documents" title="Generate Documents">
             {Object.values(this.props.generatableApplicationDocuments)
               .filter(
                 ({ now_application_document_type_code }) =>
-                  now_application_document_type_code !== "PMA" &&
-                  now_application_document_type_code !== "PMT" &&
-                  now_application_document_type_code !== "NTR"
+                  now_application_document_type_code === "CAL"
```
exzizt, line 663: Can we do this logic by exclusion instead? So if the code doesn't equal rejection letter, withdrawal letter, etc. That way when we add new generatable documents we don't need to modify this condition.

exzizt (changes requested):
Great work, just a few comments that need to be addressed. :)

services/core-web/src/components/Forms/noticeOfWork/RejectApplicationForm.js
mmumby: yes good catch!

mmumby: fixed

services/core-web/src/components/modalContent/RejectApplicationModal.js
mmumby: The end status is always rejected - the difference between withdrawn is the different letter

services/core-web/src/components/modalContent/RejectApplicationModal.js
mmumby: removed

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
mmumby: updated

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
mmumby: resolved

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
mmumby: ah correct, I had something else there origionally.

services/core-web/src/constants/NOWConditions.js
mmumby: Yes, The numbering is still on the UI in the sections, I think it is still desirable in the permit document. 
This is for the side nav - assuming we removed the lettering because it is left aligned now

services/core-web/src/components/noticeOfWork/applications/NoticeOfWorkApplication.js
mmumby: I updated it to be inclusive as this list seems to include less than it excludes - currently we're only showing the 'Acknowledge' letter here. 

The permit enclosed letter you just created will be attached to the issue permit modal, as is reject and withdrawal.

mmumby: I also included the permit enclosed letter here for now until it gets moved in a follow up PR

exzizt (approved):
💯

PR #1553, opened 2020-11-26
MDS-3199 + random fixes and tidy

exzizt:
# Feel free to squash and merge

* Fixes an issue where if an error was thrown when importing a NoW it would still navigate past the Verification tab to the Application tab
* Uses the loading effect (which also disables the button) when importing the NoW
* Puts the validation that checks to see whether a "NoW was already imported" earlier in the function
* Renamed prop `submitting` to `isImporting` because the component is a ReduxForm which overwrites and uses a prop with the same name
* Removes the not null constraint on NoW Application's received date column


services/core-web/src/components/noticeOfWork/applications/verification/ApplicationStepOne.js
```diff
@@ -3,7 +3,6 @@ import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
 import {
-  createNoticeOfWorkApplication,
   fetchImportedNoticeOfWorkApplication,
```
mmumby, line 6: was this never used?

exzizt: Yeah

services/core-api/app/api/now_applications/resources/now_application_import_resource.py
```diff
@@ -50,48 +50,47 @@ def post(self, application_guid):
...
-
-                application.contacts.append(now_party_appt)
-
         # This is a first pass but by no means exhaustive solution to preventing the now application from being saved more than once.
         # In the event of multiple requests being fired simultaneously this can still sometimes fail.
         db.session.refresh(now_application_identity)
         if now_application_identity.now_application_id is not None:
             raise BadRequest('This record has already been imported.')
 
+        application = transmogrify_now(now_application_identity)
```
JustinMacaulay, line 69: I'm not sure I understand the re-ordering of this. I think the order of it was important as it can be a longer running query and we wanted to make sure another API call hadn't modified it. so you do a check after the transmog and then you save it and check again.

JustinMacaulay (commented):
Looks good other than the one comment.

JustinMacaulay (approved):
Looks good to me, Thanks.

PR #1563, opened 2020-12-02
MDS-3194: NoW Edit - Adding a document causes the NoW to refresh

mhuseinov:
Edit mode is for editing form, and document upload is available at other times (works the same way as on the Administrative tab).

image
image
image
image
image
image


mmumby (dismissed):
🔥

mmumby (approved):

PR #1564, opened 2020-12-03
MDS-3134 - Move securities logic to NoW

mmumby:
- Reclamation security data is saved to the NoW record and transferred to the permit when the permit is issued. 
- Added the "No security required" reason to the permit amendment modal, moved options to constant
Screen Shot 2020-12-02 at 4 56 59 PM


mhuseinov (dismissed):

mhuseinov (approved):
💯

PR #1567, opened 2020-12-07
MDS-3202/MDS-3059/MDS-3118: Expose permits and bonds for regional mines on MineSpace & Only show "document generation" form errors on submit & Fix overlapping markers issue on NoW verification map

exzizt:
# Main
## MDS-3202
* Bonds are now visible to proponents on MineSpace
* Permits are now visible to regional mines on MineSpace
## MDS-3059
* Fixes an issue on the "document generation" form where the "X remaining issues" message was being displayed when the user hadn't submitted the form yet
## MDS-3118
* Fixes the issue on the NoW verification map, where identical or very close mine and NoW longitude and latitudes would make the pins difficult to see
* Fixes an issue (caused by z-index) where users could not control the zoom level on the NoW verification map

image

## Other
* Improves the styling of single and multiple permit numbers on the mine card
image

### DIFF
--- a/services/core-web/src/components/mine/NoticeOfWork/MineCard.js
+++ b/services/core-web/src/components/mine/NoticeOfWork/MineCard.js
@@ -48,7 +48,9 @@ export const MineCard = (props) => {
                   ) : (
                     <ul className="mine-list__permits">
                       {props.mine.mine_permit_numbers.map((permit_no) => (
-                        <li key={permit_no}>{permit_no}</li>
+                        <li key={permit_no}>
+                          <p>{permit_no}</p>
+                        </li>
                       ))}
                     </ul>
                   )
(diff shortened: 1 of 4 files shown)


services/core-web/src/components/mine/NoticeOfWork/MineCard.js
```diff
@@ -42,13 +42,21 @@ export const MineCard = (props) => {
...
               <div className="inline-flex padding-small">
                 <p className="field-title">Permit Number</p>
-                <ul className="mine-list__permits">
-                  {props.mine.mine_permit_numbers && props.mine.mine_permit_numbers.length > 0
-                    ? props.mine.mine_permit_numbers.map((permit_no) => (
-                        <li key={permit_no}>{permit_no}</li>
-                      ))
-                    : Strings.EMPTY_FIELD}
-                </ul>
+                {props.mine?.mine_permit_numbers?.length > 0 ? (
```
nayr974, line 45: this works in javascript?

nayr974 (approved):
image

services/core-web/src/components/mine/NoticeOfWork/MineCard.js
exzizt: Yep!... And we oughtta start using it more often! Meredith noticed it and said the same in a recent previous PR of mine

PR #1569, opened 2020-12-09
MDS-3240: Fix NoW "Incline" Error (+ Other Possible Errors)

exzizt:
# Main
* Fixes possible "has no attribute" errors for MMS underground exploration activities


nayr974 (approved):
image

PR #1574, opened 2020-12-14
MDS 3228 Permit number only set when a new permit is being issued.

JustinMacaulay:
Moved the logic for generating the permit number into the section where it changes the permit from draft status.


mmumby (approved):
🔥

mhuseinov (approved):

PR #1581, opened 2020-12-16
MDS 3254 download edited now

JustinMacaulay:
There was an issue with the logic surrounding the party address and the check to make sure a party had an address. Now the system will check to make sure the address is not an empty list.


services/core-api/app/api/now_applications/resources/now_application_export_resource.py
```diff
@@ -109,8 +109,9 @@ def is_number(s):
 
         def transform_contact(contact):
             def get_address(contact):
-                if not contact.get('party', None) and not contact['party'].get('address', None):
+                if  not contact.get('party', None) or not contact['party'].get('address', None) or not len(contact['party']['address']) > 0:
                     return ''
+                current_app.logger.debug(contact['party']['address'])
```
mhuseinov, line 114: seems like you forget to remove logger 😉

mhuseinov (approved):

PR #1584, opened 2020-12-17
Disable opening of 'historic' notice of works

nayr974:


services/core-web/src/tests/components/mine/NoticeOfWork/__snapshots__/MineNoticeOfWorkTable.spec.js.snap
```diff
@@ -62,6 +62,7 @@ exports[`MineNoticeOfWorkTable renders properly 1`] = `
     Array [
       Object {
         "document": Object {},
+        "is_historic": undefined,
```
JustinMacaulay, line 65: I think you are just missing this value in the mocked data.

JustinMacaulay (commented):
Just a comment on the tests and what we talked about on the AQL for the view.

services/core-web/src/tests/components/mine/NoticeOfWork/__snapshots__/MineNoticeOfWorkTable.spec.js.snap
nayr974: updated!

JustinMacaulay (approved):
Looks good!

PR #1585, opened 2020-12-17
Feature/mds 3239 state of the land and annual tonnage

mhuseinov:
- remapped some fields for NOW import
 - added edited tooltips
 - updated pdf template
 
image


mmumby (dismissed):
Well done!

mmumby (approved):

PR #1586, opened 2020-12-18
MDS-3243: Various Updates to NoW Letters

exzizt:
# Main
* Various fields for our NoW letters now have default values

# Other
* The Proponent Address fields for our letters is now multi-lined
* Removes some unneeded code from `now_application.py`
* The logic for retrieving the "relative data path" of fields for our "document template creator" can now handle null objects

# Screenshots
image


nayr974 (changes requested):
Quite possibly the worst code I've ever seen in my life. Please take a code bootcamp ASAP, or at least try a typing tutor because obviously you don't know how to use a keyboard properly.

services/core-api/app/api/document_generation/models/document_template.py
```diff
@@ -50,22 +51,34 @@ def _form_spec_with_context(self, primary_key):
...
-                current_app.logger.debug(f'No relative-data-path for {item["id"]}')
+            if relative_data_path:
+                current_object = source_obj_instance
+                for x in relative_data_path.split('.'):
+                    current_app.logger.debug(f'getting {current_object}.{x}')
+                    current_object = getattr(current_object, x)
+                    if current_object is None:
+                        break
+
+                current_app.logger.info(
```
nayr974, line 65: is this worth logging at this level? info will show up in our loki/grafana logs I think?

nayr974 (commented):
one small comment, the rest looks fine :)

services/core-api/app/api/document_generation/models/document_template.py
exzizt: I can change it and the others in this function to debug. They were set at info before I tabbed the code.

nayr974 (approved):

PR #1597, opened 2021-01-05
Add remaining liability when closing a permit

nayr974:
Adds the ability to set a remaining liability when closing a permit, and displays this value as the total assessed liability after closing:

image

image

image


services/core-api/app/api/mines/permits/permit/resources/permit.py
```diff
@@ -243,6 +245,10 @@ def put(self, permit_guid, mine_guid):
                 continue     # non-editable fields from put
             setattr(permit, key, value)
 
+        # TODO: Confirm desired behavior
```
mhuseinov, line 248: Is it still relevant?

mhuseinov (approved):

PR #1598, opened 2021-01-05
MDS-3166 Assessed Liability after permit closure

pdittaro:
Currently, the liability on a permit is calculated as the sum total of all liability adjustments stored on individual permit amendments (all permits have at least one permit amendment). 

As part of permit closure, the assumption by securities administrators is that in most cases the liability of a permit is regarded as unencumbered.

However, there are boundary cases where there *is* an amount of liability remaining after permit closure. In order to reflect this in our system, this PR creates a seperate `remaining_static_liability` field on the permit, which should only be populated once a permit has been closed. 

When attempting to retrieve the remaining liability, if the permit is closed it will retrieve the values from the `remaining_static_liability` field, otherwise it will behave as before.

image


services/core-web/src/components/Forms/EditPermitForm.js
```diff
@@ -36,6 +39,19 @@ export const EditPermitForm = (props) => (
             validate={[required, validateSelectOptions(props.permitStatusOptions)]}
           />
         </Form.Item>
+        {props.editPermitFormValues.permit_status_code === "C" && (
```
nayr974, line 42: This should probably also be visible if remaining_static_liability is not null

nayr974 (commented):
One small issue, otherwise good

nayr974 (approved):

PR #1603, opened 2021-01-08
Feature/mds 3201 update the acknowledgement letter

mhuseinov:
- updated template wording to include phone number etc.
- fixed acknowledgement letter form_spec

image


mmumby (approved):
nice work!

exzizt (approved):
Double approved ✅

PR #1604, opened 2021-01-11
Update NoW status to match progress/delay/end states.

mmumby:
-Logic is driven by the following chart. 
Screen Shot 2021-01-11 at 12 47 58 PM
- Status is updated on action - ie progress starts/stops, client delay starts/stops, application is imported, application is rejected/issued, etc.
- Update factories to only use active status options when creating a NoW
- Made a few options inactive
- Previous status is saved, admin can revert to previous status if application is rejected.
- Made minor updates to the status reason modal (this will work for no permit required in the future)
Screen Shot 2021-01-11 at 12 52 05 PM
Screen Shot 2021-01-11 at 12 50 01 PM


migrations/sql/V2021.01.08.15.15__add_previous_now_status.sql
```diff
@@ -0,0 +1,12 @@
+Alter table now_application add column previous_application_status_code character varying(3);
```
exzizt, line 1: There should be a foreign key relationship made for this too.

services/core-api/app/api/now_applications/models/now_application.py
```diff
@@ -59,6 +59,7 @@ class NOWApplication(Base, AuditMixin):
         db.String,
         db.ForeignKey('now_application_status.now_application_status_code'),
         nullable=False)
+    previous_application_status_code = db.Column(db.String)
```
exzizt, line 62: When the FK is made be sure to recognize it at the model level here

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
```diff
@@ -48,7 +48,16 @@ def post(self, now_application_guid):
         now_delay = NOWApplicationDelay._schema().load(request.json)
         now_app.application_delays.append(now_delay)
 
-        ##ensure this starts after most recent edit
+        # update now status
+        if now_app.now_application is not None and now_delay.delay_type_code is not None and now_delay.delay_type_code == "OAB":
```
exzizt, line 52: if `now_app.now_application` is None it will go into the else block and get a null reference exception. Is this intentional?

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
```diff
@@ -65,6 +74,11 @@ def put(self, now_application_guid, now_application_delay_guid):
         if not now_app:
             raise NotFound('Notice of Work Application not found')
 
+        # change NoW status back to previous state
```
exzizt, line 77: Does this not get triggered every time this method is called?

migrations/sql/V2021.01.08.15.15__add_previous_now_status.sql
mmumby: There was an issue with the FK - Justin and I opted against it, Can explain after standup and get your opinion?

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
mmumby: good catch, not intentional. I will add another if statement

services/core-api/app/api/now_applications/resources/now_application_progress_resource.py
```diff
@@ -45,6 +46,11 @@ def post(self, application_guid, application_progress_status_code):
                                                              progress_status_code)
             now_progress = new_now_progress
 
+        if identity.now_application is not None and identity.now_application.now_application_status_code != "REF" and progress_status_code == "REF" or progress_status_code == "CON" or progress_status_code == "PUB":
```
exzizt, line 49: A NoW can only have it's status changed to REF, CON, or PUB if it is not currently in REF?

services/core-api/app/api/now_applications/resources/now_application_progress_resource.py
```diff
@@ -75,4 +82,17 @@ def put(self, application_guid, application_progress_status_code):
         existing_now_progress.end_date = data['end_date']
         existing_now_progress.save()
 
+        # update application status if referral/consultation/Public Comment are all complete.
```
exzizt, line 85: I think the logic for determining if an application has had each of REF, CON, and PUB completed is more complex. We need to grab the **most recent** progress objects for REF, CON, and PUB. If any of those is null, then we know referral is not completed. If the end date for each of those most recent records exists, then we know that referral has completed.

exzizt (changes requested):
Good work, just a couple of comments to address.

migrations/sql/V2021.01.08.15.15__add_previous_now_status.sql
mmumby: FK has been added.

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
mmumby: fixed.

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
mmumby: yes - this method is only called when setting an end_date on the delay, so If the application record exists, revert the status back. There are no other checks needed.

services/core-api/app/api/now_applications/resources/now_application_progress_resource.py
mmumby: The below code is grabbing all the end_dates for any record of type REF/CON/PUB. 
The array could have 1-3 entries, if end_dates exist, it will update the status, if any of the end_dates are None it will not change anything. 
so if all three are in progress, the array will be [None, None, None] - Status will not be updated.

I can walk you through a demo.

exzizt: But what if there were only progress records for PUB, for example, it would still think referral was complete if PUB had an end date but REF and CON had no records yet.

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
```diff
@@ -67,6 +77,11 @@ def put(self, now_application_guid, now_application_delay_guid):
         if not now_app:
```
exzizt, line 77: This check should be moved up a bit?

services/core-api/app/api/now_applications/models/now_application.py
mmumby: updated.

services/core-api/app/api/now_applications/resources/now_application_progress_resource.py
mmumby: yes - that's what the ticket wanted. 
REF/CON/PUB is not required, if one is started it must be completed, but there is no logic to force them to all be started.

This is what the ticket stated - I will confirm with Brian/Ashley
"When all of Consulation, Referral or PC that have been started and are now complete. Anything not started does not need to be considered. If something gets started after another is complete, reset status to Referred"

services/core-api/app/api/now_applications/resources/now_application_delay_resource.py
mmumby: Can you elaborate? It is at the very top,  directly under where now_app is defined. Github is showing it weird.

mmumby: This is the PUT check, not the POST.

services/core-api/app/api/now_applications/resources/now_application_progress_resource.py
mmumby: Confirmed with Ashley - This is the workflow they would like for right now.

exzizt (dismissed):
💥

exzizt (approved):

PR #1605, opened 2021-01-12
MDS-3300, MDS-3301: template chnages and date format update

mhuseinov:
- fixed condition numbering
- fixed date alignment
- mine format changes
image
image


services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
```diff
@@ -253,10 +253,10 @@ export class NOWPermitGeneration extends Component {
       }
     }
 
-    newValues.auth_end_date = formatDate(this.props.formValues.auth_end_date);
-    newValues.application_date = formatDate(newValues.application_date);
     this.props.handleGenerateDocumentFormSubmit(this.props.documentType, {
       ...newValues,
+      auth_end_date: formatDate(this.props.formValues.auth_end_date),
+      application_dated: formatDate(newValues.application_date),
```
exzizt, line 259: Why change to application_dated from application_date?

exzizt (dismissed):
☑️

services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
mhuseinov: application_date is a field of now_application model of type DateTime, when python receives this property it treats it as a DateTime and sends it to the template in format 2021-01-01, however, we want to have it as "Jan 1 2020" ( all these data format preparations were done previously on the front end)
so this naming it is just a super easy and safe shortcut

exzizt (approved):

exzizt (approved):

exzizt (approved):

PR #1609, opened 2021-01-12
MDS 3227 MPA for permittee failing to create new permittee

JustinMacaulay:
If the permittee on the NoW matches the current permit permittee then nothing will happen. If the permittee is different then the current mine party appointment for the permittee will be ended and the new permittee appointment will be created.


services/core-api/app/api/now_applications/resources/now_application_status_resource.py
```diff
@@ -132,16 +132,20 @@ def put(self, application_guid):
...
-                            mine=now_application_identity.mine
-                            if contact.mine_party_appt_type_code != 'PMT' else None,
-                            permit=permit if contact.mine_party_appt_type_code == 'PMT' else None,
-                            party_guid=contact.party_guid,
-                            mine_party_appt_type_code=contact.mine_party_appt_type_code,
-                            start_date=datetime.utcnow(),
-                            end_date=None,
-                            processed_by=self.get_user_info())
-                        mine_party_appointment.save()
+                        current_mpa = permit.permittee_appointments[0]
```
exzizt, line 135: This will throw an index out of range error if permittee_appointments is empty

exzizt (changes requested):
See comment

services/core-api/app/api/now_applications/resources/now_application_status_resource.py
JustinMacaulay: this change is now irrelevant, I have reworked the whole flow to encompass MMG and PMT.

exzizt (approved):

nayr974 (approved):

PR #1610, opened 2021-01-13
MDS-3306:Fix NOW submission documents download for not imported files

mhuseinov:


services/core-web/src/components/noticeOfWork/applications/NOWSubmissionDocuments.js
```diff
@@ -46,7 +46,7 @@ const transformDocuments = (documents, importNowSubmissionDocumentsJob, now_appl
           })
         : null;
     return {
-      key: document.mine_document_guid,
+      key: document.mine_document_guid ?? document.id,
       now_application_guid,
```
mmumby, line 50: the syntax looks off here, should that be || ?

mhuseinov: nope, it is short for 
document.mine_document_guid !== null && document.mine_document_guid !== undefined ? document.mine_document_guid : document.id

mmumby (approved):

PR #1611, opened 2021-01-14
Bug/duration with delays calculation

mmumby:
- Updated the UI to show messages when delays exceed the time spent in each stage, previously it showed N/A
- Updated the duration text to include seconds so something is always returned/visible on the UI

This should only be visible with factory data - we should have this information on all records imported.
Screen Shot 2021-01-13 at 5 07 03 PM
Screen Shot 2021-01-13 at 5 07 15 PM
Screen Shot 2021-01-13 at 5 07 47 PM

- Also readjusted the banner height on the fixed nav status banner


nayr974 (approved):

exzizt (approved):
✔️

PR #1636, opened 2021-01-27
Changed create draft permit component

mmumby:
- Updated the "predraft form" to use more form handlers
- Only update the application type if it's been changed
- Edit type > then create permit > then start progress
- the "proceed" button doesn't trigger anything except a page change
- form maintains its data, doesn't need to be stored in the state

OTHER
- Removed required FE validation from e-referral number


nayr974 (dismissed):
Looks good!

nayr974 (approved):

PR #1637, opened 2021-01-28
MDS-3354: Fix Name Issue With Party Organizations

exzizt:
# Main
* Fixes an issue where parties of type "organization" could be created with a "first name" field, causing data quality issues
* Sets `first_name` to null for all existing parties that are an organization
# Other
* Fixes an issue where a party's "name" hybrid property and expression was not being correctly calculated

### DIFF
--- a/services/core-api/app/api/parties/party/models/party.py
+++ b/services/core-api/app/api/parties/party/models/party.py
@@ -20,8 +20,6 @@ class Party(SoftDeleteMixin, AuditMixin, Base):
 
     party_guid = db.Column(UUID(as_uuid=True), primary_key=True, server_default=FetchedValue())
     first_name = db.Column(db.String)
-    # TODO: Determine if this column is necessary; it appears to not be used by anything.
-    middle_name = db.Column(db.String)
     party_name = db.Column(db.String, nullable=False)
     phone_no = db.Column(db.String)
     phone_ext = db.Column(db.String)
--- a/services/core-api/app/api/parties/party/models/party.py
+++ b/services/core-api/app/api/parties/party/models/party.py
@@ -1,6 +1,5 @@
 from datetime import datetime
 import re
-from flask import current_app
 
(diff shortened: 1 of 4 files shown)


services/core-api/app/api/parties/party/models/party.py
```diff
@@ -1,55 +1,60 @@
...
+
 
 class Party(SoftDeleteMixin, AuditMixin, Base):
     __tablename__ = 'party'
+
     party_guid = db.Column(UUID(as_uuid=True), primary_key=True, server_default=FetchedValue())
-    first_name = db.Column(db.String, nullable=True)
-    middle_name = db.Column(db.String, nullable=True)
+    first_name = db.Column(db.String)
+    # TODO: Determine if this column is necessary; it appears to not be used by anything.
```
exzizt, line 23: Can we delete this column? It's not being used anywhere.

services/core-api/app/api/parties/party/models/party.py
```diff
@@ -100,15 +102,20 @@ def json(self, show_mgr=True, relationships=[]):
             context.update({
                 'mine_party_appt': [item.json() for item in self.mine_party_appt],
             })
+
         return context
 
+    @name.expression
```
exzizt, line 108: Funnily enough, their example uses a "name" and I copied its structure for the `expression`.
https://docs.sqlalchemy.org/en/13/orm/mapped_sql_expr.html#using-a-hybrid

services/core-api/app/api/parties/party/models/party.py
mmumby: you want to delete "first_name"  or "middle_name"?

"middle_name" yes, I'd delete - not sure why we have that.

mmumby (dismissed):
looks good, Luke!

mmumby (dismissed):

mmumby (approved):

PR #1638, opened 2021-01-28
MDS-increase max_connections value

mhuseinov:
- granted access for Metabase user to Docman schema
- increased max connection value for Postgres


exzizt (approved):
image

nayr974 (approved):

PR #1639, opened 2021-01-29
MDS-3394: Fix file uploads sometimes failing due to resumable upload feature

exzizt:
# Main
* Fixes an issue where file uploads would sometimes fail if they were being resumed and their stored cached data had expired and/or the object store responded with "not found" and we didn't properly return that status
* Implements in our TUS API the "upload expires" header to indicate to the client when our cached upload metadata expires
* Shows an error notification for file upload errors:
image

# Other
* Adds a new accepted spatial filetype


mhuseinov (approved):
💯

PR #1641, opened 2021-02-02
Added units to "estimate_rate" & updated CoreEditableTable

mmumby:
- Added 'AutoSizefield' to description columns on NoW
- Added units to "estimate_rate"
- Update CoreEditableTable to allow the addition of unit fields on edit, but display next to the value on view
- Removed disabled inputs when viewing the table, it displays as a normal table now. 
Screen Shot 2021-02-01 at 4 47 20 PM
Screen Shot 2021-02-01 at 4 47 37 PM


migrations/sql/V2021.01.28.14.28__add_units_to_details.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE water_supply_detail ADD COLUMN estimate_rate_unit_type_code character varying(3);
+
+ALTER TABLE ONLY water_supply_detail ADD CONSTRAINT estimate_rate_unit_type_code_fkey FOREIGN KEY (estimate_rate_unit_type_code) REFERENCES unit_type(unit_type_code);
+
+
+INSERT INTO unit_type
+(unit_type_code, short_description, description, active_ind, create_user, update_user)
+VALUES
+     ('MES', 'm3/s', 'Meters cubed/s', true, 'system-mds', 'system-mds'),
```
exzizt, line 9: We should be able to use the Unicode character ³ for cubed

exzizt (changes requested):
Happy to approve once 3 is changed to ³ :)

exzizt (approved):
💯

PR #1645, opened 2021-02-05
Mds 3318  contacts   tenure holder, land owner, site operator management

mhuseinov:
- Tenure Holders, Site Operators, Land Owners are linked to the permit
- transfer TH, SO, LO to a mine with a linked permit on issue permit
- the ability to add TH, SO, LO on mine page with permit assignment
- replace all TH, SO, LO linked to a specific permit on issuing an amendment via NOW
- contacts without linked permits will not be replaced


exzizt (approved):
Looks good!

PR #1646, opened 2021-02-05
Fix - added camps to reclamation summary

mmumby:
- Included logic to add camps to the reclamation summary table. 
Screen Shot 2021-02-05 at 9 19 24 AM


services/core-web/common/selectors/noticeOfWorkSelectors.js
```diff
@@ -22,20 +22,22 @@ export const getNOWReclamationSummary = createSelector(
     const reclamationList = [];
     if (options.length > 0) {
       options.forEach(({ value, label }) => {
+        // options uses the value of the code table to map to the response_model names for activities, they all match except camps.
+        const newValue = value === "camp" ? "camps" : value;
```
mhuseinov, line 26: can we rename value to camps on DB level instead?

mmumby: I was considering that - I don't want to rename the code table as "camp" is the code, but I could rename the response_model key. 

This was just less effort and no negative side effects - but I can rename the response_model key and test everything if that is preferred. 

Let me know

mmumby: Discussed: Opted to rename the response_model key.

mhuseinov (commented):
Looks good!
Happy to approve as soon as we get a second pair of eyes to review 👀

exzizt (approved):
I camp approve this PR

PR #1650, opened 2021-02-08
FIx- Add NROS/vFCBC tooltip to "historic" applications

mmumby:
# FIX
- Applications that are pushed directly to CORE will have more data than existing applications - added tooltip to older applications to address the missing data
- Added "is_pre_launch" flag to check if application should show tooltip or not
- If applications are pushed via the API to core from VFCBC, set 'originating_system'
Screen Shot 2021-02-08 at 11 48 32 AM

- Updated water_supply to expose `supply_source_type` for `type` 
- Added additional fields to the details table that we were capturing but not surfacing (pump_size and intake_location)
Screen Shot 2021-03-01 at 1 38 01 PM

- Added `quantity` to `surface_bulk_sample_activity`
- Confirmed that Placer activity_details is being populated correctly
Screen Shot 2021-03-01 at 2 25 15 PM
Screen Shot 2021-03-01 at 2 27 37 PM


mhuseinov (dismissed):
👍

mhuseinov (approved):

PR #1652, opened 2021-02-10
Double unittest timeout in sonar pod

jon-funk:
Doubles active deadline for sonar pod as a temporary fix for the blocking timeout issue.


pdittaro (approved):
2 character PR ftw

PR #1657, opened 2021-02-17
MDS-3426: "Fix permit draft incorrectly attached as the issued permit"

exzizt:
# Main
* Fixes an issue where, when issuing a permit, the generated permit PDF was not always attached as a related document to the permit as intended

# Other
* Fixes an issue where the "submitting" indicator was not correctly being set in `UpdateStatusGenerateLetterModal.js` and the user could spam click the button
* Sets the `disabled` prop for some buttons when submitting
* Minor adjustments to comments and documentation


services/core-api/app/api/now_applications/resources/now_application_status_resource.py
```diff
@@ -2,6 +2,7 @@
 from flask_restplus import Resource, reqparse, inputs
```
exzizt, line 2: In this file, I mainly just adjusted indentation to reduce nesting so its easier to read and updated some comments.

services/core-web/src/actionCreators/documentActionCreator.js
```diff
@@ -50,20 +53,29 @@ export const generateNoticeOfWorkApplicationDocument = (
...
+        .get(`${ENVIRONMENT.apiUrl + API.DOCUMENT_GENERATION(params)}`, createRequestHeader())
+        .then((response) => {
+          const mineDocument = response.data;
+          notification.success({
+            message,
+            duration: 10,
+          });
+          dispatch(success(reducerTypes.GENERATE_NOTICE_OF_WORK_APPLICATION_DOCUMENT));
+          if (downloadGeneratedDocument) {
+            notification.info({
```
exzizt, line 67: Not sure if we want this notification message or not

mmumby: Do users ever see it? Chrome opens a new tab for me until the download is successful then I get the success message.

exzizt: They do but it might not be necessary. I think it was more useful when I was testing, I'll remove it.

mmumby: I see - you removed the functionality that opens the new window - Its not typical of CORe to show "in-progress" toast messages, but I guess it doesn't hurt.

exzizt: `downloadFileFromDocumentManager` still uses the "open new window" functionality

services/core-api/app/api/document_generation/resources/now_document.py
```diff
@@ -77,7 +81,14 @@ def get(self):
         now_application_identity.now_application.documents.append(now_doc)
         now_application_identity.save()
 
-        # Return the generated document
+        now_application = NOWApplication.find_by_application_guid(now_application_guid)
+        now_application_document_type.after_template_generated(template_data, now_doc,
+                                                               now_application)
+
+        # Depending on the return_record param, return the document record or file content
```
exzizt, line 88: Everywhere in our code that calls this endpoint currently is configured to receive the record rather than the file stream, so it's possible that we could just remove this `return_record` param if we wish.

services/core-web/src/tests/actionCreators/documentActionCreator.spec.js
```diff
@@ -29,20 +29,17 @@ describe("`documentActionCreator` action creator: exported Notice of Work ", ()
...
   const url = `${ENVIRONMENT.apiUrl}${COMMON_API.NOW_APPLICATION_EXPORT_DOCUMENT_TYPE_OPTIONS}/${documentTypeCode}`;
-  it("Request successful, dispatches `success` with correct response", () => {
-    const mockResponse = { data: { token: "token" } };
-    mockAxios.onPost(url, payload).reply(200, mockResponse);
-    global.open = () => {
-      return {
-        onbeforeunload: () => jest.fn(),
-      };
-    };
+  it("Request failure, dispatches `error` with correct response", () => {
```
exzizt, line 32: I've spent roughly a few hours trying to fix this "successful" test for this action creator but I could not find a way for JEST to test methods of this shape (where the "then" of an axios request returns the promise another axios request).

services/core-web/src/actionCreators/documentActionCreator.js
```diff
@@ -50,20 +53,25 @@ export const generateNoticeOfWorkApplicationDocument = (
...
+      return CustomAxios()
+        .get(`${ENVIRONMENT.apiUrl + API.DOCUMENT_GENERATION(params)}`, createRequestHeader())
+        .then((response) => {
+          const mineDocument = response.data.mine_document;
+          notification.success({
+            message,
+            duration: 10,
+          });
+          dispatch(success(reducerTypes.GENERATE_NOTICE_OF_WORK_APPLICATION_DOCUMENT));
+          if (downloadGeneratedDocument) {
```
mmumby, line 66: Is this condition always true? I don't see any changes in this PR that would pass a false param.

mmumby (commented):
FE looks good, just one small question.

services/core-web/src/actionCreators/documentActionCreator.js
exzizt: Yeah that's correct

mmumby: Do we need the if statement then? Planning ahead?

jon-funk (commented):
BE changes look good to me. The desired behaviours of both the return_record and ordery_by changes make sense. Thanks for explaining Luke.

mmumby (approved):

PR #1658, opened 2021-02-17
Feature/mds 3456

mhuseinov:
- the ability to assign a permit to TH, LO, SO;
- fixed date range validation
image
image

### DIFF
--- a/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
+++ b/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
@@ -66,9 +66,16 @@ class MinePartyAppointment(SoftDeleteMixin, AuditMixin, Base):
         if commit:
             if not (self.permit or self.permit_id or self.mine_guid or self.mine):
                 raise AssertionError("Must have a related permit or mine")
-            if self.mine_party_appt_type_code in PERMIT_LINKED_CONTACT_TYPES and (
-                    self.mine_guid or self.mine) is not None:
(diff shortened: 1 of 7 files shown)


services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -66,9 +66,16 @@ def save(self, commit=True):
...
             if not (self.permit or self.permit_id or self.mine_guid or self.mine):
                 raise AssertionError("Must have a related permit or mine")
-            if self.mine_party_appt_type_code in PERMIT_LINKED_CONTACT_TYPES and (
-                    self.mine_guid or self.mine) is not None:
-                raise AssertionError("Permittees are not related to mines")
+
+            if self.mine_party_appt_type_code == 'PMT' and (self.mine_guid
+                                                            or self.mine) is not None:
+                raise AssertionError("Contacts linked to a permit are not related to mines")
+
```
mmumby, line 73: How does a user get this error via the UI?

mmumby (approved):
Looks good, I have a small non-blocking question.

services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
mhuseinov: This is a toast message, in theory, the user would not see this one, it is an additional step to prevent such cases when we have weird data.

PR #1662, opened 2021-02-22
Feature/mds 3374 unit type in core

mhuseinov:
This is a first part of the ticket, which includes WEB UI and Edited Notice of Work pdf changes. The DB and BE changes will be covered in the second part.

- updated titles with correct unit types on UI
- updated titles with correct unit types on pdf
- add validation on the Estimate Rate field in the "Water supply" section
- add unit type to pdf for Estimate Rate field
- minor fixes for unit type component

image
image


mmumby (approved):
Looks good, @mhuseinov 🔥

PR #1663, opened 2021-02-23
Feature/mds 3374 unit type in core

mhuseinov:
- default values for:
  - Access Roads, Trails, Helipads, Air Strips, Boat Ramps
  - Settling Ponds
  - Underground Exploration
  - Placer Operations
  - Cut Lines and Induced Polarization Survey
- default values for historical data (depth, height, width, length)


exzizt (dismissed):
💯

exzizt (approved):

PR #1664, opened 2021-02-23
MDS-3396 Permit documents should show the total security amount not the adjusted security amount

pdittaro:
Signed-off-by: Paul Dittaro <pdittaro@gmail.com>


services/core-api/app/api/now_applications/models/now_application_document_type.py
```diff
@@ -76,9 +77,19 @@ def transform_permit(template_data, now_application):
...
             template_data['longitude'] = str(now_application.longitude)
             template_data['mine_name'] = now_application.mine_name
+
+            # If amendment, get sum total security adjustment
+            if not now_application.is_new_permit:
+                permit_amendment = PermitAmendment.find_by_now_application_guid(
+                    now_application.now_application_guid)
+                associated_permit = Permit.find_by_permit_id(
+                    permit_amendment.mine_permit_xref.permit_id)
+                total_liability = now_application.liability_adjustment + associated_permit.assessed_liability_total if associated_permit.assessed_liability_total else now_application.liability_adjustment
```
exzizt, line 87: Are we sure on the OOPS here? I would add parenthesis around `(now_application.liability_adjustment + associated_permit.assessed_liability_total)`, otherwise the result could be `now_application.liability_adjustment + now_application.liability_adjustment`.

pdittaro: Good call!

pdittaro: I read through https://realpython.com/python-conditional-statements/ - it looks like conditional expressions are evaluated with all the left hand side resolved.

services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
```diff
@@ -76,8 +76,8 @@ const regionHash = {
 const getDocumentsMetadataInitialValues = (documents, guid_name) => {
   const initialValues = {};
   // eslint-disable-next-line no-unused-expressions
-  documents &&
-    documents.foreach((doc) => {
+  documents.length !== 0 &&
```
exzizt, line 79: Can throw "none type object has no attribute length". Let's use lodash "!isEmpty(documents)" instead

mmumby: updated!

jon-funk (commented):
Holy formatting batman. deployment changes look good to me.

mmumby (approved):
🔥

PR #1668, opened 2021-02-25
Bugfix/mds 3415

mhuseinov:
- removed length validation on DB level for party and address tables
- add validation on UI for first name, last name and Ext.


exzizt (approved):
✔️

PR #1669, opened 2021-02-26
MDS-3173 - GIS Spatial Files + MDS-3466 - Editing FAP Throws Error 

nayr974:
This PR addresses two tickets, one is providing some small extensions of our existing APIs for the GIS team as well as a new role for them to call our APIs with, and the other is a fix for a bug when saving edits to the FAP.


services/core-web/src/components/noticeOfWork/applications/permitGeneration/NOWPermitGeneration.js
```diff
@@ -326,10 +328,12 @@ export class NOWPermitGeneration extends Component {
         const fieldIdParts = key.split(/_(.+)/);
         const guid = fieldIdParts[0];
         const fieldName = fieldIdParts[1];
-        if (!(guid in allFileMetadata)) {
-          allFileMetadata[guid] = {};
+        if (guid && guid !== "null") {
```
exzizt, line 331: So this doesn't add the doc in the request if the guid is null, but isn't it a problem that the guid is null in the first place?

nayr974: It is, and this is a band-aid fix for sure. I didn't find the root cause as to why it's happening and got stuck at "redux forms magic".

exzizt: What if this causes a different more worse problem? :( Idk I haven't wrapped my head around it yet. Might be good to get someone else to look at it and get their opinion and maybe have a team discussion? @mhuseinov @mmumby

exzizt: *Resolved in convo with team*

exzizt (approved):
image

PR #1670, opened 2021-02-26
Bugfix/mds 3495

mhuseinov:
- added phone number validation on verify contact UI
- moved phone validation from party_model to API resources
- scrolls for contact section on NOW
image
image


services/core-web/src/styles/components/NoticeOfWork.scss
```diff
@@ -322,6 +322,11 @@
 }
 
 .contact-rows {
+  .scroll {
+    height: 1000px;
```
exzizt, line 326: We could restrict size to ~`100vh` that way it is proportional to the user's resolution?

mhuseinov: The height here is more about fitting an appropriate amount of data than a proportion. If it is not a concern then I can definitely change it. @mmumby

mmumby: yeah 100vh works fine, I just tested it.

exzizt: I think it should be responsive to the user's height. We may need to set this as `max-height` instead.

mmumby: what are your concerns with `height`? I've updated it to use 100vh and it's working well switching between my MacBook and monitor.

exzizt: It would be to use as much vertical real estate as possible before introducing the scrollbars. For some 4k monitors, like a 3840x2160, a cap at 2000px would be more suitable and responsive than setting it to 1000 px. Not a big deal but just trying to be more responsive based off of the user's device resolution.

exzizt (approved):
✔️ ✔️ ✔️

services/core-web/src/styles/components/NoticeOfWork.scss
exzizt: I think 75vh looked better but it's no problem.  🤓  I approved

PR #1674, opened 2021-03-03
Fix/Can't save empty string preamble date

exzizt:
* Fixes an issue where sending over an empty preamble date was sending over an empty string, and the backend could not parse this correctly as null

image


jon-funk (approved):
Looks good!

mmumby (approved):

PR #1676, opened 2021-03-04
Fix: update docgen api image ver to 2.1.2 to increase max request size

jon-funk:
Checked that docker-compose is able to pull the new image. We should check in test before deploying to prod.


exzizt (approved):
✅

PR #1677, opened 2021-03-04
Bugfix/mds 3420

mhuseinov:
First part
- fix for status when completing a technical review for an application
The second part with data clean up will be in a separate pr


migrations/sql/V2021.03.04.12.22__update_now_application_progress_status.sql
```diff
@@ -0,0 +1,58 @@
+/*
```
exzizt, line 1: Could you mark with a comment the difference between each `referal_complete_applications`? Also, not a big deal but "referal" should be spelled "referral". 🤓  Can we remove the commented-out code as well?

mhuseinov: The pseudo-code(the condition in each 'if' statement explains the difference in the 'with' clause) above each block of update statement explains the logic used to determine which applications should be set to referral.

exzizt (dismissed):
Nice work :)

mmumby (approved):
🔥

nayr974 (approved):

PR #1680, opened 2021-03-06
Enhancement/Modify word-wrap for toasts to show whole messages & String replace psycopg2

exzizt:
* Modifies the style of our notification messages to better fit long text
* Replaces occurrences of "psycopg2" with "Database Error" for our "custom axios" error messages
image


pdittaro:
It might be nice to move towards a header/content in our toast messages? E.g. large 'Database Error', then smaller, potentially scrollable/code block error message?

services/core-web/common/customAxios.js
```diff
@@ -10,6 +10,11 @@ promiseFinally.shim();
 const UNAUTHORIZED = 401;
 const MAINTENANCE = 503;
 
+const formatErrorMessage = (errorMessage) => {
+  errorMessage = errorMessage.replace("(psycopg2.", "(DatabaseError.");
```
nayr974, line 14: is it worth catching "psycopg2" vs "psychopg2." ? It might exist in other places in a string?

nayr974 (approved):

services/core-web/common/customAxios.js
exzizt: We can do that, I thought we may just want to change the initial mention of it though. Happy to have to replace all instances if that's what we want?

exzizt:
> It might be nice to move towards a header/content in our toast messages? E.g. large 'Database Error', then smaller, potentially scrollable/code block error message?

A good idea, but this would require a lot more time and work to figure out, of course. A good UX-debt item.

services/core-web/common/customAxios.js
nayr974: initial mention is probably preferable, so long as it's consistently followed by the period

exzizt: Yup, and by the initial "(".

PR #1681, opened 2021-03-08
Removed water_supply reclamation description

mmumby:
# Main Updates
Screen Shot 2021-03-08 at 1 50 26 PM

- Water Supply does not have reclamation descriptions or estimated costs, thus does not get included in the reclamation summary. 
- Removed these fields from the UI
- Removed 'Water supply" and "Blasting" from the technical review pdf reclamation summary table.

Before -> 
Screen Shot 2021-03-08 at 1 49 33 PM
After -> 
Screen Shot 2021-03-08 at 1 49 51 PM

# Note: 
- There are two records in production where users were using "reclamation_description" as a generic notes section for water_supply. Ashley is adding the notes to a more appropriate field. 
- ran some SQL to delete the "activity_summary" records for "water_supply", but they are connected to the "activity_detail" thus cannot be deleted.


mhuseinov (approved):
💯

PR #1686, opened 2021-03-10
Update/pdf to include more water supply details

mmumby:
# MAIN
- Updated the Water Supply table on the NoW pdf to include the new detail values
Screen Shot 2021-03-10 at 11 25 40 AM
Screen Shot 2021-03-10 at 11 26 03 AM

Original Application: 
Screen Shot 2021-03-10 at 11 23 04 AM

# Other
- updated search inputs to start searching if value >= 2


exzizt (approved):
✅

PR #1694, opened 2021-03-29
MDS-3136: updates

mhuseinov:
updates that we discussed on call:

- removed redundant types
- fix for 2 permittee
- fix of validation for already in progress amendment
- lat, long and property name from mine

image
image


mmumby (approved):
🔥

mmumby (approved):

PR #1696, opened 2021-03-30
MDS-3541: Create Admin-Only Button to Trigger NoW Document Import Job

exzizt:
# Main
* Creates an admin-only button to trigger the import-job for a NoW's submission documents
image

# Other
* Fixes an issue where we were incorrectly serializing `success_docs` of `task_result` in the import NoW submission documents job, causing the job to remain "in progress" the first time it ran even if it was successful


mhuseinov (dismissed):
👍

mhuseinov (approved):

PR #1697, opened 2021-03-31
Feature/mds 2137 collect government agency info

mhuseinov:
- government agency type code,
- default exemption fee status for the government mines,
- limited mine types to have only one (all historical data remains the same),
- conditional rendering of government agency and exception fee on mine general page
image
image
image
image


services/core-api/app/api/mines/mine/resources/mine.py
```diff
@@ -293,15 +317,22 @@ class MineResource(Resource, UserMixin):
         'exemption_fee_status_code',
         type=str,
         help='Fee exemption status for the mine.',
-        trim=True,
         store_missing=False,
+        trim=True,
         location='json')
     parser.add_argument(
         'exemption_fee_status_note',
```
mmumby, line 324: This didn't already exist on the mine resource? interesting

mmumby (approved):
one small question, but otherwise the PR looks good. Well done.

services/core-api/app/api/mines/mine/resources/mine.py
mhuseinov: nope, and it was ignoring those values on post

PR #1699, opened 2021-03-31
MDS-3575 - Add letter body content to document templates.

nayr974:
As a permitting inspector I want to be able to see all of the text (boilerplate, data and user entered) in a document that Core is generating so that i have a sense of how things are ordered and know that the information I have included makes sense in context.

A/C

static text can be defined as part of the input template
static text is displayed in the front end on the same input form as the rest of the fields
body content
Includes: Acknowledgement letter, rejection letter, withdrawal letter, and permitting enclosed letter


services/core-web/src/tests/components/common/RenderLabel.js
```diff
@@ -0,0 +1,31 @@
...
+      warning: false,
+    },
+  };
+};
+
+beforeEach(() => {
+  setupProps();
+});
+
+describe("RenderLabel", () => {
```
exzizt, line 26: 😠

exzizt (changes requested):
Looks good! Will need to update the "seed data" file because it also runs update statements that will replace the updates in this migration.

services/core-web/src/tests/components/common/RenderLabel.js
nayr974: ?

exzizt: 😠

exzizt (approved):
✅

PR #1700, opened 2021-04-01
MDS-3608: Update NoW Document Types

exzizt:
# Main
* Sets a sub-type for various NoW document types
* Adds a couple of new NoW document types (Application Export Files and Additional Application Files)
* Adjusts where certain sub types of NoW documents can be displayed/uploaded

image

image

# Other
* Fixes some typos in the description of some NoW document types


services/core-web/src/components/noticeOfWork/applications/administrative/NOWApplicationAdministrative.js
```diff
@@ -67,12 +65,12 @@ export const NOWApplicationAdministrative = (props) => {
         <ScrollContentWrapper id="generated-documents" title="Application Export Files">
           <NOWDocuments
             documents={props.noticeOfWork.documents.filter(
-              ({ now_application_document_type_code }) =>
-                exportedDocuments.includes(now_application_document_type_code)
+              ({ now_application_document_sub_type_code }) =>
+                now_application_document_sub_type_code === "AEF"
             )}
             isViewMode
             disclaimerText="This table shows all of the PDF files created from the edited Notice of Work form."
```
exzizt, line 72: Since this will also now show PMT and PMA documents, do we want to adjust this text?

exzizt:
Should we add a "not null" constraint on sub-type to ensure each NoW document type has one going forward?

services/core-web/src/components/noticeOfWork/applications/administrative/NOWApplicationAdministrative.js
mmumby: That's a good suggestion, Does this table include all the generated documents? Client acknowledgment, permit enclosed, withdrawal, etc?

exzizt: Nope, those letters are classified as GDO and will appear in the Government Documents section. They only classified NTR, PMT, and PMA as "Application Export Files".

mmumby (commented):
Looks good! 

I agree that the message on the table should be updated to reflect all documents available in that table, Happy to approve when that is updated!

services/core-web/src/components/noticeOfWork/applications/administrative/NOWApplicationAdministrative.js
mmumby: Makes sense

mmumby (dismissed):

jon-funk (approved):
Good catch!

PR #1704, opened 2021-04-08
MDS-3589: Add NoW Number to Generated NoW Document Names

exzizt:
# Main
* Adds the NoW Number to the beginning of the document name for all generated NoW documents
image

### DIFF
--- a/services/core-api/app/api/services/document_generator_service.py
+++ b/services/core-api/app/api/services/document_generator_service.py
@@ -34,10 +34,11 @@ class DocumentGeneratorService():
             cls._push_template(document_template, fileobj)
 
         # Create the document generation request body
+        document_name_extra = template_data.get('document_name_extra')
+        document_name_extra = f' {document_name_extra}' if document_name_extra else ''
         date_string = datetime.date.today().strftime("%Y-%m-%d")
-        draft_string = " DRAFT" if 'is_draft' in template_data and template_data[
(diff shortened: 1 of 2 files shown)


services/core-api/app/api/services/document_generator_service.py
```diff
@@ -34,10 +34,11 @@ def generate_document(cls, document_template, template_data):
             cls._push_template(document_template, fileobj)
 
         # Create the document generation request body
+        document_name_start_extra = template_data.get('document_name_start_extra')
+        document_name_start_extra = f'{document_name_start_extra} ' if document_name_start_extra else ''
         date_string = datetime.date.today().strftime("%Y-%m-%d")
```
mhuseinov, line 39: Should this be a UTC date?

exzizt: Good question. I'm not sure what the users are expecting. Probably the date to be in their local timezone, perhaps, but we currently don't have the user send over that information in a document generation request...

I think this is something that will be tackled in the larger datetime/timezone standardization ticket. What do you think?

mhuseinov: Is it easy to add user local date to generate document token? I think it might be confusing if you are looking for the document and the date of the newly generated one is tomorrow.

mhuseinov (approved):

PR #1706, opened 2021-04-09
MDS-added filter based on now type

mhuseinov:


mmumby (approved):
🔥

services/core-api/app/api/now_applications/resources/now_application_list_resource.py
```diff
@@ -100,10 +103,13 @@ def _apply_filters_and_pagination(self,
...
                                       submissions_only=None,
                                       import_timestamp_since=None,
-                                      update_timestamp_since=None):
+                                      update_timestamp_since=None,
+                                      application_type=None):
 
         filters = []
         base_query = ApplicationsView.query
+        if application_type:
+            filters.append((and_(ApplicationsView.application_type_code == application_type)))
```
exzizt, line 112: You can remove the `and_` function here since there is just one condition

mmumby (approved):

PR #1709, opened 2021-04-14
Feature/mds 3338 add status for withdrawn

mhuseinov:
Adds a withdrawn (WDN) to notice of work
 
image
image
image
image


services/core-api/app/api/now_applications/resources/administrative_amendment_list_resource.py
```diff
@@ -50,7 +50,7 @@ def post(self):
         has_existing_administrative_amendments = ApplicationsView.query.filter_by(
             source_permit_amendment_guid=permit_amendment.permit_amendment_guid,
             application_type_code='ADA').filter(
-                ApplicationsView.now_application_status_code.notin_(['AIA', 'REJ'])).count() > 0
+                ApplicationsView.now_application_status_code.notin_(['AIA', 'REJ', 'WDN'])).count() > 0
```
mmumby, line 54: What is the purpose of this filter?

mmumby (approved):
🔥  Looks good, I have one non-blocking question.

services/core-api/app/api/now_applications/resources/administrative_amendment_list_resource.py
mhuseinov: That is used to check that you cannot have multiple in-progress administrative amendments. The WDN status is one of the statuses for complete ADA, the same as REJ and AIA

PR #1718, opened 2021-04-26
MDS-3655: Fix "Sand & Gravel / Quarry Operations" Not Being Rendered in NoW Form

exzizt:
# Main
* Fixes an issue where the "Sand & Gravel / Quarry Operations" section was not being rendered in the NoW form document, due to inconsistent naming of the NoW activity (the activity name is "sand_gravel_quarry_operation" but we were sometimes referencing it as "sand_and_gravel", which is not the actual activity name.
image

# Other
* Removes the "Permit Application Fee Assessment" section from the NoW form document and adds in a "Maximum Annual Tonnage" section.
image


mhuseinov (approved):
💯

mmumby (commented):
Does the conditional rendering on the UI still work as expected with the name change? 
Just wanted to confirm as 'activityConditions' uses the response model to render the sections on the sideNavBar on the UI.

PR #1727, opened 2021-05-10
MDS-fix for rejection and withdrawal

mhuseinov:
- fix for rejection and withdrawal modals to open without draft permit started,
- removed redundant validation from disabled fields on generating draft permit form


mmumby (approved):
🔥

PR #1730, opened 2021-05-17
Remove Required auth_end_date for C & M Permit

mmumby:
- Major Mine permits do not set authentication end dates on their permits, removed the required field is the Permit is M or C
- If the permit type is changed In the "AddPermitForm", reset the is_exploration flag to null


services/core-web/src/components/maps/MineTailingsMap.js
```diff
@@ -271,7 +271,6 @@ export class MineTailingsMap extends Component {
   }
 
   renderPopup = (tailing) => {
-    console.log("I was clicked?S?S?S");
```
nayr974, line 274: 🤣

nayr974 (approved):

PR #1731, opened 2021-05-18
MDS-3297 no permit required

JustinMacaulay:
Adds process functionality for the "No Permit Required" Status.
    - 2 new templates for No Permit Required and No Permit Required IP
    - Changes status to no permit required when finished and allows for reverting back to previous status.


exzizt:
Look for this code in document_template.py to add support for inserting images for the new doc types:
```
        if self.document_template_code in ('PMT', 'PMA', 'NPE', 'NCL', 'NWL', 'NRL'):
            doc = docx.Document(self.os_template_file_path)
            insert_images(doc, template_data)
            ```

File: .gitignore [hunk 1]
```diff
@@ -8,6 +8,7 @@ coverage/
 .idea/
 .vscode/launch.json
 .vscode/tasks.json
+.vscode/settings.json
```
exzizt, line 11: I don't think we want to ignore this file. It's to ensure we all use the same styling plugins etc.

migrations/sql/V2021.04.06.09.53__update_document_templates.sql
```diff
@@ -1,3 +1,18 @@
+INSERT INTO document_template
```
exzizt, line 1: Changes to this migration from April 6 should be reverted? I think this is one of the changes you were going to make already.

migrations/sql/V2021.05.10.13.29__update_document_templates.sql
```diff
@@ -0,0 +1,111 @@
...
+      "id": "rc_office_mailing_address_line_2",
+      "relative-data-path": "mine.region.regional_contact_office.mailing_address_line_2",
+      "read-only": true
+    },
+    {
+      "id": "application_type_code",
+      "relative-data-path": "now_application.application_type_code",
+      "read-only": true
+    }
+  ]' WHERE document_template_code ='NWL';
```
exzizt, line 111: I think this file is revolved around another change you were going to make? Shouldn't be updating the Withdrawal Letter?

services/core-api/app/api/now_applications/resources/now_application_status_resource.py
```diff
@@ -312,7 +312,7 @@ def check_if_equal(list_1, list_2):
                 current_app.logger.info(str(ex))
 
         # Handle rejected and withdrawn statuses
-        elif now_application_status_code in ['REJ', 'WDN']:
+        elif now_application_status_code in ['REJ', 'WDN', 'NPR']:
```
exzizt, line 315: What about the NPI code here?

services/core-web/src/components/modalContent/NoPermitRequiredSelectionModal.js
```diff
@@ -0,0 +1,76 @@
...
+            type="error"
+            showIcon
+          />
+          <br />
+        </>
+      )}
+      <Select
+        virtual={false}
+        labelInValue
+        placeholder="Select a No Permit Required Letter..."
```
exzizt, line 46: Can we remove the ellipsis? Don't usually see this in our placeholder text. Maybe "Select Letter Type"?

services/core-web/src/components/modalContent/NoPermitRequiredSelectionModal.js
```diff
@@ -0,0 +1,76 @@
...
+          title="Are you sure you want to cancel?"
+          onConfirm={props.closeModal}
+          okText="Yes"
+          cancelText="No"
+        >
+          <Button className="full-mobile" type="secondary">
+            Cancel
+          </Button>
+        </Popconfirm>
+        <Button className="full-mobile" type="primary" onClick={handleNextStep} disabled={disabled}>
```
exzizt, line 65: Don't forget to set the "loading" prop when the form is submitting.

exzizt (commented):
Let me know when you've pushed the final changes and I'll take a look. Nice work!

.gitignore
JustinMacaulay: The problem is if I do then i will be pushing the location of my python interpreter to all of you which will not make sense.

migrations/sql/V2021.04.06.09.53__update_document_templates.sql
JustinMacaulay: yeah I 100% coppied that migration and then modified the original and left the copy... I will fix that

migrations/sql/V2021.05.10.13.29__update_document_templates.sql
JustinMacaulay: This was the original

services/core-web/src/components/modalContent/NoPermitRequiredSelectionModal.js
JustinMacaulay: Done.

services/core-web/src/components/modalContent/NoPermitRequiredSelectionModal.js
JustinMacaulay: this isnt a form and it doesnt submit so I dont think I can include the loading stuff.

.gitignore
exzizt: Don't commit your changes to settings.json then? 😕 
We still want this file to be consistent between all of us

services/core-web/src/components/modalContent/NoPermitRequiredSelectionModal.js
exzizt: We will need to set the loading state some way to prevent spamability of the button.

exzizt: Ah I checked out the code and understand the flow better now.

exzizt (approved):
✅

PR #1736, opened 2021-05-25
Bug/permission updates delete permit amendment

mmumby:
- Anyone with edit permit permissions can delete permit amendments via the Drafting flow, the FE permissions we correct, updated the BE to match.


exzizt (approved):
✅

PR #1737, opened 2021-05-25
Bug/last amended date fix

mmumby:
- The Permit table was showing the draft permit date for the "Last Amended Date" instead of the last issued amendment. 
- Filtered out draft permits. 

Before: (production example)
Screen Shot 2021-05-25 at 4 15 53 PM

After (Local Example) 
Screen Shot 2021-05-25 at 4 16 19 PM


services/core-web/src/components/mine/Permit/MinePermitTable.js
```diff
@@ -465,7 +465,9 @@ const transformRowData = (
   openAddPermitHistoricalAmendmentModal,
   openEditSitePropertiesModal
 ) => {
-  const latestAmendment = permit.permit_amendments[0];
+  const latestAmendment = permit.permit_amendments.filter(
+    (a) => a.permit_amendment_status_code !== draftAmendment
+  )[0];
```
exzizt, line 470: Won't this cause an index out of range error if the user goes to this page and they only have one permit amendment and it happens to be in draft?

mmumby: It shouldn't. 

cases: 
A 'new permit' is added via this page - Always has an initial amendment with an issue_date
A 'new permit' is added via the NoW - Permits with status D are filtered out and not shown on this page

exzizt (approved):

PR #1739, opened 2021-05-26
Permittee relationship incorrect on Permit model

nayr974:


services/core-api/app/api/now_applications/resources/administrative_amendment_list_resource.py
```diff
@@ -107,21 +108,32 @@ def post(self):
...
-                        mine_party_appt_type_code=mine_appt.mine_party_appt_type_code,
-                        now_application_id=new_app.now_application_id,
-                        party_guid=mine_appt.party_guid)
+            # contacts
+            application_appt = []
+            # if there is a prior imported NoW, copy contacts from it.
+            if application and application.contacts:
+                now_appts = [
+                        party for party in application.contacts
+                        if not party.end_date or party.end_date > datetime.now(timezone.utc).date() # valid?
```
mhuseinov, line 117: I think we do not need this comment anymore
"# valid?"

nayr974: I still have the question, if we are copying contacts from the prior NoW, does it matter what end date they had set?

Alternatively, was the intent to copy all current contacts from the mine? **Not** the prior NoW's contacts?

exzizt:
Would the most minimal solution be to rename "permittee_appointments" to "party appointments" (because that's what it is) and then to change the current_permittee hybrid property to filter by PMT?

nayr974:
> Would the most minimal solution be to rename "permittee_appointments" to "party appointments" (because that's what it is) and then to change the current_permittee hybrid property to filter by PMT?

Possibly, which is why I put the PR up as is and wanted to hear what the intent of the code was. 

I'm not sure the existing logic satisfies the intent, because it's all of the parties **related by permit_id,** which in the end is probably only permittees and tenure holder? Not all of the contacts that would have been on the prior NoW, if that's desired.

And if it's only permittee and tenure holder that are actually desired, permit_id may work today but not guarantee it in the future if other roles are added.

mhuseinov (approved):
💯

PR #1741, opened 2021-05-31
MDS-3722 Docgen and Document Manager running locally

JustinMacaulay:
Adds 2 scripts to modify the host file on linux, mac, and windows to allow the docgen and document manager to run when the frontend is run locally.


exzizt:
I noticed you added the Bug tag, did you need help resolving an issue?

exzizt (dismissed):
Can't wait to try this fix out!
Nice work 💯

exzizt (commented):
Let's get someone to test it on Windows before we merge it in (I'll try to do it today).

USAGE.md
```diff
@@ -122,6 +121,26 @@ To address this, you may:
 psql --dbname=mds --username=mds --host=localhost --password --port=5432
 ```
 
+### Using the Document Manager and Document Generator locally
+
+If you are running the frontend using npm run serve then you will not be able to use the document manager at the same time as the document generator. If you wish to do this then you need to make an addition to your hosts file so the browser can resolve the document_manager_backend to localhost.
+
+If you are on a windows machine or using power shell run the folowing command at the root of this project:
```
exzizt, line 128: Maybe edit the note to say that it needs to run in administrator mode in the power shell:
image

exzizt (commented):
I've spent 10 mins trying to get it to work and have been unsuccessful so far.

I think you need to update some env-examples as well, and also the "docManUrl" in environment.js files?

JustinMacaulay:
I cant find a way to force powershell to run as an administrator like I can for linux and Mac. To run that you need to make sure the powershell you run the command in is in admin mode. I will add that to the readme.

exzizt:
> I cant find a way to force powershell to run as an administrator like I can for linux and Mac. To run that you need to make sure the powershell you run the command in is in admin mode. I will add that to the readme.

Right click and run as administrator if you search for it in the Start Menu.

exzizt (approved):
image

PR #1743, opened 2021-06-02
Update/vfcbc submission data

mmumby:


services/core-api/app/api/now_applications/models/activity_summary/sand_gravel_quarry_operation.py
```diff
@@ -52,14 +54,27 @@ class SandGravelQuarryOperation(ActivitySummaryBase):
...
     reclamation_backfill_detail = db.Column(db.String)
+    proposed_activity_description = db.Column(db.String)
+    work_year_info = db.Column(db.String)
 
     details = db.relationship(
         'SandGravelQuarryOperationDetail',
         secondary='activity_summary_detail_xref',
         load_on_pending=True)
 
+    # TODO replace with value from vFCBC
```
mhuseinov, line 70: Is this comment still valid?

services/core-web/src/components/noticeOfWork/applications/review/activities/SurfaceBulkSamples.js
```diff
@@ -52,10 +53,11 @@ export const SurfaceBulkSamples = (props) => {
       />
       <br />
       <br />
+      <h4>Processing Methods</h4>
       <Row gutter={16}>
         <Col md={12} sm={24}>
           <div className="field-title">
-            Processing Methods*
+            Describe handling and on-site processing methods
```
mhuseinov, line 60: Should this field still be required? (the "*" sign)

services/core-api/app/api/now_applications/transmogrify_now.py
```diff
@@ -29,6 +29,12 @@ def code_lookup(model, description, code_column_name):
         result = None
     return result
 
+def get_boolean_value(field):
```
mhuseinov, line 32: 👍

mhuseinov: However, this changes the original logic for boolean field processing. Is it intentional?

mmumby: Yes the original logic ignoring null values, so if the string wasn't "Yes" it was defaulting to false, even if the field was null.

services/core-api/app/api/now_applications/models/activity_summary/sand_gravel_quarry_operation.py
mmumby: Yes, this value has not been sent to us. If "other" is selected, users can populate a input field. So i'm just updating the checkbox until we have that value
Screen Shot 2021-06-09 at 12 13 51 PM

services/core-web/src/components/noticeOfWork/applications/review/activities/SurfaceBulkSamples.js
mmumby: no, the field isn't required on the form, I've updated the name and removed the required validation.

services/core-web/src/components/noticeOfWork/applications/review/activities/UndergroundExploration.js
```diff
@@ -28,11 +32,198 @@ const propTypes = {
...
+                <NOWOriginalValueTooltip
+                  originalValue={
+                    props.renderOriginalValues("underground_exploration.proposed_diamond_drilling")
+                      .value
+                  }
+                  isVisible={
+                    props.renderOriginalValues("underground_exploration.proposed_diamond_drilling")
+                      .edited
+                  }
+                  style={{ marginLeft: "5px" }}
```
mhuseinov, line 113: I've noticed this styling in several places, do we want to have a css class for it?

mmumby: yes, good idea. I'll update that

mmumby: Okayy - I switched this to  a class with the same code, but since this is getting passed into a different component the className does not override correctly. So I reverted the changes back to the style

below is where the component reads this prop:
```
<span style={props.style} className="violet">
        Edited
      </span>
      ```

mhuseinov (approved):
💯

mhuseinov (approved):

PR #1748, opened 2021-06-04
Feature/mds  update the application nda status validation

mhuseinov:


JustinMacaulay:
Need to add Raise(NotImplemented) in the specific areas where this is called.

PR #1752, opened 2021-06-09
Bug/withdrawal status

mmumby:
The withdrawn status was not disabling the process button or allow admin to revert Status

Before: 
Screen Shot 2021-06-09 at 9 17 57 AM

After: 
Screen Shot 2021-06-09 at 9 18 26 AM


services/core-web/src/tests/components/modalContent/__snapshots__/AddReportModal.spec.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`AddReportModal renders properly 1`] = `
   <SlidingForms
     formContent={
       Array [
-        <Memo(Connect(ReduxForm))
+        <UNDEFINED
```
exzizt, line 8: UNDEFINED?

mmumby: ugh. hmmm... why did that happen in the snapshot...

mmumby: I've npm installed, and updated snapshots, nothing changed though. Thoughts?

exzizt (approved):

PR #1756, opened 2021-06-14
MDS 3584 preview docs before generating

JustinMacaulay:
You can now preview docs before generating them. When you click the preview button it will generate the document and download it for you to preview. It will not save it to the doc manager. You can then continue editing and then generate the document.


services/core-web/src/components/Forms/GenerateDocumentForm.js
```diff
@@ -104,6 +107,16 @@ export const GenerateDocumentForm = (props) => {
             Cancel
           </Button>
         </Popconfirm>
+        <Button
```
exzizt, line 110: I think we should return the promise from the onClick props.preview and set an "isGeneratingPreview" flag in the state here to set the loading prop of this button to prevent users from spam clicking it. Buttons should not be spammable while they are already performing the initial request.

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
```diff
@@ -531,6 +533,27 @@ export class ProcessPermit extends Component {
...
+    documentType.document_template.form_spec
+      .filter((field) => field.type === "DATE")
+      .forEach((field) => {
+        newValues[field.id] = formatDate(newValues[field.id]);
+      });
+    const payload = {
+      now_application_guid: this.props.noticeOfWork.now_application_guid,
+      template_data: newValues,
+    };
+    this.props.generateNoticeOfWorkApplicationDocument(
```
exzizt, line 548: Will need to return this function call

services/core-web/src/components/Forms/GenerateDocumentForm.js
```diff
@@ -104,6 +107,16 @@ export const GenerateDocumentForm = (props) => {
...
+          className="full-mobile"
+          type="secondary"
+          disabled={props.disabled}
+          onClick={() => {
+            props.preview(props.documentType, Object.assign({}, props.allCurrentValues));
```
exzizt, line 115: Why are you using Object.assign here?

JustinMacaulay: If I dont create a copy of the form values when I generate a preview it modifies the dates in the actual form and then they dont work

services/core-web/src/components/noticeOfWork/applications/process/ProcessPermit.js
JustinMacaulay: added

services/core-web/src/components/Forms/GenerateDocumentForm.js
JustinMacaulay: I pulled it out into a handle preview function and added some state to handle disabling the button.

services/core-web/src/components/Forms/GenerateDocumentForm.js
```diff
@@ -104,6 +108,16 @@ export const GenerateDocumentForm = (props) => {
             Cancel
           </Button>
         </Popconfirm>
+        <Button
+          className="full-mobile"
+          type="secondary"
+          disabled={props.previewGenerating}
```
exzizt, line 114: Can you change disabled to "loading"? 😄

exzizt (approved):
💯 !

PR #1757, opened 2021-06-14
MDS-3670: GIS Export Updates

exzizt:
# Main
* Adds some new columns to the GIS Export View:
  * Mine work start date
  * Mine work stop date
  * Mine last inspected date
  * Mine last inspection type
  * NoW total disturbed area
  * NoW total disturbed areas for each activity type
  * Number of days the NoW is in Client Delay
# Other
* Also adds messageid and mms_cid
* Fixes the naming of some "afterMigrate" migrations so that they run in the intended order (e.g., 100 was running before 1)


jon-funk (dismissed):

migrations/sql/R__view_now_application_gis_export.sql
```diff
@@ -1,3 +1,17 @@
+-- Creation of these two NRIS schema tables allows this migration to succeed locally.
+CREATE SCHEMA IF NOT EXISTS nris;
```
nayr974, line 2: This doesn't impact the NRIS container's ability to create the schema itself with it's code first migrations?

nayr974 (commented):
Your air fryer is stupid, and you're stupid for thinking it was a justifiable purchase. The only excuse for having an air fryer is that someone gifted it to you, and after you cut them out of your life forever you couldn't find a pawn shop that would take it, and when you tried to throw it out in the dumpster the city fined you for dumping such an eyesore, and after you were court ordered to keep it in your home where no human being would ever see it again(because nobody who is associated with you can be considered "human"). You tried to fight the court order in a legal battle, but the public defender argued against you because after representing serial killers, child rapists, and war criminals he couldn't bring himself to represent someone so detestable as to own an air fryer. With all of that, I couldn't forgive you for owning an air fryer, but I would understand, in the same way you understand why Mao Zedong made the great leap forward which killed upwards of 40 million people. The fact that you argue in favour of it is so disgusting you should be stripped of all human rights, bound and thrown off of a boat in the middle of the arctic ocean tied to that worthless waste of metal and engineering so that it can drag you to the watery grave you deserve.

migrations/sql/R__view_now_application_gis_export.sql
exzizt: image

jon-funk (approved):
A Tech. debt ticket has been created to look at resolving the cross-service dependency in the future. Approving to make this view available for the business side.

PR #1778, opened 2021-06-25
ESUPS

mmumby:
# Explosives Storage and Use Permit Applications and Historic Permit CRUD
- Ability to add an ESUP Application
- Ability to Process an ESUP Application and generate a letter, and a permit certificate
- Ability to Withdrawl/Reject an Application
- Ability to Edit an application
- Ability to Create a historic record on the permit tab
- Ability to "close" an ESUP
- Ability to upload documents with multiple categories
- Ability to View the Magazine Details
Screen Shot 2021-07-09 at 5 32 57 PM
Screen Shot 2021-07-09 at 5 33 23 PM
Screen Shot 2021-07-09 at 5 33 58 PM
Screen Shot 2021-07-09 at 5 34 20 PM
Screen Shot 2021-07-09 at 5 36 07 PM
Screen Shot 2021-07-09 at 5 37 18 PM
Screen Shot 2021-07-09 at 5 37 43 PM
Screen Shot 2021-07-09 at 5 38 11 PM
Screen Shot 2021-07-09 at 5 38 29 PM
Screen Shot 2021-07-09 at 5 38 48 PM
Screen Shot 2021-07-09 at 5 39 07 PM

- Added an Alert to the Blasting section of the NoW if an ESUP is needed
- Add a warning to the Process Tab if an ESUP is needed
Screen Shot 2021-07-09 at 5 31 53 PM
Screen Shot 2021-07-09 at 5 32 10 PM


migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -717,16 +717,19 @@ ON CONFLICT DO NOTHING;
 -- V2019.09.28.14.16
 
 INSERT INTO document_template
-(document_template_code, form_spec_json, template_file_path, active_ind, create_user, update_user)
+(document_template_code, form_spec_json, template_file_path, active_ind, create_user, update_user, source_model_name)
```
JustinMacaulay, line 720: I totally get this change but the table has a default that I am wondering if we should keep or remove

services/core-api/app/api/document_generation/models/document_template.py
```diff
@@ -48,13 +54,13 @@ def _form_spec_with_context(self, primary_key):
         source_obj_instance = source_model.query.get(primary_key)
         current_app.logger.debug(f'source_obj_instance -> {source_obj_instance}')
         if not source_obj_instance:
-            raise Exception('Context Object not found')
+            current_app.logger.warning(f'Context Object not found')
```
JustinMacaulay: why was this turned into a logging statement instead of an exception? Shouldn't this be a raised HTTP error?

services/core-api/app/api/document_generation/models/document_template.py
```diff
@@ -159,7 +165,7 @@ def delete_paragraph(paragraph):
                         break
 
         doc = None
-        if self.document_template_code in ('PMT', 'PMA', 'NPE', 'NCL', 'NWL', 'NRL'):
+        if self.document_template_code in ('PMT', 'PMA', 'NPE', 'NCL', 'NWL', 'NRL', 'ESP', 'ESL'):
```
JustinMacaulay: If we are going to keep extending this shoutld it be a constant or a field on the data table?

services/core-api/app/api/document_generation/models/document_template.py
```diff
@@ -1,16 +1,22 @@
 import json, os, docx, io, base64
 
 from flask import current_app
+from flask_restplus import inputs
```
JustinMacaulay: This import was added but I dont see anything using it.

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
```diff
@@ -0,0 +1,108 @@
...
+from app.api.mines.explosives_permit.models.explosives_permit import ExplosivesPermit
+from app.api.mines.explosives_permit.models.explosives_permit_document_type import ExplosivesPermitDocumentType
+from app.api.mines.explosives_permit.models.explosives_permit_document_xref import ExplosivesPermitDocumentXref
+
+from app.api.services.document_generator_service import DocumentGeneratorService
+from app.api.services.document_manager_service import DocumentManagerService
+from app.api.mines.explosives_permit.response_models import EXPLOSIVES_PERMIT_DOCUMENT_MODEL
+
+
+class ExplosivesPermitDocumentResource(Resource, UserMixin):
```
JustinMacaulay: the filename should reflect the class name so explosives_permit_document_resource.py

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
```diff
@@ -0,0 +1,108 @@
...
+                                            commit=True):
+        # Ensure that the token is valid
+        token_data = cache.get(EXPLOSIVES_PERMIT_DOCUMENT_DOWNLOAD_TOKEN(token))
+        cache.delete(EXPLOSIVES_PERMIT_DOCUMENT_DOWNLOAD_TOKEN(token))
+        if not token_data:
+            raise BadRequest('Valid token required for download')
+
+        # Get the template associated with the token
+        document_type_code = token_data['document_type_code']
+        explosives_permit_document_type = ExplosivesPermitDocumentType.query.unbound_unsafe().get(
```
JustinMacaulay: Is there a reason this needs to use the unbound_unsafe query?

services/core-api/app/api/now_applications/response_models.py
```diff
@@ -3,23 +3,7 @@
...
-        "context-value": fields.String,
-        "read-only": fields.Boolean(default=False),
-    })
-
-DOCUMENT_TEMPLATE_MODEL = api.model(
-    'DocumentTemplateModel', {
-        'document_template_code': fields.String,
-        'form_spec': fields.List(fields.Nested(DOCUMENT_TEMPLATE_FIELD_MODE, skip_none=True))
-    })
+from app.api.document_generation.response_models import DOCUMENT_TEMPLATE_MODEL
```
JustinMacaulay, line 6: does this actually need to be imported here?

JustinMacaulay (commented):
just wanted to get some comments on the PR still reviewing though.

services/core-api/app/api/document_generation/models/document_template.py
exzizt: It could be a constant, but it would be a constant that is only used in this file.

services/core-api/app/api/document_generation/models/document_template.py
exzizt: At the time, I wanted the presence of the context object to be optional so you were able to generate/preview an ESUP document without actually having created the ESUP yet. If the GUID doesn't exist, all of the context values are just set to null.
I think I can undo these changes now though as the flow has changed.

migrations/sql/afterMigrate__1__seed_data.sql
exzizt: The default could probably be removed at this point.

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
exzizt: Agree, but I borrowed the same name format as the `now_document.py` file in the same folder. Will change.

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
exzizt: This file is pretty much the same as the now_document.py generation resource in the same folder--I can't remember why exactly, but it might have to do with supporting being able to generate the document outside of a request.

services/core-api/app/api/now_applications/response_models.py
exzizt: Yeah, it is used below.

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
exzizt: I also think we could probably do with a tech-debt ticket to "genericize" our document_generation resources. A lot of the code each one uses is largely the same with some minor differences.

services/core-api/app/api/document_generation/models/document_template.py
JustinMacaulay: does it make sense for it to be a part of the data model?

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
JustinMacaulay: Ah, those should both be dealt with then

services/core-api/app/api/document_generation/resources/explosives_permit_document.py
JustinMacaulay: I agree

services/core-api/app/api/document_generation/models/document_template.py
jon-funk: Since it's only used here I think it makes sense to keep it to the file so it requires less navigation to find the value. I'm up for putting it in the data table, but I don't think it's quite long enough "yet"

services/core-api/app/api/document_generation/namespace.py
```diff
@@ -1,6 +1,8 @@
 from flask_restplus import Namespace
-from app.api.document_generation.resources.now_document import NoticeOfWorkDocumentResource
+from app.api.document_generation.resources.now_document_resource import NoticeOfWorkDocumentResource
+from app.api.document_generation.resources.explosives_permit_document_resource import ExplosivesPermitDocumentResource
 
 api = Namespace('documents', description='Generate documents for business objects.')
 
-api.add_resource(NoticeOfWorkDocumentResource, '')
```
jon-funk, line 6: What was this doing prior to this change? The empty string looks a bit suspect, or I'm misunderstanding.

exzizt: This used to be the only resource (for NoW document generation) and was just accessed via just `documents/`, weirdly enough. But now it is `documents/notice-of-work` as we had to add another resource `documents/explosives-permit`.

exzizt: I mentioned above too in a reply to Justin that we should try to generalize the document generation endpoint in a tech debt ticket.

services/core-api/app/api/document_generation/models/document_template.py
exzizt: Y'know what, I will adjust it so it's not even necessary 😄 It will just attempt to insert images into all templates.

It's too bad document insertion is not an open feature with Carbone and we had to implement it ourselves. We have to do the image insertion into the templates "dynamically" before sending the template document file to the document generator, which prevents us from using their caching feature (which is why this check was put in in the first place).

migrations/sql/afterMigrate__1__seed_data.sql
exzizt: Dropped

services/core-api/app/api/mines/explosives_permit/resources/explosives_permit_list.py
```diff
@@ -0,0 +1,179 @@
...
+        explosives_permits = ExplosivesPermit.find_by_mine_guid(mine_guid)
+        return explosives_permits
+
+    @api.doc(
+        description='Create a new Explosives Permit.',
+        params={'mine_guid': 'The GUID of the mine to create the Explosives Permit for.'})
+    @api.expect(parser)
+    @requires_role_edit_explosives_permit
+    @api.marshal_with(EXPLOSIVES_PERMIT_MODEL, code=201)
+    def post(self, mine_guid):
```
jon-funk, line 154: This is the list resource, shouldn't this be in the base resource since we're posting a single record?
On second thought - is that from the mapping between namespace URLs and the resource classes?

tasks/now-etls/mms_now_import.py
```diff
@@ -1182,17 +1182,17 @@ def ETL_MMS_NOW_schema(connection, tables):
 
         applications = etl.leftjoin(applications, timber_cutting, key='mms_cid')
 
-        explosive_permits = etl.fromdb(
+        EXPLOSIVES_PERMITS = etl.fromdb(
```
jon-funk: Reason for capitals?

services/core-web/src/components/noticeOfWork/applications/review/activities/Blasting.js
```diff
@@ -331,6 +336,43 @@ export const Blasting = (props) => {
...
+                    <li>
+                      Navigate to the `Permits & Approvals` tab on the{" "}
+                      <Link
+                        to={router.MINE_NOW_APPLICATIONS.dynamicRoute(props.mineGuid)}
+                        title="Mine"
+                      >
+                        Mine Record
+                      </Link>
+                    </li>
+                    {/* <li>Click On Permits and Approvals</li>
```
jon-funk: Remove this comment?

jon-funk (changes requested):

services/core-web/src/components/noticeOfWork/applications/review/activities/Blasting.js
mmumby: removed! good catch

services/core-api/app/api/mines/explosives_permit/resources/explosives_permit_list.py
exzizt: We have been putting the POST resource in the "list" resources in our project. It's because of the naming scheme for accessing the resource.

tasks/now-etls/mms_now_import.py
exzizt: Oh, that looks like a mistake. Not sure how that was changed. Great catch!

exzizt: @mmumby ?

services/core-api/app/api/document_generation/namespace.py
jon-funk: Nice, do we have a link to the ticket we can post here for historical purposes?

jon-funk (approved):
huge  (hyo͞oj)
adj. hug·er, hug·est
1. Of exceedingly great size, extent, or quantity. See Synonyms at enormous.
2. Of exceedingly great scope or nature: the huge influence of the Hellenic world.
3. Informal Contributing in a major way to success; very important: The defensive line was huge in the second half.

PR #1789, opened 2021-07-09
MDS-3639 Update the HSRC document and MEMP documents for download in Core on the landing page

areyeslo:
Ticket:  https://bcmines.atlassian.net/browse/MDS-3639

- Added new PDF documents
- Updated references to the new PDF documents recently added.
- Updated tests/components/naviation/__snapshots__/NavBar.spec.js.snap file.


exzizt (commented):
Good work :) Don't forget to delete the two old files these ones replaced.

mmumby (dismissed):

PR #1791, opened 2021-07-13
vFCBC data bug

mmumby:
- Updated logic for showing addresss on NoW contacts - These columns in the db are nullable, so 'and' was preventing the values to show up

before -> 
Screen Shot 2021-07-13 at 8 26 16 AM

After -> 
Screen Shot 2021-07-13 at 8 26 47 AM

- Added a boolean check for a field


exzizt (dismissed):
🏣 💯

exzizt (approved):

PR #1804, opened 2021-07-20
MDS-3816: Fix Site Properties Bug

exzizt:
# Main
- Fixes an issue where new site properties for permits were not being created


jon-funk (approved):
🔍 🐛

PR #1806, opened 2021-07-21
MDS 3772 issue date on permits

JustinMacaulay:
Changes the variables and logic on the permit template to set the issue original issue date based on if it is an Draft permit or amendment.


mmumby (dismissed):
Looks good! 
Looks like you have a failing BE test though

mmumby (approved):

PR #1808, opened 2021-07-23
Feature/mds 3793 - Update NoW PDF with new vFCBC data

mmumby:
# Main
- Updated the NoW Pdf with the additional vFCBC data fields
- Added some conditional rendering to a few sections
- Updated the document logic to only include documents in the permit package - merged both the submission and uploaded documents together
- Updated the styling of a few 'Edited' tooltips to bring them closer to the title

Screen Shot 2021-07-23 at 9 00 36 AM


services/core-api/app/api/mines/tailings/models/tailings.py
```diff
@@ -32,8 +32,22 @@ def json(self):
         }
 
     @classmethod
-    def create(cls, mine, mine_tailings_storage_facility_name, latitude, longitude, consequence_classification_status_code, has_itrb, tsf_operating_status_code,  add_to_session=True):
-        new_tsf = cls(mine_tailings_storage_facility_name=mine_tailings_storage_facility_name, latitude=latitude, longitude=longitude, consequence_classification_status_code=consequence_classification_status_code, has_itrb=has_itrb, tsf_operating_status_code=tsf_operating_status_code)
+    def create(cls,
```
mmumby, line 35: Accidentally formatted this file. Added nothing.

services/core-web/src/components/noticeOfWork/applications/review/ReviewNOWApplication.js
```diff
@@ -473,6 +473,7 @@ export const ReviewNOWApplication = (props) => {
             Access presently gated
             {props.isPreLaunch && <NOWFieldOriginTooltip />}
             <NOWOriginalValueTooltip
+              style={{ marginLeft: "5%" }}
               originalValue={props.renderOriginalValues("is_access_gated").value}
```
mmumby, line 477: There is a reason I didn't use a class for this - 
NOWOriginalValueTooltip expects a style prop, and then passes that prop through as `style={{props.style.marginLeft}}`, This needs to be a style prop to override the default class on the component. 

I could put the '5%' into a constant, but it would still be passed through in the same way 
`style={{ marginLeft: CONSTANT_FIVE_PERCENT }}` doesn't seem to clean it up much.

services/core-api/app/api/mines/tailings/models/tailings.py
jon-funk: In the future, you can `git checkout <file>` to revert that file to the state of the commit HEAD is pointing at (usually the last commit of the current branch) and remove all changes. But there's nothing wrong with consistent formatting :D

Warning: using that on a directory will change all contents of a directory.

services/core-api/app/api/now_applications/resources/now_application_export_resource.py
```diff
@@ -320,7 +388,20 @@ def transform_documents(now_application):
                     doc['now_application_document_type_code'])
                 doc['now_application_document_type_description'] = get_description(
                     document_type, doc['now_application_document_type_code'])
-                included_docs.append(doc)
+                if doc['is_final_package']:
```
jon-funk, line 391: Are there any potential side effects to other docs without this key?

mmumby: Not that I could see, all Docs will have this key, but it may be null or false, in those cases its ignored.

jon-funk (approved):
💯 📝 👍

PR #1809, opened 2021-07-26
Set node_env in pipeline for each namespace

jon-funk:
# Main
- Sets NODE_ENV explicitly per namespace. This will enable the test banner in the test namespace.


jon-funk:
The event change was done to address an edge case where tests are required but no code changes were made under the applicable services/ directories so they don't fire. It doesn't appear the branch settings allow this conditional behavior so I feel it's safer to always do tests and require them, rather than make them not required.

exzizt (approved):
💯

PR #1810, opened 2021-07-26
MDS-3860: Fix issue where Permit Package docs not editable unless Tech Review active

exzizt:
# Main
* Fixes an issue where permit package documents were not editable on the Draft tab unless the NoW was in technical review (editing on the Manage Documents or Administrative tab still worked, however)

# Other
* Updates some text to be more correct
* Fixes an issue where some action creator calls were not being returned


services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
```diff
@@ -110,7 +110,7 @@ export class FinalPermitDocuments extends Component {
           showPreambleFileMetadata={this.props.showPreambleFileMetadata}
           editPreambleFileMetadata={this.props.editPreambleFileMetadata}
           isFinalPackageTable
-          isAdminView={this.props.adminView}
+          isAdminView
```
exzizt, line 113: The culprit

mmumby (approved):
🔥

PR #1811, opened 2021-07-27
Add missing dev infra workflow event trigger

jon-funk:
# Main
- Adds a on-push event for if the dev infra workflow file is changed, otherwise infra changes only occur currently if you edit dc templates directly.


exzizt (approved):
✔️

PR #1812, opened 2021-07-27
MDS-3826: Fix bond transfer on mine permits

jon-funk:
# Main
- Explicitly sets the mine context while working with the permit


services/core-api/app/api/securities/resources/bond.py
```diff
@@ -143,6 +143,9 @@ def put(self, bond_guid):
         if not permit_guid:
             raise BadRequest('permit_guid is required.')
         permit = Permit.find_by_permit_guid(permit_guid)
+        # Set mine context
+        Permit._context_mine = permit._all_mines[0]
```
jon-funk, line 149: I tried to look for an applicable class API but didn't see one. I'm a bit suspect about directly assigning a class member like this.

mmumby (approved):

PR #1814, opened 2021-07-28
Bug - TSF

mmumby:
- Added null check to has_itrb before converting to a string

Modal can open now:
Screen Shot 2021-07-28 at 2 59 17 PM


exzizt (approved):
🌮

PR #1815, opened 2021-07-29
Updates/vfcbc now data

mmumby:
- Mapped a now_submission field to now_applications


JustinMacaulay (approved):
The fact that there not the same bothers be but looks good otherwise.

PR #1819, opened 2021-08-04
MDS-3844: Add github runner build for frontend plus pipeline improvements

jon-funk:
# Main
- Adds build for frontend in github runner `Note:` pushes to a testing imagestream for now while it's being manually tested
- Adds polling behaviour to post-image-promotion in pipeline to ensure rollouts are a success. Polling occurs per service every 10 seconds until a success or a 15min timeout limit is hit, resulting in a error

# Other
- Renames workflows for infrastructure updates to make them less confusing


exzizt (dismissed):
Amazing, Jon! 💯 😄 Nice work

exzizt (dismissed):

PR #1825, opened 2021-08-09
MDS-3760: Sort the PP

exzizt:
# Main
- Allows users to sort documents in the PP by using a handy "drag-n-drop" feature:
image
  - Sorting is only allowed in the Draft tab and while the draft is in progress
  - The order starts at "1.2" in the table because 1.1 is currently always the Application Form in the permit
 - Accounts for various edge cases, e.g., when a user uploads a document and sets it to "Is In Final Application Package" or when we automatically add the NoW form to the PP we need to add it to the bottom of the PP


services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -1,11 +1,17 @@
-import React from "react";
+import React, { Component } from "react";
```
exzizt, line 1: I turned this component into a class component to make it easier to use the state (+ it's what the Ant Design demo was using).

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
```diff
@@ -111,6 +117,7 @@ export class FinalPermitDocuments extends Component {
           editPreambleFileMetadata={this.props.editPreambleFileMetadata}
           isFinalPackageTable
           isAdminView
+          isSortingAllowed={!this.props.adminView && draftInProgress}
         />
```
mmumby, line 121: Is sorting is enabled on the manage documents tab, draft tab, but not the admin tab?

services/core-api/app/api/now_applications/resources/now_application_export_resource.py
```diff
@@ -185,20 +185,47 @@
 ]
```
mmumby, line 185: I should get the same formatter as you and/or figure out why mine isn't working. (considering I was the last to update this file)

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -188,7 +306,20 @@ export const NOWDocuments = (props) => {
...
+    const sortColumn = {
+      title: "Order",
+      dataIndex: "index",
+      className: "drag-visible",
+      render: (text) => (
+        <>
+          <DragHandle />
+          {/* NOTE: We are adding 2 here because "1.1" in the issued permits is currently always the application form document. */}
+          &nbsp; 1.{text + 2}
+        </>
```
mmumby, line 318: order starts at 0?

mmumby (dismissed):
Looks great! 
🔥 

I just have a few questions but nothing jumped out at me.

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
exzizt: Enabled just on the Draft tab. The adminView prop is set to true for the Manage Docs and Admin Tab.

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
exzizt: We use the index (starting from 0) and then need to add 2 to get "1.2".

exzizt:
> Looks great!
> 🔥
> 
> I just have a few questions but nothing jumped out at me.

Thank you for reviewing! 😃

mmumby (dismissed):

PR #1826, opened 2021-08-09
View only route for NoW applications

mmumby:
# Main
- Added a "View only" route for now applications for both historic and current NoW
- Added a check to the ApplicationGuard that only allows historic applications to be viewed on this page as they are not imported. 
- Update the GET to set the "is_historic" flag 
- Updated the transmgrify to read mms.status as well as now_app.status

- Regular header of application on the main flow
Screen Shot 2021-08-09 at 4 42 03 PM
- simplified header on the view only for the same application
Screen Shot 2021-08-09 at 4 52 23 PM
Screen Shot 2021-08-09 at 4 54 51 PM

# Other
- Updated the Clear Notice of work Action in the reducer to include more data that should be cleared
- Removed some unused props from the NoWTables


services/core-api/app/api/now_applications/resources/now_application_resource.py
```diff
@@ -42,9 +43,13 @@ def get(self, application_guid):
         if now_application_identity.now_application_id and not original:
             application = now_application_identity.now_application
             application.imported_to_core = True
+            application.is_historic = ApplicationsView.query.filter_by(
```
exzizt: This can be moved out of the if/else block and just done once.

exzizt (changes requested):
Nice work on this one; good minimal solution! Just one minor comment to address! 😄

services/core-api/app/api/now_applications/resources/now_application_resource.py
mmumby: oh good point

exzizt (approved):

PR #1829, opened 2021-08-12
Add job to verify db migrations

jon-funk:
# Main
- Add job during integration tests to verify return code of flyway container
# Other
- Remove unnecessary coverage wrapper for tests and directly exec pytest


JustinMacaulay (approved):
looks good to me.

PR #1832, opened 2021-08-16
Explicitly set ns in oc watch rollout action

jon-funk:


exzizt (approved):
Makes perfect sense 💯 How could you miss that?

PR #1835, opened 2021-08-17
Changing 'due' to 'due_date' and 'status' to 'order_status' to correspond with the data records.

dragancolic-bcgov:


exzizt (approved):
Nice work 💯 😄

PR #1850, opened 2021-08-26
MDS-3602: Show the NoW/AA Number When Header is Fixed

exzizt:
# Main
* Shows the Notice of Work or Administrative Amendment number on the dashboard when the header is in fixed mode
image


services/core-web/src/components/noticeOfWork/applications/NOWTabHeader.js
```diff
@@ -44,6 +46,15 @@ export const NOWTabHeader = (props) => (
...
               {props.tabName}
             </Popover>
+            {props.fixedTop && (
+              <div className="view--header--sub">
+                {props.noticeOfWork?.application_type_code == "NOW"
+                  ? "Notice of Work"
+                  : "Administrative Amendment"}
+                :
+                <br />#{props.noticeOfWork.now_number}
+              </div>
```
mmumby, line 57: can you add a default for those that have null now_numbers

mmumby: Ideally, this field shouldn't be null, but there are close to 400 in test with null values. We should look into this...

exzizt: Should we display "N/A"? Or just remove this altogether if it is null?

mmumby: I vote for "N/A", the same as the main header.

mmumby (approved):

PR #1854, opened 2021-08-30
adding 'is None' condition to filter out None final_package_order(s)

dragancolic-bcgov:
# Main

- Keep all headers and just leave a single list item 'N/A' if there's no content for that header, otherwise remove N/A and add your details!

# Other

- N/A

# How to test

- N/A

# Notes

- N/A

### DIFF
--- a/services/core-api/app/api/now_applications/resources/now_application_export_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_export_resource.py
@@ -438,7 +438,7 @@ class NOWApplicationExportResource(Resource, UserMixin):
                     }
                     included_docs.append(new_doc)
 
-            included_docs.sort(key=lambda doc: (doc['final_package_order'] is not None))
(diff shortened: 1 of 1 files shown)


mmumby (dismissed):
tested - well done! 🚢 -it

exzizt (changes requested):
This breaks the sorting logic, it will now sort by final_package_order not being None, when it should be sorting by the integer value (excluding null values, I suppose, although it should never happen where a FAP document doesn't have its final_package_order set).

exzizt:
> # Main
> * What did you focus on in this PR?
> * Keep all headers and just leave a single list item 'N/A' if there's no content for that header, otherwise remove N/A and add your details!
> 
> # Other
> * What other side fixes, tweaks did you add on the side?
> * N/A
> 
> # How to test
> * N/A
> 
> # Notes
> * Any notes/screenshots for this PR?
> * N/A

You forgot to fill out the template 😛

exzizt (approved):

PR #1855, opened 2021-09-02
Delete Draft permits that are New Permits.

areyeslo:
# Main
Delete a draft permit that is a New Permit  with the format {NOW_type_code}-DRAFT-{NOW_no} as permit_no. If this draft permit is soft deleted, then a new permit has to be created and the permit_no has to include a suffix.  The next draft permits created must contain a suffix with the format {NOW_type_code}-DRAFT-{NOW_No}-{dd} where dd ranges from 01-99.

# Other
The logic of suffix in the permit is added in the permit resource. In addition, the Front End needed to enable the button to delete the permit.

# How to test
Follow the following steps to setup the test case:
1. Browse a Notice Of Work in "Received" status and open it:
image
2.  Go to Draft tab and click on Start Draft Permit:
image
3.  On the modal window select New Permit as Application Type and Yes for Exploration Permit. Then, click on Proceed. Finally, click on Start Draft Permit:
image
4. Confirm the Permit Number follows the pattern {NOW_type_code}-DRAFT-{NOW_no}:
image

The following steps test that the draft permit is deleted, and during creation of a new draft permit the suffix is added to avoid issues related to unique constraint for permit_no:
1.  Click on "Delete Draft" button  to delete the draft permit
2.  Create again a Draft Permit by clicking on "Create Draft Permit" button
3.  Confirm the newly created Draft permit is created with the suffix made of two digits:
image
Optionally, you can delete and create more draft permits. It is important to notice that the first created draft permit has the pattern {NOW_type_code}-DRAFT-{NOW_no} and the subsequent created draft permits would have the pattern {NOW_type_code}-DRAFT-{NOW_no}-{dd} starting with 01 in {dd} up to 99.

# Notes
https://bcmines.atlassian.net/browse/MDS-3735

### DIFF
--- a/services/core-api/app/api/utils/helpers.py
+++ b/services/core-api/app/api/utils/helpers.py
@@ -8,11 +8,10 @@ def clean_HTML_string(raw_html):
 
 
 """
-Add sufix to permit number adding one to versioning.
+Add two digits suffix to new draft permit numbers modifying the permit_no to the pattern:
+{NOW_type_code}-DRAFT-{NOW_No}-{dd}
 """
-
-
-def generate_permit_no_sufix(permit, separator, filling=2):
+def generate_draft_permit_no_suffix(permit, separator, filling=2):
     result = permit.split(separator + "-")
     version_str = '0' if len(result) < 2 else result[1]
     version = str(int(version_str) + 1).zfill(filling)
--- a/services/core-api/app/api/utils/helpers.py
+++ b/services/core-api/app/api/utils/helpers.py
@@ -1,6 +1,21 @@
 import re
 
+
 def clean_HTML_string(raw_html):
-  cleanr = re.compile('<.*?>')
-  cleantext = re.sub(cleanr, '', raw_html)
-  return cleantext
+    cleanr = re.compile('<.*?>')
+    cleantext = re.sub(cleanr, '', raw_html)
+    return cleantext
+
+
+"""
+Add sufix to permit number adding one to versioning.
+"""
+
+
+def generate_permit_no_sufix(permit, separator, filling=2):
+    result = permit.split(separator + "-")
+    version_str = '0' if len(result) < 2 else result[1]
+    version = str(int(version_str) + 1).zfill(filling)
+    permit_no = separator + "-" + version
+
+    return permit_no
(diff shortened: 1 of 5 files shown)


services/core-api/app/api/utils/helpers.py
```diff
@@ -1,6 +1,21 @@
...
+    cleantext = re.sub(cleanr, '', raw_html)
+    return cleantext
+
+
+"""
+Add sufix to permit number adding one to versioning.
+"""
+
+
+def generate_permit_no_sufix(permit, separator, filling=2):
```
exzizt: Typo, sufix --> suffix

services/core-api/app/api/utils/helpers.py
```diff
@@ -1,6 +1,21 @@
...
-  cleanr = re.compile('<.*?>')
-  cleantext = re.sub(cleanr, '', raw_html)
-  return cleantext
\ No newline at end of file
+    cleanr = re.compile('<.*?>')
+    cleantext = re.sub(cleanr, '', raw_html)
+    return cleantext
+
+
+"""
```
exzizt, line 10: Can you please move this comment into or closer to the function?

services/core-api/app/api/utils/helpers.py
exzizt: So, this is used just for determining what a draft permit's permit number should be? I would document that maybe, otherwise it could be used for issuing permits or cause other confusion.

exzizt (commented):
Thank you for being so thorough in documenting your PR, great job! I would like some more eyes on this PR before we approve it, though. Good work :)

services/core-api/app/api/utils/helpers.py
areyeslo: Sure Luke :)

services/core-api/app/api/utils/helpers.py
areyeslo: Absolutely Luke , fixing the typo. The function is generic to be reused if we have similar need, it is not connecting to database, just taking two strings, using one of these parameters to split the other string and increasing in one the version number.

services/core-api/app/api/utils/helpers.py
```diff
@@ -1,6 +1,21 @@
...
+Add sufix to permit number adding one to versioning.
+"""
```
mmumby, line 13: Can you update this message to specify that it's for the 'Draft permit Number' only?

areyeslo: Sure Meredith, this function is very generic, to be reused in similar cases. I will add further comments on it.

exzizt (approved):
Thanks for making those changes! Great job, Arturo! 💯

PR #1856, opened 2021-09-02
MDS-3891: Email Notifications

exzizt:
# Main
- Creates an "email service" for sending emails by using the Common Services "Common Hosted Email Service" (CHES) API
  - Uses the email service by sending an email to the corresponding regional office when a mine's start/stop info is updated in MineSpace
image

# Other
- N/A

# How to test

- Emails only get sent out in Production; we will have to wait for a MineSpace user to update their start/stop info. Otherwise, make adjustments locally so that the email gets sent to yourself, for example.

# Notes
- N/A


services/core-api/app/api/services/email_service.py
```diff
@@ -0,0 +1,176 @@
...
+    <p>This is a no-reply email address. If you need to contact the MDS team, please email us at: <a href="mailto: {Config.MDS_EMAIL}">{Config.MDS_EMAIL}</a>.</p>
+    <br />
+    <img src="{CORE_PURPLE_LOGO_BASE64_ENCODED}" width="320" height="106" alt="Core Logo">
+</div>
+'''
+
+
+# NOTE: Manage our Common Services application access here: https://getok.nrs.gov.bc.ca/app/apps/MDS
+# NOTE: See here for details: https://ches.nrs.gov.bc.ca/api/v1/docs
+class EmailService():
```
exzizt, line 40: There are just 3 methods implemented in this service, one to interface with each of the 3 CHES resources we need at the moment: authentication, health check, and send an email.

services/core-api/app/config.py
```diff
@@ -52,6 +53,10 @@ class Config(object):
     NRIS_USER_NAME = os.environ.get('NRIS_USER_NAME', None)
     NRIS_PASS = os.environ.get('NRIS_PASS', None)
     ENVIRONMENT_NAME = os.environ.get('ENVIRONMENT_NAME', 'dev')
+    CORE_PRODUCTION_URL = os.environ.get('CORE_PRODUCTION_URL',
```
exzizt, line 56: I did not set up these 3 env vars. I figure they are the same in each namespace (dev/test/prod). Do we want to do that @jon-funk?

openshift4/templates/backend.dc.yaml
```diff
@@ -386,6 +386,26 @@ objects:
                     secretKeyRef:
                       key: database-nris-password
                       name: postgresql-nris
+                - name: COMMON_SERVICES_CLIENT_ID
```
jon-funk: These are under the "flyway-migration" container,  https://github.com/bcgov/mds/pull/1856/files#diff-e89a9cf81abf84592135e97e8b2e257f16b74e89c28aff020b987e605a2fb387R327

You'll want to move them to https://github.com/bcgov/mds/pull/1856/files#diff-e89a9cf81abf84592135e97e8b2e257f16b74e89c28aff020b987e605a2fb387R136 so they're under the "app" container

exzizt:
Regarding the code climate issue, maybe we should increase this to 15? Something higher than 10, anyway.
image
We could get around this by making a class and passing an **EmailRequest** object or something but... yeah.

services/core-api/app/api/services/email_service.py
```diff
@@ -0,0 +1,176 @@
...
+            message = 'Common Services authentication request did not return an access token!'
+            current_app.logger.debug(resp_data)
+            current_app.logger.error(message)
+            return
+
+        return auth_token
+
+    # NOTE: See here for details: https://ches.nrs.gov.bc.ca/api/v1/docs#tag/Health
+    @classmethod
+    def perform_health_check(cls):
```
areyeslo, line 76: Should not return failure or success the health check function called in send_email function?

areyeslo:
Awesome job in setting up the email service :)

services/core-api/app/api/services/email_service.py
exzizt: We still want to attempt to send the email even if a dependency in the health check is not healthy (in my opinion). So there is no use at the moment for this to return anything. We just call it before we send the email and then it logs any issues.
For example, maybe the SMTP server dependency is unhealthy, but their queue service is healthy. We still want to send the email request and it will probably get sent out later.

openshift4/templates/backend.dc.yaml
jon-funk: It won't let me comment on the line since it's out of the committed of changes, but if you change https://github.com/bcgov/mds/pull/1856/files#diff-e89a9cf81abf84592135e97e8b2e257f16b74e89c28aff020b987e605a2fb387R253
to ${ENVIRONMENT_NAME} it should pass through the value

services/core-api/app/config.py
jon-funk: Should be fine as is.

jon-funk (changes requested):
Super exciting to have notifications, and the implementation is really nice! Just a few small tweaks for the backend yaml :)

openshift4/templates/backend.dc.yaml
exzizt: Good catch!

openshift4/templates/backend.dc.yaml
```diff
@@ -315,6 +315,16 @@ objects:
                     secretKeyRef:
                       key: common-services-email-host
                       name: common-services
+                - name: EMAIL_ENABLED
+                  valueFrom:
+                    secretKeyRef:
+                      key: email-enabled
+                      name: common-services
+                - name: EMAIL_RECIPIENT_OVERRIDE
```
jon-funk, line 323: Is this meant to be empty? Or is an override always provided?

openshift4/templates/backend.dc.yaml
```diff
@@ -315,6 +315,16 @@ objects:
                     secretKeyRef:
                       key: common-services-email-host
                       name: common-services
+                - name: EMAIL_ENABLED
```
jon-funk, line 318: Not sure if we need a secret for a bool flag

exzizt: No, but it's nicer to control it via OpenShift rather than having to make a code change and redeploy to change the value?

openshift4/templates/backend.dc.yaml
exzizt: If you want to override all email recipients, then it should not be empty. If it is empty then it will not override the email recipients.

openshift4/templates/backend.dc.yaml
jon-funk: Fair, the current setup is a clunky developer experience.

jon-funk (approved):

PR #1867, opened 2021-09-16
Bug/employees permissions

mmumby:
# Main
- Updated the permissions on the mine PUT to allow proponents to update number of employees, number of contractors


jon-funk (approved):
🐛 🔨

PR #1875, opened 2021-10-12
MDS-3886 Allow Draft Permit Issue Date to be in the future/or not required

areyeslo:
# Main

- Only issue date in Draft permits are allowed to be in the future and the date is not required.
- Issue date in issued permits is not allowed.

# Other
- N/A

# How to test
## Set date in the future
- Set a date in the future in a Draft permit and save:
image

- Confirm is successfully updated:
image

- Solve all the issues in Process tab.

- Click on issue a Permit:
image

-  Confirm issue date cannot be set in the future:
image

- Have to be unable to proceed to issue the permit when clicking Issue Permit button.

## Issue date is not set:

- Under Draft, fill with blank as issue date and set authorization end date. Then save:
image

- Confirm is successfully updated:
image

- Solve all the issues in Process tab.

- Click on issue a Permit:
image

-  Confirm issue date cannot be empty:
image

- Have to be unable to proceed to issue the permit when clicking Issue Permit button.

# Notes
https://bcmines.atlassian.net/browse/MDS-3886

### DIFF
--- a/services/core-web/src/components/Forms/permits/GeneratePermitForm.js
+++ b/services/core-web/src/components/Forms/permits/GeneratePermitForm.js
@@ -6,11 +6,7 @@ import { Field, reduxForm, getFormValues } from "redux-form";
 import { Form } from "@ant-design/compatible";
 import "@ant-design/compatible/assets/index.css";
 import { Col, Row, Descriptions } from "antd";
-import {
-  required,
-  dateNotAfterOther,
-  dateNotBeforeOther,
-} from "@common/utils/Validate";
+import { required, dateNotAfterOther, dateNotBeforeOther } from "@common/utils/Validate";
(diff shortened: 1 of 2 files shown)


services/core-web/src/components/Forms/permits/GeneratePermitForm.js
```diff
@@ -220,8 +219,6 @@ export const GeneratePermitForm = (props) => {
                 label="Issue Date*"
                 component={renderConfig.DATE}
                 validate={[
-                  required,
-                  dateNotInFuture,
                   dateNotAfterOther(props.formValues.auth_end_date),
```
mmumby: I think you have to also remove the validation on the 'auth_end_date' as well, Neither should be required on this form as they don't always have the data at this point and could prevent them from editing other things.

mmumby (approved):
🔥

PR #1878, opened 2021-10-15
Adding lead and issuing inspector name column in the Applications page.

areyeslo:
# Main
Add Lead and Issuing Inspector columns in Application page.

# Other
- N/A

# How to test
- Under Permit & Approvals select Applications for a specific mine:
- Two new columns are shown under the Notice of Work Applications:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-3806


mmumby (changes requested):
Just tested the search for "issuing_inspector" and it does not return any results. 


This may need another look to ensure the search is working for both inspector types - I believe you will have to add the filter to the BE as this is a BE search. 

look into 'now_application_list_resource.py' and the if statement with 'lead_inspector_name', 


Also, add 'issuing_inspctor_name' to the defaultParams on 'NoticeOfWorkHomePage.js'

mmumby (approved):
Thanks for making the updates! Looks good.

PR #1880, opened 2021-10-18
Removing limit of mines returned.

areyeslo:
# Main
- Minespace list every mine that a proponent has access on the "My Mines" page.

# Other
- N/A

# How to test
- Adding mines to proponent.

1. Get the maximum number of mines associated to a proponent:
```
select user_id, count(user_id)
from minespace_user_mds_mine_access
group by user_id
order by 1 desc
```

2. Delete data for your user in minespace_user_mds_mine_access. I am using mining@bceid which is user_id= 253
```
delete 
from minespace_user_mds_mine_access
where user_id= 253

```
3. Insert new records considering as a base the records with the maximum number of mines found in step 1:
```
select 'INSERT INTO minespace_user_mds_mine_access values (253, '''||mine_guid||''')'
from minespace_user_mds_mine_access
where user_id=237 // the user_id with the maximum number of mines from step 1.
```
4. Ensure you commit the changes in database if you did not auto-commit.

- **Validation**: 
Get the same number of mines in the first query and in minespace UI:
image

Get the 50 mines limit in core UI:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-3721

### DIFF
--- a/services/core-api/app/api/mines/mine/models/mine.py
+++ b/services/core-api/app/api/mines/mine/models/mine.py
@@ -226,7 +226,6 @@ class Mine(SoftDeleteMixin, AuditMixin, Base):
 
     @classmethod
     def find_by_mine_name(cls, term=None, major=None):
-        MINE_LIST_RESULT_LIMIT = 50
         if term:
             name_filter = Mine.mine_name.ilike('%{}%'.format(term))
             mines_q = Mine.query.filter(name_filter).filter_by(deleted_ind=False)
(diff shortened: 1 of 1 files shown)


mmumby (changes requested):
Tested - Having issues with the Admin tab returning data, lets discuss before merging.

mmumby (approved):
🔥  

well done, Arturo! Thanks for making those updates.

PR #1881, opened 2021-10-19
MDS-3925: Add linting workflow for manifests

jon-funk:
# Main

- Adds workflow to check kubernetes manifests during PRs

# Other

- Increases select quotas to better handle spiking demand for key services (namely FE, BE)
- Adjusts backend readiness & liveness probes to be more forgiving during deploy. Should lead to a smoother prod deploy.
- As per the PS request: update legacy scalingTargetRef blocks on all HPAs to specify the API endpoint of DCs as this is required syntax in new OCP versions.
- Fixes pr checklist spam. Had to collapse comment to single line and include extra whitepace at the end of every line??!?!??

# How to test

- N/A

# Notes

- N/A

### DIFF
--- a/.github/workflows/pr_checklist.yaml
+++ b/.github/workflows/pr_checklist.yaml
@@ -11,10 +11,5 @@ jobs:
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         with:
-          msg: |
-            Definition of Done Reminders:
-            - [ ] Did you update associated documentation?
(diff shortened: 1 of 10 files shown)


JustinMacaulay (approved):
Looks good.

jon-funk:
Historical note: The failing linting checks are warnings related to legacy syntax. They are harmless and we've been using them for months so we'll just keep them in mind moving forward and incrementally correct them.

PR #1885, opened 2021-10-26
Use new version number with new deployment of vc issuer service

Jsyro:
# Main

- The MDT team needed to update the VC issuer controller and the schema version. CORE needs the corresponding change.


jon-funk (approved):
👍

PR #1891, opened 2021-11-12
MDS-4001: resource audit

jon-funk:
# Main

- Performed a resource audit across all our kubernetes resources and adjusted them according to historical data
- This is important because we can't waste our quota and need to make room for new services in the future
- Using https://app.sysdigcloud.com/#/dashboard-template/view.kubernetes.cluster.rightsizing?from=1636070400&to=1636675200

# Other

- N/A

# How to test

- N/A

# Notes

- Overall, the trend is that most services were over provisioned for CPU and had just enough or under provisioned amounts of ram
- Additionally, if I've done the math correctly we should see better stability


JustinMacaulay (approved):
Looks good.

PR #1892, opened 2021-11-15
Feature/mds 3995 feature flagging in minespace

JustinMacaulay:
# Main

- Adds feature flagging to MineSpace through the AuthorizationWrapper Component just as it is in CORE.

# Other

- Added the flagging to the Record Incident button in MineSpace to shoe in test.

# How to test

- See that the button is available in test and then push to prod and confirm the button is not there.


services/minespace-web/src/constants/permissions.js
```diff
@@ -1,2 +1,3 @@
 export const IN_DEVELOPMENT = "inDevelopment";
 export const IN_TESTING = "inTesting";
+export const ADMIN = "role_admin";
```
mmumby: is ADMIN being used?

mmumby (commented):
Just one small comment regarding unused const, otherwise looks good!

mmumby (approved):

PR #1913, opened 2021-12-01
MDS-4042: Tweaks to Minespace project summary display

BenHarrington:
# Main

- Repositioned/renamed "Project Summaries" tab to "Applications" and hid it for non Major mines
- Updated Project Summaries table columns to match current mockup
- Added guidance text for Applications tab
- Removed project summary date field on AddEditProjectSummaryForm component

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


services/core-api/app/api/mines/project_summary/response_models.py
```diff
@@ -17,7 +17,9 @@
         'status_code': fields.String,
         'project_summary_lead_party_guid': fields.String,
         'project_summary_lead_name': fields.String,
-        'documents': fields.List(fields.Nested(PROJECT_SUMMARY_DOCUMENT_MODEL))
+        'documents': fields.List(fields.Nested(PROJECT_SUMMARY_DOCUMENT_MODEL)),
+        'update_user': fields.String,
+        'update_timestamp': fields.DateTime
     })
```
mmumby, line 23: We don't usually return the audit columns to the FE, we've created hybrid properties or created duplicate columns in the past that we expose to the UI. Our current pattern is not good though and given this is an expanding need of the users I think we should re-assess how we are doing it. 

@JustinMacaulay thoughts?

services/minespace-web/src/components/Forms/projectSummaries/AddEditProjectSummaryForm.js
```diff
@@ -49,13 +49,6 @@ export class AddEditProjectSummaryForm extends Component {
...
     return (
       <Form layout="vertical" onSubmit={this.props.handleSubmit}>
-        <Field
-          id="project_summary_date"
-          name="project_summary_date"
-          label="Date"
-          placeholder="Please select date"
-          component={renderConfig.DATE}
-        />
         <Field
```
mmumby: I thought we weren't going to remove this field in this ticket?

BenHarrington: Hmm, I think perhaps I got mixed up with what the end result was supposed to be. I can easily revert this if its a problem.

mmumby: yeah lets either: 
1. add it back
2. remove it, but populate the received date on the BE so we can still submit project summaries. 

I vote for 1

BenHarrington:
Two small issues I have found since the last update:

1. If you try to create a contact without an email the POST fails even though the form does not indicate email is required
2. Completing the POST with an email included results in a "AttributeError: 'NoneType' object has no attribute 'emli_contact_type_code'" error from the API

services/minespace-web/src/components/Forms/projectSummaries/AddEditProjectSummaryForm.js
BenHarrington: Alright, I will bring it back in, thanks for the feedback

mmumby (approved):
🔥  Looks good!

PR #1915, opened 2021-12-02
Fix/docker compose upgrade

jon-funk:
# Main

- Correct volume paths for docker compose V2
- Correct some makefile file paths/links

# Other

- N/A

# How to test

- You need to `make clean` so your volumes continue to work
- Run `make be` `make getdb` `make seeddb`

# Notes

- N/A


mmumby (approved):
Tested! Thanks for making this update 🔥

PR #1925, opened 2021-12-15
Updated the now_submission tables to include the new fields

mmumby:
# Main

-  Updated the following now_submission tables:
- application
- exp_access_activity
- camps
- buildings
- stagingareas
- exp_surface_drill_activity
- surface_bulk_sample_activity
- sand_grv_qry_activity

# Other

- Mapping to the core now_application tables will happen in a follow up PR - This PR is to ensure we get the data when vFCBC pushes their updated to prod.

# How to test

- Submit a now_submission to the local API with the new fields and ensure it populates in the db.

# Notes
 N/A

### DIFF
--- a/migrations/sql/V2021.12.14.15.52__update_now_submission_data.sql
+++ b/migrations/sql/V2021.12.14.15.52__update_now_submission_data.sql
@@ -1,9 +1,9 @@
 ALTER TABLE now_submissions.application
 ADD COLUMN IF NOT EXISTS maxannualcubicmeters   numeric,
 ADD COLUMN IF NOT EXISTS proposedstartmonth     varchar(100),
-ADD COLUMN IF NOT EXISTS proposedstartday       varchar(300),
-ADD COLUMN IF NOT EXISTS proposedendmonth       varchar(300),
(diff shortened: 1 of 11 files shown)


migrations/sql/V2021.12.14.15.52__update_now_submission_data.sql
```diff
@@ -0,0 +1,42 @@
+ALTER TABLE now_submissions.application
+ADD COLUMN IF NOT EXISTS maxannualcubicmeters   numeric,
+ADD COLUMN IF NOT EXISTS proposedstartmonth     varchar(100),
```
areyeslo, line 3: Just one overall comment, wondering if there is a special case to assign 100 or 300 length to day/month for the fields related to month and day. Would it be enough to add length 2 or 3 if you are handling days from 01-31 and months from 01-12?

mmumby: discussed on a call!

BenHarrington (approved):
Great job Meredith!

PR #1926, opened 2021-12-15
Minespace vc authn 0.1

Jsyro:
# Main

- Add small configuration for minespace website to use vc-authn OIDC Identity Provider as a new method for authentication

# Other

- I've added a new Identity Provider to the MDS TEST Keycloak Realm, see that configuration. 

# How to test
- ~~Once this PR is merged and deployed. https://github.com/bcgov/a2a-trust-over-ip-configurations/pull/72.~~ This PR's Minespace instance allows for login with bcgov-minespace-access:0.1 credentials issued by (https://mcad-bpa-dev.apps.silver.devops.gov.bc.ca/)

# Notes

~~- Unsure if this will work exactly as is, please advise on how to deploy this instance before merging this PR. Or to run this locally and connect with your TEST KC realm~~

- Works locally, will produce video to showcase, I have videos, unsure how to share them. 

# Future Work
**- If you want to make custom buttons that redirect automatically to the two given login screens, can you do that via idp_hint's. It looks like the prod log-in button does this, so even deployed to PROD, the login button will take users directly to the BCeID login until you change that KC client or create a new button with a different idp hint.**


openshift4/templates/minespace.dc.yaml
```diff
@@ -81,6 +81,10 @@ parameters:
     displayName: AppName
     description: Application name for grouping pods in Openshift
     value: Stand-Alone
+  - name: VCAUTHN_PRES_REQ_CONF_ID
+    displayName: VC-AUTHN-PRES-REQ_CONF_ID
+    description: Presentation Requestion Configuration Id for KC to foward to vc-authn Identity Provider
+    value: minespace-access-0.1-dev
```
Jsyro, line 87: @jon-funk How/Where do we define the parameters into the DC's? this should be different for each environment.

.github/workflows/deploy-prod.yaml
```diff
@@ -74,5 +74,6 @@ jobs:
           -p MS_KEYCLOAK_CLIENT_ID=minespace-prod \
           -p SITEMINDER_URL=https://logon.gov.bc.ca \
           -p MATOMO_URL=https://matomo-4c2ba9-prod.apps.silver.devops.gov.bc.ca/ \
+          -p VCAUTHN_PRES_REQ_CONF_ID=minespace-access-0.1 \
```
Jsyro, line 77: NOTE, this pres_req_conf_id doesn't exist yet. https://github.com/bcgov/a2a-trust-over-ip-configurations/tree/master/proof-configurations/minespace-access

If this version/solution needs to work in production, just copy paste the `test` one and change the id value to match this line.

Jsyro: Oh, actually we would need to configure the VC-AUTHN installation to have a client for the PROD MDS Keycloak Realm. 

Let me know if you want me to do this before merging. But there are two steps before this can work in prod (permanent connection between vc-authn and mds Keycloak, and the proof configuration being added to the vc-authn deployment).

jon-funk:
I'm happy with the pipeline, param, and manifest changes. Might just want to have Justin to sign off on the decision for the frontend login / flagging approach.

Jsyro:
How the FE works. 'Log in with Verifiable Credentials' buttons are feature flagged on only visible locally and in test. 

image

Jsyro:
Lastly, the vcauthn service is vulnerable to being spoofed (the pres_req_conf_id can be modified by the user). This means minespace should inspect the JWT token, and ensure the `pres_req_conf_id` matches the same one that the user was redirected to. 

I've configured the IDP in TEST KC to map the `pres_req_conf_id` to the client which maps it to the JWT, so it is visible to the application as the `pres_req_conf_id` attribute the JWT and AuthorizationGuard should probably be modified to handle this case (if `pres_req_conf_id` exists, compare it to `env.vcauthn_pres_req_conf_id`).

services/minespace-web/src/components/pages/LandingPage.js
```diff
@@ -55,13 +56,29 @@ export const LandingPage = (props) => (
...
+            <AuthorizationWrapper inTesting>
+              <Button type="primary" size="large" className="login">
+                <a
+                  href={`${COMMON_ENV.KEYCLOAK.loginURL}${MINESPACE_ENV.BCEID_LOGIN_REDIRECT_URI}&kc_idp_hint=${COMMON_ENV.KEYCLOAK.vcauthn_idpHint}`}
+                >
+                  Log in with Verifiable Credentials
+                </a>
+              </Button>
+            </AuthorizationWrapper>
+          </div>
```
mmumby, line 81: Is it possible to use the default log in screen with VC as an option like core does? (but remove the IDIR option) ?
Screen Shot 2021-12-21 at 9 06 07 AM

Jsyro: That is an option, if that is preferred, but that is KC configuration that is out of code (not in GH), and would need to be replicated in TEST and PROD. 

If this is preferred by your team, I can look into it, but the two buttons with kc_idp_hint was easier.  

I don't know exactly how to pick and choose which IDP methods come up for each client.

Jsyro: I can see if any of these options will give us what we want, but it's not 'customizable' beyond that as far as I know. 

image

mmumby: @Jsyro confirmed with the team, the two-button approach works for now! well done.

.github/workflows/deploy-prod.yaml
jon-funk: I don't think we want to enable this on prod right now, so no need to set that up.

mmumby (approved):
Tested the BCeId log-in flow locally and it still works as expected.

PR #1927, opened 2021-12-16
Add cronjob manifest for gis refresh

jon-funk:
# Main

- Adds manifest for cronjob definition that generates a generic psql client pod that executes a arbitrary psql command
- sql command for this manifest is to refresh the gis export materialized view nightly

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


JustinMacaulay (approved):
Looks good to me lets have it run in test over the weekend and then check back on monday.

PR #1928, opened 2021-12-16
Mds 3852

wilson496:
# Main

- Retired openshift tusd build
- Implemented github action for tusd build

# Other

- Removed tusd-bc.yaml from migrations/
- removed tusd references from build.yaml in migrations/

# How to test

- Trigger GitHub action "Build tusd" and ensure it succeeds. 

# Notes

- N/A


jon-funk (changes requested):
Nice work, this is looking great. Just a few nit picky / format things:

- update tusd.yaml filename to build-tusd.yaml so it's ordered along side build-flyway.yaml in file lists
- Don't forget to update your PR info!
- We should look at putting it into our test environment manually when you're available just to ensure there are no issues with the runtime before merging (we can sidestep the pipeline for manual testing using oc)
- A note regarding the manifest lint checks, they are an experimental addition and are picking up on old syntax. They wont block the PR

jon-funk (approved):
Awesome work! 🎉

PR #1930, opened 2021-12-17
Exemption status

mmumby:
# Main

- Updated the Exemption fee on permits where it is null
- Added mine_types for those that do not already exist

Logic for the status: 
Is the permit closed? Y
Is the permit Placer? Y
Is the permit exploration? AND Does it have ONLY a surface disturbance? Y
Is the Permit Mineral or Coal? MIM
Is the Permit Quarry, or Sand and Gravel? MIP

# Other

- N/A

# How to test

- Run the scrip on your local env and ensure the permits were updated correctly, and the correct number of mine_types was created (Happy to test with anyone)

# Notes

- N/A


areyeslo (approved):
Amazing job. I liked the anonymous block and the use of "returning *" in the cursor. I love it.

mmumby:
> 

that was all @BenHarrington :)

PR #1938, opened 2021-12-23
MDS-3825 - add max_old_size param to .env file to prevent js heap issues

wilson496:
# Main

 Changing the maximum memory allocation in NODE_OPTIONS to allow for hot reloading while running frontend in docker.

# Other

Maximum memory allocation is increased from 512M to 2GB
Update Dockerfile in services/core-web to use node:14.8.0 instead of node:10

# How to test

Add NODE_OPTIONS --max_old_space_size param to your environment file, spin up the mds_frontend container, and make a change while the container is running to trigger a hot reload. If the hot reload is successful (i.e. the changes are reflected in the UI), testing passes. 

# Notes

- N/A


jon-funk:
Good addition. Are we wanting to add the serve command change to this PR as well? Me and Justin bumped the local dockerfile node version to 14.8.0 and had better performance from that as well.

wilson496:
Confirmed that updating the node docker container from 10 to 14.8.0 dramatically increased loading times. Was able to hot reload in a few seconds compared to a few minutes before.

services/minespace-web/.env-example
```diff
@@ -14,3 +14,4 @@ BASE_PATH=
 ASSET_PATH=/
 
 VCAUTHN_PRES_REQ_CONF_ID=minespace-access-0.1-dev
+NODE_OPTIONS=--max_old_space_size=2048
```
jon-funk, line 17: Just a note for this change, since devs manage their own copies of .env's, you'll want to make an announcement to the team to make sure they have these lines in their core/minespace .env's after the PR is merged.

jon-funk (approved):
Excellent work! Super happy to see this fixed. Don't forget to remove the commit comments  when you squash and merge in the UI :)

PR #1941, opened 2021-12-31
MDS-3846 - separate docman from build.yaml into separate github action

wilson496:
# Main

Creating a separate GitHub action for document-manager.

# Other

document-manager has been removed from the monolithic build.yaml file and is a separate job that can be viewed as an individual GitHub action. 

# How to test

Trigger GitHub action "Build docman" and ensure it succeeds.

NOTE: As of the time of this PR, lint manifests may fail even if the build is successful. Linting for these builds are a W.I.P and do not work correctly. Thus, these can be ignored for the time being according to @jon-funk .  

# Notes

- N/A


.github/workflows/build-docman.yaml
```diff
@@ -0,0 +1,34 @@
+name: Build docman
+
+on:
+  pull_request:
+  push:
+    branches:
+      - develop
+    path:
+      - services/document-manager/backend/**
+      - .github/workflows/build-tusd.yaml
```
jon-funk: Should be `build-docman.yaml`

.github/workflows/build-docman.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+env:
+  INITIAL_TAG: latest
+  TAG: dev
+  NAME: docman
+  CONTEXT: services/document-manager/backend/
+
+jobs:
+  build-tusd:
```
jon-funk: Rename job to `build-docman`

jon-funk:
We'll want to update https://github.com/bcgov/mds/blob/develop/openshift4/templates/docman.dc.yaml#L344 as well for the worker definition to match the new script location in the image. (the current path is for the redhat image)

jon-funk:
Confirmed that NoW doc uploads are functioning with the new build in test.

.github/workflows/build-docman.yaml
wilson496: Oops! I'll get on that.

jon-funk (approved):
Nice work! 🎉

PR #1942, opened 2022-01-05
MDS-3857: refactor compose

jon-funk:
# Main

- Remove unused container stanzas from docker-compose file
- Rework major dockerfiles to use a non-root user during runtime so that file permissions do not hamper containerized development
- Add healthchecks to major container definitions to regularly poll container health and not mark it as ready until it's actually ready
`Note:`  These healthchecks are only as helpful as the check method itself, which are generally lacking, but we've got the connective tissue now so we'll get dividends in the future for improving them.
- Add bind mount for flyway container, allowing hot reload for sql scripts
`Note:` we don't have rollback scripts so you still might need to wipe your database if you want to undo changes and re-apply them due to mismatched checksums
- VSCode DevContainer integration: allows developers to launch vscode directly inside a container for a fully sandboxed development experience. The integration overlays on top of our compose file. Refer to `Notes` section for benefits.

# Other

- Rework filesystem-provider build to work correctly, bringing back containerized dev for our dotnet app
- Add flyway sql template script output to gitignore so it doesn't clutter local dev
- Remove vscode configs from gitignore so the team can start managing debugger configurations - it looks like these were initially ignored because we thought they were an annoyance and related only to dotnet

# How to test

- For most changes just use the `make lite` command and interact with the containers as you normally would. Be on the lookout for permissions issues, although I hopefully squashed them all.

- To test the DevContainer integration:
`vscode should now suggest https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers to you to install`
`start containers with make lite`
`open vscode command pallette`
`Select Remote-Containers: open folder in container... from list (only this option works with our repo)`
`navigate to the directory containing the new configs and hit ok, eg: services/core-api`
`vscode will restart in the container's app/ directory and install the specified extensions for that service`

# Notes

- The DevContainers don't have git installed so you'll have to have a separate terminal/vscode window for git commands running outside the containers. If the team would like git installed in containers that's something we can look at adding.

`Benefits of DevContainers:` 
- By fully sandboxing vscode inside the container we can control extensions, packages, modules per service that your IDE uses
- vscode understands the dependencies/modules inside the container, so there's no need to locally install dependencies/tools/software on your machine to stop linting from freaking out or other errors. Just get docker and create the containers
- Eliminates any developer need for python virtual environments
- Consistent developer experience
- Support for debuggers
- Support for web-browser based development and live codesharing (develop inside the container with other developers in Realtime)


services/core-api/Dockerfile
```diff
@@ -13,6 +15,19 @@ RUN pip install --no-cache-dir -r requirements.txt
...
+# Fix log path permissions
+# This is a bit hacky but other fixes require changing app code
+RUN mkdir -p /var/log/core-api
+RUN chmod -R 700 /var/log
+RUN chown -R 1000:0 /var/log
+
+# Setup user to represent developer permissions in container
+ARG USERNAME=python
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
```
JustinMacaulay: is there a reason that these two args are being assigned like this?

jon-funk:
Currently investigating codeclimate error in workflow

services/core-api/Dockerfile
jon-funk: Updated: explicitly set USER_GID to 1000

services/core-api/Dockerfile
```diff
@@ -1,18 +1,28 @@
-FROM python:3.6
+FROM bitnami/python:3.6
```
wilson496: does this change the provider for the container image? if so, why?

jon-funk: Good catch - this was from an attempt at fixing permissions issues in the container. Bitnami as a provider tends to be better with setting their permissions. Ultimately I fixed this by using https://github.com/bcgov/mds/pull/1942/files#diff-ae943a242e11cae5e334eb2a09b112dd40418caea9454ef04a2dda6aac807740R88 so I've reverted the change back to the old python:3.6

services/core-api/requirements.txt
```diff
@@ -10,7 +10,6 @@ GeoAlchemy2==0.6.3
 ldap3==2.6
 names==0.3.0
 petl==1.6.8
-psycopg2==2.8.5
```
wilson496, line 13: Nice. Removing unused packages sometimes goes unnoticed. Good stuff.

wilson496 (approved):
Looks good!

PR #1944, opened 2022-01-06
Fix: Update app path for new build

jon-funk:
# Main

- Corrects prehook script path and makes volume mounts path consistent (they still functioned but were confusing)
- Issue was not caught in original PR due to prehook command not raising an error

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


wilson496:
Looks good to me.

wilson496 (approved):

PR #1954, opened 2022-01-15
MDS-4102: Update the AuthWrapper in Minespace to consider an unauthenticated state

areyeslo:
# Main
In test an an unauthenticated user render "Log in with Verifiable Credentials"  wrapped by the AuthorizationWrapper.

# Other

# How to test
In test, the button "Log in with Verifiable Credentials" is rendered as in Dev environment:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-4102


JustinMacaulay (approved):
Looks good to me.

PR #1955, opened 2022-01-19
Filter Draft permit amendments in permit table under Permit tab.

areyeslo:
# Main
- In minespace, the user should NOT be able to see Draft amendment permits in permit table under Permit tab.

# Other

# How to test
Draft permit amendments are not visible under Permit tab.

# Notes
https://bcmines.atlassian.net/browse/MDS-4115


JustinMacaulay (approved):
Looks Good

PR #1956, opened 2022-01-19
Allow permition deletion if the permit is not generated in Core.

areyeslo:
# Main

-Allow users in Core to delete original and permit amendments if they are not generated in Core.  Before the user was getting this message informing that he/she was not able to delete the permit:
image

# Other

# How to test
- Create a permit amendment related to a NoW:
  a) Go to Mines and filter by Status Pending Verification. Click on Open
  b) Change the name for the mine you want to use o leave the existing one
  c) Solve all the issues in Process tab and issue a permit amendment

- Confirm you have the permit amendment under Permit tab which is assign to the original permit:
image
  a) Update is_generated_in_core to false in permit_amendment table for the record associated with the mine and permit_id 

- Delete the permit amendment:
image

- Confirm soft delete
  a) Confirm in permit_amendment table the field deleted_ind = True
  
image

# Notes
https://bcmines.atlassian.net/browse/MDS-3669


JustinMacaulay (approved):
All looks good

PR #1957, opened 2022-01-21
MDS-4118: Change verbiage in mine incident follow up.

areyeslo:
# Main

Change labels in Add incident - Follow up:

- Replace "Recommendations" for  "Mine Manager Recommendations"
- The text within the box should read - ‘Write in each individual Mine Manager Recommendation here’

# Other

# How to test
In Core, when creating an incident the labels should be updated:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-4118


JustinMacaulay (approved):
Looks Good.

PR #1958, opened 2022-01-23
MDS-4060 factory issues

JustinMacaulay:
# Main

- Tried to get debugging working for factories but could not get any output to print.
- When you generate 5 mines you now only get 5 mines instead of 15 mines.
- permitee and now submission now properly link to the mine and permit in question.

# How to test

- pull down the branch and run the flask command in the back end pod "flask create-data <num>"


areyeslo:
Looks good to me. It is creating just the requested number of mines and related now, variances, incidences, permits.

areyeslo (approved):

PR #1960, opened 2022-01-26
MDS-3848 - init build refactor for filesystem-provider

wilson496:
# Main

Refactoring build process for filesystem-provider service.

# Other
- N/A

# How to test

- Tag and build through oc tag command (ensure you're logged into the OC console)
- Check if the container is running
- Boom, done. 

# Notes

- N/A


services/filesystem-provider/Dockerfile.ci
```diff
@@ -0,0 +1,21 @@
+FROM mcr.microsoft.com/dotnet/sdk:3.1
```
jon-funk, line 1: Just a note, not requesting change: As per the norm, if the resulting build doesn't work you may need to look at https://github.com/bcgov/mds/blob/develop/openshift4/docker-images/dotnet-31/Dockerfile as that's what we had to use for permissions to get it to work in the past, however that is a redhat image.

.github/workflows/build-filesystem-provider.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - develop
+    paths:
+      - migrations/**
+      - .github/workflows/build-filesystem-provider.yaml
+
+env:
+  INITIAL_TAG: latest
+  TAG: dev
+  NAME: filesystem-provider
+  CONTEXT: migrations/
```
jon-funk: Update context path

.github/workflows/build-filesystem-provider.yaml
```diff
@@ -0,0 +1,34 @@
+name: Build filesystem-provider
+
+on:
+  pull_request:
+  push:
+    branches:
+      - develop
+    paths:
+      - migrations/**
```
jon-funk: update path trigger to match where devs would be making changes

wilson496: fixed

jon-funk (approved):
Great work!!

PR #1961, opened 2022-01-27
MDS-4063: Send notification to a proponent when creating a new Project Description in Minespace.

areyeslo:
# Main

- Send an email to the main proponent when creating a new Project Description in Minespace.

# Other

# How to test
- Login in Core, and add mines to the BCEID user through.
- Login in Minespace to a major mine and select Applications tab
- Click on Start a new application 
-  Fill out all the required information and under Project Contacts indicate the email of the proponent: 
image
- Receive notification of the creation of the Project Description in the email you indicated: 
image

# Notes
https://bcmines.atlassian.net/browse/MDS-4063


services/core-api/app/api/services/email_service.py
```diff
@@ -142,7 +152,9 @@ def send_email(cls,
         EmailService.perform_health_check()
 
         # If the sender is the MDS no-reply email address, add the MDS no-reply signature to the email body.
-        if sender == Config.MDS_NO_REPLY_EMAIL:
+        if proponent:
+            body += MINESPACE_NO_REPLY_SIGNATURE
+        elif sender == Config.MDS_NO_REPLY_EMAIL:
             body += MDS_NO_REPLY_SIGNATURE
```
mmumby, line 159: hmmmm, I'm not sure about this logic. The trigger for all emails at this point comes from the proponent.

mmumby: Ohhh, nvm, I see what you did. 'Proponent' will default to false. Okay, this works! 

Can we update the param name as it's a little confusing since the proponent is the trigger for all emails. Maybe 'send_minespace_email' or 'minespace_email', or 'send_to_proponent'... something like that?

services/core-api/app/api/mines/project_summary/resources/project_summary_list.py
```diff
@@ -116,6 +116,7 @@ def post(self, mine_guid):
             project_summary.save()
             if is_minespace_user():
                 project_summary.send_project_summary_email_to_ministry(mine)
+                project_summary.send_project_summary_email_to_proponent(mine)
         except Exception as e:
```
mmumby, line 121: We should only send the email when it goes from draft -> open. Can you add that additional logic for both emails?

mmumby (commented):
Few comments regarding when to send the emails, name change. 

Otherwise, they both sent correctly and the email looks good!

services/core-api/app/api/mines/project_summary/resources/project_summary_list.py
```diff
@@ -116,6 +116,8 @@ def post(self, mine_guid):
             project_summary.save()
             if is_minespace_user():
                 project_summary.send_project_summary_email_to_ministry(mine)
```
mmumby: This should have the same logic - only send if it goes from draft - open for both emails. 

The ministry cannot see applications that are in draft - that logic was added after the original email service was created. 

So came logic as the put: 

```
if prev_status == 'DFT' and project_summary.status_code == 'OPN':
project_summary.send_project_summary_email_to_ministry(mine)
            project_summary.send_project_summary_email_to_proponent(mine)
            ```

File: services/core-api/app/api/mines/project_summary/resources/project_summary.py [hunk 4]
```diff
@@ -121,6 +128,10 @@ def put(self, mine_guid, project_summary_guid):
             data.get('documents', []), data.get('contacts', []), data.get('authorizations', []))
 
         project_summary.save()
+
+        if prev_status == 'DFT' and project_summary.status_code == 'OPN':
+            project_summary.send_project_summary_email_to_proponent(mine)
+
```
mmumby, line 136: can you add the other email to this check? 

`project_summary.send_project_summary_email_to_ministry(mine)`

mmumby (dismissed):
well done! 🚢  it

mmumby (approved):

PR #1962, opened 2022-01-27
MDS-4037: Display Additional Project Summary form fields in Minespace

areyeslo:
# Main
In Minespace, under Project Description table conduct the following changes
- Add two fields:
 -  Project Title
 -  First Submitted: N/A if the status is Draft
- Remove Project # field 

# Other
- N/A

# How to test
- Create Project Descriptions with Draft and Open status.
- Confirm the table is showing the fields in correct order:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-4037

### DIFF
--- a/services/minespace-web/src/components/dashboard/mine/projectSummaries/ProjectSummariesTable.js
+++ b/services/minespace-web/src/components/dashboard/mine/projectSummaries/ProjectSummariesTable.js
@@ -28,8 +28,10 @@ export class ProjectSummariesTable extends Component {
       mine_guid: projectSummary.mine_guid,
       project_summary_guid: projectSummary.project_summary_guid,
       project_summary_id: projectSummary.project_summary_id,
+      project_title: projectSummary.project_summary_title,
       status_code: codeHash[projectSummary.status_code],
       update_user: projectSummary.update_user,
+      create_timestamp: formatDate(projectSummary.create_timestamp),
       update_timestamp: formatDate(projectSummary.update_timestamp),
       documents: projectSummary.documents,
       handleDeleteDraft,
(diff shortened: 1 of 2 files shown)


services/minespace-web/src/components/dashboard/mine/projectSummaries/ProjectSummariesTable.js
```diff
@@ -54,6 +56,15 @@ export class ProjectSummariesTable extends Component {
...
       sorter: (a, b) => (a.update_user > b.update_user ? -1 : 1),
     },
+    {
+      title: "First Submitted",
+      dataIndex: "create_timestamp",
+      sorter: dateSorter("create_timestamp"),
+      render: (text, record) => {
+        if (record.status_code !== "Draft") return <div title="First Submitted">{text}</div>;
+        return "N/A";
+      },
```
mmumby, line 66: I don't think this date will work for "First Submitted" - What happens if the draft was created 2 months ago, and just submitted now? It will show the created_date, not the submitted_date. 

I think we will need to create a new date field and only set it when the status goes from Draft -> Open.

I'm going to be working on the status update ticket now, would you like me to add that logic and come back to this ticket once that is completed?

areyeslo: Totally agreed. Thanks for raising this point.

BenHarrington: We still  have an old field left over from previous migrations called project_summary_date(never deleted the field), perhaps we can re-purpose that?

mmumby (dismissed):
I'm thinking we can get this approved and merged in, The Core side requires the same update regarding submitted date. I'll address both in my PR, but at least we can close this ticket.

mmumby (approved):

PR #1965, opened 2022-01-28
MDS-4058: View additional project data in Core

BenHarrington:
# Main

- https://bcmines.atlassian.net/jira/software/c/projects/MDS/boards/1?modal=detail&selectedIssue=MDS-4058

# How to test

- Visit Core -> Mines -> Select a mine with project descriptions ->Permits & Approvals - Pre-applications - > Open on a project description
- Verify side menu navigation works and that all information entered from the Minespace side shows up


services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
```diff
@@ -0,0 +1,263 @@
...
+  projectSummary: CustomPropTypes.projectSummary.isRequired,
+  initialValues: PropTypes.objectOf(PropTypes.any).isRequired,
+  projectSummaryDocumentTypesHash: PropTypes.objectOf(PropTypes.string).isRequired,
+  projectSummaryAuthorizationTypes: PropTypes.objectOf(PropTypes.any).isRequired,
+  projectSummaryPermitTypesHash: PropTypes.objectOf(PropTypes.string).isRequired,
+};
+
+const transformAuthorizationTypes = (types) => {
+  const transformedObject = {};
+  types.forEach((type) => {
```
mmumby: should this be let not const?

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
```diff
@@ -0,0 +1,263 @@
...
+                <p className="bold padding-sm--bottom">Types of permits</p>
+                {a.project_summary_permit_type.map((pt) => {
+                  return <p className="padding-md--left">{projectSummaryPermitTypesHash[pt]}</p>;
+                })}
+                <br />
+                <p className="bold padding-sm--bottom">Existing permit numbers involved</p>
+                {a.existing_permits_authorizations.map((epa) => {
+                  return <p className="padding-md--left">{epa}</p>;
+                })}
+                <br />
```
mmumby, line 114: can you add an "N/A" if 'existing_permits_authorizations' is empty

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
```diff
@@ -0,0 +1,263 @@
...
+          })}
+      </div>
+    );
+  };
+
+  const renderContacts = () => {
+    const {
+      projectSummary: { contacts },
+    } = props;
+    return (
```
mmumby, line 126: destructuring will fail the linter - is the linter running? We can revisit that rule and disable it if required, I can't remember the reasons why we enforced it.

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
```diff
@@ -0,0 +1,263 @@
...
+        <p className="bold">Primary project contact</p>
+        <p>{contacts[0]?.name}</p>
+        {contacts[0]?.job_title && <p>{contacts[0]?.job_title}</p>}
+        {contacts[0]?.company_name && <p>{contacts[0]?.company_name}</p>}
+        <a href={`mailto: ${contacts[0]?.email}`}>{contacts[0]?.email}</a>
+        <div className="inline-flex">
+          <p>{contacts[0]?.phone_number}</p>
+          {contacts[0]?.phone_extension && <p> ({contacts[0]?.phone_extension})</p>}
+        </div>
+        <br />
```
mmumby, line 139: For the primary contact, we'll have to add some safety for if it's null. I think we should add a null view here, but also ensure the form fails if there is no primary contact (This could all happen in a follow-up) Thoughts?

services/core-web/src/components/mine/ProjectSummaries/MineProjectSummary.js
```diff
@@ -68,7 +68,4 @@ const mapDispatchToProps = (dispatch) =>
 
 MineProjectSummary.propTypes = propTypes;
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(AuthorizationGuard(Permission.IN_TESTING)(MineProjectSummary));
+export default connect(mapStateToProps, mapDispatchToProps)(MineProjectSummary);
```
mmumby: this should still be wrapped in the 'AuthorizationGuard' otherwise we'll be releasing when this pr gets merged haha

services/core-web/src/components/mine/ProjectSummaries/MineProjectSummaryTable.js
```diff
@@ -15,40 +18,70 @@ const propTypes = {
...
+      projectSummary,
+      mine_guid: projectSummary.mine_guid,
+      project_stage: projectSummary.status_code,
+      documents: projectSummary.documents,
+      project_summary_id: projectSummary.project_summary_id || Strings.EMPTY_FIELD,
+      project_name: projectSummary.project_summary_title,
+      project_proponent_id: projectSummary.proponent_project_id || Strings.EMPTY_FIELD,
+      project_contact: contact?.name || Strings.EMPTY_FIELD,
+      first_submitted_date: formatDate(projectSummary.create_timestamp),
+      last_updated_date: formatDate(projectSummary.update_timestamp),
```
mmumby, line 36: I chated with Arturo re: 'first_submitted_date' on his PR, I'm going to add a new date field that is set when the status goes from draft - open. I'll update this field in my PR, no action needed but something to be aware of.

services/core-web/src/components/mine/ProjectSummaries/MineProjectSummaryTable.js
```diff
@@ -15,40 +18,70 @@ const propTypes = {
...
-      title: "Last updated",
-      dataIndex: "update_timestamp",
-      sorter: true,
-      render: (text) => <div title="Last updated">{text}</div>,
+      key: "first_submitted_date",
+      title: "First submitted date",
+      dataIndex: "first_submitted_date",
+      render: (text) => <div title="First submitted date">{text}</div>,
+      sorter: dateSorter("first_submitted_date"),
+    },
```
mmumby, line 77: with the addition of sorters, what's the default?

services/minespace-web/src/styles/components/ProjectSummaries.scss
```diff
@@ -0,0 +1,3 @@
+.fixed-header {
+  // position: fixed;
+}
```
mmumby: this can be deleted, my bad.

mmumby (commented):
few comments - but otherwise looks good!

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
BenHarrington: Const works here because we are only re-assigning object properties, not the object itself. If I were to re-assign the whole object than you would be correct.

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
BenHarrington: Yeah I have been fighting with that, not sure why destructuring fails here as it is valid ES6 syntax. I think the rules may need to be updated, and perhaps we can add in null defaults on that to make it more secure.

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
BenHarrington: This is a little bit lazy admittedly, however the backend will never allow for 0 contacts so I wasn't sure the best way to manage this. I believe above I had a destructuring default that included the object so  that this reference would never hard fail, but I am open to suggestions.

services/core-web/src/components/mine/ProjectSummaries/MineProjectSummaryTable.js
BenHarrington: Updated date is the default as it is the most fresh, if that doesn't seem appropriate please let me know

services/minespace-web/src/styles/components/ProjectSummaries.scss
BenHarrington: I got rid of this in a previous PR so not sure how it got here, but will remove

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
mmumby: I think its okay for this PR - but lets create a followup as I was able to POST this am without any contacts.

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryForm.js
mmumby: That works for me - do you want to add those linting rules in this PR or disable the rule for this situation?

services/core-api/app/api/mines/project_summary/resources/project_summary.py
```diff
@@ -105,6 +105,7 @@ def get(self, mine_guid, project_summary_guid):
             'mine_guid': 'The GUID of the mine the Project Description belongs to.',
             'project_summary_guid': 'The GUID of the Project Description to update.'
         })
+    @requires_any_of([MINE_ADMIN, MINESPACE_PROPONENT])
     @api.marshal_with(PROJECT_SUMMARY_MODEL, code=200)
```
mmumby, line 109: 'MINE_ADMIN' is fine for now, but let's remember to update it when we implement the editing on the core side with the new role

services/core-api/tests/auth/test_expected_auth.py
```diff
@@ -44,98 +44,93 @@
...
+     (StandardPermitConditionsListResource, 'post', [EDIT_STANDARD_PERMIT_CONDITIONS]),
+     (StandardPermitConditionsListResource, 'get', [EDIT_STANDARD_PERMIT_CONDITIONS]),
+     (StandardPermitConditionsResource, 'put', [EDIT_STANDARD_PERMIT_CONDITIONS]),
+     (StandardPermitConditionsResource, 'delete', [EDIT_STANDARD_PERMIT_CONDITIONS]),
+     (ProjectSummaryListResource, 'get', [VIEW_ALL, MINESPACE_PROPONENT]),
+     (ProjectSummaryListResource, 'post', [MINE_ADMIN, MINESPACE_PROPONENT]),
+     (ProjectSummaryResource, 'get', [VIEW_ALL, MINESPACE_PROPONENT]),
+     (ProjectSummaryResource, 'put', [MINE_ADMIN, MINESPACE_PROPONENT]),
+     (ProjectSummaryResource, 'delete', [MINE_ADMIN, MINESPACE_PROPONENT])])
 def test_endpoint_auth(resource, method, expected_roles):
```
mmumby, line 134: thanks for adding these!

mmumby (approved):
looks good!

PR #1976, opened 2022-02-07
Bug/mds 4127 contact on proj sum

JustinMacaulay:
# Main

- If a project description does not have a contact it cannot be submitted and an error message indicating that appears under the submit button.
- if no contacts are present then in core a message is displayed indicating so.

# Other

- one small linting fix

# How to test

- Go to minespace and create a new project description on a mine, skip adding a contact entirely and attempt to submit.


mmumby (approved):
looks good!

PR #1986, opened 2022-02-15
bug fix - updated AuthWrapper logic

mmumby:
# Main

Core
- Updated the Admin override so features that should only exist in test, stay in test and admin can no longer see in prod. 
- removed the feature flagging on "delete contacts", its only visible to admin, and with the above bug they've been using it for over a year in prod thinking it was available in prod. 

MineSpace
- Ensured that every button was wrapped in the generic "AuthorizationWrapper" as it will hide the actions when IDIR users log into MineSpace (They should only be visible to proponents)

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


services/core-web/src/tests/components/common/wrappers/AuthorizationWrapper.spec.js
```diff
@@ -42,7 +42,7 @@ describe("AuthorizationWrapper", () => {
 describe("AuthorizationWrapper", () => {
   it("renders properly admin overrides is major mine", () => {
     props.userRoles.push(USER_ROLES[PERMISSIONS.ADMIN]);
-    props.isMajorMine = false;
+    props.isMajorMine = true;
     const component = shallow(<AuthorizationWrapper {...props} />);
```
mmumby, line 46: updated the test case to actually match what it was testing.

areyeslo:
Thanks for these changes. They look good to me.

areyeslo (approved):
Looks good the last change. Final PR ready to prod!! 
Awesome job. Thanks for your hard work not just on this, but in the project.
It was amazing working with you.

PR #1991, opened 2022-02-24
fix/MDS-4151:Update permit amendments in Draft related NOW 

areyeslo:
# Main

- Update permit_amendment table for draft permits related to NoW, is_generated_in_core field has to be always equal to true.

# Other
- The added SQL update is aimed to update the following highlighted records: 
13 records in prod:
image

25 records in test:
image

# How to test
- Execute the following query to get the updates:
`select nai.application_type_code, pa.is_generated_in_core, count(pa.is_generated_in_core)
from permit_amendment pa, now_application_identity nai
where pa.permit_amendment_status_code = 'DFT' and
      pa.now_application_guid = nai.now_application_guid
group by nai.application_type_code, pa.is_generated_in_core`

# Notes
https://bcmines.atlassian.net/browse/MDS-4150

### DIFF
--- a/migrations/sql/V2022.02.23.20.20__fix_update_preambule_text_column_to_existing_draft_permit_amendment_permit.sql
+++ b/migrations/sql/V2022.02.23.20.20__fix_update_preambule_text_column_to_existing_draft_permit_amendment_permit.sql
@@ -1,3 +1,13 @@
+update permit_amendment
+set is_generated_in_core = true
+from (select pa.permit_amendment_guid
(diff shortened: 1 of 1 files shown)


jon-funk:
This looks fine to me, assuming you've tested and confirmed the data change is intended. The prior sql was run in dev and I've deleted the checksum there so it will run without a flyway error. Test never received the prior merge's changes so we're fine on that front.

JustinMacaulay (approved):
Looks good

PR #2004, opened 2022-03-09
Enable Preview Documents to open PDF documents and speed up.

areyeslo:
# Main

- Preview documents in Core and speed up preview

# Other
- N/A

# How to test
-  Download a document from any of the following sections from test/prod:
   -  NoW
   -  Variances
   -  Incidents

- Go to Oversight > Incidents and Investigation
- Create a Mine Incident Record and add a document under Initial Report Documents 
- Click on Preview:
image
- Preview the document successfully:
image

# Notes
https://bcmines.atlassian.net/browse/MDS-4158


services/core-web/package.json
```diff
@@ -11,6 +11,22 @@
     "@babel/runtime-corejs3": "^7.5.5",
     "@datapunt/matomo-tracker-react": "^0.1.5",
     "@hot-loader/react-dom": "^16.13.0",
+    "@syncfusion/ej2-base": "^19.1.57",
```
jon-funk, line 14: were these caused by a formatter in vscode? looks like the versions didn't change

openshift4/templates/filesystem-provider.dc.yaml
```diff
@@ -121,13 +121,13 @@ objects:
...
+    value: 120m
   - name: MEMORY_LIMIT
-    value: 4Gi
+    value: 6Gi
   - name: CPU_REQUEST
-    value: 10m
+    value: 30m
   - name: MEMORY_REQUEST
-    value: 1024Mi
+    value: 2Gi
```
jon-funk, line 130: I'm happy with these resource adjustments, considering the nature of loading large PDFs. Maybe get Juxtin to look at the frontend changes?

services/core-web/package.json
areyeslo: Yes, it was formatted in that order after upgrading syncfusion/ej2-react-pdfviewer from 19.1.57 to 19.4.55. That's the only change. The reason to update was to get FormDesigner object and be able to inject it.

JustinMacaulay: My PR for cypress did the same thing it just moved them up higher in the list.

JustinMacaulay (approved):
Looks good to me.

PR #2006, opened 2022-03-10
MDS-4147: Add notes to each mine incident

BenHarrington:
# Main

- Ticket; https://bcmines.atlassian.net/jira/software/c/projects/MDS/boards/1?modal=detail&selectedIssue=MDS-4147
- Add "Communication bubble" to the Incidents and Investigation table
- Create new MineIncidentNotes modal to handle creation/deletion and display of notes for an incident
- Add new role(core_edit_incidents) and applied it to certain groups on Test/Prod(Health & Safety, Major Mines Permitting, Regional Mines Permitting, Compliance & Enforcement)

# How to test

- Pre-requisite: Navigate to any mine that has an incident/create a new incident
- On the incident table ensure every record has an action icon(far right) that is a chat bubble
- Click the Chat Bubble icon and be presented with this: 
image
- Type in a new note and click "Add Comment" -> ensure the comment panel updates
- Add another note and ensure the messages are ordered from newest to oldest(top to bottom)
- Delete a note and ensure it gets removed correctly


services/core-api/tests/incidents/resources/test_mine_incident_notes_resource.py
```diff
@@ -0,0 +1,58 @@
+import json
+from flask_restplus import marshal
+
+from tests.status_code_gen import *
+from tests.factories import MineIncidentNoteFactory, MineIncidentFactory, MineFactory
+
+
+class TestGetIncidentNotes:
```
areyeslo, line 8: FrontEnd is not allowing to create an empty note. Should we have a test_post_empty_notes which would assert that the comments cannot be empty? Should we check that on the POST resource?

areyeslo (dismissed):
Amazing having the comments for incidents. Very well done Ben!!!

services/core-api/tests/factories.py
```diff
@@ -354,6 +355,22 @@ def documents(obj, create, extracted, **kwargs):
             size=extracted, incident=obj, mine_document__mine=None, **kwargs)
 
 
+class MineIncidentNoteFactory(BaseFactory):
```
JustinMacaulay, line 369: Do you want some fake notes generated when you mock mine data cause if you would need to add an after generate to the incident to add some.

JustinMacaulay (commented):
Just wanted confirmation that all of the comment stuff still works in the code required reports as it looks like you used the components that were made for that? Happy to approve though so just let me know.

services/core-api/tests/factories.py
BenHarrington: I've tested MineComments, and ReportComments as well and they behave as expected. I will add in the post generate mock stuff, thanks Justin!

JustinMacaulay (approved):
Looks good thanks Ben

PR #2008, opened 2022-03-10
Feat/mds 41419 replace reported columns

dragancolic-bcgov:
# Main

- Replaced 'Reported By' columns with 'Inspector Responsible' including the filter.

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -119,6 +119,13 @@ class MineIncident(SoftDeleteMixin, AuditMixin, Base):
     mine_name = association_proxy('mine_table', 'mine_name')
     mine_region = association_proxy('mine_table', 'mine_region')
     major_mine_ind = association_proxy('mine_table', 'major_mine_ind')
+    party_table = db.relationship('Party', lazy='joined', foreign_keys=[reported_to_inspector_party_guid])
+
+    @hybrid_property
+    def reported_to_inspector_party(self):
+        if self.party_table is not None:
+            return self.party_table.party_name
```
JustinMacaulay: if you look on the party table there is a hybrid property, 'name' I would use that here as it will display the first and last for a person and the company name for an organization.

JustinMacaulay (commented):
I am happy to approve after this change.

services/core-api/app/api/incidents/models/mine_incident.py
dragancolic-bcgov: Make sense. Thanks. Fixed.

areyeslo: Would be nice to use this existing hybrid property in Party model:
```
@hybrid_property
    def name(self):
        return f'{self.first_name} {self.party_name}' if self.first_name and self.party_type_code == 'PER' else self.party_name
        
```
in a call from mine_incident model. Something similar to this:
```
@hybrid_property
    def reported_to_inspector_party(self):
        if self.reported_to_inspector_party_guid:
            party = Party.find_by_party_guid(self.reported_to_inspector_party_guid)
            return party.name
        return None
 ```
Could you please include this change?

Comment (areyeslo) [reply services/core-api/app/api/incidents/models/mine_incident.py]: Would be nice to use this existing hybrid property in Party model:
```
@hybrid_property
    def name(self):
        return f'{self.first_name} {self.party_name}' if self.first_name and self.party_type_code == 'PER' else self.party_name
        
```
in a call from mine_incident model. Something similar to this:
```
@hybrid_property
    def reported_to_inspector_party(self):
        if self.reported_to_inspector_party_guid:
            party = Party.find_by_party_guid(self.reported_to_inspector_party_guid)
            return party.name
        return None
 ```
Could you please include this change? Thanks

dragancolic-bcgov: It should be fixed now.

JustinMacaulay (approved):
Looks good, Thank you.

PR #2009, opened 2022-03-14
hotfix/Update livenessProbe and readinessProbe path for 2.4.1 version of docgen.

areyeslo:
# Main
- Update path in probes for docgen after upgrade.

# Other
- N/A

# How to test
- N/A

# Notes
-N/A

### DIFF
--- a/openshift4/templates/docgen.dc.yaml
+++ b/openshift4/templates/docgen.dc.yaml
@@ -120,7 +120,7 @@ objects:
               livenessProbe:
                 failureThreshold: 3
                 httpGet:
-                  path: /health
+                  path: /api/v2/health
                   port: ${{PORT}}
                   scheme: HTTP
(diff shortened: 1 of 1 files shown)


jon-funk (dismissed):
Thank you for the fix Arturo!

openshift4/templates/docgen.dc.yaml
```diff
@@ -120,7 +120,7 @@ objects:
               livenessProbe:
                 failureThreshold: 3
                 httpGet:
-                  path: /health
+                  path: api/v2/health
```
BenHarrington: I think there needs to be a leading slash for both the paths

areyeslo: thanks Ben!!

BenHarrington (approved):
Thanks for updating that Arturo!

PR #2010, opened 2022-03-15
Remove experimental fe build

jon-funk:
# Main

- Remove a experimental github action build for the frontend. We're removing it now because it's being formalized for all services & we need to reduce the spread of images we're building (this had its own image stream).

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


JustinMacaulay (approved):
Looks good

PR #2017, opened 2022-03-26
MDS-4205: Error when loading permits searched by now_application_guid

areyeslo:
# Main

- Error when referring to _context_mine property in an empty permit object.

# Other
- N/A

# How to test
- Query this mine_guid in Metabase and recreate the scenario in your local:
```
select *
from permit_amendment pa
where pa.mine_guid='558524a0-94b1-41d5-8b7b-65225c44f2b6' and 
      pa.permit_amendment_status_code= 'DFT'
```
-  Create two permit amendments in Draft, one is generated in Core and the other not.

# Notes
https://bcmines.atlassian.net/browse/MDS-4205


JustinMacaulay (approved):
I get your comment on not really wanting to do it this way but I think its OK for now if it solves our issue.

PR #2018, opened 2022-03-30
Add revision limits to dcs

jon-funk:
# Main

- Explicitly set the revision limits on dcs
- This controls the number of replicationcontrollers that can exist under the dc, which keep images around while they persist

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


JustinMacaulay (approved):
Looks good

PR #2019, opened 2022-03-30
Add customresource for argocd management

jon-funk:
# Main

- Adds manifest for version controlling argocd team space
- Previously, this was manually managed by platform services, but now they have an available custom resource we can provision to control things ourselves

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


wilson496 (approved):
Looks good to me.

PR #2021, opened 2022-03-30
MDS-3843 - remove backend build from build.yaml, add Dockerfile.ci, remove build config.

wilson496:
# Main

Pull backend / core-api build out of build.yaml (monolith build GHA for building in Openshift) and put it into a separate build file. 

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


services/core-api/Dockerfile.ci
```diff
@@ -0,0 +1,29 @@
...
+
+COPY . .
+
+# Fix log path permissions
+# This is a bit hacky but other fixes require changing app code
+RUN mkdir -p /var/log/core-api
+RUN chmod -R 700 /var/log
+RUN chown -R 1000:0 /var/log
+
+# Setup user to represent developer permissions in container
```
vyasworks, line 20: Do we need to represent developer permissions in remote environments?

services/core-api/Dockerfile.ci
```diff
@@ -0,0 +1,29 @@
...
+# Update installation utility
+RUN apt-get update
+
+# Install the requirements
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+# Fix log path permissions
```
vyasworks, line 14: These permissions allow for volume mount locally 

in the cluster the logs must go via stdout to log aggregator. Since we use the default logger. 

If that is not the case, we should pipe all our logs to stdout in dev in proper JSON parable format. 


https://github.com/bcgov/mds/blob/ba754e73cf603752718cc05b27d6b0d61bea0f5d/services/core-api/app/config.py#L16-L43


Good for now but I would like to revisit this later

vyasworks:
Loooks good, few questions on dockerfile.ci 

We should also remove the s2i code in the repo.

services/core-api/Dockerfile.ci
wilson496: It appears that the user is set up with the same permissions as local dev, and is also used to run the API server.

Unless there's some sort of security / permissions problem that I don't see here, I don't see this as an issue. If having the generated 'python' user run the service with similar permissions to a dev makes it work and work securely, this is fine to leave as is.

services/core-api/Dockerfile.ci
wilson496: Definitely worth revisiting and pointing out. Looking at the logs in Openshift, it outputs the health and liveness checks as well as API calls or errors. Let's keep this as is for now, but I'll create a ticket to address this.

wilson496: Following up with created ticket: https://bcmines.atlassian.net/browse/MDS-4338

vyasworks (approved):

PR #2025, opened 2022-03-31
Update core-api liveness and readiness probes

vyasworks:
# Main

Update Liveness and Readiness Check. 

https://bcmines.atlassian.net/browse/MDS-4215
https://bcmines.atlassian.net/browse/MDS-4216

# Other

Improve generic health check 
Update the timeout and frequency of healthcheck settings of containers 

# How to test

Test endpoint locally / in dev

### DIFF
--- a/openshift4/templates/backend.dc.yaml
+++ b/openshift4/templates/backend.dc.yaml
@@ -335,20 +335,20 @@ objects:
                   memory: ${MEMORY_REQUEST}
               livenessProbe:
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/live
                   port: 5000
                   scheme: HTTP
                 initialDelaySeconds: 160
-                timeoutSeconds: 15
-                periodSeconds: 30
+                timeoutSeconds: 1
+                periodSeconds: 1
               readinessProbe:
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/ready
                   port: 5000
                   scheme: HTTP
                 initialDelaySeconds: 160
-                timeoutSeconds: 15
-                periodSeconds: 30
+                timeoutSeconds: 5
+                periodSeconds: 5
               terminationMessagePath: /dev/termination-log
               terminationMessagePolicy: File
               imagePullPolicy: IfNotPresent
--- a/openshift4/templates/backend.dc.yaml
+++ b/openshift4/templates/backend.dc.yaml
@@ -338,7 +338,7 @@ objects:
                   path: ${BASE_PATH}/health/live
                   port: 5000
                   scheme: HTTP
-                initialDelaySeconds: 160
+                initialDelaySeconds: 80
                 timeoutSeconds: 1
                 periodSeconds: 1
               readinessProbe:
@@ -346,7 +346,7 @@ objects:
                   path: ${BASE_PATH}/health/ready
                   port: 5000
                   scheme: HTTP
-                initialDelaySeconds: 160
+                initialDelaySeconds: 80
                 timeoutSeconds: 5
                 periodSeconds: 5
               terminationMessagePath: /dev/termination-log
--- a/services/core-api/app/__init__.py
+++ b/services/core-api/app/__init__.py
@@ -1,4 +1,6 @@
+import datetime
 import logging
+import requests
 from logging.config import dictConfig
 from flask import Flask, request
 from flask_cors import CORS
@@ -30,7 +32,7 @@ from app.config import Config
 #alias api to avoid confusion with api folder (speifically on unittest.mock.patch calls)
 from app.extensions import db, jwt, api as root_api_namespace, cache
 from app.api.utils.setup_marshmallow import setup_marshmallow
-
+from sqlalchemy.sql import text
 
 def create_app(test_config=None):
     """Create and configure an instance of the Flask application."""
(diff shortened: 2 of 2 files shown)


services/core-api/app/__init__.py
```diff
@@ -97,11 +99,94 @@ def register_routes(app):
...
+                app.logger.error("Docgen health check failed " + str(error))
+                service['docgen'] = False
+                status = 503
+
+            try:
+                service['docman'] = get_service_status('DOCUMENT_MANAGER_URL')
+            except Exception as error:
+                app.logger.error("Document Manager health check failed " + str(error))
+                service['docman'] = False
+                status = 503
```
vyasworks, line 148: Having a 503 error if any of the dependencies are not available will be useful for alerting by external monitoring services.

openshift4/templates/backend.dc.yaml
```diff
@@ -335,20 +335,20 @@ objects:
...
-                periodSeconds: 30
+                timeoutSeconds: 1
+                periodSeconds: 1
               readinessProbe:
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/ready
                   port: 5000
                   scheme: HTTP
                 initialDelaySeconds: 160
```
will-freshworks: Does it really take 160s for the backend to come up? It's a python runtime so I would expect more like 5s.

openshift4/templates/backend.dc.yaml
```diff
@@ -335,20 +335,20 @@ objects:
...
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/live
                   port: 5000
                   scheme: HTTP
                 initialDelaySeconds: 160
-                timeoutSeconds: 15
-                periodSeconds: 30
+                timeoutSeconds: 1
+                periodSeconds: 1
```
will-freshworks, line 343: This is a bit intense. Maybe 5 secs? Maybe with 2 for timeout?

services/core-api/app/__init__.py
```diff
@@ -97,11 +99,94 @@ def register_routes(app):
...
+            return service, status
+
+        def get(self):
+            return self.get_health()
+
+    # Liveness Endpoint to make sure that python server is ready to accept connections and container / server is running. 
+    @root_api_namespace.route('/health/live')
+    class Livenesscheck(Resource):
+        def get(self):
+            return {'live': True}
```
will-freshworks, line 159: Is this PR _just_ readiness? If so, the name of the PR needs updating

services/core-api/app/__init__.py
will-freshworks: Yeah I like the approach of checking every upstream service. If core-api won't work without them it should check them. Be careful we don't end up with a cycle in our liveness graph though!

will-freshworks (changes requested):
I think the `initialDelaySeconds` should be smaller and the period should be higher, but perhaps I'm missing something.

openshift4/templates/backend.dc.yaml
vyasworks: I had the same question, but I believe there is a reason this has been set to 160 @jon-funk . Can we reduce it?

openshift4/templates/backend.dc.yaml
vyasworks: This is liveness check. So it should be ok.

vyasworks: https://bcmines.atlassian.net/browse/MDS-4216

openshift4/templates/backend.dc.yaml
jon-funk: The backend is quite slow to come up, I recall it taking 40-60 seconds minimum back when I made this change. Plus variance from how openshift is doing. If you'd like to try tighten it up though, go for it!

vyasworks: I am going to experiment with 80 seconds as a safe middle ground. This will be reviewed going forward

openshift4/templates/backend.dc.yaml
```diff
@@ -335,20 +335,20 @@ objects:
...
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/live
                   port: 5000
                   scheme: HTTP
-                initialDelaySeconds: 160
-                timeoutSeconds: 15
-                periodSeconds: 30
+                initialDelaySeconds: 80
+                timeoutSeconds: 1
```
will-freshworks, line 342: This still feels too regular for a liveness check but happy for you to try it.

openshift4/templates/backend.dc.yaml
```diff
@@ -335,20 +335,20 @@ objects:
...
               livenessProbe:
                 httpGet:
-                  path: ${BASE_PATH}/health
+                  path: ${BASE_PATH}/health/live
                   port: 5000
                   scheme: HTTP
-                initialDelaySeconds: 160
-                timeoutSeconds: 15
-                periodSeconds: 30
+                initialDelaySeconds: 80
```
will-freshworks, line 341: I just don't understand how a flask app can take this long to start-up, but happy to start with this and move forwards

services/core-api/app/__init__.py
```diff
@@ -97,11 +99,94 @@ def register_routes(app):
...
+    def get_database_status():
+        return db.session.query("up").from_statement(text("SELECT 1 as up")).all()[0][0] == 1;
+
+    def get_cache_status():
+        now = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+        cache.set('ping', now)
+        cached_now = cache.get('ping')
+        return now == cached_now
+
+    def get_service_status(URI):
```
will-freshworks: variable names are lower camel case, so just `uri` here

services/core-api/app/__init__.py
```diff
@@ -97,11 +99,94 @@ def register_routes(app):
...
+        def get(self):
+            return {'live': True}
+
+    # Readiness Endpoint to make sure that the container is ready to process the request it receives. Dependencies must be available 
+    @root_api_namespace.route('/health/ready')
+    class Readinesscheck(Resource):
         def get(self):
-            return {'status': 'pass'}
+            try:
+                get_database_status()
```
will-freshworks, line 166: are you meaning to not check the result here?

services/core-api/app/__init__.py
```diff
@@ -97,11 +99,94 @@ def register_routes(app):
     root_api_namespace.add_namespace(orgbook_api)
     root_api_namespace.add_namespace(EMLI_contacts_api)
 
-    # Healthcheck endpoint
+    # General Service status
     @root_api_namespace.route('/health')
     class Healthcheck(Resource):
+        def get_health(self):
```
will-freshworks, line 105: This isn't being used any more

will-freshworks (changes requested):

services/core-api/app/__init__.py
vyasworks: It will be used by external monitoring services - Uptime robot

services/core-api/app/__init__.py
vyasworks: the check is done in function.

will-freshworks: If it fails, the function would return `False`, which then gets ignored.

will-freshworks (changes requested):

will-freshworks (approved):

PR #2030, opened 2022-04-11
MDS-4196: Create Project namespace to support IRT development

BenHarrington:
# Main

- Ticket: https://bcmines.atlassian.net/jira/software/c/projects/MDS/boards/1?modal=detail&selectedIssue=MDS-4196
- Please see ticket for refactor details

# Other

- Project_summary_title, proponent_project_id, project_summary_lead_party_guid have been removed from project_summary and placed at the project level
- Project_summary_contact is not currently in use but still exists in the API, this will be addressed as more Project requirements become available

# How to test

- Verify Project Descriptions CRUD still works in Minespace
- Verify Project Descriptions RU still works in Core

# Notes

- UI is still currently the same, changes will be coming in a upcoming sprint


migrations/sql/V2022.03.2916.56__create_project_tables.sql
```diff
@@ -0,0 +1,75 @@
+CREATE TABLE IF NOT EXISTS project (
+  project_guid            uuid DEFAULT gen_random_uuid()        PRIMARY KEY,
+  project_id              serial                                   NOT NULL,
+  mine_guid               uuid                                     NOT NULL,
```
JustinMacaulay, line 4: Does this also need a reference to permit? Not as a hard link but can some projects not be linked to a permit as well because they need an amendment?

migrations/sql/V2022.03.2916.56__create_project_tables.sql
```diff
@@ -0,0 +1,75 @@
...
+
+COMMENT ON TABLE project_contact IS 'Project contact details related to a project. ';
+
+-- Add foreign keys to project on children(Summary)
+ALTER TABLE project_summary ADD COLUMN IF NOT EXISTS project_guid uuid;
+ALTER TABLE project_summary ADD CONSTRAINT project_guid_fkey FOREIGN KEY (project_guid) REFERENCES project(project_guid);
+
+-- Data Fix:
+-- Copy data from project_summary to project table(create a project for every project summary)
+INSERT INTO project (mine_guid, project_title, project_lead_party_guid, proponent_project_id, create_user,update_user)
```
JustinMacaulay, line 56: Should the insert and data copying be on a separate migration? I think in the past we have tried to keep the addition of new tables and data migrations on separate migrations.

services/minespace-web/common/constants/API.js
```diff
@@ -116,15 +116,25 @@ export const VARIANCE_DOCUMENT = (mineGuid, varianceGuid, documentManagerGuid) =
...
-export const PROJECT_SUMMARY = (mineGuid, projectSummaryGuid) =>
-  `/mines/${mineGuid}/project-summaries/${projectSummaryGuid}`;
-export const PROJECT_SUMMARY_DOCUMENTS = (mineGuid, projectSummaryGuid) =>
-  `/mines/${mineGuid}/project-summaries/${projectSummaryGuid}/documents`;
-export const PROJECT_SUMMARY_DOCUMENT = (mineGuid, projectSummaryGuid, documentManagerGuid) =>
-  `/mines/${mineGuid}/project-summaries/${projectSummaryGuid}/documents/${documentManagerGuid}`;
+// Project & Project Summaries
+export const PROJECT_PROJECT_SUMMARIES = (projectGuid, params = {}) =>
+  `/projects/${projectGuid}/project-summaries?${queryString.stringify(params)}`;
+export const NEW_PROJECT_SUMMARY = () => "/projects/new/project-summaries/new";
```
JustinMacaulay, line 122: I am not the biggest fan of using new in place of a GUID, not sure if there is a better way to do this or if we should chat about it a bit.

services/minespace-web/common/actionCreators/projectSummaryActionCreator.js
```diff
@@ -62,7 +66,10 @@ export const fetchProjectSummariesByMine = ({ mineGuid }) => (dispatch) => {
   dispatch(request(reducerTypes.GET_PROJECT_SUMMARIES));
   dispatch(showLoading());
   return CustomAxios({ errorToastMessage: Strings.ERROR })
-    .get(ENVIRONMENT.apiUrl + API.MINE_PROJECT_SUMMARIES(mineGuid), createRequestHeader())
+    .get(
+      ENVIRONMENT.apiUrl + API.PROJECT_PROJECT_SUMMARIES(null, { mine_guid: mineGuid }),
```
JustinMacaulay, line 70: I feel like this URL should be something more like /Projects?mine_guid=<guid> and not using the same project URL because your not fetching a specific Project summary, you are getting all projects for a specific mine. the project summary is just a piece of the project.

services/core-web/src/components/mine/ProjectSummaries/ProjectSummary.js
```diff
@@ -94,8 +94,9 @@ export class ProjectSummary extends Component {
   handleUpdate = (message) => {
     const mineGuid = this.props.match?.params?.mineGuid;
     const projectSummaryGuid = this.props.match?.params?.projectSummaryGuid;
+    const projectGuid = this.props.formValues.project_guid;
```
JustinMacaulay, line 97: Is there a reason this has to come from form values when the other two come from params? just not sure why there is a disconnect.

services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
```diff
@@ -1,23 +1,53 @@
...
-from app.api.mines.project_summary.models.project_summary import ProjectSummary
-from app.api.mines.project_summary.models.project_summary_contact import ProjectSummaryContact
-from app.api.mines.project_summary.models.project_summary_authorization import ProjectSummaryAuthorization
+from app.api.projects.project_summary.response_models import PROJECT_SUMMARY_MODEL
+from app.api.projects.project.models.project import Project
+from app.api.projects.project_summary.models.project_summary import ProjectSummary
 
 
-class ProjectSummaryListResource(Resource, UserMixin):
+class ProjectSummaryListGetResource(Resource, UserMixin):
```
JustinMacaulay, line 17: I would also like to chat on Tuesday around these resources and why there is a distinct difference between the get and post resource.

services/core-api/app/api/projects/namespace.py
```diff
@@ -0,0 +1,29 @@
...
+api.add_resource(ProjectResource, '/<string:project_guid>')
+
+# Project Description(Summary)
+api.add_resource(ProjectSummaryResource,
+                 '/<string:project_guid>/project-summaries/<string:project_summary_guid>')
+api.add_resource(
+    ProjectSummaryDocumentUploadResource,
+    '/<string:project_guid>/project-summaries/<string:project_summary_guid>/documents')
+api.add_resource(ProjectSummaryListGetResource, '/<string:project_guid>/project-summaries')
+api.add_resource(ProjectSummaryListPostResource, '/new/project-summaries/new')
```
JustinMacaulay, line 25: Would like to have a discussion and maybe find a better version of this.

JustinMacaulay (commented):
A few comments and would love to have a convo on tuesday about some of them.

services/core-web/src/components/mine/ProjectSummaries/ProjectSummary.js
BenHarrington: Because we don't have Project setup in minespace yet I've had to do some work arounds until that work is done. This is meant only as a temporary measure.

migrations/sql/V2022.03.2916.56__create_project_tables.sql
BenHarrington: If that is true it was never communicated to me. Conversely a permit could have an optional FK to project perhaps?

services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
BenHarrington: Much like most of your other comments, the reason some of this looks odd is because Minespace doesn't have  a Project space yet in the UI meaning we don't have access to the project_guid in many circumstances, these are all temporary measures and will be addressed when the work on https://bcmines.atlassian.net/browse/MDS-4261 begins. Anywho I can walk you through that tomorrow.

migrations/sql/V2022.03.2916.56__create_project_tables.sql
BenHarrington: We can definitely do that if you would prefer. I kept it in here because the datafix needs to be done directly after the DDL and they are related, but I can also create another one and use a name that indicates the relationship between the two.

areyeslo (approved):
Amazing job on getting things ready to start with IRT.

PR #2031, opened 2022-04-14
MDS-4229: CORE: Updating Assign Inspector Role Date = 400 Bad Request Error

BenHarrington:
# Main

- Ticket: https://bcmines.atlassian.net/browse/MDS-4229

# Other

- Updated toast error message to show correct context(project lead vs. inspector)

# How to test

- Update project lead or inspector role that already has a start date set and no end date to a start date earlier than the current one
- Update project lead or inspector role that already has a start date set and no end date to a start date later than the current one


areyeslo (approved):
Great job on this fix. With this change the user can modify the dates without any issue.
Very well done!!

PR #2041, opened 2022-04-20
feature/MDS-IRT-backend

areyeslo:
# Main

- Database implementation for Information Requirements Table (IRT)
- CRUD endpoints for IRT and requirements
- Pytests

# Other
- N/A

# How to test
- N/A

# Notes
https://bcmines.atlassian.net/browse/MDS-4209


migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
```diff
@@ -0,0 +1,322 @@
...
+);
+ALTER TABLE information_requirements_table_status_code OWNER TO mds;
+--
+-- Name: TABLE information_requirements_table_status_code; Type: COMMENT; Schema: public; Owner: mds
+--
+
+COMMENT ON TABLE information_requirements_table_status_code IS 'Information Requirements Table (IRT) status. ';
+
+
+insert into information_requirements_table_status_code(status_code, description, create_user, update_user)
```
JustinMacaulay, line 41: all of these can be done in a single insert statement instead

migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
```diff
@@ -0,0 +1,322 @@
...
+ insert into information_requirements_table_status_code(status_code, description, create_user, update_user)
+   values('UNR','Under review - with reviewers', 'system-mds','system-mds');
+  
+ insert into information_requirements_table_status_code(status_code, description, create_user, update_user)
+   values('APV','Approved', 'system-mds','system-mds');
+
+  
+insert into requirements (requirement_id, description, display_order, create_user, update_user)
+  values
+    (1, 'Introduction and Project Overview', 1, 'system-mds','system-mds'),
```
JustinMacaulay, line 50: is the sql and the serial requirements_id ok with you adding these id's manually? I dont think it will care when your adding them but if there is a situation when one gets added without this defined is the serial ID going to be in the right place?

migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
```diff
@@ -0,0 +1,322 @@
...
+
+COMMENT ON TABLE irt_requirements_xref IS 'Reference table between Information Requirements Table (IRT) and associated requirements. ';
+
+
+
+
+
+-- Before droping these columns, need to run data fix to copy data into project table
+
+-- ALTER TABLE project_summary
```
JustinMacaulay: does this all have to be run manually?

services/core-api/tests/projects/requirements/models/test_requirements_model.py
```diff
@@ -0,0 +1,17 @@
+# from tests.factories import RequirementsFactory
+# from app.api.projects.information_requirements_table.models.requirements import Requirements
```
JustinMacaulay: is this needed?

JustinMacaulay (commented):
The only worry I have is if the requirements themselves need to be editable that will require a small rework of how they work.

migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
areyeslo: changing to one.

migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
areyeslo: you are right, the same adjustment as permit_standard_conditions.

migrations/sql/V2022.04.1316.56__create_IRT_tables.sql
areyeslo: I have to delete this.

services/core-api/tests/projects/requirements/models/test_requirements_model.py
areyeslo: Not needed

areyeslo:
When the requirements would need to be editable themselves then we update them invoking the associated PUT method passing the requirement GUID:
http://localhost:5000/projects/requirements/1b9916df-b747-4074-a2ef-88b8921cabd0

with the specified requirement:
{
	"requirement_guid": "1b9916df-b747-4074-a2ef-88b8921cabd0",
        "requirement_id": 10,
        "parent_requirement_id": 1,
	"description": "Application Background - Updated",
	"display_order": 1

}

Should be the same as we are doing for standard_permit_conditions where the user can modify a condition.  Are thinking in a different scenario?

BenHarrington (dismissed):
Fantastic work Arturo! This should set us up nicely moving forward.

BenHarrington (approved):

PR #2042, opened 2022-04-20
Mds 4186 update validation check

matbusby-fw:
# Main

- Validation wasn't fully working as expected
- Refactored form as functional component (personal preference)
- Improved the validation flow to trigger properly with field level errors and not allow the submission of the select field with random text entered
- reset form on submit

# Other

- N/A

# How to test

- N/A

# Notes

image


hitankar-freshworks:
looks like you moved functional components over a class component.

hitankar-freshworks (approved):

matbusby-fw:
> looks like you moved functional components over a class component.

Yeah, I don't see any benefit in using a class component and the functional feels a lot cleaner.  Now that I have a better grip on what's happening it seemed like a better choice.

PR #2045, opened 2022-04-20
Hotfix/fix project refactor migration

BenHarrington:
# Main

- Keep all headers and just leave a single list item 'N/A' if there's no content for that header, otherwise remove N/A and add your details!

# Other

- N/A

# How to test

- N/A

# Notes

- N/A


areyeslo (approved):
Good catch Ben!! Thanks for your hard work on this

areyeslo (commented):
How about if we create a new migration for the data fix?

areyeslo (changes requested):
How about if we create a new migration for the data fix?

areyeslo (approved):
Looks great, thanks for working on this.
Go to Prod!!

PR #2047, opened 2022-04-22
Create new role edit TSF

areyeslo:
# Main

- Create new role to edit TSF

# Other
- Creation of EDIT_TSF role in Keycloak 
- Change of MINE_EDIT to EDIT_TSF in PUT & POST method
- Change permission in Core-web for Add button from EDIT_PERMITS to EDIT_TSF
- Change permission in Core-web for Edit button from EDIT_REPORTS to EDIT_TSF
- Update permissions in tests
- Update snapshots
  

# How to test
- Login to Core-web with your IDIR and confirm you cannot Add/Edit a TSF.
- Login to minespace-web with BCEID and confirm that as proponent you can edit a TSF
- Assign the new role EDIT_TSF to your account in keycloak and login to Core-web with your IDIR account.
- Confirm you CAN Add/Edit a TSF

# Notes
https://bcmines.atlassian.net/browse/MDS-4094


services/core-api/tests/auth/test_expected_auth.py
```diff
@@ -72,7 +72,7 @@
      (MineReportListResource, "post", [EDIT_REPORT, MINESPACE_PROPONENT]),
      (MineStatusXrefListResource, "get", [VIEW_ALL]),
      (MineTailingsStorageFacilityListResource, "get", [VIEW_ALL]),
-     (MineTailingsStorageFacilityListResource, "post", [MINE_EDIT]),
+     (MineTailingsStorageFacilityListResource, "post", [EDIT_TSF]),
```
BenHarrington, line 76: Should we also have another entry for the "put" given that it requires the same permissions?

BenHarrington (changes requested):
One other thing that would be nice is to add the new role to the Production Keycloak, just so it is easier for Colin, Harp or whomever to add that role fast to select users.

Otherwise I've tested it and things are behaving as expected. Great work!

JustinMacaulay (commented):
Looks good and my only comment was the same as bens so I removed it.

BenHarrington (approved):
Thanks for making those changes!

PR #2062, opened 2022-05-02
MDS-4350-Message Tool in Communication related to a Mine not appearing

areyeslo:
## Objective 

https://bcmines.atlassian.net/browse/MDS-4350

Currently, Core users with no Admin privileges cannot create new comments under Communication for a mine. To post comments should be widely available among all Core Users.

Login in Core and select a mine. Then, at the top right click on Options > Communication to display a dialog box to insert comments:
image

With the current change, Core users with no Admin privileges can post comments:
image

The sections affected are the following:
- Create comment for a mine (explained above)
- Create comment for incident (requires "core_edit_incidents" permission to get access to the button to open comment dialog)
- Create comment for report (requires "core_edit_reports" permission to get access to open comment dialog)

The changes in the PR apply for invokations to CommentPanel component not having value in createPermission, so if there ir no special privilege, the user could see the comment dialog.

### DIFF
--- a/services/core-web/src/components/common/comments/CommentPanel.js
+++ b/services/core-web/src/components/common/comments/CommentPanel.js
@@ -40,7 +40,7 @@ const defaultProps = {
 export const CommentPanel = (props) => {
   const createPermission = props.userRoles.includes(USER_ROLES[props?.createPermission])
     ? props.createPermission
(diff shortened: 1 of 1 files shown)


BenHarrington (approved):
Good catch on that conditional, thanks Arturo!

PR #2065, opened 2022-05-03
Get back backend tag in github actions for prod deployment.

areyeslo:
## Objective 
Deploy to Production

Why are you making this change? 
Getting back backend tag to deploy to Production.

Alternative workflow considering ArgoCd is not fully ready for deployment.


areyeslo:
No need to use this change. Deployment to Prod was done successfully.

PR #2059, opened 2022-04-27
MDS-4217 - remove frontend references from build.yaml, add separate build file for frontend

wilson496:
# Main

- What did you focus on in this PR?
- Keep all headers and just leave a single list item 'N/A' if there's no content for that header, otherwise remove N/A and add your details!

# Other

- What other side fixes, tweaks did you add on the side?
- N/A

# How to test

- N/A

# Notes

- Any notes/screenshots for this PR?
- N/A

PR #2068, opened 2022-05-05
Revert "MDS-4217 - remove frontend references from build.yaml, add separate build file for frontend"

wilson496:
Reverts bcgov/mds#2059


areyeslo (approved):
Getting the original front end looks good to me Cameron. Thanks for making this change.

PR #2073, opened 2022-05-10
MDS-4217 - patch to add deploy files for other environments

wilson496:
Patch to add deploy files for other environments.


.github/workflows/core-web.deploy.test.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install oc
+        uses: redhat-actions/openshift-tools-installer@v1
+        with:
+          oc: "4.7"
+
+      - name: oc login
+        run: |
+          oc login --token=${{ secrets.BUILD_TOKEN }} --server=${{ secrets.CLUSTER_API }}
+
+      - name: Promote core-web from dev to test
```
vyasworks, line 24: How does the roll out happen after promotion? 

The gitops projects / trigger are missing. So is the roll out.

wilson496: Right. Adding that now.

vyasworks (approved):

PR #2077, opened 2022-05-10
Add recognized default for maxFiles

BenHarrington:
## Objective 

- Fix no default of maxFiles causing upload to not work


matbusby-fw (approved):
Good catch!  Sorry about that!

PR #2089, opened 2022-05-17
add nod statuses and ability to set status from core

hitankar-freshworks:
## Objective 

Add nod statuses and the ability to set status in core


matbusby-fw (approved):
Looks good!  We'll get you enjoying the front end soon enough! 😜

PR #2091, opened 2022-05-17
[fix] Remove erroring and unwanted after migrate hook

vyasworks:
## Objective 

This migration file causes issues on local env on fresh installation / clone 

There seems to be no clear motive to have this file. 

We are connecting to a "" null remote db to create `mds_test` for no reason.


migrations/sql/afterMigrate__87__create_test_db.sql
```diff
@@ -1,13 +0,0 @@
-DO
-$$
-DECLARE
-BEGIN
-  CREATE EXTENSION IF NOT EXISTS dblink;
-  IF (EXISTS (SELECT 1 FROM pg_database WHERE datname = 'mds_test')) THEN
-    RAISE NOTICE 'Database already exists';
-  ELSE
-    PERFORM dblink_connect('');
-    PERFORM dblink_exec('CREATE DATABASE mds_test OWNER mds');
```
vyasworks, line 10: Commenting out these two lines make the migration a success, so I suppose there is no need for this file.

hitankar-freshworks: This is necessary for setting up `test` database locally. I think migration will fail otherwise!

vyasworks: does local testing run against test database?

hitankar-freshworks: I believe so. we would like to run integration tests locally to make sure they pass locally before pushing it for a pull request

BenHarrington: We use mds_test to run integration tests locally as Hitankar mentioned so we should keep this.

vyasworks:
This PR is not relevant anymore. Fixes done on postgres migration.

PR #2093, opened 2022-05-18
MDS-3845 - test build file for docgen api

wilson496:
## Objective 

[MDS-](https://bcmines.atlassian.net/browse/MDS-3845)

Refactor build for docgen-api service


.github/workflows/docgen.build.deploy.dev.yaml
```diff
@@ -0,0 +1,43 @@
+name: Build docgen
+
+on: 
+  pull_request:    
+  push:
+    branches:
+      - develop
+    path:
+      - services/core-api/**
```
vyasworks: core-api reference

.github/workflows/docgen.build.deploy.dev.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - .github/workflows/docgen.build.yaml
```
vyasworks: Name not referenced correctly

.github/workflows/docgen.build.deploy.dev.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+env:
+  INITIAL_TAG: latest
+  TAG: dev
+  NAME: docgen
+  CONTEXT: services/core-api/
```
vyasworks: wrong context

.github/workflows/docgen.build.deploy.dev.yaml
```diff
@@ -0,0 +1,43 @@
...
+  image: docker.io/bcgovimages/common-document-generation-service
+  version: 2.4.1
+
+jobs:
+  build-docgen:
```
vyasworks: Do we need to build anything at all if we are using standard image that is already available? 

We could use it directly on the Kustomize files.

.github/workflows/docgen.deploy.prod.yaml
```diff
@@ -0,0 +1,56 @@
+name: DOCGEN - Promote PROD
```
vyasworks: Same question applies for test / prod..

vyasworks (changes requested):

.github/workflows/docgen.build.deploy.dev.yaml
```diff
@@ -0,0 +1,43 @@
...
+    path:
+      - services/core-api/**
+      - .github/workflows/docgen.build.yaml
+
+env:
+  INITIAL_TAG: latest
+  TAG: dev
+  NAME: docgen
+  CONTEXT: services/core-api/
+  image: docker.io/bcgovimages/common-document-generation-service
```
vyasworks: We do we need to build on top of this?

vyasworks (changes requested):

vyasworks:
Looks like some commits from master are also in this, which are not relavant to the PR, rebasing + cherry pick should help

.github/workflows/docgen.build.deploy.dev.yaml
wilson496: That was intentional. It's a service that ties into the core-api, so I wanted to trigger builds based on changes to core-api.

.github/workflows/docgen.build.deploy.dev.yaml
wilson496: This file is essentially just taking the Dockerfile and pulling it to build. A little redundant if we don't update it, but I figured this way gave us the opportunity to have a specified way to build this service, as well as specify which version we pull.

.github/workflows/docgen.build.deploy.dev.yaml
wilson496: This can be removed since it's not needed. Good catch.

.github/workflows/docgen.build.deploy.dev.yaml
vyasworks: Why do we need to build if there are changes to core-api? Changes to core-api does not affect docgen service in anyway

.github/workflows/docgen.build.deploy.dev.yaml
vyasworks: The version we pull can be marked in the kustomize deployment file too. 

Something like `docker.io/bcgovimages/common-document-generation-service:VERSION`

If we have nothing to append to this image/ custom config stuff that needs to be done on top of base image, then we do not need a build.

wilson496: Fair enough. I'll just remove the references for this stuff and put the image into the kustomize deployments then.

vyasworks (approved):

PR #2094, opened 2022-05-18
MDS-4359: Import IRT in MineSpace - Parse and save step

BenHarrington:
## Objective 

[MDS-4359](https://bcmines.atlassian.net/browse/MDS-4359)

- Renamed resources to be consistent with project
- Implemented function to parse, sanitize, and validate spreadsheet before creating associated models(POST)
- Disabled PUT resource test until fully implemented
- Renamed redux associated files and constants for consistency
- Extended FileUpload component to support additional actions "afterSuccess" of document upload has occured
- Fixed styling on Step 2 of IRT workflow
- Added feature flag on ProjectStages row for IRT
- Created component test spec and snapshot for new IRT main page


services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table_list.py
```diff
@@ -1,21 +1,134 @@
...
+            is_row_top_level_category = True if active_requirement[
+                0].parent_requirement_id is None else False
+            # If the "Information" requirement provided does not match DB, an empty row is provided, or the row contains a top level category do not create a requirement
+            if active_requirement and is_empty_row is False and is_row_top_level_category is False:
+                new_requirement_dict = {
+                    'information': sanitized_information_cell,
+                    'requirement_guid': active_requirement[0].requirement_guid,
+                    'required': required_cell,
+                    'methods': methods_cell,
+                    'comments': comments_cell
```
areyeslo: should be 'comment' as in irt_requirements_xref table to keep the value in the database.

services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table_list.py
```diff
@@ -1,21 +1,134 @@
...
+                continue
+            # If "Required" or "Methods" cell entry is true and "Comments" are 'None' add error
+            if (required_cell is True or methods_cell is True) and comments_cell == 'None':
+                import_errors.append(
+                    f'Row {row_number} - "Required" or "Methods" cells is checked off and requires "Comments" to be provided.'
+                )
+                continue
+
+            is_empty_row = required_cell is False and methods_cell is False and comments_cell == 'None'
+            active_requirement = [
```
areyeslo: if we check just for the description, we might fall into the issue with similar descriptions. How to ensure to get the correct requirement_guid for Summary in section 2 instead of Summary in section 5? Would we need to consider the level as well to be able to differentiate between similar descriptions?

BenHarrington: So far in my tests this hasn't become an issue but that is a fair point. Perhaps as we continue to iron out the spreadsheet and feature we can enhance this check if it becomes problematic, as it stands it isn't a wildcard comparison and I think should suffice but let me know if you have serious concerns about that explanation.

areyeslo (approved):
The implementation looks amazing. Thanks for all your hard work on this Ben!!! Excellent job on this one.

PR #2095, opened 2022-05-18
MDS-4198 download import final irt minespace

areyeslo:
## Objective 

[MDS-4198](https://bcmines.atlassian.net/browse/MDS-4198)

-  Front end work for IRT to download template and import final IRT.


services/minespace-web/src/components/pages/Project/ProjectPage.js
```diff
@@ -97,6 +99,11 @@ export class ProjectPage extends Component {
                 <Tabs.TabPane tab="Overview" key="overview">
                   <ProjectOverviewTab />
                 </Tabs.TabPane>
+                {!IN_PROD() && (
```
areyeslo, line 102: Just available in PROD

services/core-api/app/api/projects/information_requirements_table/resources/Information_requirements_table_import.py
```diff
@@ -0,0 +1,31 @@
...
+from werkzeug.datastructures import FileStorage
+from app.extensions import api
+from app.api.utils.resources_mixins import UserMixin
+from app.api.utils.custom_reqparser import CustomReqparser
+from app.api.projects.project.models.project import Project
+from app.api.mines.mine.models.mine import Mine
+from app.api.services.document_manager_service import DocumentManagerService
+
+
+class InformationRequirementsTableListImportResource(Resource, UserMixin):
```
BenHarrington: This looks to be the resource that relates to the document upload for the spreadsheet. We may want to rename the class to reflect that and also update the API doc description to be a bit clearer to the developers what this is doing.  Unless this is doing something else that is not obvious?

Additionally we may want to perform checks on both project and mine to make sure they exist before using the DocumentManagerService.

Lets also add "params" to the @api.doc object to be consistent with our other endpoints.

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -10,7 +10,7 @@
 ALLOWED_DOCUMENT_CATEGORIES = [
     'tailings', 'permits', 'variances', 'incidents', 'reports', 'mine_party_appts', 'noticeofwork',
     'bonds', 'reclamation_invoices', 'explosives_permits', 'project_summaries',
-    'notices_of_depature'
+    'notices_of_depature', 'irt'
```
BenHarrington: Do we want to use the full name here rather than the abbreviation? It would be consistent with the rest. Would that be much work?

services/core-api/app/config.py
```diff
@@ -157,6 +159,13 @@ def JWT_ROLE_CALLBACK(jwt_dict):
     EMAIL_ENABLED = os.environ.get('EMAIL_ENABLED', False)
     EMAIL_RECIPIENT_OVERRIDE = os.environ.get('EMAIL_RECIPIENT_OVERRIDE')
 
+    #Templates
+    TEMPLATE_FOLDER_BASE = os.environ.get('TEMPLATE_FOLDER_BASE', '/app/app/templates')
```
BenHarrington: Is "/app/app" a typo here? I notice the templates folder is "/app/templates".

services/core-api/app/config.py
```diff
@@ -157,6 +159,13 @@ def JWT_ROLE_CALLBACK(jwt_dict):
...
+
+    TEMPLATE_FOLDER_IRT = os.environ.get('TEMPLATE_FOLDER_IRT', f'{TEMPLATE_FOLDER_BASE}/projects/')
+    TEMPLATE_IRT = os.environ.get('TEMPLATE_IRT',
```
BenHarrington: Should the default here be "IRT_Template.xlsx" like I see the file named in the templates folder?

services/core-api/app/config.py
```diff
@@ -157,6 +159,13 @@ def JWT_ROLE_CALLBACK(jwt_dict):
     EMAIL_ENABLED = os.environ.get('EMAIL_ENABLED', False)
     EMAIL_RECIPIENT_OVERRIDE = os.environ.get('EMAIL_RECIPIENT_OVERRIDE')
 
+    #Templates
+    TEMPLATE_FOLDER_BASE = os.environ.get('TEMPLATE_FOLDER_BASE', '/app/app/templates')
+
+    TEMPLATE_FOLDER_IRT = os.environ.get('TEMPLATE_FOLDER_IRT', f'{TEMPLATE_FOLDER_BASE}/projects/')
```
BenHarrington, line 165: I notice we have both "project" and "projects" folders in the template folder. Can we clean one of those up that we aren't using?

services/minespace-web/src/components/dashboard/mine/projects/ProjectStagesTable.js
```diff
@@ -32,28 +32,51 @@ export class ProjectStagesTable extends Component {
...
-            record.stage?.payload?.project_guid,
-            record.stage?.payload?.project_summary_guid
-          )}
-        >
-          <Button className="full-mobile margin-small" type="secondary">
-            Edit
-          </Button>
-        </Link>
-      ),
+      render: (text, record) => {
```
BenHarrington, line 49: Could we add the links for each stage to the projectStages payload(begins line 77) in the ProjectOverviewTab component? Then we can just return the record link in this render function and be done with it.

I think this will clean things up, especially when we begin to add further stages..

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
```diff
@@ -0,0 +1,193 @@
...
+  prev = () => this.setState((prevState) => ({ current: prevState.current - 1 }));
+
+  handleFetchData = () => {
+    const { projectGuid, tab } = this.props.match?.params;
+    if (projectGuid) {
+      return this.props
+        .fetchProjectById(projectGuid)
+        .then(() => this.setState({ isLoaded: true, activeTab: tab }));
+    }
+    return this.props.fetchProjectById(projectGuid).then(() => {
```
BenHarrington: Can we remove this duplicate fetch call?

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
```diff
@@ -0,0 +1,193 @@
...
+      this.setState({ isLoaded: true, activeTab: tab });
+    });
+  };
+
+  render() {
+    const title = this.state.isEditMode
```
BenHarrington, line 123: This title is referencing the project summary title, I think we want it to reference the project title, right?

services/minespace-web/src/components/pages/Project/ProjectPage.js
BenHarrington: To confirm, this means available everywhere except Prod :)

BenHarrington (changes requested):
Great work on this Arturo. There are a few things to clean up but this will set us up well as we begin to enhance this feature further.

services/core-api/app/config.py
areyeslo: relative path is considered.

services/core-api/app/config.py
areyeslo: Having just projects folder.

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
areyeslo: Removed.

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
areyeslo: Good catch, changed.

BenHarrington (approved):
Great work on this Arturo. There are a few things to clean up but this will set us up well as we begin to enhance this feature further.

PR #2097, opened 2022-05-20
Mds 4420 nod update timestamp 🐛 

hitankar-freshworks:
## Objective 

[MDS-4420](https://bcmines.atlassian.net/browse/MDS-4420)

Fixed issue with update)timestamp not changing when a nod doc as attached to a nod. 

### DIFF
--- a/services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
+++ b/services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
@@ -95,8 +95,8 @@ class NoticeOfDeparture(SoftDeleteMixin, AuditMixin, Base):
         return query.all()
 
     def save(self, commit=True):
-        self.updated_by = User().get_user_username()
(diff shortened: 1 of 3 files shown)


services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
```diff
@@ -95,8 +95,8 @@ def find_all_by_permit_guid(cls, __guid, mine_guid=None):
         return query.all()
 
     def save(self, commit=True):
-        self.updated_by = User().get_user_username()
-        self.updated_timestamp = datetime.utcnow()
+        self.update_user = User().get_user_username()
```
matbusby-fw, line 98: 🤦‍♂️

matbusby-fw (dismissed):
How did we not see this?  😆   Good catch.

matbusby-fw (approved):

PR #2106, opened 2022-05-27
fixed submit disable while uploading

matbusby-fw:
## Objective 

[MDS-4442](https://bcmines.atlassian.net/browse/MDS-4442)

- Fixed the check for a self-assessment checklist that disables/enables the submit button


hitankar-freshworks:
@matbusby-fw can we discuss your solution here. I am trying to understand if all the scenarios are met.

hitankar-freshworks (approved):

PR #2109, opened 2022-05-30
Bugfix/mds 4450 4445 4451

matbusby-fw:
## Objective 

[MDS-4450](https://bcmines.atlassian.net/browse/MDS-4450)
[MDS-4445](https://bcmines.atlassian.net/browse/MDS-4445)
[MDS-4451](https://bcmines.atlassian.net/browse/MDS-4451)

- Disabled edit nod form submit on Minespace if the form is pristine
- Enabled nod form submit on Minespace after dismissing error file (this was causing the submit button to be unavailable until refresh).  There is a recent open issue regarding this on Stack Overflow.  The solution I used sets `uploading` to false when a file is removed.  This was the best option I could find based on the current capabilities, but it has the additional effect of setting `uploading` to false when _any_ file is removed.  It won't upload the errored file at any rate, so it shouldn't cause much of any issue, but there doesn't seem to be a better solution within `filepond` at this point in time.
- Added `popConfirm` to CoreWeb on cancelling the `ViewNoticeOfDepartureModal`

image
image

### DIFF
--- a/services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
+++ b/services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
@@ -31,7 +31,7 @@ const propTypes = {
 };
 
 // eslint-disable-next-line import/no-mutable-exports
-let AddNoticeOfDepartureForm = (props) => {
+let EditNoticeOfDepartureForm = (props) => {
   const { onSubmit, closeModal, handleSubmit, mineGuid, noticeOfDeparture, pristine } = props;
   const { permit, nod_guid } = noticeOfDeparture;
   const [submitting, setSubmitting] = useState(false);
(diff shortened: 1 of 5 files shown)


services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
```diff
@@ -25,13 +25,14 @@ const propTypes = {
   onSubmit: PropTypes.func.isRequired,
   closeModal: PropTypes.func.isRequired,
   handleSubmit: PropTypes.func.isRequired,
+  pristine: PropTypes.bool.isRequired,
   mineGuid: PropTypes.string.isRequired,
   noticeOfDeparture: CustomPropTypes.noticeOfDeparture.isRequired,
 };
 
 // eslint-disable-next-line import/no-mutable-exports
 let AddNoticeOfDepartureForm = (props) => {
```
hitankar-freshworks: Just noticed this should be been
```suggestion
let EditNoticeOfDepartureForm = (props) => {
```
right?

services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
```diff
@@ -25,13 +25,14 @@ const propTypes = {
...
-  const { onSubmit, closeModal, handleSubmit, mineGuid, noticeOfDeparture } = props;
+  const { onSubmit, closeModal, handleSubmit, mineGuid, noticeOfDeparture, pristine } = props;
```
hitankar-freshworks, line 35: How is `pristine` manupulated?

matbusby-fw: It's a bool provided directly from Redux-Form that reports weather any of the fields have been touched.  I haven't played with manipulating it yet though.

services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
matbusby-fw: :man_facepalming: Indeed.  All of the references on this page are for `AddNoticeOfDepartureForm`.  I'll update them all and test quickly.

matbusby-fw: Good catch!

services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
hitankar-freshworks: Ah, its an inbuilt functionality. thanks

services/minespace-web/src/components/Forms/noticeOfDeparture/EditNoticeOfDepartureForm.js
matbusby-fw: Okay, this has been updated now.

hitankar-freshworks (approved):

PR #2112, opened 2022-05-31
Feature/mds 4219 change insp during delay

JustinMacaulay:
## Objective 

[MDS-4219](https://bcmines.atlassian.net/browse/MDS-4219)

image

Changed the logic in the NOWActionWrapper so that you can choose to ignore the delay logic if you need to.

All other NoW Logic will still apply to this section.


BenHarrington (approved):
Looks good Justin!

PR #2119, opened 2022-06-02
Bugfix/Mds 4458 4459 nod update forms

matbusby-fw:
## Objective 

[MDS-4458](https://bcmines.atlassian.net/browse/MDS-4458)
[MDS-4459](https://bcmines.atlassian.net/browse/MDS-4459)

- Refactored ViewNoticeOfDepartureModal in Core to be a form to facilitate better validation
- Made checklist required on EditNoticeOfDepartureForm in Minespace

### DIFF
--- a/services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
+++ b/services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
@@ -1,6 +1,6 @@
-import React from "react";
+import React, { useEffect } from "react";
 import PropTypes from "prop-types";
-import { Button, Col, Popconfirm, Row, Select } from "antd";
+import { Button, Col, Form, Popconfirm, Row } from "antd";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
 import {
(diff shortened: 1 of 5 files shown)


services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
```diff
@@ -44,25 +54,30 @@ export const ViewNoticeOfDepartureModal = (props) => {
...
-  // eslint-disable-next-line no-return-assign
-  const handleChangeNodStatus = ({ value }) => (props.noticeOfDeparture.nod_status = value);
-
-  const updateNoticeOfDepartureStatus = async () => {
-    await props.updateNoticeOfDeparture(
-      { mineGuid: mine.mine_guid, nodGuid: nod_guid },
-      noticeOfDeparture
-    );
+  const updateNoticeOfDepartureSubmit = async (values) => {
+    await props.updateNoticeOfDeparture({ mineGuid: mine.mine_guid, nodGuid: nod_guid }, values);
```
hitankar-freshworks, line 80: nitpick. probably a good idea to fully destructure props for the sake of consistency.

hitankar-freshworks (approved):

services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
matbusby-fw: That's actually a quirk of redux.  It raises an error if you try to destructure the methods passed in through redux due to it seeing them as shadowed properties.  When I was first working on the forms I played with a few different options, and this felt like the least ugly of them. 🤷

matbusby-fw: I think that's there to prevent you from trying to call the redux methods directly from an import instead of wiring them up through the `connect` function in the export.

hitankar-freshworks: Okay yes. make sense

PR #2122, opened 2022-06-03
IRT Template - Fix default file path for template in API

BenHarrington:


areyeslo (approved):
Thanks for this fix Ben.

PR #2128, opened 2022-06-07
MDS-4353-nod-no-added

hitankar-freshworks:
## Objective 

NOTE: -- `notice_of_departure and notice_of_departure_document-xref table needs to be truncate before we can merge this ticket.`

[MDS-4353](https://bcmines.atlassian.net/browse/MDS-4353)

- Adds Nod No to the NOD

### DIFF
--- a/services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
+++ b/services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import lazyload
 from app.extensions import db
 from app.api.constants import *
 from app.api.utils.include.user_info import User
-from sqlalchemy import desc
+from sqlalchemy.sql import text
 
 
 class NodType(Enum):
@@ -29,6 +29,7 @@ class NoticeOfDeparture(SoftDeleteMixin, AuditMixin, Base):
     __tablename__ = 'notice_of_departure'
 
     nod_guid = db.Column(UUID(as_uuid=True), server_default=FetchedValue(), primary_key=True)
+    nod_no = db.Column(db.String(36), nullable=False)
     mine_guid = db.Column(UUID(as_uuid=True), db.ForeignKey('mine.mine_guid'), nullable=False)
     permit_guid = db.Column(UUID(as_uuid=True), db.ForeignKey('permit.permit_guid'), nullable=False)
     nod_title = db.Column(db.String(50), nullable=False)
@@ -69,7 +70,10 @@ class NoticeOfDeparture(SoftDeleteMixin, AuditMixin, Base):
             nod_title=nod_title,
             nod_description=nod_description,
             nod_type=nod_type,
-            nod_status=nod_status)
+            nod_status=nod_status,
+            nod_no=text(
+                f'\'NOD-\' || (SELECT permit_no FROM permit WHERE permit_guid = \'{permit.permit_guid}\') || \'-\' || (SELECT TO_CHAR((SELECT COUNT(*) FROM notice_of_departure nod WHERE permit_guid = \'{permit.permit_guid}\') + 1, \'fm00\'))'
+            ))
 
         if add_to_session:
             new_nod.save(commit=False)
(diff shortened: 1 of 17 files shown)


services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
```diff
@@ -69,7 +70,10 @@ def create(cls,
             nod_title=nod_title,
             nod_description=nod_description,
             nod_type=nod_type,
-            nod_status=nod_status)
+            nod_status=nod_status,
+            nod_no=text(
+                f'\'NOD-\' || (SELECT permit_no FROM permit WHERE permit_guid = \'{permit.permit_guid}\') || \'-\' || (SELECT TO_CHAR((SELECT COUNT(*) FROM notice_of_departure nod WHERE permit_guid = \'{permit.permit_guid}\') + 1, \'fm00\'))'
```
matbusby-fw, line 75: Nice!

services/core-web/src/components/mine/NoticeOfDeparture/MineNoticeOfDeparture.js
```diff
@@ -49,7 +49,7 @@ export const MineNoticeOfDeparture = (props) => {
     event.preventDefault();
     const detailedNoticeOfDeparture = await props.fetchDetailedNoticeOfDeparture(
       mineGuid,
-      selectedNoticeOfDeparture.nod_id
+      selectedNoticeOfDeparture.nod_guid
```
matbusby-fw, line 52: Good catch!

services/core-web/src/components/mine/NoticeOfDeparture/MineNoticeOfDepartureTable.js
```diff
@@ -43,13 +43,15 @@ export class MineNoticeOfDepartureTable extends Component {
...
         update_timestamp,
         nod_guid,
+        nod_no,
         nod_type,
         nod_status,
         ...other
       }) => ({
         ...other,
         key: nod_guid,
-        nod_id: nod_guid,
```
matbusby-fw, line 52: Ah, I see where that nod_id was coming from

services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
```diff
@@ -292,8 +292,8 @@ let ViewNoticeOfDepartureModal = (props) => {
         </Form.Item>
         <Row justify="space-between" className="padding-md--top" gutter={24}>
           <Col span={12}>
-            <p className="field-title">Technical Operations Director</p>
-            <p className="content--light-grey padding-md">{EMPTY_FIELD}</p>
+            <p className="field-title">Updated Date</p>
```
matbusby-fw, line 295: Was there a design change here?  I don't see any change on Figma.

matbusby-fw (dismissed):
LGTM.  Just one question about the changes to the core modal.

services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
hitankar-freshworks: No, but there is no actual use of those dummy fields so decided to remove them for now.

services/core-api/app/api/mines/notice_of_departure/models/notice_of_departure.py
areyeslo: Sorry, just checking this. It seems a SQL injection can be used and become dangerous. We rely always on ORM, the use of SQL Alchemy would be very advisable to avoid this security vulnerability.

hitankar-freshworks: This is not directly getting any values from the frontend. Instead, we are generating a value from within the database. the passed value is coming from the application itself. My thought the risk of sql injection is minimal but I am open to seeing how we can use sqlalchemy to improve this.

areyeslo: The real risk of SQL injection is that the hacker can replace your SELECT for something else such as DELETE or UPDATE. This is the real danger here.

hitankar-freshworks: After chatting with @areyeslo we realized that the chance of SQL injection is low as we are dealing with generating queries using validated data. However, we realize it can be uncomfortable to have raw queries. I will do a new PR to use SQL alchemy helper functions.

matbusby-fw (approved):

PR #2129, opened 2022-06-07
[MDS-4200] Submit-final-IRT-Minespace

areyeslo:
## Objective 

https://bcmines.atlassian.net/browse/MDS-4200

Ability to submit a final IRT.

image

Currently, we have the following IRT states:
REC- Received
UNR - Under review
APV - Approved

When importing Information Requirements Table (IRT) requirements, the IRT is created with REC status. When the user clicks on submit IRT, then the state is changed to UNR (Under review). 
Following ticket would add functionality to send an email to Reviewer to notify there is a new IRT for review.


BenHarrington (commented):
Not related to the work done in this ticket, but most likely the previous iteration: I notice the first time you enter the IRT workflow and try to advance from step 2 -> 3 it reloads the page with a different route and goes back to step 1. After experiencing this you are able to navigate to step 3 with no problem. Would be nice if we could get that sorted out. 

I think it may have to do with step 1/2 sharing the same frontend route /new and possibly is using the wrong link or something to that effect.

BenHarrington:
Also, noted one other thing. We should not  be able to navigate to Step 3(via nav buttons or clicking the link) before we have submitted an IRT, looks like it breaks the frontend.

services/minespace-web/src/components/common/FileUpload.js
```diff
@@ -77,13 +79,19 @@ class FileUpload extends React.Component {
...
-          onSuccess: () => {
+          onSuccess: async () => {
             const documentGuid = upload.url.split("/").pop();
             load(documentGuid);
             this.props.onFileLoad(file.name, documentGuid);
             // Call an additional action on file blob after success(only one use case so far, may need to be extended/structured better in the future)
             if (this.props?.afterSuccess?.action) {
-              this.props.afterSuccess.action(this.props.afterSuccess?.actionGuid, file);
+              try {
+                var resp = await this.props.afterSuccess.action(
```
BenHarrington: The result of this call doesn't need to be assigned to a variable since we are not referencing "resp" anywhere.

JustinMacaulay: agreed

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
```diff
@@ -126,31 +152,34 @@ const StepForms = (props, state, next, prev, close, handleTabChange, handleIRTSu
...
+            >
+              Submit IRT
+            </Button>
+          </Link>
+        ) : (
+          <Button
+            type="primary"
+            style={{ display: "inline", float: "right" }}
+            htmlType="submit"
+            disabled={state.submitting}
```
JustinMacaulay: does this button also need the handleIRTUpdate() on click?

JustinMacaulay (commented):
just the 2 comments

services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
areyeslo: To be removed. Need to work on Update in future work.

JustinMacaulay (dismissed):
Looks good.

services/core-api/tests/factories.py
```diff
@@ -1239,7 +1239,7 @@ class Params:
         permit = factory.SubFactory('tests.factories.PermitFactory')
 
     nod_guid = GUID
-    nod_no = factory.LazyAttribute(lambda a: f'NOD-X-45564456-{randrange(1,10)}')
+    nod_no = factory.LazyAttribute(lambda a: f'NOD-X-45564456-{randrange(1,1000)}')
```
areyeslo, line 1242: Thank you

JustinMacaulay (approved):

PR #2133, opened 2022-06-09
MDS 3776 now work during delay

JustinMacaulay:
## Objective 

[MDS-3776](https://bcmines.atlassian.net/browse/MDS-3776)

This allows for the addition of documents, securities and drafting of the permit while the NoW is client delayed.

image


BenHarrington (approved):
Looks good Justin!

PR #2134, opened 2022-06-09
fix nod edit modal, replace sql query with sqlachemy builder functions

hitankar-freshworks:
## Objective 

[MDS-4474](https://bcmines.atlassian.net/browse/MDS-4474)

- Fixes nod edit modal
- Replace SQL query with SQLAlchemy builder functions


matbusby-fw (approved):
Nice enhancement!

PR #2137, opened 2022-06-13
bugfix/mds-4475 updated core nod view file validation

matbusby-fw:
## Objective 

[MDS-4475](https://bcmines.atlassian.net/browse/MDS-4475)

- Added an `uploading` boolean to the Core Nod view modal to flag disabling the `update` button.
- Updated `FileUpload` component to accept functions to manage the uploading boolean.

### DIFF
--- a/services/core-web/src/components/common/FileUpload.js
+++ b/services/core-web/src/components/common/FileUpload.js
@@ -27,6 +27,8 @@ const propTypes = {
   chunkSize: PropTypes.number,
   allowRevert: PropTypes.bool,
   allowMultiple: PropTypes.bool,
+  onprocessfiles: PropTypes.func,
+  onAbort: PropTypes.func,
 };
 
 const defaultProps = {
@@ -38,6 +40,8 @@ const defaultProps = {
   chunkSize: 1048576, // 1MB
   allowRevert: false,
   allowMultiple: true,
+  onprocessfiles: () => {},
+  onAbort: () => {},
 };
 
 class FileUpload extends React.Component {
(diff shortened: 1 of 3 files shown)


services/core-web/src/components/common/FileUpload.js
```diff
@@ -27,6 +27,8 @@ const propTypes = {
   chunkSize: PropTypes.number,
   allowRevert: PropTypes.bool,
   allowMultiple: PropTypes.bool,
+  onprocessfiles: PropTypes.func,
```
hitankar-freshworks: for consistency we can us camel case here

services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
```diff
@@ -275,6 +277,9 @@ let ViewNoticeOfDepartureModal = (props) => {
               id="fileUpload"
               name="fileUpload"
               component={FileUpload}
+              addFileStart={() => setUploading(true)}
+              onAbort={() => setUploading(false)}
+              onprocessfiles={() => setUploading(false)}
```
hitankar-freshworks: Same as my previous comment

hitankar-freshworks (dismissed):
nitpicks only, otherwise LGTM

services/core-web/src/components/common/FileUpload.js
matbusby-fw: This is coming from filepond, internally it uses this case.  To keep from having a confusing case mismatch I mirrored the naming on ours.

hitankar-freshworks (approved):

PR #2143, opened 2022-06-14
[MDS-3744] - Abandoned mines banner

matbusby-fw:
## Objective 

[MDS-3744](https://bcmines.atlassian.net/browse/MDS-3744)

- Updated logic for displaying the banner only if either of the new roles is present (rather than if a mine is flagged as abandoned)
- Made banner text dynamic based on combination of those new roles added.
- Small update to the default contact styling to keep the `See History` button from being cut off.

### DIFF
--- a/services/core-web/src/components/mine/MineDashboard.js
+++ b/services/core-web/src/components/mine/MineDashboard.js
@@ -2,6 +2,7 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
 import { bindActionCreators } from "redux";
+import moment from "moment";
 import { Menu, Button, Dropdown, Popconfirm, Tooltip, Drawer } from "antd";
 import {
   DownOutlined,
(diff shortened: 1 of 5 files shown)


services/core-web/src/components/mine/MineDashboard.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
 import { bindActionCreators } from "redux";
+import moment from "moment";
```
hitankar-freshworks: moment is deprecated. suggest using 'date-fns'

hitankar-freshworks (commented):
1 important comment. didn't approve because we need to discuss.

services/core-web/src/components/mine/MineDashboard.js
matbusby-fw: Sounds good to me.  I was unsure about introducing a new date library at this point in the project, but I much prefer date-fns.

hitankar-freshworks: just moment is deprecated and the new code should use actively maintained libraries. Plus moment.js is not immutable.

matbusby-fw: 💯 Updating it now

hitankar-freshworks (approved):

PR #2146, opened 2022-06-17
[MDS-4393]: Core - NoW client delay date dates being overridden

BenHarrington:
## Objective 

[MDS-4393](https://bcmines.atlassian.net/browse/MDS-4393)

## How to Test

- Complete a Technical Review and Referral on a NoW and have their end dates set in the past
- Complete a delay(end date in the past) on one step, and leave another delay open(do not complete)
- Navigate to process and set the status to Rejected/Withdrawn/No Permit needed
- Navigate back to Adminstrative->Application Progress Tracker
- Ensure all the application progress steps have not had their end dates changed to the decision date
- Ensure the completed delay has not had its end date changed
- Ensure the incomplete delay had its end date set to the decision date
- DEV NOTE: The logic for these end dates being set are if an Application Progress Step/Delay has no end date or it's end date is past the decision date then it will be updated to the decision date, otherwise they should remain untouched,


JustinMacaulay (approved):
Looks good.

PR #2161, opened 2022-06-24
[MDS-4182] - send email on non-substantial nod create

hitankar-freshworks:
## Objective 

[MDS-4182](https://bcmines.atlassian.net/browse/MDS-4182)

- Send an email if a non-substantial nod was created. 

- [x] Send an email if a non-substantial nod was created. 
~~- [ ] Add a config map or secret for MAJOR_MINES_OFFICE_EMAIL env var (discuss with @v-y-a-s)~~

### DIFF
--- a/services/core-api/app/__init__.py
+++ b/services/core-api/app/__init__.py
@@ -117,13 +117,15 @@ def register_routes(app):
 
         def get_health(self):
             service = {
-                'database': False,
-                'cache': False,
-                'nris': False,
-                'docgen': False,
-                'docman': False
+                'database': True,
+                'cache': True,
+                'nris': True,
+                'docgen': True,
+                'docman': True
             }
             status = 200
+            if (Config.ENVIRONMENT_NAME == 'local'):
+                return service, status
 
             try:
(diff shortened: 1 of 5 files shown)


hitankar-freshworks:
@v-y-a-s @vyasworks we need to add an env var. let's sync up how to add that to our infra.

services/core-api/app/__init__.py
```diff
@@ -117,13 +117,15 @@ class Healthcheck(Resource):
 
         def get_health(self):
             service = {
-                'database': False,
-                'cache': False,
-                'nris': False,
-                'docgen': False,
-                'docman': False
+                'database': True,
```
hitankar-freshworks, line 120: @v-y-a-s @vyasworks made some updates to the health check method. This skips `healthchecks` to fail when running the application locally

vyasworks: Looks good

matbusby-fw (dismissed):
LGTM!

matbusby-fw (approved):

PR #2162, opened 2022-06-27
[MDS-4490]: MineSpace IRT - Change the IRT template hyperlink from PDF to Excel

BenHarrington:
## Objective 

[MDS-4490](https://bcmines.atlassian.net/browse/MDS-4490)


JustinMacaulay (approved):
Looks good

PR #2163, opened 2022-06-27
[MDS-4484] NoW app status changed on delay edit.

JustinMacaulay:
## Objective 

[MDS-4484](https://bcmines.atlassian.net/browse/MDS-4484)

Removing the portion of the now delay put that sets the status of the NoW app to the previous status when edited.


BenHarrington (commented):
Do we want to add a check at the start of the PUT endpoint for an existing NowApplicationDelay? Just to be consistent with how we do everything else.

BenHarrington (approved):
Looks good Justin!

PR #2165, opened 2022-06-28
[MDS-4487] Minespace - IRT - Add ability to re-upload/overwrite an existing IRT

areyeslo:
## Objective 

https://bcmines.atlassian.net/browse/MDS-4487

Add ability to re-upload an existing IRT.


BenHarrington (commented):
Not a huge thing, but I noticed on Step 2 when we have an existing IRT uploaded I think we should update the upload header "Import final IRT file" to something like "Re-import final IRT file" or something like that so that the user knows this will override the existing one. Dealer's choice.

Maybe add some clarifying text when is this state as well to saying this import will replace the existing one, if that title isn't clear enough. This is optional.

BenHarrington (changes requested):
On Step 2, when an IRT has already been imported the "Next" navigation button is enabled but the Step link for "Review & Submit" is not.

image

services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table_list.py
```diff
@@ -137,17 +132,12 @@ def post(self, project_guid):
...
-                sanitized_irt_requirements
-            })
-            new_information_requirements_table.save()
-            return new_information_requirements_table, 201
+            sanitized_irt_requirements = InformationRequirementsTableListResource.build_irt_payload_from_excel(
+                import_file)
+
+            new_irt = InformationRequirementsTable.save_new_irt(project, sanitized_irt_requirements,
+                                                                import_file, document_guid)
+            return new_irt
```
BenHarrington: Could we add a 201 status code onto this return for consistency?

BenHarrington (changes requested):

BenHarrington (approved):
Great work Arturo!

PR #2170, opened 2022-07-05
[MDS-4520]: New gate location pop-up in a NoW

BenHarrington:
## Objective 

[MDS-4520](https://bcmines.atlassian.net/browse/MDS-4520)

- Add new fields to db schema, models, and response models
- The UI will not be implementing these fields yet


JustinMacaulay (approved):
All looks good to me.

PR #2171, opened 2022-07-06
[MDS-4547] better email validation and show contact for all NODs

hitankar-freshworks:
## Objective 

[MDS-4547](https://bcmines.atlassian.net/browse/MDS-4547)

- [x]  better email validation
- [x]  show contact for all NOD types

### DIFF
--- a/services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
+++ b/services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
@@ -55,12 +55,8 @@ const propTypes = {
 };
 
 export const renderContacts = (props) => {
-  const { fields, formValues } = props;
-  if (fields.length > 0 && formValues.nod_type === NOD_TYPE_FIELD_VALUE.NON_SUBSTANTIAL) {
-    fields.pop();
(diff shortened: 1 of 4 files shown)


services/core-web/src/components/modalContent/ViewNoticeOfDepartureModal.js
```diff
@@ -55,12 +55,8 @@ const propTypes = {
...
 
 export const renderContacts = (props) => {
-  const { fields, formValues } = props;
-  if (fields.length > 0 && formValues.nod_type === NOD_TYPE_FIELD_VALUE.NON_SUBSTANTIAL) {
-    fields.pop();
-  } else if (fields.length < 1 && formValues.nod_type !== NOD_TYPE_FIELD_VALUE.NON_SUBSTANTIAL) {
-    fields.push({ is_primary: true });
-  }
+  const { fields } = props;
+
```
matbusby-fw, line 58: This is the only chunk where we were using the `formValues` prop, so we can remove it from the rest of this page.
That would be:
`MapStateToProps`
`propTypes`
the destructured props
`FieldArray`

matbusby-fw (changes requested):
Looks good.  Just need to remove the `formValues` in a couple places since they're not being used anymore.

matbusby-fw (approved):

PR #2176, opened 2022-07-13
[MDS-3454] - Removed limit on seached mines for minespace user access

matbusby-fw:
## Objective 

- Removed limit on mines returned for mine users access
- Reduced extra calls to fetch the mine list
- Fixed a styling issue where the input in the multi select field was below where it should have been
- Cleaned up some errors in the themOverride.scss file

image

### DIFF
--- a/services/core-web/src/components/admin/UpdateMinespaceUser.js
+++ b/services/core-web/src/components/admin/UpdateMinespaceUser.js
@@ -1,15 +1,12 @@
-import React, { Component } from "react";
-import { bindActionCreators } from "redux";
+import React from "react";
 import { connect } from "react-redux";
 import PropTypes from "prop-types";
 import { getMineNames } from "@common/selectors/mineSelectors";
-import { fetchMineNameList } from "@common/actionCreators/mineActionCreator";
 
 import CustomPropTypes from "@/customPropTypes";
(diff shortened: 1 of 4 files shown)


services/core-web/src/components/admin/UpdateMinespaceUser.js
```diff
@@ -21,59 +18,34 @@ const defaultProps = {
...
-            onSubmit={this.props.handleSubmit}
-            handleChange={this.handleChange}
-            handleSearch={this.handleSearch}
-          />
-        )}
-      </div>
-    );
-  }
-}
+export const UpdateMinespaceUser = (props) => {
```
hitankar-freshworks, line 21: 💯

hitankar-freshworks (approved):
Looks good. One question. How are we fetching the mines data, i.e. hitting `/mines/search` API

matbusby-fw:
> Looks good. One question. How are we fetching the mines data, i.e. hitting `/mines/search` API

It's called as soon as you navigate to the `MinespaceUserManagement` component which is parent to these edit components.

```  componentDidMount() {
    this.props.fetchMineNameList().then(() => {
      this.setState({ isLoaded: true });
    });
    this.refreshUserData();
  }```

PR #2181, opened 2022-07-15
[MDS-4552] Minespace - IRT - Importing & Import successful and Import failed message (missing field)

BenHarrington:
## Objective 

[MDS-4552](https://bcmines.atlassian.net/browse/MDS-4552)

_Key changes_
- When an IRT import is successful a success modal will pop up
- When an IRT import has failed with parsing errors an error modal will pop up with a list of sections and rows that produced the errors(two types of import errors(Methods checked, no comments & Information cell invalid)) - please test various combinations
- When an IRT import has failed with a unhandled server error an error modal will popup with that error

_Other changes_
- Updated CustomAxios to allow for suppressing of error notifications when passing a flag through an actionCreator


areyeslo (approved):
Amazing job!! 🚀

PR #2183, opened 2022-07-15
[MDS-3454] - Added existing mine names to core user access edit modal

matbusby-fw:
## Objective 

- Added the options for the user's existing selection in the mine user access management edit modal
- Fixed some stylesheet bugs
- Fixed styling on multi-select input
- reinstated the limit on the api call for mine names.

### DIFF
--- a/services/core-api/app/api/mines/mine/models/mine.py
+++ b/services/core-api/app/api/mines/mine/models/mine.py
@@ -237,7 +237,7 @@ class Mine(SoftDeleteMixin, AuditMixin, Base):
                            column('mine_location_description'), column('mine_name'), column('mine_no'),
                            column('deleted_ind'), column('major_mine_ind'))
 
-        mines_q = select([mine_table]).where(mine_table.c.deleted_ind == False)
+        mines_q = select([mine_table]).where(mine_table.c.deleted_ind == False).limit(50)
 
         if term:
             mines_q = mines_q.where(mine_table.c.mine_name.ilike('%{}%'.format(term)))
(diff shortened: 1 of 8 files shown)


services/core-api/app/api/mines/mine/models/mine.py
```diff
@@ -237,7 +237,7 @@ def find_by_mine_name(cls, term=None, major=None):
                            column('mine_location_description'), column('mine_name'), column('mine_no'),
                            column('deleted_ind'), column('major_mine_ind'))
 
-        mines_q = select([mine_table]).where(mine_table.c.deleted_ind == False)
+        mines_q = select([mine_table]).where(mine_table.c.deleted_ind == False).limit(50)
```
hitankar-freshworks: I think the old code has the additional requirement to skip the check if minespace_user. We can add a check here, eg.

```
if not (is_minespace_user()):
            mines_q = mines_q.limit(100)
```

hitankar-freshworks (changes requested):
LGTM. one important change needed.

services/core-api/app/api/mines/mine/models/mine.py
matbusby-fw: Good catch!  Done!

services/core-web/src/components/admin/UpdateMinespaceUser.js
```diff
@@ -36,16 +35,24 @@ export class UpdateMinespaceUser extends Component {
...
         <h3>Edit Proponent</h3>
         {this.props.mines && (
           <EditMinespaceUser
-            mines={this.props.mines.map((mine) => ({
-              value: mine.mine_guid,
-              label: `${mine.mine_name} - ${mine.mine_no}`,
-            }))}
+            mines={this.parseMinesAsOptions([
+              ...this.props.mines,
+              ...this.props.minespaceUserMines,
```
hitankar-freshworks: I am wondering if this is going to lead to a performance issue..

hitankar-freshworks (approved):

PR #2184, opened 2022-07-19
[MDS-4467] update frontend builds

vyasworks:
## Objective 

[MDS-4467](https://bcmines.atlassian.net/browse/MDS-4467)

Should reduce build size from ~750 to ~150 mb


services/core-web/.eslintrc.js
```diff
@@ -15,7 +15,6 @@ module.exports = {
     "airbnb",
     "prettier",
     "prettier/react",
-    "plugin:cypress/recommended",
```
vyasworks, line 18: ES Lint was causing some issues here, I think we don't need this now since we are not running cypress yet.

sidsayani (approved):

PR #2195, opened 2022-07-22
[MDS-4551]Update in styling and fixing issue in requirements 

areyeslo:
## Objective 

[MDS-4551](https://bcmines.atlassian.net/browse/MDS-4551)

Changes to keep interface the same as mockups:

-Render Back button in step 2 to take the user to step 1
image

- Remove text "Review imported data before submission. Check the requirements and comments fields that are required for the project." included in callout to be just text in step 3. Remove from other steps:
image

- Update IRT table to match mockups.
image

-Fix children requirements being added to wrong parent:
image


areyeslo:
Closing PR

PR #2197, opened 2022-07-22
[MDS-4551] Fix issue in requirements and update sytling

areyeslo:
## Objective 

- Render always "Back" button in step 2 to take the user to step 1
image

- Remove text "Review imported data before submission. Check the requirements and comments fields that are required for the project." included in callout to be just text in step 3. Remove from other steps:
image

- Fix issue in requirement 4.10 having the children of 4.11 
image

- Additional changes in styling for IRT table.


BenHarrington (approved):
Great work polishing things up!

PR #2199, opened 2022-07-22
[MDS-3454] - bugfix

matbusby-fw:
## Objective 

- Updated the mine basic info endpoint to employ a simpler query if the parameter `simple` is passed from the front end. 

### DIFF
--- a/services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
+++ b/services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
@@ -1,10 +1,7 @@
-import decimal, uuid
+from flask_restplus import Resource, reqparse
+from sqlalchemy import table, column, select
 
-from flask import request
-from flask_restplus import Resource, reqparse, inputs
-from sqlalchemy_filters import apply_sort, apply_pagination
-
-from app.extensions import jwt, api
+from app.extensions import jwt, api, db
 from app.api.mines.response_models import MINES_MODEL
 from app.api.mines.mine.models.mine import Mine
 from app.api.utils.resources_mixins import UserMixin
@@ -16,6 +13,10 @@ class MineBasicInfoResource(Resource, UserMixin):
                         type=list,
                         location="json",
                         help='List of guids for mines to get basic info for.')
+    parser.add_argument('simple',
+                        type=bool,
+                        location="json",
+                        help='Boolean indicator to determine if simple of full info is returned')
 
     @api.expect(parser)
     @api.marshal_with(MINES_MODEL, code=200)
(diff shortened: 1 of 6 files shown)


services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
```diff
@@ -25,6 +26,23 @@ def post(self):
...
+
+            mines_q = select([mine_table]).where(mine_table.c.deleted_ind == False)
+
+            if mines:
+                mines_q = mines_q.where(mine_table.c.mine_guid.in_(mines))
+
+            connection = db.engine.connect()
+            results = connection.execute(mines_q).fetchall()
+
+            return results
```
hitankar-freshworks, line 47: One thing I noticed here is if mines = [] it will return all mine data instead of empty. Is that true?

services/minespace-web/common/actionCreators/minespaceActionCreator.js
```diff
@@ -60,13 +60,20 @@ export const fetchMinespaceUserMines = (mine_guids) => (dispatch) => {
   dispatch(showLoading());
   dispatch(request(reducerTypes.GET_MINESPACE_USER_MINES));
   return CustomAxios()
-    .post(ENVIRONMENT.apiUrl + API.MINE_BASIC_INFO_LIST, { mine_guids }, createRequestHeader())
+    .post(
+      ENVIRONMENT.apiUrl + API.MINE_BASIC_INFO_LIST,
+      { mine_guids, simple: true },
```
hitankar-freshworks, line 65: shouldn't this be parameterized?

hitankar-freshworks: or perhaps be removed?

hitankar-freshworks (commented):
left some questions

services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
matbusby-fw: I don't think so since `mines_q = mines_q.where(mine_table.c.mine_guid.in_(mines))` should dictate finding only the mines who's guids are within the `mines` array.

services/minespace-web/common/actionCreators/minespaceActionCreator.js
matbusby-fw: I don't really see why.  We were already passing the mine_guids as part of the payload in a post, so adding the boolean here makes sense to me.  If we removed it, it wouldn't send the simple flag at all.

matbusby-fw: FYI This isn't actually being used in minespace anywhere.  It's just duplicated for the sake of the common folders.

hitankar-freshworks: I thought minespace needs everything related to a mine...

matbusby-fw: No that was a different function call (within Core)

services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
```diff
@@ -25,6 +26,23 @@ def post(self):
 
         data = self.parser.parse_args()
         mines = data.get('mine_guids', [])
+        simple = data.get('simple', False)
+
```
hitankar-freshworks, line 30:
```suggestion
return [] if not mines
```

services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
```diff
@@ -25,6 +26,23 @@ def post(self):
 
         data = self.parser.parse_args()
         mines = data.get('mine_guids', [])
+        simple = data.get('simple', False)
+return [] if not mines
```
hitankar-freshworks:
```suggestion
       return [] if not mines
```

services/core-api/app/api/mines/mine/resources/mine_basicinfo.py
```diff
@@ -25,6 +26,23 @@ def post(self):
 
         data = self.parser.parse_args()
         mines = data.get('mine_guids', [])
+        simple = data.get('simple', False)
+       return [] if not mines
```
hitankar-freshworks, line 30:
```suggestion
        return [] if not mines
```

hitankar-freshworks (dismissed):

hitankar-freshworks (approved):

PR #2200, opened 2022-07-25
[MDS-4582] Fixes in UX for IRT when status is Approved

areyeslo:
## Objective 

[MDS-4582](https://bcmines.atlassian.net/browse/MDS-4582)

When IRT is approved there is no stepper neither Resubmit IRT button:
image


BenHarrington (approved):
Great work on the quick fixes!

PR #2202, opened 2022-07-27
[MDS-4581] Project Description Application 

areyeslo:
## Objective 

[MDS-4581](https://bcmines.atlassian.net/browse/MDS-4581)

- Save draft with default value False for MRC Review Required 
- Remove validation for back button. User is always able to go back. However, Next button is still validating the user has all information required to change to next section.


JustinMacaulay (dismissed):

BenHarrington (approved):
Thanks for keeping on top of these little changes, good work!

PR #2205, opened 2022-07-28
[MDS-4569] Major Mine Application Confirmation Page - MineSpace

BenHarrington:
## Objective 

[MDS-4569](https://bcmines.atlassian.net/browse/MDS-4569)

_Changes made_
- Created a new success page which is the user is redirected to after successful submission of a Major Mine Application: 

image

### DIFF
--- a/services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
+++ b/services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
@@ -6,10 +6,7 @@ import { change, submit, getFormSyncErrors, reset, touch } from "redux-form";
 import { Row, Col, Typography } from "antd";
 import { ArrowLeftOutlined } from "@ant-design/icons";
 import PropTypes from "prop-types";
-import customPropTypes from "@/customPropTypes";
 import { flattenObject, cleanFilePondFile } from "@common/utils/helpers";
-import * as FORM from "@/constants/forms";
-import { EDIT_PROJECT } from "@/constants/routes";
 import { getProject } from "@common/reducers/projectReducer";
 import { getMines } from "@common/selectors/mineSelectors";
 import {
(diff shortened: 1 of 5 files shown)


services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
```diff
@@ -72,13 +73,23 @@ export class MajorMineApplicationPage extends Component {
...
 
-  handleSaveData = (e, values, message) => {
+  handleSaveData = async (e, values, message) => {
     e.preventDefault();
     this.props.submit(FORM.ADD_MINE_MAJOR_APPLICATION);
     this.props.touch(FORM.ADD_MINE_MAJOR_APPLICATION);
     const errors = Object.keys(flattenObject(this.props.formErrors));
     if (errors.length === 0) {
-      return this.handleCreateMajorMineApplication(values, message);
+      await this.handleCreateMajorMineApplication(values, message);
```
JustinMacaulay, line 82: does this not work in the lifecycle hooks? or .then()?

JustinMacaulay (commented):
Just one comment since we don't use the async/await pattern very often

services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
BenHarrington: .Then() supports it because it is promise based., and it also works in lifecycle methods but may have undesired side effects due to the blocking nature(used sparingly).

I had to use it here because I needed the majorMineApplicationGuid which would only be available after creating a MMA and re-hydrating the store with the fetch.

JustinMacaulay: .then() should handle it then but I am not against this either. We have used .then to handle this type of situation in the past but this is the same effect just different implementation.

JustinMacaulay (approved):

services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
areyeslo: https://nodejs.dev/learn/modern-asynchronous-javascript-with-async-and-await#the-code-is-much-simpler-to-read

PR #2208, opened 2022-07-29
[MDS-4460] Final Status for DO only with final report

JustinMacaulay:
## Update the DO form to prevent setting the status to final unless a final report has been uploaded.

[MDS-4460](https://bcmines.atlassian.net/browse/MDS-4460)

Modifies the logic to prevent saving a DO as final if no final report has been uploaded.

image

image


areyeslo (approved):
Looks good Justin!! Thanks for going with me through the changes.

PR #2211, opened 2022-08-03
[MDS-4589] User can resubmit an IRT when in "In review" status

areyeslo:
## Objective 

[MDS-4589](https://bcmines.atlassian.net/browse/MDS-4589)

IRT cannot be resubmitted when it is in status "In Review". Resubmit IRT disappears:
image

### DIFF
--- a/services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
+++ b/services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
@@ -220,25 +220,26 @@ const StepForms = (
           </>
         ) : (
           <>
-            {props.project.information_requirements_table?.status_code !== "APV" && (
-              <Button
-                type="secondary"
-                htmlType="submit"
-                style={{ marginRight: "24px" }}
-                onClick={() => {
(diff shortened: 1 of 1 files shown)


services/minespace-web/src/components/pages/Project/InformationRequirementsTablePage.js
```diff
@@ -220,25 +220,26 @@ const StepForms = (
...
-                    )}`,
-                    state: { current: 1 },
-                  });
-                }}
-                disabled={state.submitting}
-              >
-                Resubmit IRT
-              </Button>
-            )}
+            {props.project.information_requirements_table?.status_code !== "APV" &&
```
BenHarrington, line 223: This could be minimized to: !["APV", "UNR"].includes(props.project.information_requirements_table?.status_code), but looks good otherwise.

JustinMacaulay: agreed, happy either way though.

BenHarrington (approved):
Looks good

PR #2214, opened 2022-08-04
[REFACTOR] MDS-4344 - refactor build script for nris to support gitops, remove references from deploy scripts.

wilson496:
## Objective 

[MDS-4344](https://bcmines.atlassian.net/browse/MDS-4344)

Build and deploy refactor for nris. Small tweaks for other services (removing references from deploy scripts)


gitops/commit.sh
```diff
@@ -45,7 +45,7 @@ else
     MDS_GIT_HASH=$(cat $TARGET_APP/overlays/$SOURCE_ENV/*deployment.patch.yaml | grep git-commit -m1 | cut -f2 -d ":" | xargs | cut -d "-" -f 3)
     echo
     # Replace the commit sha in env vars in overlay patch
-    sed -i "s^git-commit.*^git-commit-$MDS_GIT_HASH^" $TARGET_APP/overlays/$TARGET_ENV/deployment.patch.yaml
+    sed -i "s^git-commit.*^git-commit-$MDS_GIT_HASH^" $TARGET_APP/overlays/$TARGET_ENV/*deployment.patch.yaml
```
vyasworks, line 48: Do we need the * wildcard. Does it work as intended?

vyasworks:
Looks good, just one clarification.

gitops/commit.sh
wilson496: We will if we want to include different patches for multi resource configurations (e.g. docman, postgres).

vyasworks (approved):

PR #2215, opened 2022-08-04
[MDS-4084] Permit Clean up

JustinMacaulay:
## Objective 
Removed the NoW application, Permit amendment, and all other references from permit G-10-63 on mine 1000798 Lang's Pit. This was an error from MMS and was just data clean up. This is the script that was used and then just adding in the command to run it as a history of when it was used and on what.

[MDS-4084](https://bcmines.atlassian.net/browse/MDS-4084)


areyeslo (approved):
Fantastic idea saving the record of deletions.

PR #2216, opened 2022-08-04
[MDS-4602] "None" displayed in the comment section of IRT review page

areyeslo:
## Objective 

[MDS-4602](https://bcmines.atlassian.net/browse/MDS-4602)

Comments cells are saved as string "None" when there is no comment. The comment should be NULL in database and not being shown:
image


BenHarrington (approved):
Great catch!

PR #2223, opened 2022-08-12
[MDS-4628] Default to showing max 20 notifications in activity feed

simensma-fresh:
## Objective 

[MDS-4528](https://bcmines.atlassian.net/browse/MDS-4528)

Display max 20 notifications in activity feed instead of 25.


matbusby-fw (approved):
Magical!

simensma-fresh:
:magic_wand:

PR #2225, opened 2022-08-12
[MDS-4529] View a major mine application submission, in minespace

BenHarrington:
## Objective 

[MDS-4529](https://bcmines.atlassian.net/browse/MDS-4529)

_What Changed_

- See ticket for for specifics in AC

_How to Test_

- When a MMA has a status of ["APV", "REC", "UNR"]:
  - Clicking the View button on the project stages table navigates to Step 3 - read only (no nav, controls)
  - Clicking the Tab navigates to Step 3(tabbed view) which shows content from the previous point
- When a MMA has a status of ["CHR", "DFT"]:
  - Clicking the Tab navigates to entry page with appropriate instructions and links(CHR has it's own entry graphic and message)
  - Clicking the Resume button navigates to Step 2
- When a MMA has not been created yet both Tab and Button navigate to entry screen in tabbed view
- Added a dynamic callout for Steps 2/3 based on status

### DIFF
--- a/services/minespace-web/src/components/Forms/projects/majorMineApplication/MajorMineApplicationReviewSubmit.js
+++ b/services/minespace-web/src/components/Forms/projects/majorMineApplication/MajorMineApplicationReviewSubmit.js
@@ -1,7 +1,7 @@
 import React from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
-import { CALLOUT_SEVERITY } from "@common/constants/strings";
+import { withRouter } from "react-router-dom";
 import { Checkbox, Row, Col, Typography, Descriptions, Card, Input } from "antd";
 import PropTypes from "prop-types";
 import * as Strings from "@common/constants/strings";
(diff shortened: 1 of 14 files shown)


services/minespace-web/src/components/Forms/projects/majorMineApplication/MajorMineApplicationReviewSubmit.js
```diff
@@ -59,25 +82,34 @@ export const MajorMineApplicationReviewSubmit = ({
   );
   const documentColumns = [uploadDateColumn("upload_date")];
 
+  const columnStyleConfig = tabbedView ? { style: { maxWidth: "67%", margin: "0 auto" } } : {};
```
JustinMacaulay, line 85: not a comment for a change just pointing put that if there are a lot of style tags in a component then we probably want to move them to there own SCSS and interact with them as classes.

JustinMacaulay (approved):
Just the one comment and it is just a comment so happy to approve.

services/minespace-web/src/components/Forms/projects/majorMineApplication/MajorMineApplicationReviewSubmit.js
BenHarrington: I agree, it is a bit of fine line sometimes trying to determine when it is appropriate to break things out into classes. In this case if this style was used in another component(s) as well I definitely would have. Good observation, as we don't want to get bogged down with too many repetitive style props.

PR #2227, opened 2022-08-15
[MDS-4594] Add ability to add/edit referrals and consultations while a NoW is on client delay

BenHarrington:
## Objective 

[MDS-4594](https://bcmines.atlassian.net/browse/MDS-4594)


JustinMacaulay (approved):
Looks good.

PR #2228, opened 2022-08-15
[MDS-4519] navigate from notification CORE

matbusby-fw:
## Objective 

[MDS-4519](https://bcmines.atlassian.net/browse/MDS-4519)

- Added a query param to the MineNoticeOfDeparture page to open the selected nod when present
- Made activities clickable to navigate to the MineNoticeOfDeparture page with the appropriate queryParam
- Did an initial pass of the markAsRead functionality.
- This will be refined in an upcoming ticket as per this mornings discussion with Rebecca and Tanya

For the markAsRead functionality on the backend, feel free to look at it for any optimizations.  I was unsure about fetching the activities on the backend.  Currently I fetch each one and mark it as read, but I'm sure there's a better way of doing them as a batch.

### DIFF
--- a/services/core-api/app/api/activity/resource/activities_mark_as_read.py
+++ b/services/core-api/app/api/activity/resource/activities_mark_as_read.py
@@ -1,0 +1,22 @@
+from flask_restplus import Resource, reqparse
+
+from app.api.activity.models.activity_notification import ActivityNotification
+from app.api.utils.access_decorators import EDIT_DO, requires_any_of, MINESPACE_PROPONENT
+from app.api.utils.resources_mixins import UserMixin
+
+
+class ActivityMarkAsReadResource(Resource, UserMixin):
+    @requires_any_of([EDIT_DO, MINESPACE_PROPONENT])
+    def patch(self):
+        parser = reqparse.RequestParser()
+        parser.add_argument(
+            'activity_guids',
+            type=list,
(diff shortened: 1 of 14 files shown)


services/core-api/app/api/activity/resource/activities_mark_as_read.py
```diff
@@ -0,0 +1,22 @@
...
+            'activity_guids',
+            type=list,
+            location='json',
+            required=True,
+            help='List of activity guids to mark as read')
+        args = parser.parse_args()
+        activity_guids = args.get('activity_guids')
+        for activity_guid in activity_guids:
+            activity = ActivityNotification.find_by_guid(activity_guid)
+            ActivityNotification.update(activity)
```
simensma-fresh: We can probably add in an `mark_as_read_many` method on `ActivityNotification` with a query that would be something like this instead of looking up each individually

```python
ActivityNotification.filter(ActivityNotification.notification_guid.in_(notification_guids))
    .update({'notification_read': True})
```

simensma-fresh (changes requested):
Just one comment on the notification read update, otherwise looks great! 

One other thing we'll need to tackle is that you can only see / mark your own notifications as read, but happy to take a look at that part as part of my PR

simensma-fresh (approved):

PR #2229, opened 2022-08-15
[MDS-4629] made notification bell only show when logged in.

matbusby-fw:
## Objective 

[MDS-4629](https://bcmines.atlassian.net/browse/MDS-4629)

Passed in `isAuthorized` as prop from App.js and used it to flag the display of the notification bell.
Cleaned up missing props validation in App.js 🏕️ 🔥 

Note:  Importing `isAuthorized` from the state wasn't working as expected and updating dynamically.  Passing it from App.js works correctly.

image
image

### DIFF
--- a/services/minespace-web/src/App.js
+++ b/services/minespace-web/src/App.js
@@ -1,6 +1,5 @@
 import React, { Fragment, Component } from "react";
-import { compose } from "redux";
-import { bindActionCreators } from "redux";
+import { compose, bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import { BrowserRouter } from "react-router-dom";
 // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


services/minespace-web/src/App.js
```diff
@@ -1,31 +1,42 @@
...
 
+const propTypes = {
+  loadBulkStaticContent: PropTypes.func.isRequired,
+  isAuthenticated: PropTypes.bool,
+  staticContentLoadingIsComplete: PropTypes.bool,
+};
+
+const defaultProps = {
+  isAuthenticated: false,
+  staticContentLoadingIsComplete: false,
```
simensma-fresh, line 37: Is this one necessary?

simensma-fresh (approved):

services/minespace-web/src/App.js
matbusby-fw: Without it, it breaks the current eslint rules:

`ESLint: propType "isAuthenticated" is not required, but has no corresponding defaultProps declaration.(react/require-default-props)`

simensma-fresh (approved):

PR #2232, opened 2022-08-16
[MDS-4630] updated notification mark as read behaviour

matbusby-fw:
## Objective 

[MDS-4630](https://bcmines.atlassian.net/browse/MDS-4630)

- Notifications are now marked as read when clicked on rather than when the drawer is opened
- Added a `Mark all as read` button to the bottom of the drawer which marks all unread notifications as read and closes the drawer
- Updated the Notice of Departure pages to remove the query once loaded so returning to this place in the history/refreshing won't re-open the modal once it's been dismissed.

image


simensma-fresh (approved):
👍

PR #2237, opened 2022-08-19
[MDS-4631] Added endpoint to create TSF EOR appointment

simensma-fresh:
## Objective 

[MDS-4631](https://bcmines.atlassian.net/browse/MDS-4631)

Added endpoint to create TSF EOR appointment, for use by MineSpace web. This endpoint creates a `party` with the given info, and associates it with a TSF through a `MinePartyAppointment`. The previous EOR appointments term will be updated to not overlap with the new appointment.

Why a new endpoint for this instead of using the existing party creation + TSF update endpoints? The creation of a party is currently not exposed to MineSpace users at all. We still only want the MS users to be able to create a `party` within the context of a TSF EOR appointment - so this will allow us to restrict that.

### DIFF
--- a/services/core-api/app/api/mines/tailings/party_appt/resources/party_appt_list.py
+++ b/services/core-api/app/api/mines/tailings/party_appt/resources/party_appt_list.py
@@ -1,0 +1,134 @@
+import uuid
+
+from datetime import datetime, timezone
+from dateutil.tz import UTC
+
+import dateutil.parser
+
+from flask import request, current_app
+from flask_restplus import Resource
+from sqlalchemy import or_, exc as alch_exceptions
+from werkzeug.exceptions import BadRequest, InternalServerError, NotFound, Forbidden
+
+from app.api.parties.party.models.address import Address
+from app.extensions import api
+from app.api.utils.access_decorators import requires_role_view_all, requires_role_mine_edit
+from app.api.utils.resources_mixins import UserMixin
+from app.api.utils.custom_reqparser import CustomReqparser
+
+from app.api.mines.mine.models.mine import Mine
+from app.api.mines.permits.permit.models.permit import Permit
+from app.api.parties.party.models.party import Party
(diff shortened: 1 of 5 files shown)


services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
```diff
@@ -118,21 +118,13 @@ def post(self, mine_party_appt_guid=None):
...
-                current_mpa = MinePartyAppointment.find_current_appointments(
-                    mine_party_appt_type_code=mine_party_appt_type_code, permit_id=permit.permit_id)
-            else:
-                current_mpa = MinePartyAppointment.find_current_appointments(
-                    mine_guid=mine_guid, mine_party_appt_type_code=mine_party_appt_type_code)
-            if len(current_mpa) != 1:
-                raise BadRequest('There is currently not exactly one active appointment.')
-            current_mpa[0].end_date = start_date - timedelta(days=1)
-            current_mpa[0].save()
+            MinePartyAppointment.end_current_appointment(
```
matbusby-fw, line 121: Nice!

matbusby-fw (approved):
Looks great!

services/core-api/app/api/mines/tailings/party_appt/resources/party_appt_list.py
```diff
@@ -0,0 +1,134 @@
...
+    parser.add_argument(
+        'sub_division_code',
+        type=str,
+        store_missing=False,
+        help='The region code where the party is located. Ex: BC')
+    parser.add_argument(
+        'post_code',
+        type=str,
+        store_missing=False,
+        help='The postal code of the party address. Ex: A0B1C2')
```
matbusby-fw, line 73: Actually, I just noticed we don't have an email field here.   There is one on the form.  Can we add that in?

PR #2239, opened 2022-08-23
[MDS-4631] Open up endpoint to create Mine Party Appt (EORs only) to Minespace proponents

simensma-fresh:
## Objective 

[MDS-4631](https://bcmines.atlassian.net/browse/MDS-4631)

Opened up endpoint to create Mine Party Appt (EORs only) to Minespace proponents.
- (for now) they should only be able to create EOR appointments
- (for now) The party being appointed as an EOR must already be associated with the mine

This will be used on the Minespace side to assign EORs to a TSF


matbusby-fw (approved):
Awesome!

PR #2240, opened 2022-08-23
[MDS-4632] create tsf page

matbusby-fw:
## Objective 

[MDS-4632](https://bcmines.atlassian.net/browse/MDS-4632)

- Created new TSF add/edit page
- changed navigation on edit TSF to new page
- set up fields for the `Basic Information` and `Engineer of Record` tabs
- connected the save as draft functionality to update the engineer of record
- wrapped components/elements in appropriate production flags

image
image
image
image


services/minespace-web/src/components/Forms/tailing/tailingsStorageFacility/TailingsSummaryForm.js
```diff
@@ -0,0 +1,219 @@
...
+      "qualified-person": <div />,
+      "registry-document": <div />,
+      reports: <div />,
+      summary: <div />,
+    }[tab]);
+  const isFirst = tabIndex === 0;
+  const isLast = tabs.length - 1 === tabIndex;
+  const errors = Object.keys(flattenObject(formErrors));
+  const disabledButton = errors.length > 0;
+  return (
```
simensma-fresh, line 94: I see that this is very similar to the project summary do you think it would be a lot of effort to make this into a reusable component? E.g. you would be able to pass along the tabs you want, and it internally manages which tab is active, whether or not to show the back button etc. I get the feeling that we'll have more pages like these popping up in MineSpace going forwards

services/minespace-web/src/components/Forms/tailing/tailingsStorageFacility/BasicInformation.js
```diff
@@ -0,0 +1,86 @@
...
+      id="mine_tailings_storage_facility_name"
+      name="mine_tailings_storage_facility_name"
+      label="Tailing Storage Facility Name"
+      component={renderConfig.FIELD}
+      validate={[maxLength(300)]}
+    />
+    <Row gutter={16}>
+      <Col span={12}>
+        <Field
+          id="latitude"
```
simensma-fresh, line 43: More of a general question, are there any validation for lat/long other places in the app? If so, we should probably match it up

simensma-fresh (commented):
Looks good! Just two comment I'd like to get your thoughts on before approving

services/minespace-web/src/components/Forms/tailing/tailingsStorageFacility/TailingsSummaryForm.js
matbusby-fw: Good thought.  It could certainly be done.  I'd say it would be a moderate effort to set that up, but it could definitely be worth it.

services/minespace-web/src/components/Forms/tailing/tailingsStorageFacility/BasicInformation.js
matbusby-fw: Ah yes.  Indeed there is a validator for that.  Will add it in.

services/minespace-web/src/components/Forms/tailing/tailingsStorageFacility/TailingsSummaryForm.js
simensma-fresh: Sounds good! Maybe let's check in with Roop to see if he imagines it being used elsewhere, and we can do the refactor with one of the stories this sprint

simensma-fresh (approved):

PR #2242, opened 2022-08-25
[MDS-4660] - removed add new eor button.

matbusby-fw:
## Objective 

- Removed `Assign New EOR` button
- triggered validation before save to prevent some server message error responses

image


simensma-fresh (approved):
Approved!

PR #2246, opened 2022-08-26
[MDS-4563] Added prompt before uploading new nod checklist and warning text

matbusby-fw:
## Objective 

[MDS-4563](https://bcmines.atlassian.net/browse/MDS-4563)

- Added text warning before nod self assessment upload field (only when one exists)
- Added window prompt to confirm overwrite of the previous file when uploading a new nod self-assessment

image
image


simensma-fresh (approved):
⚠️

PR #2247, opened 2022-08-30
[MDS-4656] - TSF Basic Information

matbusby-fw:
## Objective 

WORK IN PROGRESS 
Still working on some of the other A/C of the ticket

[MDS-4656](https://bcmines.atlassian.net/browse/MDS-4656)

- Created "SteppedForm" component for reuse
- refactored logic for TSF form to live as much as possible within this form

image


services/minespace-web/src/components/pages/Tailings/TailingsSummaryPage.js
```diff
@@ -161,31 +164,23 @@ export const TailingsSummaryPage = (props) => {
...
-        </Tabs>
+        <SteppedForm
+          tabs={tabs}
+          initialValues={initialValues}
+          handleSaveData={handleSaveData}
+          handleTabChange={handleTabChange}
+          handleSaveDraft={handleSaveData}
+          errors={errors}
+          match={match}
+          form={
```
simensma-fresh: This looks pretty clean! Thoughts on modelling a bit more on the `<Tab` component? So you would pass in the tabs as subcomponents themselves. That way the conditional rendering of the different steps could also be internal to the `SteppedForm` component

```js
<SteppedForm
          currentStep={tab}
          initialValues={initialValues}
          handleSaveData={handleSaveData}
          handleTabChange={handleTabChange}
          handleSaveDraft={handleSaveData}
          errors={errors}
          match={match}>
    <Step key="basic-information">
          <BasicInformation initialValues={initialValues} />
    </Step>
    <Step key="engineer-of-record">
             <EngineerOfRecord initialValues={initialValues} /
     </Step>
    ...
</SteppedForm>
```

matbusby-fw: That's a good idea.  I had been keeping passing in the form, but it's been stripped of any specific usefulness at this point so we can easily wrap this current component in the redux form wrapper and pass things in this way.

migrations/sql/V2022.09.01.11.43__update_mine_tailings_storage_facility.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TYPE storage_location AS ENUM ('above_ground', 'below_ground');
+CREATE TYPE facility_type AS ENUM ('tailings_storage_facility');
+CREATE TYPE tailings_storage_facility_type AS ENUM ('conventional', 'dry_stacking', 'pit', 'lake', 'other');
+
+ALTER TABLE
+    mine_tailings_storage_facility
+ADD COLUMN storage_location storage_location,
+ADD COLUMN facility_type facility_type,
```
simensma-fresh: Maybe we can add in a `NOT NULL` constraint for `facility_type` and a default value of `tailings_storage_facility` so that we can populate this for the existing records.

`ADD COLUMN facility_type facility_type NOT NULL DEFAULT 'tailings_storage_facility'`

services/core-api/app/api/mines/tailings/models/tailings.py
```diff
@@ -23,6 +50,10 @@ class MineTailingsStorageFacility(AuditMixin, Base):
     itrb_exemption_status_code = db.Column(db.String)
     tsf_operating_status_code = db.Column(db.String)
     notes = db.Column(db.String)
+    storage_location = db.Column(db.Enum(StorageLocation), nullable=True)
+    facility_type = db.Column(db.Enum(FacilityType), nullable=True)
```
simensma-fresh: As per my other comment, we can make this `nullable=False` if we add in a `default` here

services/core-api/app/api/mines/tailings/resources/tailings_list.py
```diff
@@ -61,6 +62,34 @@ class MineTailingsStorageFacilityListResource(Resource, UserMixin):
...
+        location='json',
+        choices=list(TailingsStorageFacilityType),
+        store_missing=False)
+    parser.add_argument(
+        'mines_act_permit_no',
+        type=str,
+        help='Mines Act Permit Number',
+        location='json',
+        store_missing=False)
+
```
simensma-fresh, line 92: Should all of these be optional? 
If not we should try to enforce it here

services/core-web/common/reducers/partiesReducer.js
```diff
@@ -39,9 +41,19 @@ export const partiesReducer = (state = initialState, action) => {
         partyIds: createItemIdsArray([action.payload], "party_guid"),
       };
     case actionTypes.STORE_PARTY_RELATIONSHIPS:
+      const eors = action.payload.reduce((acc, curr) => {
```
simensma-fresh: Thoughts on using a filter + map instead of a reduce?

For readability I tend to prefer a `filter + map` then reduce in situations like these, just makes it easier at a glance to figure out what's going on - and performance isn't really a concern.

```javascript
const eors = action.payload
     .filter(p => p.mine_party_appt_type_code === "EOR")
     .map(p => ({
         value: p.party_guid,
         label: p.party.name
     });
```

services/core-web/src/tests/mocks/dataMocks.js
```diff
@@ -7157,46 +7157,3 @@ export const ACTIVITIES = {
     total: 3,
   },
 };
-
-export const PROJECT_DECISION_PACKAGE = {
```
simensma-fresh, line 7161: Was this intended?

services/minespace-web/src/components/dashboard/mine/tailings/TailingsTable.js
```diff
@@ -7,12 +7,11 @@ import {
   getITRBExemptionStatusCodeOptionsHash,
   getTSFOperatingStatusCodeOptionsHash,
 } from "@common/selectors/staticContentSelectors";
-import { Link } from "react-router-dom";
 import { detectProdEnvironment as IN_PROD } from "@common/utils/environmentUtils";
 import * as Strings from "@/constants/strings";
 import { EDIT_PENCIL } from "@/constants/assets";
 import AuthorizationWrapper from "@/components/common/wrappers/AuthorizationWrapper";
-import { EDIT_TAILINGS_STORAGE_FACILITY } from "@/constants/routes";
+// import { EDIT_TAILINGS_STORAGE_FACILITY } from "@/constants/routes";
```
simensma-fresh: Can remove this

services/minespace-web/src/components/pages/Tailings/TailingsSummaryPage.js
```diff
@@ -161,43 +161,47 @@ export const TailingsSummaryPage = (props) => {
...
         </Row>
         <Divider />
-        <Tabs
-          tabPosition="left"
-          activeKey={activeTab}
-          defaultActiveKey={tabs[0]}
-          onChange={(tab) => handleTabChange(tab)}
-          className="vertical-tabs"
+        <SteppedForm
+          tabs={tabs}
```
simensma-fresh: Looks great 🙌 

Just one nitpick, do we need to pass in the tabs here any more? I think we can achieve the same by reconstructing it inside the `SteppedForm` component instead (`children.map(c => c.key)`). Just one less thing you have to pass along 🙃

services/minespace-web/src/components/pages/Tailings/TailingsSummaryPage.js
```diff
@@ -161,43 +161,47 @@ export const TailingsSummaryPage = (props) => {
...
+          handleSaveData={handleSaveData}
+          handleTabChange={handleTabChange}
+          handleSaveDraft={handleSaveData}
+          errors={errors}
+          match={match}
```
simensma-fresh: One suggestion would be to instead of passing along `match` we pass in `match.params.tab` here (`activeTab= {match.params.tab}`, that way we can decouple the component itself from the router completely

simensma-fresh (changes requested):
Overall looks great! Left a couple of suggestions, but overall really like how the Stepped Form component turned out 🎆

services/core-api/app/api/mines/tailings/resources/tailings_list.py
simensma-fresh: On second thought, that might cause issues on the core side? Anyhow, we should create a ticket to add these fields to the Core TSF form if there's not already one

migrations/sql/V2022.09.01.11.43__update_mine_tailings_storage_facility.sql
matbusby-fw: Sounds good!

services/core-api/app/api/mines/tailings/resources/tailings_list.py
matbusby-fw: Yeah, I was trying not to interfere with the current process as much as possible for the time being.  Agree, we should make sure we have those tickets for adding the fields (I know there are designs for the updated Core pages at least)

services/core-web/common/reducers/partiesReducer.js
matbusby-fw: Sure, I'm happy with that

services/core-web/src/tests/mocks/dataMocks.js
matbusby-fw: Hmmm, no it wasn't.  Must have missed this in the rebase and accepted the other side.  I'll put it back.  Good catch!

services/minespace-web/src/components/pages/Tailings/TailingsSummaryPage.js
matbusby-fw: Ooooh, good thinking!

simensma-fresh (dismissed):

simensma-fresh (approved):

PR #2249, opened 2022-08-31
[MDS-4597] major mine application all projects view

areyeslo:
## Objective 

[MDS-4597](https://bcmines.atlassian.net/browse/MDS-4597)

Create a Search page to search for projects by stage. If a project has different stages (Project Summary, IRT, Major Mine Application), then the most updated stage is shown.

image

Status code changes:
- Project Summary:
  -Deleted status_codes: OPN- Open, CLD- Closed, IAT-Inactive, UNP- Under review- with proponents
  -Updated description for UNR- Under review to reviewers to Under review

- IRT:
 -Updated PRG- In progress to DFT- Draft, REC- Pending Review to SUB- Submitted
 -Updated description for APV- Review Complete to APV- Approved
 
- Major Mine Application:
 -Updated REC- Pending Review to SUB- Sbumitted
 -Updated UNR- In review with reviewers to Under review

Sorters:
Currently, there are some fields that cannot be sorted on the table:
- Stage
- Review Status
- Updated Date
- Commodity

Each project can have multiple stages (Project Summary, IRT, MMA) and when Front End sends the request to sort by specific sort_field, sqlalchemy includes it in the query. However, it is until getting the results that we can obtain the most updated stage and pick this one from the three records (Project Summary, IRT, MMA) returned by the query. Then, sorting by stage, review_status or updated_date is not possible at the moment with the current logic using sqlalchemy.
Commodity field is obtained during transformation of row data.


services/core-api/app/api/projects/project/resources/project.py
```diff
@@ -72,3 +81,211 @@ def get(self):
...
+        sort_models = {
+            'project_title': 'Project',
+            'project_id': 'Project',
+            'mine_commodity_code': 'MineTypeDetail',
+            'status_code':'ProjectSummary',
+            'status_code':'InformationRequirementsTable',
+            'status_code': 'MajorMineApplication',
+            'update_timestamp': 'ProjectSummary',
+            'update_timestamp': 'InformationRequirementsTable',
+            'update_timestamp': 'InformationRequirementsTable'
```
BenHarrington: I think this third "update_timestamp" was meant to be MajorMineApplication?

services/core-api/app/api/projects/response_models.py
```diff
@@ -214,3 +214,32 @@ def format(self, value):
...
+    'ProjectMineList',{
+        'stage': fields.String,
+        'id': fields.Integer,
+        'guid': fields.String,
+        'project_title': fields.String,
+        'project_id': fields.String,
+        'project_guid': fields.String,
+        'mrc_review_required': fields.Boolean,
+        'status_code': fields.String,
+        'contacts': fields.List(fields.Nested(PROJECT_SUMMARY_CONTACT_MODEL)),
```
BenHarrington: There is a PROJECT_CONTACT_MODEL I believe that should be used here. I think the PS Contact has been retired, but would need to double check that.

services/core-web/common/constants/actionTypes.js
```diff
@@ -101,6 +101,7 @@ export const STORE_REQUIREMENTS = "STORE_REQUIREMENTS";
 export const CLEAR_REQUIREMENTS = "CLEAR_REQUIREMENTS";
 export const STORE_MAJOR_MINES_APPLICATION = "STORE_MAJOR_MINES_APPLICATION";
 export const CLEAR_MAJOR_MINES_APPLICATION = "CLEAR_MAJOR_MINES_APPLICATION";
+export const STORE_PROJECT_VIEW_ALL_TABLE = "STORE_PROJECT_VIEW_ALL_TABLE";
```
BenHarrington, line 106: Any reason a "CLEAR..." variable wasn't provided as well here?

services/core-web/src/components/navigation/NavBar.js
```diff
@@ -314,6 +314,18 @@ export class NavBar extends Component {
               </Link>
             </Col>
           </Row>
+          <Row>
+            <Col span={24}>
+              <Link to={router.MAJOR_PROJECTS_DASHBOARD.route}>
```
BenHarrington: Are we putting this dashboard behind a feature/ENV wall first for testing or is it straight to Prod?

services/core-web/src/components/navigation/NavBar.js
```diff
@@ -413,6 +425,11 @@ export class NavBar extends Component {
           <button type="button">Notices of Work</button>
         </Link>
       </Menu.Item>
+      <Menu.Item key="browse-major-projects" className="navbar-dropdown-menu-item">
+        <Link to={router.MAJOR_PROJECTS_DASHBOARD.route}>
```
BenHarrington: Same as previous comment.

services/core-api/app/api/projects/project/resources/project.py
areyeslo: That's correct. Thank you for noticing

services/core-api/app/api/projects/response_models.py
areyeslo: Agree, contacts has to be the same as PROJECT_MODEL.

services/core-web/src/components/navigation/NavBar.js
areyeslo: Good idea, going to put the behind feature flag

BenHarrington (commented):
Pagination controls do not seem to be setup correctly: 
image

The previous/next buttons are disabled even though we have more than 1 page of results. Additionally the result count is 0.

BenHarrington (commented):
I notice only a couple columns on the table offer sort: 
image

Was this a miss or intentional? It seems a bit inconsistent between the screenshot on the ticket and mockup, but the AC in the ticket does say these fields should be sortable(unless those reqs changed).

BenHarrington (commented):
I notice on the "Application Statuses" filter control that we have over 10 eligible statuses. Were we not consolidating those to simplify things? Perhaps there is another static content selector that needs to be updated, unless we are still intending to have all of those statuses as eligible.

services/core-api/app/api/projects/namespace.py
```diff
@@ -1,6 +1,6 @@
 from flask_restplus import Namespace
 
-from app.api.projects.project.resources.project import ProjectResource, ProjectListResource
+from app.api.projects.project.resources.project import ProjectResource, ProjectListResource, MajorProjectListDashboardResource
```
JustinMacaulay: should the name of the new resource be inline with the others? ProjectListDashboardResource?

JustinMacaulay (commented):
Just a comment around the naming.

BenHarrington (approved):
Great job on this huge task Arturo. Very good work handling the complicated sort/filtering on the backend.

PR #2253, opened 2022-09-08
[MDS-4647] NoW Delay end reverts status

JustinMacaulay:
When a delay is ended the status of the NoW is set back to the status from before the delay.

[MDS-4647](https://bcmines.atlassian.net/browse/MDS-46478)

image

image


BenHarrington (approved):
Looks good.

PR #2254, opened 2022-09-09
[MDS-4646] Notice of Work delay dates off by 1 day

BenHarrington:
## Objective 

[MDS-4646](https://bcmines.atlassian.net/browse/MDS-4646)

_Changes_

- Fix PUT payload on both Progress Dates and Delay Dates to cast start/end dates as a moment objects before reaching the API for consistent date formatting in and out

NOTE: I updated the Progress Dates as well because I discovered it had the same issue as delays.


JustinMacaulay (commented):
Looks good

JustinMacaulay (approved):

PR #2255, opened 2022-09-09
[FEATURE] MDS-4684 - init dir for discord_sysdig alerts

wilson496:
## Objective 

[MDS-4684](https://bcmines.atlassian.net/browse/MDS-4684)

Creating AWS resources in Terraform for microservice to webhook msgs from Sysdig to Discord


vyasworks (approved):
Looks good, 

I would use deterministic name for S3 bucket instead of `random_pet` suffix, but that should be ok.

PR #2263, opened 2022-09-15
[MDS-4669] Core-web: Add new project description

areyeslo:
## Objective 

[MDS-4669](https://bcmines.atlassian.net/browse/MDS-4669)

Add ability to create a Project Description in Core-web:
image


services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
```diff
@@ -125,10 +131,14 @@ def post(self):
         if mine is None:
             raise NotFound('Mine not found')
 
+        current_app.logger.debug(f'mine: {mine_guid}')
```
BenHarrington: Do we need this here?

services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
```diff
@@ -125,10 +131,14 @@ def post(self):
...
         new_project = Project.create(mine, data.get('project_summary_title'),
                                      data.get('proponent_project_id'),
                                      data.get('mrc_review_required', False),
-                                     data.get('contacts', []))
+                                     data.get('contacts', []),
+                                     data.get('project_summary_lead_party_guid'))
+
+        current_app.logger.debug(f'new project: {new_project}')
```
BenHarrington: Same as previous comment.

services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
```diff
@@ -125,10 +131,14 @@ def post(self):
         if mine is None:
             raise NotFound('Mine not found')
 
+        current_app.logger.debug(f'mine: {mine_guid}')
         new_project = Project.create(mine, data.get('project_summary_title'),
                                      data.get('proponent_project_id'),
                                      data.get('mrc_review_required', False),
-                                     data.get('contacts', []))
+                                     data.get('contacts', []),
+                                     data.get('project_summary_lead_party_guid'))
```
BenHarrington: Not a big issue, but do you think we should change the name of that to "project_lead_party_guid", I realize the frontend would need a small change as well. I wonder if leaving it like this could be confusing to future devs without the context. Dealer's choice.

BenHarrington (commented):
I noticed this(no header info) when entering the create new project page with an empty state: 
image

BenHarrington (commented):
Noticed the Project Details section is only half width(looks to be full width in mocks):
 
image

There are other sections that are as well(Contacts, Dates), I think Dates is fine because the input will never get very large, however in Contacts I notice the phone extension field is misaligned slightly so you may want to bump that col-12 up a bit on that section:

image

BenHarrington (commented):
After creating a project description we get a success toast and the URL updates to "/pre-applications/175eedbc-7336-44ef-9d0f-9cdc78b3e66e/project-description/f8f975eb-2d21-4d8d-a261-99473d41801c"(for example) but all the fields are empty(need to fetch appropriate entity after creation in the submit handler most likely). This page should behave as the edit version after successful creation.

BenHarrington (commented):
This is more of a UI issue(may exist on Minespace as well), I notice if you fill out the form and do not interact with the MRC radio input the Create button at the bottom is disabled but there is no feedback that that field needs to be selected. Maybe we add some text there, or pre-validate it on initial page load, something like that. 

This isn't a blocker for this ticket, just a thought.

services/core-api/app/api/projects/project_summary/resources/project_summary_list.py
JustinMacaulay: theres also a logging statement above that either needs more info in it or to be removed.

JustinMacaulay (commented):
a few small comments

JustinMacaulay (approved):
Looks good.

PR #2264, opened 2022-09-15
[MDS-3544] Move Permit MX-2-16 to correct mine

BenHarrington:
## Objective 

[MDS-3544](https://bcmines.atlassian.net/browse/MDS-3544)

- Created a reusable function to be used in place of running manual SQL statements to transfer mine_permit_xref and related entities from one mine to another
- Excluded deleted mine_permit_xref from response model


JustinMacaulay (approved):
Looks good, will run tomorrow.

PR #2269, opened 2022-09-22
[MDS-4670] Edit project description in CORE

JustinMacaulay:
## Objective 

[MDS-4670](https://bcmines.atlassian.net/browse/MDS-4670)

image

image

image

adds the functionality to the CORE side or project descriptions for them to be editable.


BenHarrington (commented):
When setting a Project Lead we still get this callout:

image

BenHarrington (commented):
When setting Authorizations Involved, the PUT payload for authorizations is empty. Perhaps look at the create handler to see if it is doing anything different.

Update: This only happens on EDIT not create.

BenHarrington (approved):
Changes look good, thanks Justin!

PR #2273, opened 2022-09-27
[FIX] Update transfer_mine_permit_xref_record function

BenHarrington:
## Objective 

[MDS-3544](https://bcmines.atlassian.net/browse/MDS-3544)

- Fixed race condition with mine_report
- Added comments/special use instructions


JustinMacaulay (approved):
Looks good.

PR #2274, opened 2022-09-27
[MDS-4613] Allow creation / editing of a TSF Qualified Person through MineSpace

simensma-fresh:
## Objective 

[MDS-4613](https://bcmines.atlassian.net/browse/MDS-4613)
[MDS-4725](https://bcmines.atlassian.net/browse/MDS-4725)
[MDS-4726](https://bcmines.atlassian.net/browse/MDS-4726)

Added support for MineSpace users to assign/create update Qualified Persons to a TSF. Fixed various bugs in the contact creation modal, TSF flow.

image
image


services/core-api/app/api/mines/tailings/models/tailings.py
```diff
@@ -63,11 +63,25 @@ class MineTailingsStorageFacility(AuditMixin, Base):
...
+
     @hybrid_property
     def engineer_of_record(self):
         if self.engineer_of_records:
             return self.engineer_of_records[0]
 
+    @hybrid_property
+    def qualified_person(self):
+        if self.qualified_persons:
+            return self.qualified_persons[0]
```
matbusby-fw, line 83: Are we only allowing a single qualified person?  I thought there could be more than one, but I may be wrong about that.

services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -64,9 +64,9 @@ def validate_permit():
                     'Permits cannot be associated with non permit-linked contact types.')
 
         def validate_tsf():
-            if self.mine_party_appt_type_code == 'EOR':
+            if self.mine_party_appt_type_code in TSF_ALLOWED_CONTACT_TYPES:
                 if not self.mine_tailings_storage_facility:
-                    raise AssertionError('The associated TSF is required for Engineer of Records.')
+                    raise AssertionError('The associated TSF is required for Engineer of Records or Qualified Persons.')
```
matbusby-fw: Probably not a big deal, but since Qualified Persons can exist for other Mine contacts, maybe we could change the last part to something like 'TSF related qualified persons'?

services/minespace-web/src/components/common/SteppedForm.js
```diff
@@ -48,7 +48,8 @@ const SteppedForm = (props) => {
 
   const handleNextClick = (evt, tab) => {
     evt.preventDefault();
-    handleTabClick(tab);
+    handleSaveData(null, tab);
```
matbusby-fw, line 52: Hmmm, we'll need to think how we want to handle saves on tab changes.  I think this is going to trigger a double save for 'Basic Information' and 'EOR' since those saves are triggered within `SteppedForm`

matbusby-fw (changes requested):
Looks great!  Just a couple of questions.  I added a couple of questions.  I had thought we were doing multiple QPs, but it's quite possible that I have that wrong.

services/minespace-web/src/components/common/SteppedForm.js
simensma-fresh: Good point! My opinion would be that saves should only happen when you hit "next" and not when you switch tabs, if anything we could reset the form with a "are you sure you want to discard your current changes?" prompt if you switch tabs without hitting next. 

Will check in with Roop/Raj on this and update it across the board

matbusby-fw: Yeah, that would work for me!

services/core-api/app/api/mines/tailings/models/tailings.py
simensma-fresh: Currently, a mine can have multiple Qualified Persons since they are not tied to a TSF at all, but my understanding is that a TSF will only have one QP going forward

matbusby-fw: Maybe you're right.  The designs certainly support that look.  I just had a vague memory of discussing multiple TSF QPs, but I think that's either not what's needed or at least not MVP.

simensma-fresh:
@matbusby-fw I've updated the PR with some of the changes if you want to take another look
- All tabs except the basic-info tab are now disabled until the TSF exists instead of basing it on form errors
- Saving/creating of TSF/EOR/QP now only happens when you hit "next", not when you click on the tabs
- The state now fully reloads after you add a new EOR/QP, to avoid any inconsistencies (e.g. EOR document not showing up when you hit back button until you refresh the page). 
- Fixed the related issues in MDS-4725 / MDS-4726

matbusby-fw (approved):
Awesome!  Looks great!

PR #2275, opened 2022-09-29
[MDS-4702] Read only review submit with success page

areyeslo:
## Objective 

[MDS-4702](https://bcmines.atlassian.net/browse/MDS-4702)

- Ability to review a new incident and submit:
image

- Success page is shown upon successful submission:
1) No dangerous occurrence:
image
2) Dangerous occurrence:
image


services/minespace-web/src/components/pages/Incidents/IncidentPage.js
```diff
@@ -187,34 +280,48 @@ export class IncidentPage extends Component {
...
-    return this.props
-      .updateMineIncident(mineGuid, mineIncidentGuid, values)
-      .then(() => this.handleFetchData())
-      .then(() => this.setState({ isLoaded: true }));
+
+    let message;
+    if (isDraft) {
+      message = "Successfully updated draft incident.";
+    } else if (this.props.location.state.current === 2) {
+      message = "Submitted successfully a new incident.";
```
BenHarrington: Should this copy read "Successfully submitted a new incident."? To be consistent with the previous message structure.

services/minespace-web/src/components/pages/Incidents/IncidentSuccessPage.js
```diff
@@ -0,0 +1,106 @@
...
+  const renderContent = () => {
+    const incident = props.location.state?.incident || {};
+    const { mine_incident_guid, mine_guid, mine_determination_type_code } = incident;
+    return (
+      <div style={{ textAlign: "center" }}>
+        <>
+          <Row>
+            <Col span={24}>
+              <CheckCircleOutlined
+                style={{
```
BenHarrington: I notice IRT and Major Mines Application success pages also use this same style. Does it make sense to create a class for it at this point? Not sure if we are going to use have other features using this pattern moving forward, but there may be a chance of it. Dealer's choice.

services/minespace-web/src/components/pages/Incidents/IncidentPage.js
```diff
@@ -144,16 +170,82 @@ const StepForms = (props, state, navigation, handlers, formatInitialValues) => [
...
+        setConfirmedSubmission={setConfirmedSubmission}
+        confirmedSubmission={state.confirmedSubmission}
+        incident={props.incident}
+        formValues={props.formValues}
+      />
+    ),
+    buttons: [
+      <React.Fragment key="step-3-buttons">
+        <Button
+          id="step-back2"
```
BenHarrington: Not a big issue, but this id seems to reference step 2. Wouldn't worry about it unless you are updating other stuff in this component.

JustinMacaulay (commented):
Just what we went over in a call but looks good otherwise.

JustinMacaulay (dismissed):
Looks good to me and thanks for making the change.

services/minespace-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -47,28 +56,71 @@ const documentColumns = [
...
-      <Form.Item label="Incident type(s)">
-        <Field
-          id="categories"
-          name="categories"
-          placeholder="Select incident type(s)"
-          component={renderConfig.MULTI_SELECT}
-          validate={[requiredList]}
-          data={props?.incidentCategoryCodeOptions}
+      {!props.isReviewSubmitStage && (
+        <Form.Item label="Incident type(s)">
```
BenHarrington: Is there a reason we can't just use the <Field> component and set disabled on that. Why do we need an <Input> when it is in review?

BenHarrington (approved):
Great work Arturo, thanks for digging into the comments I left.

PR #2276, opened 2022-10-03
[MDS-4706] Internal ministry comments component.

areyeslo:
## Objective 

[MDS-MDS-4706](https://bcmines.atlassian.net/browse/MDS-4706)

_Create new common component for Internal Ministry comments.

### DIFF
--- a/services/core-web/src/components/mine/Incidents/MinistryInternalComments.js
+++ b/services/core-web/src/components/mine/Incidents/MinistryInternalComments.js
@@ -1,0 +1,95 @@
+import React, { Component } from "react";
+import { bindActionCreators } from "redux";
+import { connect } from "react-redux";
+import PropTypes from "prop-types";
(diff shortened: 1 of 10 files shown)


services/core-web/src/components/mine/Incidents/MinistryInternalComments.js
```diff
@@ -0,0 +1,95 @@
...
+  fetchMineIncidentNotes: PropTypes.func.isRequired,
+  mineIncidentGuid: PropTypes.string,
+};
+
+const defaultProps = {
+  mineIncidentGuid: null,
+};
+
+export class MinistryInternalComments extends Component {
+  state = { loading: true };
```
BenHarrington, line 26: Should loading be false initially? I notice we set it to true at the start of the fetch which would negate setting this prematurely.

BenHarrington (approved):
Good work Arturo!

PR #2277, opened 2022-10-04
[MDS-4750] Updated TSF Flow success messages, fixed edge case errors in TSF Flow

simensma-fresh:
## Objective 

[MDS-4750](https://bcmines.atlassian.net/browse/MDS-4750)

Updated success messages for TSF flow steps to be more relevant. Fixed a couple of bugs found during testing.
- When you hit next, the next button gets disabled and you wait until the action (create tsf/assign eor/qp) finishes before continuing to the next tab. This resolves things being in an inconsistent state (e.g. if you click the back button too quickly
- Fixed a DB error that popped up for new TSFs if you entered text in the contact select dropdown, but it didn't have any contacts populated -> This case was not caught by form validation
- Next button of stepped form is now disabled when there are form errors, but tabs are not.

image


services/core-web/common/utils/Validate.js
```diff
@@ -168,9 +168,9 @@ export const validateIncidentDate = memoize((reportedDate) => (value) =>
 );
 
 // eslint-disable-next-line consistent-return
-export const validateSelectOptions = memoize((data) => (value) => {
-  if (value && data.length > 0) {
-    return data.find((opt) => opt.value === value) !== undefined
+export const validateSelectOptions = memoize((data, allowEmptyData = false) => (value) => {
+  if (value && (data?.length > 0 || allowEmptyData)) {
```
matbusby-fw, line 172: Good enhancement for the whole site!

matbusby-fw (approved):
Looks great!

PR #2281, opened 2022-10-05
[FEATURE] MDS-4685 - add sysdig promql alerts

wilson496:
## Objective 

[MDS-4685](https://bcmines.atlassian.net/browse/MDS-4685)


terraform/modules/variables.tf
```diff
@@ -48,3 +48,14 @@ variable "configs" {
 variable "storage_buckets" {
   description = "Contains a list of s3 buckets for this workspace"
 }
+
+## Sysdig ##
+
+data "aws_secretsmanager_secret" "secrets" {
```
vyasworks: do we need to mention full arn? could secret be imported by name?

vyasworks (dismissed):

terraform/modules/variables.tf
wilson496: Fixed!

PR #2285, opened 2022-10-06
[MDS-4702] hotfix- Updating obsolete status code

areyeslo:
## Objective 

[MDS-4702](https://bcmines.atlassian.net/browse/MDS-4702)
Update status code from "PRE"-Preliminary to "IRS"- Initial Report Submitted in payload when creating a new incident


JustinMacaulay (approved):
sorry I missed this

PR #2288, opened 2022-10-06
[MDS-4707] Core View and Edit an Incident

BenHarrington:
## Objective 

[MDS-4707](https://bcmines.atlassian.net/browse/MDS-4707)

image

_Changes Included_

- Created new route component (MineIncident)
- Created new form component
- Updated MineIncidents/MineIncidentsTable to use workspace flow when not IN_PROD(modal view will remain for now)
- Created migration to add new mine_incident_document_type_code ("INM", "Internal Ministry Document")
- Added support to DocumentTable to allow providing custom columns

_Not Included_

- Status change widget(will be addressed in a future ticket)


services/core-web/src/components/mine/Incidents/MineIncident.js
```diff
@@ -0,0 +1,319 @@
...
+import * as Permission from "@/constants/permissions";
+import Loading from "@/components/common/Loading";
+import customPropTypes from "@/customPropTypes";
+import IncidentForm from "@/components/Forms/incidents/IncidentForm";
+import ScrollSideMenu from "@/components/common/ScrollSideMenu";
+import * as routes from "@/constants/routes";
+
+const propTypes = {
+  // eslint-disable-next-line react/no-unused-prop-types
+  incident: customPropTypes.incident.isRequired,
```
JustinMacaulay, line 38: Is this ignore still needed cause I can see this prop being used.

services/core-web/src/components/mine/Incidents/MineIncident.js
```diff
@@ -0,0 +1,319 @@
...
+    }
+  };
+
+  handleFetchData = () => {
+    const { mineGuid, mineIncidentGuid } = this.props.match.params;
+    if (mineGuid && mineIncidentGuid) {
+      this.setState({ isNewIncident: false });
+      return this.props.fetchMineIncident(mineGuid, mineIncidentGuid);
+    }
+    return Promise.resolve();
```
JustinMacaulay, line 100: Is this just what you have to do if you do not fetch data?

services/core-web/src/components/mine/Incidents/MineIncident.js
```diff
@@ -0,0 +1,319 @@
...
+  };
+
+  handleCreateMineIncident = (formattedValues) => {
+    this.setState({ isLoaded: false });
```
JustinMacaulay, line 104: should this be isLoaded? in the past we have split these up between loading and submitting.

JustinMacaulay (commented):
Just a couple small comments.

services/core-web/src/components/mine/Incidents/MineIncident.js
BenHarrington: This is a known bug with ESLint where it complains it's not being used when it is, we've had to add this in several areas throughout the project because of "reasons".

services/core-web/src/components/mine/Incidents/MineIncident.js
BenHarrington: If you look at the ComponentDidMount method you see that we call handleFetchData and then use a "thenable" to set state, in order for it to work handleFetchData needs to return a resolved promise. It could be restructured but it does the trick. In handleFetchData we need to return something(ESLint rule) when it's not an existing incident and null would not work with the "thenable".

services/core-web/src/components/mine/Incidents/MineIncident.js
BenHarrington: From what I can see in the project we use submitting to disable buttons/elements during form submission and we manually set isLoaded when loading initially, creating, updating, etc.

JustinMacaulay (approved):
Looks good

PR #2289, opened 2022-10-12
[MDS-4755] core updates major projects table

areyeslo:
## Objective 

[MDS-4755](https://bcmines.atlassian.net/browse/MDS-4755)

- Add filters for MRC and EMLI Project Lead.
- Click on Open button redirects to Overview Page.

image


services/core-api/app/api/projects/project/resources/project.py
```diff
@@ -276,11 +285,36 @@ def _apply_filters_and_pagination(self, args):
...
+            if len(project_lead_conditions) == 2:
+                conditions.append({'or': project_lead_conditions})
+
         projects_query = apply_filters(query, conditions)
 
         # Apply sorting
         if sort_model and args['sort_field'] and args['sort_dir']:
-            sort_criteria = [{'model': sort_model, 'field': args['sort_field'], 'direction': args['sort_dir']}]
+            sort_criteria=[]
+            if sort_model == 'Party':
```
BenHarrington, line 313: I notice when using the column sorter that all entries which do not have a project_lead set do not get returned in the sorted result, is that to be expected?

BenHarrington (commented):
Other than my one comment, this appears to look ready to go.

services/core-api/app/api/projects/project/resources/project.py
areyeslo: Fixed.

BenHarrington (approved):
Thanks for making that change, looks good!

PR #2290, opened 2022-10-12
[MDS-4761] Fixed intermittent validation error that pops up when assigning EoR/QP

simensma-fresh:
## Objective 

[MDS-4761](https://bcmines.atlassian.net/browse/MDS-4761)

After assigning a new EoR / QP to a TSF, the validation message indicating an overlap of start dates with previous EoRs briefly flashes. The cause of this is that we do a full reload of all party relationships for the TSF after submission as to populate dropdowns / the EoR pages etc, this causes the validation to be performed against the newly added record for a brief second before React has had a chance to navigate to the next page.


matbusby-fw (approved):
Nice fix!

PR #2291, opened 2022-10-12
[MDS-4686] Added script to delete a party relationship from a mine/tsf

simensma-fresh:
## Objective 

[MDS-4686](https://bcmines.atlassian.net/browse/MDS-4686)

Added a script to delete party relationships from a mine/tsf. This can be used to delete any party relationship as long as they're not listed as a party on an ESUP, as that will require some more considerations.


matbusby-fw (dismissed):
Nice!

matbusby-fw (approved):

PR #2292, opened 2022-10-12
[MDS-4712] Awaiting Final Report Notifications

JustinMacaulay:
## Objective 

[MDS-4712](https://bcmines.atlassian.net/browse/MDS-4712)

This PR adds in the notifications to the proponent and the OCI for when an incident gets to the awaiting final report stage.


services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -28,6 +30,11 @@ def getYear():
     return datetime.datetime.utcnow().year
 
 
+def format_incident_date(datetime_string):
+    # datetime_object = dateutil.parser.isoparse(datetime_string)
```
BenHarrington: Does this need to be here?

JustinMacaulay: removed

BenHarrington (approved):
Looks good

PR #2296, opened 2022-10-13
[MDS-4510] Core and MineSpace: Delete Maps column from Permits Table

BenHarrington:
## Objective 

[MDS-4510](https://bcmines.atlassian.net/browse/MDS-4510)

_Changes_

- Removed "Maps" column from Core - Mine Permits table
- Removed "Maps" column from Minespace - Mine Permits table
- Updated styling on Minespace - Mine Permits table to remove vertical scroll on files based columns and make expanded row columns more responsive


services/minespace-web/src/styles/components/Tables.scss
```diff
@@ -5,6 +5,11 @@
   overflow-y: auto;
 }
 
+.wrapped-text {
+  word-break: break-word;
```
areyeslo: Based on the documentation `break-word` is deprecated. `normal` is prefered:
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#break-word

areyeslo (approved):
Looks great Ben!! Thanks for addressing the concerns in the PR. 🚀

PR #2297, opened 2022-10-14
[MDS-4727] Direct link to EMLI Inspection Mapper

BenHarrington:
## Objective 

[MDS-4727](https://bcmines.atlassian.net/browse/MDS-4727)

_Changes_

- Added map inspector link to Mine Information tab in Core(links using lat/lng)
- Added map inspector link to NoW Application -> Application tab -> Mine Information(links using NoW guid)

_Developer Note_

- You cannot test this due to needing a specific login to access the Inspection Mapper, in order to verify the two different links please make sure they are in the same format as the ticket
    - In the NoW link, match it with the first link in the ticket and make sure the string after NOW_APPLICATION_GUID updates with the one you are visiting, the rest will be the same as in the ticket
    - In the Mine Information link, match it with the last link(the one that uses level=10) and make sure what follows center changes(lat, lng), the rest will be the same as in the ticket


areyeslo (approved):
Looks great Ben!! 🚀

PR #2300, opened 2022-10-14
[MDS-4601] Project Summary form bug on minespace

JustinMacaulay:
## Objective 
When updating a project summary in Minespace the form was never being reset after the submission so the dirty flag was never being cleared. This caused a warning to pop up when the user would try to go back to the overview after updating.

[MDS-4601](https://bcmines.atlassian.net/browse/MDS-4601)


BenHarrington (approved):
Thanks for making those changes.

PR #2304, opened 2022-10-18
[MDS-4717] Final report received notifications

JustinMacaulay:
## Objective 

When a final report is received on an incident emails are sent to the OCI and the proponent to indicate this has occured as well as Core and Minespace notifications are triggered.

[MDS-4717](https://bcmines.atlassian.net/browse/MDS-4717)

### DIFF
--- a/services/core-api/app/api/mines/incidents/resources/mine_incidents.py
+++ b/services/core-api/app/api/mines/incidents/resources/mine_incidents.py
@@ -325,6 +325,10 @@ class MineIncidentResource(Resource, UserMixin):
                     incident.send_awaiting_final_report_email(True)
                     incident.send_awaiting_final_report_email(False)
                     trigger_notifcation(f'A new Mine Incident has been created for ({incident.mine_name})', incident.mine_table, 'MineIncident', incident.mine_incident_guid, {})
+                if value == 'FRS':
+                    incident.send_final_report_received_email(True)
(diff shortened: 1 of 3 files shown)


services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -325,6 +325,10 @@ def put(self, mine_guid, mine_incident_guid):
                     incident.send_awaiting_final_report_email(True)
                     incident.send_awaiting_final_report_email(False)
                     trigger_notifcation(f'A new Mine Incident has been created for ({incident.mine_name})', incident.mine_table, 'MineIncident', incident.mine_incident_guid, {})
+                if value == 'FRS':
```
BenHarrington, line 328: Will these status based emails get sent every time it is one of those stauses(ie. updating another field with the same status hits this endpoint). Should it only be when the existing status code does not match the request one?

JustinMacaulay: at the moment it is whenever that status is set. I think when we actually wire in all the logic around the statuses this might need to be changed cause we need that logic in first.

BenHarrington (approved):
Looks good Justin. Let's make sure we don't lose track of the status change tracking/management  work that is still outstanding. I just want to be sure we aren't sending out repeat emails on Prod while the remaining work is being done.

PR #2306, opened 2022-10-19
[MDS-4745] Minespace Version Fixes - Part 2

vyasworks:
## Objective 

[MDS-4745](https://bcmines.atlassian.net/browse/MDS-4745)


package.json
```diff
@@ -0,0 +1,18 @@
+{
```
vyasworks, line 1: Basic pkg json, not used for anything yet.

simensma-fresh (approved):

PR #2308, opened 2022-10-19
[MDS-4737] major mines error final application save draft

areyeslo:
## Objective 

[MDS-4737](https://bcmines.atlassian.net/browse/MDS-4737)

- Ability to save a draft for a Major Mine Application
- Added Document Table: 
image

- Add validation in form incidents: 
image

### DIFF
--- a/services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
+++ b/services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
@@ -2,7 +2,16 @@ import React, { Component } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import { Link, withRouter } from "react-router-dom";
-import { change, submit, getFormSyncErrors, getFormValues, isDirty, reset, touch } from "redux-form";
+import {
+  change,
+  submit,
+  getFormSyncErrors,
+  getFormValues,
+  isDirty,
+  destroy,
+  reset,
+  touch,
+} from "redux-form";
 import { Button, Row, Col, Popconfirm, Steps, Typography } from "antd";
 import { ArrowLeftOutlined } from "@ant-design/icons";
 import PropTypes from "prop-types";
@@ -46,6 +55,7 @@ const propTypes = {
   history: PropTypes.shape({ push: PropTypes.func }).isRequired,
   touch: PropTypes.func.isRequired,
   submit: PropTypes.func.isRequired,
+  destroy: PropTypes.func.isRequired,
   formErrors: PropTypes.objectOf(PropTypes.string),
   // eslint-disable-next-line react/no-unused-prop-types
   formValues: PropTypes.objectOf(PropTypes.any).isRequired,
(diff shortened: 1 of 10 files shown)


services/minespace-web/src/routes/Routes.js
```diff
@@ -33,6 +33,21 @@ const Routes = () => (
...
+    <Route
+      exact
+      path={routes.EDIT_MAJOR_MINE_APPLICATION.route}
+      component={AuthenticationGuard()(routes.EDIT_MAJOR_MINE_APPLICATION.component)}
+    />
+    <Route
+      exact
+      path={routes.ADD_MAJOR_MINE_APPLICATION.route}
+      component={AuthenticationGuard()(routes.ADD_MAJOR_MINE_APPLICATION.component)}
+    />
```
JustinMacaulay, line 50: I know the routes were giving you trouble, did moving them up make a difference?

services/minespace-web/src/constants/routes.js
```diff
@@ -102,6 +103,12 @@ export const MAJOR_MINE_APPLICATION_SUCCESS = {
   component: MajorMineApplicationSuccessPage,
 };
 
+export const ADD_MAJOR_MINE_APPLICATION = {
```
JustinMacaulay, line 106: Same here with the re-ordering.

services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
```diff
@@ -441,6 +468,7 @@ const mapDispatchToProps = (dispatch) =>
       reset,
       touch,
       change,
+      destroy,
```
JustinMacaulay, line 471: Do we use this anywhere else? I dont know that I have ever seen destroy.

JustinMacaulay (commented):
Just a few comments

services/minespace-web/src/routes/Routes.js
areyeslo: Yes, the issue was related to the order of the routes. Sorting from more specific to more general plus adding `/edit` to Route for edit made the trick.

services/minespace-web/src/components/pages/Project/MajorMineApplicationPage.js
areyeslo: Just in Incidents in minespace and core-web

services/minespace-web/src/routes/Routes.js
areyeslo: The routes for Major Mine Application are as follow:
/projects/:projectGuid/major-mine-application/:mmaGuid/success

/projects/:projectGuid/major-mine-application/:mmaGuid/review

/projects/:projectGuid/major-mine-application/:mmaGuid/edit

/projects/:projectGuid/major-mine-application/new

Before the edit Route was `/projects/:projectGuid/major-mine-application/:mmaGuid`. Somehow it was not found in the routes eventhough it was placed after review Route (keeping the most specific first, then more general). Then, the only way to make it work was to add `/edit` suffix to the route to make it similar to success and review.

JustinMacaulay: Interesting but good to know.

BenHarrington (approved):
Thanks for digging in and getting those validation issues sorted out!

PR #2310, opened 2022-10-20
[MDS-4738] Npm to yarn - Part 3

vyasworks:
## Objective 

[MDS-4738](https://bcmines.atlassian.net/browse/MDS-4738)


simensma-fresh (dismissed):
Alright, that's a big one! Overall this looks great - I like the `@mds/common` abbreviation 💯 

There's one DS_Store file that can be removed (`.yarn/.DS_Store`), but otherwise this looks good to me.

PR #2314, opened 2022-10-25
[MDS-3892] Delete now script modifications

JustinMacaulay:
## Objective 

Removed a duplicate NoW and in the process I modified the script to do it. if a notice of work is not imported then it will not have an ID so the ID and guid needed to be gotten at the same time.

[MDS-3892](https://bcmines.atlassian.net/browse/MDS-3892)


BenHarrington (approved):

matbusby-fw (approved):
Ah!  Good enhancement!

PR #2316, opened 2022-10-26
[MDS-4728] Added TSF Basic Information page and skeleton of the rest of the TSF pages to core

simensma-fresh:
## Objective 

[MDS-4728](https://bcmines.atlassian.net/browse/MDS-4728)

- Added Basic Info tab of a TSF to core, this is pretty much identical to the TSF page in Minespace.
- `SteppedForm` and TSF `BasicInformation` components have now been moved into the common folder and are "shared" between core and MS

This also serves as a bit of an example of what sharing components between MS and Core could look like. A couple of learnings/things to keep in mind for the future 
- It was fairly simple to refactor the components and move them into the common folder. The only thing I had to take care of was injection of the `renderConfig` which differs between the projects.
- I decided to keep completely separate stylesheets for core and minespace for the `SteppedForm` component, but think we could easily keep a shared stylesheet in the common folder as pretty much the only styling difference it needed were around colors
- I introduced a separate TailingsSummaryPage component to core, which is a minimal subset of the one in MS to get the basic info page working. I think it probably makes sense to make this one a shared a component as well as pretty much all of the logic applies in both places, but it fell out of scope for what I could achieve with this ticket. 
-  The custom prop types *could* (should?)  fairly easily be shared between the projects

image
image
image


services/minespace-web/common/components/Step.js
```diff
@@ -0,0 +1,16 @@
+// import React from 'react';
```
matbusby-fw: Can we remove this comment?

matbusby-fw (dismissed):
Awesome!  Great thoughts on the implementation of this as a common component.  I love the idea of having mostly shared style sheets with minimal overrides for colour and whatnot in the individual packages.

matbusby-fw (approved):

PR #2319, opened 2022-10-27
[MDS-4815] TSF Bugs

matbusby-fw:
[MDS-4815](https://bcmines.atlassian.net/browse/MDS-4815)

- set action columns on tsf tables as fixed so they don't disappear on smaller screens (as this is a wide table, this wasn't taking much)
- Updated styling of the nested dams tables in Core to match design
- Changed the Dams dto type for lat and long in the BE from fixed(2,7) to a float.  Fixed was causing zero values in these fields to be returned in notation (0e7)
- Updated TSF consequence classification on the TSF table in core to match the highest consequence of a child dam.

image
image
image


services/core-web/src/components/mine/Tailings/MineTailingsTable.js
```diff
@@ -38,6 +37,7 @@ const defaultProps = {};
 const MineTailingsTable = (props) => {
   const {
     TSFOperatingStatusCodeHash,
+    // eslint-disable-next-line no-unused-vars
     consequenceClassificationStatusCodeHash,
```
taraepp: Is this being used or not? Wondering why the eslint line was necessary.

matbusby-fw: Updated the PR to remove these for this page.

You'll find a lot more of these in the system than I'd like, but we can work on clearing them out as we go.

There are some interactions with Redux that seem to cause some props to be flagged as unused when they are not and I think often this is why those flags were set.

taraepp (approved):

PR #2320, opened 2022-10-28
[MDS-4821] Fixed reloading of TSFs on navigation

simensma-fresh:
## Objective 

[MDS-4821](https://bcmines.atlassian.net/browse/MDS-4821)

Fixed an issue where the TSF page in core would not reload when navigation changed, it would always display info of the first TSF you navigated to

### DIFF
--- a/services/core-web/src/components/mine/Tailings/MineTailingsTable.js
+++ b/services/core-web/src/components/mine/Tailings/MineTailingsTable.js
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
+import { withRouter } from "react-router-dom";
 import { EyeOutlined } from "@ant-design/icons";
 import { Button, Tooltip, Typography } from "antd";
 import {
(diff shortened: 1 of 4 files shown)


services/core-web/src/components/mine/Tailings/MineTailingsTable.js
```diff
@@ -237,4 +238,4 @@ const mapStateToProps = (state) => ({
   itrmExemptionStatusCodeHash: getITRBExemptionStatusCodeOptionsHash(state),
 });
 
-export default connect(mapStateToProps)(MineTailingsTable);
+export default withRouter(connect(mapStateToProps)(MineTailingsTable));
```
matbusby-fw, line 241: Interesting.  Was `connect` causing problems here?

matbusby-fw (approved):
LGTM!

services/core-web/src/components/mine/Tailings/MineTailingsTable.js
simensma-fresh: @matbusby-fw Nope! `withRouter` adds `history` as a prop to the component which is used when navigating to the TSF details

PR #2321, opened 2022-10-31
[MDS-3704] US Postal Codes not accepted

taraepp:
## Use appropriate validation for postal code depending on country (CAN/USA)

[MDS-3704](https://bcmines.atlassian.net/browse/MDS-3704)

- five-four zip codes were not accepted at all
- validation did not catch when US zip code entered for address with Canadian province or Canadian postal code for US state
- backend validation differed from frontend validation, making some validation errors show as a toast

Summary of changes:
- db migration: added address_type_code field to sub_division_code table to link prov/state to its country, linked existing entries to CAN/USA (ex: Alberta has address type code of CAN, California has USA)
- modified the return of of province dropdown options to include the country code as subType
- found/created separate US/CAN postal code regex patterns- put on both FE & BE
   - US pattern accepts formats: 12345, 123451234, 12345-1234
   - Canadian pattern accepts: A1A1A1, does not accept A1A-1A1 or A1A 1A1
- modified FE & BE to validate postal code based on country
- added address_type_code to data being parsed before creating address
- when there is no prov/state (and therefore country) entered: both BE & FE will assume Canadian and apply Canadian rules
- merge_resource functionality: include country in address merge

Didn't include:
- clean up of address_type_code field in address table- debatable if this field is still necessary
- modification to address_etl function
- did not change error message to specify correct postal code field name depending on country (this would be easy- is it desired? ex "invalid _zip_ code format" or "invalid _postal_ code format")

### DIFF
--- a/services/core-web/common/utils/Validate.js
+++ b/services/core-web/common/utils/Validate.js
@@ -60,9 +60,8 @@ class Validator {
   checkPostalCode(code, country = "CAN") {
     if (country === "USA") {
       return this.US_POSTAL_CODE_REGEX.test(code);
-    } 
-      return this.CAN_POSTAL_CODE_REGEX.test(code);
-    
+    }
+    return this.CAN_POSTAL_CODE_REGEX.test(code);
   }
 
   checkCurrency(number) {
--- a/services/core-web/common/utils/Validate.js
+++ b/services/core-web/common/utils/Validate.js
@@ -58,15 +58,11 @@ class Validator {
   }
 
   checkPostalCode(code, country = "CAN") {
-    switch (country) {
-      case "CAN":
-        return this.CAN_POSTAL_CODE_REGEX.test(code);
-      case "USA":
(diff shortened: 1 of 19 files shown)


services/core-web/common/utils/Validate.js
```diff
@@ -123,8 +128,14 @@ export const lonNegative = (value) =>
 export const phoneNumber = (value) =>
   value && !Validate.checkPhone(value) ? "Invalid phone number e.g. xxx-xxx-xxxx" : undefined;
 
-export const postalCode = (value) =>
-  value && !Validate.checkPostalCode(value) ? "Invalid postal code or zip code" : undefined;
+export const postalCode = (value, allValues, formProps) => {
```
matbusby-fw, line 131: 🔥

matbusby-fw (approved):
Awesome!  This looks super solid (and not quite as small of a ticket as we originally thought).  This will be a great enhancement for features down the road!

PR #2327, opened 2022-11-03
[MDS-4802] [MS] [Core] Reporter phone and email are required fields

BenHarrington:
## Objective 

[MDS-4802](https://bcmines.atlassian.net/browse/MDS-4802)

_Changes_

- Added "required" to form inputs
- Updated MineIncidentFactory to populate these new required fields

_Dev Notes_

- I did not update the model/table due to supporting existing legacy data where these fields are not required


JustinMacaulay (approved):
Looks good

PR #2329, opened 2022-11-03
[MDS-4846] - Core fetch NOD data

matbusby-fw:
## Objective 
Fix refreshing of mine data in core from causing an error

[MDS-4846](https://bcmines.atlassian.net/browse/MDS-4846)

- mine_guid was being incorrectly passed within an object to the `fetchNoticesOfDeparture` action creator ( ie. `{mineGuid: id}`).  Updated this to passing the parameter directly.

image


simensma-fresh (approved):

taraepp (approved):
👍

PR #2330, opened 2022-11-04
[MDS-4842] Added FILE_UPLOAD component to core render config

simensma-fresh:
## Objective 
This fixes the issue where navigating to the EOR tab in core results in a blank screen as the component tries to render the FileUpload component, which did not exist

[MDS-4842](https://bcmines.atlassian.net/browse/MDS-4842)


taraepp (approved):
👍

PR #2331, opened 2022-11-04
[MDS-4848] - TSF Operating Status

matbusby-fw:
## Objective 

[MDS-4848](https://bcmines.atlassian.net/browse/MDS-4848)

The `Closed` operating status no longer exists for TSFs, but previous TSFs may already have this assigned.  Added in logic to add it as an option when it already exists (to avoid showing "CLO" in the dropdown).

Also removed the code disabling the `Next` button in the stepped form if errors in the form validation were present, and instead properly triggered the validation messages on fields without navigating to the next page.

image

### DIFF
--- a/services/core-web/common/components/tailings/BasicInformation.js
+++ b/services/core-web/common/components/tailings/BasicInformation.js
@@ -5,7 +5,7 @@ import {
   TSF_INDEPENDENT_TAILINGS_REVIEW_BOARD,
   TSF_OPERATING_STATUS_CODE,
   TSF_TYPES,
-} from "@common/constants/strings";
+} from "@mds/common";
 import { Col, Row, Typography } from "antd";
 import React, { useEffect, useState } from "react";
 import {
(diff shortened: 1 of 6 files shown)


services/core-web/common/components/tailings/BasicInformation.js
```diff
@@ -141,10 +146,10 @@ export const BasicInformation = (props) => {
         id="tsf_operating_status_code"
         name="tsf_operating_status_code"
         label="Operating Status"
-        data={TSF_OPERATING_STATUS_CODE}
+        data={includeClosedStatus}
```
taraepp: I don't like this variable name, it sounds like a boolean.

taraepp (commented):
Nit about the variable name includeClosedStatus (is it intended to be all available status options?).
What happens when there are errors on the form and the user clicks the next button? Do they advance to the next page with errors on the first?

services/core-web/common/components/tailings/BasicInformation.js
matbusby-fw: Fair enough, I see your point.  How about `parsedStatusCodeOptions` or something similar?

taraepp (approved):
👍

PR #2333, opened 2022-11-07
[MDS-4819] [MS] Update Incident homepage content

BenHarrington:
## Objective 

[MDS-4819](https://bcmines.atlassian.net/browse/MDS-4819)

- Updating copy and styles


areyeslo (approved):
Looks great Ben!! 🚀

PR #2334, opened 2022-11-07
[MDS-3704] Fix backend bug when there is no postal code submitted

taraepp:
## Objective 
Bug fix- validation was being run on the postal code when the postal code was NoneType. Added a null check, now it accepts no postal code as valid postal code.

[MDS-3704](https://bcmines.atlassian.net/browse/MDS-3704)


matbusby-fw (approved):
👍

PR #2336, opened 2022-11-08
[FIX]updating to sass and image-webpack-loader

JustinMacaulay:
## Objective 

Updates our sass package from node-sass to sass as well as upgrading the version of our image-webpack-loader which wasnt working with sass.


BenHarrington (commented):
Do we want to lock down those sass/image loader versions without the caret ^ like was done with all other dependencies?

services/common/package.json
```diff
@@ -31,7 +31,7 @@
     "copy-webpack-plugin": "5.1.1",
     "hard-source-webpack-plugin": "0.13.1",
     "html-webpack-plugin": "3.2.0",
-    "image-webpack-loader": "5.0.0",
+    "image-webpack-loader": "^8.1.0",
```
BenHarrington: Still has a caret

BenHarrington (approved):

PR #2338, opened 2022-11-08
[MDS-4856] Fix postal code error on NoW pages

simensma-fresh:
## Objective 

[MDS-4856](https://bcmines.atlassian.net/browse/MDS-4856)

Relaxed the postal code validation of an address to fix errors that have been popping up where the validation fails when you do not know the address_type_code of an address (as in the case of some MMS applications). This still keeps support for both Canadian and US postal codes

### DIFF
--- a/services/core-api/app/api/parties/party/models/address.py
+++ b/services/core-api/app/api/parties/party/models/address.py
@@ -87,17 +87,15 @@ class Address(SoftDeleteMixin, AuditMixin, Base):
         return address
 
     # will be called for both fields, in the order defined in model, 1st call will be missing 2nd value
-    @validates('post_code', 'address_type_code')
-    def validate_address_code(self, key, value):
-        if key == 'address_type_code' and self.post_code is not None:
-            maxLength = 6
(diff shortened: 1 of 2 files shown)


services/core-api/app/api/parties/party/models/address.py
```diff
@@ -87,17 +87,15 @@ def create(cls,
...
-            if len(self.post_code) > maxLength:
-                raise AssertionError(f'post_code must not exceed {maxLength} characters.')
-            if not validPostalCode.match(self.post_code):
-                current_app.logger.error(f'Failed post_code validation for address {self.address_id}. post_code: {self.post_code}, address_type_code: {self.address_type_code}, value: {value}',)
-                raise AssertionError(f'Invalid post_code format.')
-        return value
+    @validates('post_code')
+    def validate_address_code(self, key, post_code):
+        if post_code and len(post_code) > 10:
+            raise AssertionError('post_code must not exceed 6 characters.')
```
matbusby-fw, line 93: Since we're just checking the length once for both Canadian and US, maybe we should make the error more generic.

Something like "post_code exceeds the maximum allowed characters"?

matbusby-fw:
```suggestion
            raise AssertionError('post_code exceeds the maximum allowable characters.')
```

matbusby-fw (dismissed):

matbusby-fw (dismissed):

PR #2339, opened 2022-11-09
[MDS-4633] Added EoR history table to MS and Core

simensma-fresh:
## Objective 

[MDS-4633](https://bcmines.atlassian.net/browse/MDS-4633)
[MDS-4608](https://bcmines.atlassian.net/browse/MDS-4608)

Added Historical EoR tables in MS and core.

Note: I ended up adding the new component to the old common folder as to not having to deal with the `antd` error that popped up in the new package. Might pick that back up towards the end of the sprint (time allowing)

image
image


matbusby-fw (approved):
👍

PR #2341, opened 2022-11-14
[MDS-4859] Filter out empty mine_documents

simensma-fresh:
## Objective 

[MDS-4859](https://bcmines.atlassian.net/browse/MDS-4859)

Hide NoW documents that don't have an associated `mine_document_guid`. This happens when a NoW document is soft-deleted. The document reference still exists, and does not respect the deleted flag


matbusby-fw (approved):
Works for me!

PR #2342, opened 2022-11-14
[MDS-4781] Remove all project table Feature Flag

JustinMacaulay:
## Objective 

Removes the feature flagging for the all projects view table.

[MDS-4781](https://bcmines.atlassian.net/browse/MDS-4781)


BenHarrington (approved):
Looks good

PR #2343, opened 2022-11-14
[MDS-4859] Added soft-delete support for now_document_xref

simensma-fresh:
## Objective 

[MDS-4859](https://bcmines.atlassian.net/browse/MDS-4859)
This PR fixes the issue where deleted documents causes errors when modifying consultation packages/permit packages in core. 
Added soft delete functionality to the document xref to match the document itself so we can properly exclude it when deleted


matbusby-fw (approved):
🤞

PR #2347, opened 2022-11-15
[MDS-4038] Drop down styling 2

taraepp:
## Objective 
Redoing previous PR that has since been reverted.

[MDS-4038](https://bcmines.atlassian.net/browse/MDS-4038)

Fixed additional bug created where certain instances of Select boxes that had their option list rendering outside of the parent were not showing correctly, by attaching the list to its parent.
Otherwise the same as https://github.com/bcgov/mds/pull/2332


matbusby-fw (approved):
Looks solid to me!

simensma-fresh (approved):
🎨

PR #2348, opened 2022-11-15
[MDS-4806] [MS] [Core] Has an inspector already been informed of the reportable incident?

BenHarrington:
## Objective 

[MDS-4806](https://bcmines.atlassian.net/browse/MDS-4806)

image

_Changes_

- Created migration to add new fields
- Updated entity model, resource, and response models
- Updated frontend components/forms to use new fields
- Added conditional validation for contacted, contact methods fields


services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -169,6 +177,11 @@ def post(self, mine_guid):
                     )
                 incident.dangerous_occurrence_subparagraphs.append(sub)
 
+        reported_to_inspector_party = Party.query.filter_by(
+            party_guid=data.get('reported_to_inspector_party_guid')).first()
+        if reported_to_inspector_party and 'INS' in reported_to_inspector_party.business_roles_codes:
```
JustinMacaulay, line 182: What is the fail case for this? is it silent fail?

JustinMacaulay (commented):
Just one comment but looks good otherwise.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
BenHarrington: That snippet used to be in this block:

`if is_minespace_user() is not True:
            incident.followup_inspection = data.get('followup_inspection')
            incident.followup_inspection_date = data.get('followup_inspection_date')

            # lookup and validated inspector party relationships
            tmp_party = Party.query.filter_by(
                party_guid=data.get('reported_to_inspector_party_guid')).first()
            if tmp_party and 'INS' in tmp_party.business_roles_codes:
                incident.reported_to_inspector_party_guid = tmp_party.party_guid
            tmp_party = Party.query.filter_by(
                party_guid=data.get('responsible_inspector_party_guid')).first()
            if tmp_party and 'INS' in tmp_party.business_roles_codes:
                incident.responsible_inspector_party_guid = tmp_party.party_guid
            tmp_party = Party.query.filter_by(
                party_guid=data.get('determination_inspector_party_guid')).first()
            if tmp_party and 'INS' in tmp_party.business_roles_codes:
                incident.determination_inspector_party_guid = tmp_party.party_guid`

I just pulled it out of that block and renamed the temporary party variable. I do not believe there ever was any additional handling so I imagine it is silent.

JustinMacaulay: K, might be something we want to revisit at some point

JustinMacaulay (approved):
Looks good

PR #2352, opened 2022-11-18
[MDS-4823] New incident submission email

JustinMacaulay:
## Objective 

Changes the email flow to use HTML templates and adds a new function to the email service to use the CHES emailMerge endpoint with can populate string templates.

[MDS-4823](https://bcmines.atlassian.net/browse/MDS-4823)

Dark Mode:

image

image

Light Mode:

image

image


services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -106,8 +105,7 @@ def post(self, mine_guid):
                         'Dangerous occurrences require one or more cited sections of HSRC code 1.7.3'
                     )
 
-        reported_timestamp_default = datetime.utcnow(
-        ) if not data['reported_timestamp'] and is_minespace_user() else data['reported_timestamp']
+        reported_timestamp_default = data.get('reported_timestamp', datetime.utcnow())
```
BenHarrington: I was wrong about this when we spoke about it before, you were doing this the right way the first time: `reported_timestamp_default = datetime.utcnow(
        ) if not data['reported_timestamp'] else data['reported_timestamp']` I think should be fine.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -248,16 +247,48 @@ def validate_reported_timestamp(self, key, reported_timestamp):
...
-
-        link = f'{Config.CORE_PRODUCTION_URL}/mine-dashboard/{self.mine.mine_guid}/oversight/incidents-and-investigations'
-        body += f'<p>View updates in Core: <a href="{link}" target="_blank">{link}</a></p>'
-        EmailService.send_email(subject, recipients, body)
+        emli_recipients = [INCIDENTS_EMAIL]
+        cc = [MDS_EMAIL]
+        minespace_recipients = [self.reported_by_email]
+        duration = self.reported_timestamp - self.incident_timestamp
+        duration_in_s = duration.total_seconds()
+        days = divmod(duration_in_s, 86400)
```
BenHarrington, line 268: I think there is a timedelta.days, but this looks to get to the same result.

JustinMacaulay: That works if I only want days but I need days and hours. I used divmod cause it gets the days and the remainder of the time is then used to get hours.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -248,16 +247,48 @@ def validate_reported_timestamp(self, key, reported_timestamp):
...
+        hours = divmod(days[1], 3600)
+
+        emli_body = open("app/templates/email/incident/emli_incident_email.html", "r").read()
+        minespace_body = open("app/templates/email/incident/minespace_incident_email.html", "r").read()
```
areyeslo, line 272: Thinking this is going to be reuse lots of times. Do you think we can include email path in config.py?

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
JustinMacaulay: merged yours in and dealt with this.

BenHarrington (approved):

PR #2353, opened 2022-11-18
[MDS-4901] - Setup Github Codespaces

matbusby-fw:
## Objective - Create a default configuration for spinning up an MDS Github Codespace

[MDS-4901](https://bcmines.atlassian.net/browse/MDS-4901)

- Created a devcontainer.json with configuration settings for MDS
  - Installs all dependencies of the project in a Debian environment
  - Forwards all necessary ports to local
  - Generates a `.env` file on creation
  - After creation runs `make be && make seeddb && yarn install`
  - Does not currently serve any of the frontend packages.  This can be added if we want, or left to be run as normal.
  - Seems to fail at setting up celery.  This isn't crucial to most local development, but will look into this further
- Updated `senenv.sh` to accept an optional `INPUT` variable and skip the question if it is provided


taraepp (approved):
👍

PR #2357, opened 2022-11-22
[MDS-4867] Mine Incident Re-categorization - DB/BE

BenHarrington:
## Objective 

[MDS-4867](https://bcmines.atlassian.net/browse/MDS-4867)
[MDS-4890](https://bcmines.atlassian.net/browse/MDS-4890)

_Changes_

- Added "is_historic" and "parent_mine_incident_category_code" field to DB/models(mine_incident_category_code)
- Updated existing category codes to historic
- Adjusted static content endpoint to return only "is_historic" records until FE work has begun
- Created new categories with parent/child relationships


JustinMacaulay (approved):
Looks good to me

PR #2359, opened 2022-11-23
[MDS-4891] Send notifications when an EOR is expiring/has expired

simensma-fresh:
## Objective 

Added functionality to trigger notifications for EORs that
- Expire in 60 days
- Have already expired

To enable this, the following changes have been made:
- Introduced an explicit `status` column to MinePartyAppointment (active | inactive | pending) to replace the current dynamic deduction. PRs will follow from @matbusby-fw / me to actually make sure these statuses are set properly
- Added an `idempotency_key` to ActivityNotification that can be used to ensure that a notification is only sent once for a user, a notification with the same `idempotency_key` will only be sent once even on subsequent retries. This enables us to safely retry the notification sending jobs in case it fails halfway through

*Note:* This does not include work around actually triggering the job on a regular basis, a separate PR will be made for that.


taraepp (dismissed):

taraepp (dismissed):

taraepp (dismissed):
👍

taraepp (dismissed):

taraepp (approved):

PR #2360, opened 2022-11-23
[MDS-4911] Enables Editing of TSFs in core using the new TSF pages

simensma-fresh:
## Objective 

[MDS-4911](https://bcmines.atlassian.net/browse/MDS-4911)

- Enables Editing of TSFs in core using the new TSF pages so you can fix any issues 
- Make a newly created party appear first in dropdown list of search results.

image
image


taraepp (approved):
👍

PR #2361, opened 2022-11-23
[MDS-4832] [Core] Connect change to Major Mine Application Status to Alert feed

BenHarrington:
## Objective 

[MDS-4832](https://bcmines.atlassian.net/browse/MDS-4832)

_Changes_

- Modified activity_notification to handle "project" metadata
- Added notifications for Project Summary, IRT, MMA submissions


services/core-web/src/components/mine/Projects/Project.js
```diff
@@ -45,6 +45,16 @@ export class Project extends Component {
     this.handleScroll();
   }
 
+  // Handle clicking an activity notification link(MMA tab) while Project component is mounted
+  componentDidUpdate(prevProps) {
+    const currentTab = this.props.match.params.tab;
```
JustinMacaulay, line 50: is this specific to the project page in that the link from the notification drawer wasn't working?

JustinMacaulay (commented):
Just the one comment on the project page

services/core-web/src/components/mine/Projects/Project.js
BenHarrington: What was happening is the URL gets updated to the correct one using the link however since this page is in a tab the activeTab(local state) never gets refreshed since the component is still mounted. I'm not super happy doing it this way but no other clean options presented themselves.

JustinMacaulay (approved):
Looks good

PR #2365, opened 2022-11-25
[MDS-4895] Notification when TSF has new EoR or QP

taraepp:
## Objective 

[MDS-4895](https://bcmines.atlassian.net/browse/MDS-4895)

- when a TSF has a new EoR or QP, notify users subscribed to the mine
- clicking on the notification directs user to view the TSF, and the tab of the new person

- should be resolved first:
  - my current user has the permissions to view the page the notification directs them to, but they _do not_ have the permissions to navigate to the same page through the UI
  - my suggestion is to change the condition that shows the "View" button for the TSF on the mine page


simensma-fresh (approved):
Looks great 🙌

PR #2371, opened 2022-11-28
[MDS-4911] Fixed EoR / QFP validation in TSF flow when start/end dates are disabled

simensma-fresh:
## Objective 

[MDS-4911](https://bcmines.atlassian.net/browse/MDS-4911)

Fixed EoR / QFP validation in TSF flow when start/end dates are disabled, they now don't prevent you from hitting next if the fields are disabled.

Fixed an issue where you wouldn't see a validation message when the file upload component failed validation. Added a file upload wrapper components to make them work nicely with redux-form. 

_Why are you making this change? Provide a short explanation and/or

image

image


taraepp (dismissed):
👍

PR #2372, opened 2022-11-28
[MDS-4905] Tooltip "floating": disable background scroll

taraepp:
## Objective 

[MDS-4905](https://bcmines.atlassian.net/browse/MDS-4905)

- a rule was once added to set popover position to fixed
- this was because tooltips within modals (ex: are you sure you want to close? type popover) would move their positions around a lot without the fixed rule when the background was scrolled, even off of the page
- this problem, and also issues with option lists within select, is resolved when the background is disabled on the <body> when a modal is visible
- enabling me to remove the position:fixed rule that was interfering in non-modal contexts
- the popover on a modal still wiggles a little bit because the modal is a little too tall for the page (same behaviour as before, doesn't affect functionality)

### DIFF
--- a/services/core-web/src/components/common/wrappers/ModalWrapper.js
+++ b/services/core-web/src/components/common/wrappers/ModalWrapper.js
@@ -51,6 +51,7 @@ const ModalWrapper = (props) => {
     propCloseModal();
   };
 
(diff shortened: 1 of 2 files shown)


simensma-fresh (approved):
📜

PR #2375, opened 2022-11-29
[MDS-4297: OHSC Worker & Management Representative dates have to be after the incident

areyeslo:
## Objective 

[MDS-4297](https://bcmines.atlassian.net/browse/MDS-4817)

- Add validation for Inspectors, OHSC, Local Union or Worker Representatives dates with respect to incident date.

Time validation:
image

image

Date Validation:
image

### DIFF
--- a/services/core-web/common/utils/Validate.js
+++ b/services/core-web/common/utils/Validate.js
@@ -172,13 +172,12 @@ export const dateNotBeforeStrictOther = memoize((other) => (value) =>
 );
 
 export const timeNotBeforeOther = memoize(
-  (comparableDate, baseDate, baseTime) => (comparableTime) =>
-    baseTime &&
-    baseDate &&
-    comparableDate &&
-    comparableTime &&
-    baseDate === comparableDate &&
-    comparableTime < baseTime
+  (comparableDate, baseDate, baseTime) => (comparableTime) => baseTime &&
+      baseDate &&
+      comparableDate &&
+      comparableTime &&
+      baseDate === comparableDate &&
+      comparableTime < baseTime
       ? `Time cannot be before ${baseTime.format("H:mm")} hrs.`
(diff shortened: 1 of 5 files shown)


services/core-web/common/utils/Validate.js
```diff
@@ -167,6 +167,22 @@ export const dateNotBeforeOther = memoize((other) => (value) =>
...
+  value && other && new Date(value) < new Date(other) ? `Date cannot be before ${other}` : undefined
+);
+
+export const timeNotBeforeOther = memoize(
+  (comparableDate, baseDate, baseTime) => (comparableTime) =>
+    comparableDate &&
+    comparableTime &&
+    baseDate &&
+    baseTime &&
+    new Date(baseDate) === new Date(comparableDate) &&
```
JustinMacaulay: for this to work as intended the following line 180 and 181 need to be a single clause cause it only maters if the the time is before if the dates match

services/minespace-web/common/utils/Validate.js
```diff
@@ -167,6 +167,22 @@ export const dateNotBeforeOther = memoize((other) => (value) =>
...
+  value && other && new Date(value) < new Date(other) ? `Date cannot be before ${other}` : undefined
+);
+
+export const timeNotBeforeOther = memoize(
+  (comparableDate, baseDate, baseTime) => (comparableTime) =>
+    comparableDate &&
+    comparableTime &&
+    baseDate &&
+    baseTime &&
+    new Date(baseDate) === new Date(comparableDate) &&
```
JustinMacaulay: same here

JustinMacaulay (commented):
Just the one comment really everything else looks good.

JustinMacaulay (approved):
Looks good

PR #2376, opened 2022-11-30
[MDS-4610] Acknowledgement of EORs in Core

simensma-fresh:
## Objective 

[MDS-4610](https://bcmines.atlassian.net/browse/MDS-4610)

Added functionality for acknowledging EORs in core.
This extends the current TSF assignment flow, where an EOR assigned through MineSpace starts out with a status of 'Pending', after which ministry staff has to "Acknowledge" the EOR change in Core before it becomes the active one.
In order to make this happen, a couple of notable changes:
- Added an explicit `status` column for EORs instead of the current derived status AND migration to backfill these statuses
- Added a `mine_party_acknowledgement_status` column
- Added the ability for Core users to change the status of an EOR
- Changed the redux config for TSFs a bit to make the PartyAppointment table work better with `redux-form`


taraepp (approved):
👍

taraepp (approved):
👍

PR #2379, opened 2022-11-30
[MDS-4732] Fix error when assigning party through core

simensma-fresh:
## Objective 

[MDS-4732](https://bcmines.atlassian.net/browse/MDS-4732)

Fixes and error that pops up when trying to add a new contact in core, preventing the contact from being created -introduced with the eor/qfp changes.

image


taraepp (approved):
👍

PR #2380, opened 2022-12-01
[MDS-4831] Husky hooks with new common setup

taraepp:
## Objective 
- get our commits linted and tested when committing like before

[MDS-4831](https://bcmines.atlassian.net/browse/MDS-4831)
Bonus:
[MDS-4790](https://bcmines.atlassian.net/browse/MDS-4790)

"pre-commit":
- triggers lint-staged
- will lint file types: js, css, json, md
**- do we want to add scss, py?**
image

"pre-push":
- triggers lint-prepush
- for js files will run related tests
image

Issues:
- eslint couldn't resolve the @common imports and it was easier to disable the rule (failed imports will at least be caught by tests)
- don't think I caught everything that should be removed in the nested package.json files


.eslintrc.js
```diff
@@ -9,13 +9,7 @@ module.exports = {
...
   },
   plugins: ["jest"],
-  extends: [
-    "eslint:recommended",
-    "plugin:react/recommended",
-    "airbnb",
-    "prettier",
-    "prettier/react",
-  ],
+  extends: ["eslint:recommended", "plugin:react/recommended", "airbnb", "prettier"],
```
simensma-fresh, line 12: intensional removal of `prettier/react`? 🙃

simensma-fresh (approved):

.eslintrc.js
taraepp: It was! It's been merged into prettier, I removed it when I got an error, I think when I updated prettier.

simensma-fresh: Oh cool! Makes sense 👍

vyasworks (approved):
Looks good !

PR #2383, opened 2022-12-08
[MDS-4949] In Minespace reviewing a submitted incidents report

areyeslo:
## Objective 

[MDS-4949(https://bcmines.atlassian.net/browse/MDS-4949)

- Passing in payload reported_timestamp, johsc_worker_rep_contact_timestamp johsc_management_rep_contact_timestamp values when the incident is created.

Incident Review Page with dates not saved (error):
image

Incident Review Page with dates saved (fix):
image


BenHarrington (approved):
Good catch Arturo!

PR #2384, opened 2022-12-08
[MDS-4891] Fix error when running job through celery

simensma-fresh:
## Objective 

Fixes an error popping up when running jobs that inserts records into tables that autofills the `updated_by` columns. These cases look up the username of the person form the Auth token, which in the case of Celery jobs will not exist as it is run asynchronously. In this case, we set the `updated_by` to `system`

image


taraepp (approved):
👍

PR #2390, opened 2022-12-16
[MDS-4980] Codespaces Makefile changes

taraepp:
## Objective 
- make running codespaces a little smoother

[MDS-4980](https://bcmines.atlassian.net/browse/MDS-4980)

- project has to be re-initialized if you ever make a new codespace, can use "make init" now
- should also work in a regular local environment
- codespaces shuts down more frequently to preserve resources: added "make restart" command to rebuild backend and serve up frontend
   - it "should" work to run ```make be``` while your containers are running to rebuild them, but when the CS automatically pauses, I have found **that this is not the case**. In this scenario, ```make stop``` must be run before ```make be``` (or ```make restart```)
   - in order for the command to also be useful for local dev, I have omitted ```make stop``` from restart

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -5,13 +5,13 @@
   "recommendations": [
     "aaron-bond.better-comments",
     "editorconfig.editorconfig",
-    "coenraads.bracket-pair-colorizer",
     "eamodio.gitlens",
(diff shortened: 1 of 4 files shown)


matbusby-fw (approved):
Wooo!  Codespaces!

PR #2393, opened 2022-12-19
[MDS-4926] [MS] Major Project Submission Spatial File Upload failing for XML

BenHarrington:
## Objective 

[MDS-4926](https://bcmines.atlassian.net/browse/MDS-4926)


JustinMacaulay (approved):
Looks good

PR #2395, opened 2022-12-20
[MDS-4960] Fixed updating status of EORs through core

simensma-fresh:
## Objective 

[MDS-4960](https://bcmines.atlassian.net/browse/MDS-4960)

Fixed an issue where an EoR created through core would not have a status assigned. The issue was that the flag to end the term of the previous EoR was not passed along correctly, preventing the status update from happening. In addition to that, I uncovered a couple of errors to fix


taraepp (approved):
👍

PR #2396, opened 2022-12-20
[MDS-4227] Inspector filter on the global incidents table doesn't work

BenHarrington:
## Objective 

[MDS-4227](https://bcmines.atlassian.net/browse/MDS-4227)


services/core-web/src/components/mine/Incidents/MineIncidentTable.js
```diff
@@ -221,11 +221,14 @@ export class MineIncidentTable extends Component {
...
         ),
-        onFilter: (value, record) => record.incident.responsible_inspector_party === value,
+        onFilter: (value, record) => record.incident?.responsible_inspector_party === value,
         filters: _.reduce(
           this.props.incidents,
           (reporterList, incident) => {
-            if (!reporterList.map((x) => x.value).includes(incident.responsible_inspector_party)) {
+            if (
+              !reporterList.map((x) => x.value).includes(incident.responsible_inspector_party) &&
+              incident?.responsible_inspector_party !== null
```
JustinMacaulay: If the second one has a null safe check shouldn't the first one as well

JustinMacaulay (commented):
Just one comment on a null safe check

JustinMacaulay (dismissed):
Looks good

JustinMacaulay (approved):
re-approve

PR #2397, opened 2022-12-23
[MDS-4891] Fixed expiring / expired notifications sending one day early

simensma-fresh:
## Objective 

Fixed an issue where the 60 day expiry / expired alerts for EoRs were triggering one day early compared to the alerts in the app themselves. 


taraepp (approved):
👍

PR #2400, opened 2023-01-06
[MDS-4883] consequence classification order

taraepp:
## Objective 

[MDS-4883](https://bcmines.atlassian.net/browse/MDS-4883)

- update test data, enums, any other defined constants of consequence classification status
- migrate any instances of "No Dam" to "Not Rated"
- modify the display order of consequence classification statuses
- remove the "after migrate" sql INSERT & UPDATE that was undoing changes to consequence classification


matbusby-fw (approved):
👍

PR #2402, opened 2023-01-09
[MDS-4933] HSRC conditions

matbusby-fw:
## Objective 

[MDS-4933](https://bcmines.atlassian.net/browse/MDS-4933)

- Update records in the `compliance_article` table of the database
- Load expired articles, but append `(Repealed)` to those entries
- Remove repealed articles from the options when creating a new variance on Minespace

image
image
image

### DIFF
--- a/services/minespace-web/common/selectors/staticContentSelectors.js
+++ b/services/minespace-web/common/selectors/staticContentSelectors.js
@@ -96,10 +96,15 @@ export const getIncidentFollowupActionOptions = (state, isShowActiveOnly) =>
     isShowActiveOnly
   );
 
-// removes all expired compliance codes from the array
-export const getCurrentComplianceCodes = createSelector([getComplianceCodes], (codes) =>
-  codes.filter((code) => code.expiry_date === null || new Date(code.expiry_date) > new Date())
-);
+// marks all expired compliance codes as 'Repealed'
+export const getCurrentComplianceCodes = createSelector([getComplianceCodes], (codes) => {
+  return codes.map((code) => {
+    if (new Date(code?.expiry_date) < new Date()) {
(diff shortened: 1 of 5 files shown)


services/minespace-web/common/selectors/staticContentSelectors.js
```diff
@@ -96,10 +96,15 @@ export const getIncidentFollowupActionOptions = (state, isShowActiveOnly) =>
...
 
-// removes all expired compliance codes from the array
-export const getCurrentComplianceCodes = createSelector([getComplianceCodes], (codes) =>
-  codes.filter((code) => code.expiry_date === null || new Date(code.expiry_date) > new Date())
-);
+// marks all expired compliance codes as 'Repealed'
+export const getCurrentComplianceCodes = createSelector([getComplianceCodes], (codes) => {
+  return codes.map((code) => {
+    if (new Date(code?.expiry_date) < new Date()) {
+      code.description = `${code.description} (Repealed)`;
```
taraepp, line 103: Could we have added a boolean property so we're not reliant on substrings?

taraepp (commented):
The frontend validation _should_ be sufficient to prevent repealed sections from being submitted with variances, but a backend check would also be nice.

services/minespace-web/common/selectors/staticContentSelectors.js
matbusby-fw: I thought about that, and it's possible.  
The snag I found there that I didn't like was that the data is eventually being transformed into label/value options before being passed into the table, so they would no longer contain that boolean (originally I thought of just considering the `expiry_date` in the table to conditionally render the `(Repealed)` text).

taraepp (approved):
👍

PR #2403, opened 2023-01-10
[CHORE] - Codespaces setup

matbusby-fw:
## Objective 

Added devcontainer.json file to base folder with project specific setup.

updated setenv.sh file to accept `INPUT` parameter so `make INPUT=true env` can be used to auto run the script and bypass the confirmation check.

With this in place, spinning up a codespace will set up the entire environment and be ready to go.
It will:

- forward all necessary ports to local
- install all requirements for the application to run (node, python etc.)
- On initial creation it will (it will not run these commands again unless the container is rebuilt)
    - run `make env`
    - run `make be`
    - run `make seeddb`
    - run `yarn install`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,49 +1,4 @@
 {
-	"name": "Debian",
-	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
-	"features": {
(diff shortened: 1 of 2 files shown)


taraepp (approved):
Nice

PR #2404, opened 2023-01-12
[CHORE] removed devcontainer

matbusby-fw:
## Objective 

devcontainer.json seemed to be causing issues with celery within codespaces (could not find the path to ./celery.sh)
Removed the devcontainer for now.


taraepp (approved):
🤞

PR #2406, opened 2023-01-16
[MDS-4989] NOW Consultation page not loading

taraepp:
## Objective 

[MDS-4989](https://bcmines.atlassian.net/browse/MDS-4989)
Bug where specific notice of work consultation & referral pages were not loading. I believe it's related to deleted documents as part of the NoW application. Making front-end not break if bad data comes through, but back-end changes to not send back the data in the first place is probably the next step.

### DIFF
--- a/services/core-web/common/utils/helpers.js
+++ b/services/core-web/common/utils/helpers.js
@@ -188,7 +188,7 @@ export const normalizeExt = (value) => (value ? value.slice(0, 6) : value);
 
 export const upperCase = (value) => value && value.toUpperCase();
 
-export const truncateFilename = (filename, max = 40) => {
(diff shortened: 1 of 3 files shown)


services/core-web/common/utils/helpers.js
```diff
@@ -188,7 +188,7 @@ export const normalizeExt = (value) => (value ? value.slice(0, 6) : value);
 
 export const upperCase = (value) => value && value.toUpperCase();
 
-export const truncateFilename = (filename, max = 40) => {
+export const truncateFilename = (filename = "", max = 40) => {
```
matbusby-fw, line 191: Nice solution!

matbusby-fw (approved):

PR #2408, opened 2023-01-18
[MDS-5000] Celery Log Level

matbusby-fw:
## Objective 

Changed Celery log level to debug to get more info on error


taraepp (approved):
👍

PR #2410, opened 2023-01-18
[MDS-5000] - updates to celery worker command

matbusby-fw:
## Objective 

### DIFF
--- a/services/core-api/celery.sh
+++ b/services/core-api/celery.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
-cd /app
+# -n is the number of tasks to consume
+# -A is the name of the app to run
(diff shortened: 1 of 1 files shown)


services/core-api/celery.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
-cd /app
+# -n is the number of tasks to consume
+# -A is the name of the app to run
+# -Q is the name of the queue to consume from
+# -concurrency is the number of child processes processing the queue
+# -B is the Beat
+# --scheduler is the scheduler class to use
+# -E Enable sending task-related events that can be captured by monitors
```
taraepp, line 9: I like this handy reference

taraepp (approved):
🤞

PR #2411, opened 2023-01-20
[MDS-5000] CELERY!! ✊ ✊ 

matbusby-fw:
## Objective 

Explicitly added the folders for both the pid file and the schedule file in the `celery.sh` command  


taraepp (approved):
❗

PR #2412, opened 2023-01-20
[MDS-5000] Celery - redbeat lock key

taraepp:
## Objective 

I think they're both trying to acquire the same lock- adding in a lock prefix that should be different from the default (redbeat).


matbusby-fw (approved):
🤞

PR #2413, opened 2023-01-23
[MDS-4992] Manual trigger for scheduled notifications

taraepp:
## Objective 
Make testing easier by being able to trigger scheduled notifications from the command line

[MDS-4992](https://bcmines.atlassian.net/browse/MDS-4992)

- Shell into backend with `flask <command>`


matbusby-fw (approved):
Awesome.  Nice thinking!

PR #2415, opened 2023-02-02
[MDS-4956] incidents tables

matbusby-fw:
## Objective 

Add server side handling of pagination/filters/sorting for the incidents tables.  Previously, all records were being fetched each time and causing pod crashes in some instances.

- Refactored various incident components from class to functional
- Universally utilized the incidents endpoint optimized for pagination etc. 
- Wired the flow for server side pagination from:
   - The incidents table within a mine on Minespace
   - The incidents table from the browse menu on Core
   - The incidents table from the mine menu on Core
- ~~Disabled some of the exports from the common package as they were causing eslint duplicate export errors~~ This ended up causing errors, so I restored them
- Added common pagination utils to the common package

image
image


services/common/src/constants/index.js
```diff
@@ -1,6 +1,13 @@
+/** This is the source of the eslint import errors.
+ * These files contain many exports with identical names.
+ * This will need some cleaning, but the commented out files are
+ * not currently being imported anywhere, so this will at least
+ * stop the errors for now.
+ */
```
taraepp, line 7: comment out of date?

services/common/src/constants/utils.js
```diff
@@ -0,0 +1,26 @@
...
+export const parseServerSideSearchOptions = (pagination, filters, sorter) => {
+  Object.keys(filters).forEach((key) => !filters[key] && delete filters[key]);
+  return {
+    per_page: pagination.pageSize,
+    page: pagination?.current,
+    sort_field: sorter?.column?.dataIndex || undefined,
+    sort_dir: parseSortDir(sorter.order),
+    ...filters,
+  };
+};
```
taraepp, line 26: Nice- this looks like it could be used all over the place!

taraepp (commented):
I think that one comment I pointed out should be revised but otherwise LGTM

services/common/src/constants/index.js
matbusby-fw: Good catch.  Updated!

services/common/src/constants/utils.js
matbusby-fw: Indeed!  Tried to make it as reusable as possible.

taraepp (approved):
👍

PR #2416, opened 2023-02-03
[MDS-5037] EOR/QP appointment

matbusby-fw:
[MDS-5037](https://bcmines.atlassian.net/browse/MDS-5037)

A check was in place in the API before making a mine party appointment to make sure that if the user was a Minespace user the `party` was already assigned to that mine somewhere else.  This was designed to prevent Minespace user access to the entire user base.  This didn't work though for a newly created user, so I added a check before this to see if the creating user of the party was the same as the current user.

Also fixed the updating of Info on already assigned EORs and QPs.  Permission to edit these was lost with new SSO, so created logic to check if the current user has access to any mines that the party is appointed to.

### DIFF
--- a/services/core-api/app/flask_jwt_oidc_local/jwt_manager.py
+++ b/services/core-api/app/flask_jwt_oidc_local/jwt_manager.py
@@ -166,7 +166,7 @@ class JwtManager:  # pylint: disable=too-many-instance-attributes
             return True
         return False
 
-    def get_user_name(self):
+    def get_bceid_user_name(self):
         """Get the bceid_username from the token."""
         token = self.get_token_auth_header()
         unverified_claims = jwt.get_unverified_claims(token)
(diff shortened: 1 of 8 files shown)


services/core-api/app/flask_jwt_oidc_local/jwt_manager.py
```diff
@@ -166,6 +166,12 @@ def contains_role(self, roles):
             return True
         return False
 
+    def get_user_name(self):
```
taraepp: What would happen here if this was run for an idir user?
If it's just intended for bceid I'm fine with renaming, otherwise I'd expect this function to work for either, especially because it's a bit obscured by username() in access_decorators.

taraepp (commented):
There's a test failing that should be pretty quick to fix, and a function I'm not so sure about.

services/core-api/app/flask_jwt_oidc_local/jwt_manager.py
matbusby-fw: This is only intended to be used for bceid since it just references the `bceid_username` within the token.  I think renaming it makes sense.

access_decorators was the most appropriate place I could find for this to live since it's pulling info from the JWT, but with the rename it should be clearer.

taraepp (approved):
👍

PR #2417, opened 2023-02-06
[MDS-4956] decrease default per page count from 25 to 10

matbusby-fw:
## Objective 

Certain large queries are timing out with the base page count of 25 and failing.  I think we'll need to look into those queries to see if we can make them more efficient, but this should get things running a little smoother in the short term.


taraepp (dismissed):
👍

taraepp (approved):

PR #2418, opened 2023-02-06
[MDS-4956] - Incidents optimizations

matbusby-fw:
## Objective 

- Cleaned up repeat calls to incidents in Core
- Added table loading to incidents for a mine in core
- Changed the incidents query to only do a join on Mine when it is actually needed


taraepp (approved):
👍

PR #2420, opened 2023-02-07
[MDS-5017] Refresh token- bug fix

taraepp:
## Objective 
No bug ticket has been written yet, but this was bug from 5017 PR. 

Making sure to append `@bceid` to bceid usernames. Notification function in particular broke when it wasn't able to query users properly, with this fix my notifications are back.

[MDS-5017](https://bcmines.atlassian.net/browse/MDS-5017)


matbusby-fw (approved):
Good catch!

PR #2421, opened 2023-02-08
[MDS-4969] incident fields

matbusby-fw:
## Objective 

Modify fields in the incidents form for both Minespace and Core.
DB Migration for mine_incidents table
API update

Side Note:  Not a big fan of the way these forms are structured.  Having multiple child functional components within the same file sharing the same props being passed forward isn't a great flow in my opinion.  Didn't want to go down a rabbit hole, so I fixed it enough that eslint won't complain, but would like to see these components distributed to their own files, or at least properly passing down props.

Side Note 2:  Turns out we have two separate incident endpoints with some overlapping(ish) functionality...also don't love this.

[MDS-4969](https://bcmines.atlassian.net/browse/MDS-4969)

image
image

### DIFF
--- a/migrations/sql/V2023.02.06.11.30__update_mine_incidents.sql
+++ b/migrations/sql/V2023.02.06.11.30__update_mine_incidents.sql
@@ -1,0 +1,4 @@
+ALTER TABLE
+    mine_incident
+    ADD COLUMN verbal_notification_provided boolean,
+    ADD COLUMN verbal_notification_timestamp timestamp with time zone;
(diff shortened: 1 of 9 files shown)


taraepp (approved):
I am somewhat concerned about timezone ambiguity on this one but am willing to let it slide until we pick up timezones ticket.

matbusby-fw:
Yeah, that will need some looking into for sure.  All of the DB DateTime fields are of the type "DateTime with Timezone".  Just worth noting.  I haven't looked at what the specific need is here.

PR #2423, opened 2023-02-09
[MDS-4969] Incidents Form

matbusby-fw:
## Objective 

[MDS-4969](https://bcmines.atlassian.net/browse/MDS-4969)

Removed required validator from verbal confirmation timestamp field
Made this field conditionally appear.


taraepp (approved):
👍

PR #2425, opened 2023-02-09
[MDS-5002] Incident notification emails

matbusby-fw:
## Objective 

[MDS-5002](https://bcmines.atlassian.net/browse/MDS-5002)

Added a check for the "DFT" status before sending an incident email on creation.


taraepp (approved):
👍

PR #2426, opened 2023-02-10
[MDS-5032] Missing Incident status banners

taraepp:
## Objective 
Add banners for RSS and IMS statuses. I removed the condition that only displayed the callout if "is final review stage"- verified that the default in the switch-case returning null didn't render something strange.

[MDS-5032](https://bcmines.atlassian.net/browse/MDS-5032)

Used https://titlecaseconverter.com/ for consistency in headings, APA style.


matbusby-fw (approved):
👍

PR #2428, opened 2023-02-13
[MDS-5027] - Incident link/refresh loading

matbusby-fw:
## Objective 

[MDS-5027](https://bcmines.atlassian.net/browse/MDS-5027)

- Fixed loading of the correct part of the form when navigated to directly via a link or when refreshing
- Refactored the components surrounding this
- Updated styles around the final review stage to make the fixed header less janky
- Fixed scroll to on final review stage (some stages had no id to navigate to)


taraepp (approved):
👍

PR #2429, opened 2023-02-15
[MDS-5040] - mine manager appointment

matbusby-fw:
## Objective 

[MDS-5040](https://bcmines.atlassian.net/browse/MDS-5040)

filtered out null values in party action creator before sending them to the api.
Updated `custom_reqparser.py` to return actual error message

### DIFF
--- a/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
+++ b/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
@@ -57,10 +57,7 @@ class MinePartyApptResource(Resource, UserMixin):
         'mine_party_acknowledgement_status',
         type=MinePartyAcknowledgedStatus,
         choices=list(MinePartyAcknowledgedStatus),
-        help='Indicator of status of acknowledgement.',
-        default=MinePartyAcknowledgedStatus.not_acknowledged,
-        location='json',
-        store_missing=False)
+        help='Indicator of status of acknowledgement.'),
     parser.add_argument(
         'status',
         type=MinePartyAppointmentStatus,
--- a/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
+++ b/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timedelta
-from flask import request
+from flask import request, current_app
 from app.api.activity.models.activity_notification import ActivityType
 from app.api.utils.access_decorators import is_minespace_user
 from flask_restplus import Resource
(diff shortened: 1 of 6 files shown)


services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
```diff
@@ -56,13 +56,15 @@ class MinePartyApptResource(Resource, UserMixin):
...
+        choices=list(MinePartyAcknowledgedStatus) + list([None]),
+        help='Indicator of status of acknowledgement.',
+        default=MinePartyAcknowledgedStatus.not_acknowledged,
+        location='json',
         store_missing=False)
     parser.add_argument(
         'status',
         type=MinePartyAppointmentStatus,
-        choices=list(MinePartyAppointmentStatus),
+        choices=list(MinePartyAppointmentStatus) + list([None]),
```
taraepp: I just don't think that a status of None should be valid, if we're getting null status we should change it, not allow it. Acknowledgement of appointments makes sense as None because not every appointment _needs_ to be acknowledged, but every appointment _should_ be in a state of active/inactive/pending. If we're getting null status it should probably be set to active.

Down around line 284 we're only setting status to active if the appointment is acknowledged, assuming that appointments that require acknowledgement are sent over properly as "not_acknowledged", I think the logic should be flipped to value !== not_acknowledged.

matbusby-fw: Conversely, since these aren't required fields and are in fact nullable, we don't want the api to fail if None is passed for whatever reason.

I'm pretty sure both of these fields are only utilized in the TSF roles section.  So, it wouldn't matter to any other assignments whether they are "active" or not.  I'd rather not set values for entries that aren't relevant.

services/core-web/common/actionCreators/partiesActionCreator.js
```diff
@@ -56,18 +56,20 @@ export const updateParty = (payload, partyGuid) => (dispatch) => {
...
-  dispatch(showLoading("modal"));
-  return CustomAxios()
-    .get(ENVIRONMENT.apiUrl + API.PARTIES_LIST_QUERY(params), createRequestHeader())
-    .then((response) => {
-      dispatch(success(reducerTypes.GET_PARTIES));
-      dispatch(partyActions.storeParties(response.data));
-    })
-    .catch(() => dispatch(error(reducerTypes.GET_PARTIES)))
-    .finally(() => dispatch(hideLoading("modal")));
-};
```
matbusby-fw, line 70: A lot of this is just prettier reformatting

services/core-web/common/actionCreators/partiesActionCreator.js
```diff
@@ -85,49 +87,55 @@ export const fetchPartyById = (id) => (dispatch) => {
...
-      throw new Error(err);
-    })
-    .finally(() => dispatch(hideLoading("modal")));
-};
+export const updatePartyRelationship =
+  (payload, successMessage = null) =>
+  (dispatch) => {
+    dispatch(request(reducerTypes.UPDATE_PARTY_RELATIONSHIP));
+    dispatch(showLoading("modal"));
+    const sanitizedPayload = removeNullValues(payload);
```
matbusby-fw, line 117: this is the actual change

taraepp (approved):
👍

PR #2431, opened 2023-02-15
[MDS-4930] Incident Notifications

matbusby-fw:
## Objective 

[MDS-4930](https://bcmines.atlassian.net/browse/MDS-4930)

Updated the checks for sending incident notifications from sending any time an incident was updated and the status was x -> send the notification

to any time an incidient was updated and the status has _changed_ to x -> send the notification

Updated activity feed wording for these as well


taraepp (approved):
🎉

PR #2432, opened 2023-02-15
[MDS-5066] Can't upload a file

taraepp:
## Objective 
Make uploading a file possible for incidents with RSS/IMS status.

[MDS-5066](https://bcmines.atlassian.net/browse/MDS-5066)

### DIFF
--- a/services/minespace-web/src/components/Forms/incidents/IncidentForm.js
+++ b/services/minespace-web/src/components/Forms/incidents/IncidentForm.js
@@ -627,40 +627,38 @@ const renderUploadInitialNotificationDocuments = (
           {title}
         </Typography.Title>
       )}
-      {notInitialDocumentsInForm &&
-        (!formValues.status_code || formValues.status_code === "DFT") &&
-        !isFinalReviewStage && (
-          <>
-            <Col span={24}>
-              <Typography.Title level={4}>
-                Upload Supporting Notification Documentation
-              </Typography.Title>
-              <Typography.Paragraph>
-                Please upload any documents that support this written incident notification. You may
-                return later to upload additional documents as needed.
-              </Typography.Paragraph>
-            </Col>
-            <Col span={24}>
-              <Form.Item>
-                <Field
-                  id={INITIAL_INCIDENT_DOCUMENTS_FORM_FIELD}
-                  name={INITIAL_INCIDENT_DOCUMENTS_FORM_FIELD}
-                  onFileLoad={(document_name, document_manager_guid) =>
-                    handlers.onFileLoad(
-                      document_name,
-                      document_manager_guid,
-                      Strings.INCIDENT_DOCUMENT_TYPES.initial,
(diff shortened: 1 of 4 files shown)


services/minespace-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -627,40 +627,38 @@ const renderUploadInitialNotificationDocuments = (
...
-                  onRemoveFile={parentHandlers?.deleteDocument}
-                  mineGuid={match.params?.mineGuid}
-                  component={IncidentFileUpload}
-                  labelIdle='<strong class="filepond--label-action">Supporting Document Upload</strong><div>Accepted filetypes: .kmz .doc .docx .xlsx .pdf</div>'
-                />
-              </Form.Item>
-            </Col>
-          </>
-        )}
+      {notInitialDocumentsInForm && !isFinalReviewStage && (
```
taraepp, line 630: I recommend turning on the "Hide whitespace differences" here!
This is mostly formatting- the only real changes were removing the condition `(!formValues.status_code || formValues.status_code === "DFT")` in 2 spots where we have `!isFinalReviewStage`. isFinalReviewStage is already false when status_code is undefined or DFT, so it was unnecessary to duplicate that logic.

services/minespace-web/src/components/pages/Incidents/IncidentPage.js
```diff
@@ -76,7 +76,6 @@ const defaultProps = {
   formErrors: {},
 };
 
-export const POST_SUBMISSION_INCIDENT_STATUSES = ["WNS", "AFR", "FRS", "UNR", "INV", "MIU", "CLD"];
```
taraepp, line 79: This is an export that wasn't imported anywhere else, it was only referenced in the useEffect below. I thought it made more sense to check against something constant (our _pre_-submission statuses of DFT and no status) than to keep adding to this list.

services/minespace-web/src/components/pages/Incidents/IncidentPage.js
```diff
@@ -146,7 +145,7 @@ export const IncidentPage = (props) => {
 
   useEffect(() => {
     if (incident.status_code) {
-      setIsFinalReviewStage(POST_SUBMISSION_INCIDENT_STATUSES.includes(incident.status_code));
+      setIsFinalReviewStage(incident.status_code !== "DFT");
```
matbusby-fw, line 148: Won't this set `IsFinalReviewStage` to be true at the stage before it's been saved as draft?  I know a few things are dependent on that (including the side menu that appears in the final stage)

taraepp: Nope! By default it's just undefined, so it won't pass the `if (incident.status_code)` guard.

matbusby-fw (approved):
Just left one question about the change in the `IsFinalReview` check.  If that's working though, then I'm all good!

PR #2433, opened 2023-02-16
[MDS-5073] Generate list of roles <> classes with python ast

vyasworks:
Script that generates a list for roles<>api methods and resources. 

This is temp script, we must find ways to document this using swagger or other means..

### DIFF
--- a/scripts/roledoc.py
+++ b/scripts/roledoc.py
@@ -2,7 +2,7 @@ import ast
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


scripts/roledoc.py
```diff
@@ -0,0 +1,115 @@
+import ast
+import os
+import sys
+
+def find_decorated_functions(folder_path, decorator_names):
```
taraepp, line 5: can we get a docstring on this function? Otherwise 👌

PR #2434, opened 2023-02-16
[MDS-5033] Remove Incidents Feature Flag

taraepp:
## Objective 
WIP! For visibility
- remove the feature flags for incidents
  - there appears to be a significant amount of dead code resulting from this
-  put comments on other feature flags so we can check if we want to unflag them yet and find them easier in the future
- I _did not_ fix any of the eslint errors, but I took a screenshot for later
- might be formatting issues

[MDS-5033](https://bcmines.atlassian.net/browse/MDS-5033)

image


services/minespace-web/src/components/dashboard/mine/incidents/IncidentsTable.js
```diff
@@ -156,16 +152,7 @@ export const IncidentsTable = (props) => {
                   record.mine_incident_guid
                 ),
                 state: { current: 1 },
-              });
-            }
-            return props.openModal({
```
taraepp, line 161: This is where the dead code is rooted. The above (that was previously guarded by if !IN_PROD) will always have a return, so my IDE detected the return props.openModal block as dead code, and after that, more and more started showing as dead. I have _not_ confirmed that this all still works

matbusby-fw (approved):
APPROVAL!!!!

PR #2435, opened 2023-02-17
[MDS-5084] - Incident Activity Feed Text

matbusby-fw:
## Objective 

[MDS-5084](https://bcmines.atlassian.net/browse/MDS-5084)

Updated text in the activity feed to include the incident number when a new incident is submitted.
image


taraepp (dismissed):
👍 🎨

taraepp (approved):

PR #2436, opened 2023-02-17
[MDS-5085] Mine GUID undefined when clicking on Incident link in notification

taraepp:
## Objective 
Notification drawer bug. Was originally intending to fix 5081 but this one was blocking it.

[MDS-5085](https://bcmines.atlassian.net/browse/MDS-5085)


matbusby-fw (approved):
👍

PR #2437, opened 2023-02-17
[MDS-5081] Incident page reloading bugs

taraepp:
## Objective 
The component wasn't updating when it was getting new data.
Refactored Core MineIncident component, the only real change should be the same as in MS- adding the location prop to the useEffect dependency array.

[MDS-5081](https://bcmines.atlassian.net/browse/MDS-5081)

### DIFF
--- a/services/core-web/src/components/Forms/incidents/IncidentForm.js
+++ b/services/core-web/src/components/Forms/incidents/IncidentForm.js
@@ -942,7 +942,7 @@ const renderEditSaveControls = (childProps, isEditMode, isNewIncident) => (
         <Popconfirm
           placement="topLeft"
           title="Are you sure you want to cancel this submission? All unsaved changes will be lost."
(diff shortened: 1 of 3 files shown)


services/core-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -942,7 +942,7 @@ const renderEditSaveControls = (childProps, isEditMode, isNewIncident) => (
         <Popconfirm
           placement="topLeft"
           title="Are you sure you want to cancel this submission? All unsaved changes will be lost."
-          onConfirm={() => childProps.handlers.handleCancelEdit(isNewIncident)}
```
taraepp, line 945: eslint was angry that the original definition of this from MineIncident was reusing a variable name already declared in upper scope, and I thought that we didn't need a parameter at all.

matbusby-fw (approved):
Yay more refactoring.  Looks great!

PR #2438, opened 2023-02-21
[MDS-5017] Reduce refresh token frequency

taraepp:
## Objective 
Was using a force refresh (value of -1 in updateToken function) when checking whether to update token. Multiple instances seemed to be getting created and so it was refreshing the token multiple times in quick succession. Put in a value for it to check against before refreshing so that it will only do it once.

[MDS-5017](https://bcmines.atlassian.net/browse/MDS-5017)


matbusby-fw (approved):
Ah, nice!

PR #2441, opened 2023-02-22
[MDS-5083] updated mine alert logic and validation

matbusby-fw:
## Objective 

[MDS-5083](https://bcmines.atlassian.net/browse/MDS-5083)

- Added frontend validation to match backend for alert form
- Removed database constraint preventing users from creating new alerts that overlap with previous inactive alerts (this is the requested behaviour, and the constraint wasn't properly ignoring records with `deleted_ind=true`

image


taraepp (dismissed):
👍

taraepp (approved):

PR #2446, opened 2023-02-23
[MDS-5070] Incident Documents

matbusby-fw:
## Objective 

Incident documents uploaded from Core were removing previously uploaded documents.  Stopped this from happening.

[MDS-5070](https://bcmines.atlassian.net/browse/MDS-5070)

Removed a function in the put endpoint for incidents that checked every document attached to the incident and deleted it if it wasn't passed along with the `updated_documents`.  
This was weird.  I went through all of the scenarios I could find and can see no reason for this code to have existed.  I'm wondering if this was part of some legacy code that didn't get removed for some reason.


taraepp (approved):
👍 
Yeah it makes no sense for this code to be in there.

PR #2448, opened 2023-03-01
[MDS-5058] incident notifications

matbusby-fw:
## Objective 

[MDS-5058](https://bcmines.atlassian.net/browse/MDS-5058)

Decision was made to remove the email portion of notifications when an incident is updated.


taraepp (approved):
👍

PR #2449, opened 2023-03-01
[MDS-5061] Bug: can edit fields when in view mode

taraepp:
## Objective 
- fixed some bonus bugs relating to routing as they were disruptive during development
   - refreshing the page won't take the user out of edit mode anymore
   - sidebar menu should always work now! Did not fix that the scroll amount is slightly off, but did fix bugs in create mode: undefined never gets put in the URL and the mine name doesn't go away
   - both isEditMode and the mine name (for new incidents) were being passed along in the state and this was not working very well- now edit has its own route and the mine name is being passed as a query string (maybe not best practice...)
- clarified that the add ministry comments should still be available in view mode
   - there was a major bug with this component/set of components that I also fixed, result was that it was always giving permission on the FE, even if the user did not have the required permission (would still fail in BE)
   - did not get clarification on whether they should be viewable when the user lacks the core_edit_incidents role (required to make a comment) to _view_ the comments, so right now it's still viewable (didn't change). 
- removed the button to switch from view to edit mode
   -  and the Cancel edit button to switch to view mode
   - and the functions required for both buttons

[MDS-5061](https://bcmines.atlassian.net/browse/MDS-5061)

### DIFF
--- a/services/core-web/src/components/mine/Incidents/MineIncident.js
+++ b/services/core-web/src/components/mine/Incidents/MineIncident.js
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from "react";
 import { bindActionCreators } from "redux";
 import { flattenObject } from "@common/utils/helpers";
 import { connect } from "react-redux";
-import { Link, withRouter } from "react-router-dom";
+import { Link, withRouter, useParams, useLocation } from "react-router-dom";
 import { change, submit, getFormSyncErrors, getFormValues, touch, isDirty } from "redux-form";
 import { Tag } from "antd";
 import { ArrowLeftOutlined, EnvironmentOutlined } from "@ant-design/icons";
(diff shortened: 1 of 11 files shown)


services/core-web/src/components/mine/Incidents/MineIncident.js
```diff
@@ -42,20 +34,7 @@ const propTypes = {
   updateMineIncident: PropTypes.func.isRequired,
   clearMineIncident: PropTypes.func.isRequired,
   removeDocumentFromMineIncident: PropTypes.func.isRequired,
-  match: PropTypes.shape({
```
matbusby-fw, line 45: love this.  Not a fan of how this has been implemented in the past

matbusby-fw (approved):
Nice!

PR #2450, opened 2023-03-01
[MDS-5024] - Added fields to NoW list

matbusby-fw:
## Objective 

[MDS-5024](https://bcmines.atlassian.net/browse/MDS-5024)

Added to the information returned by the NoW list resource:

- NoW contacts
- lat and long of the Mine the NoW is attached to


taraepp (approved):
👍

PR #2451, opened 2023-03-02
[MDS-5061] Bug fixes incidents editing & reloading

taraepp:
## Objective 
- test environment was doing funky things to the URL so I changed it to use the matchPath function intended for that purpose (tests for prefix, allowing extra / at end of URL or state, etc) instead of String.endsWith (not reliable).
- missed making status not editable before- fixed that
- I was noticing that incidents were really slow due to an internal server error caused by backend trying to get NRIS data for non-prod data (always returned 403 Unauthorized). I'd made the change before to speed things up but now I decided to commit it.
  - took this out because it's bigger than I thought, new ticket: https://bcmines.atlassian.net/browse/MDS-5115

[MDS-5061](https://bcmines.atlassian.net/browse/MDS-5061)

### DIFF
--- a/services/core-api/app/api/mines/compliance/resources/compliance.py
+++ b/services/core-api/app/api/mines/compliance/resources/compliance.py
@@ -12,7 +12,6 @@ from app.api.utils.access_decorators import requires_role_view_all
 from app.api.services import NRIS_API_service
 from app.api.mines.response_models import MINE_COMPLIANCE_RESPONSE_MODEL
 from app.api.mines.mine.models.mine import Mine
-from app.config import Config
 
 
 class MineComplianceSummaryResource(Resource, UserMixin):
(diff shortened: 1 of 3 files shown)


services/core-api/app/api/mines/compliance/resources/compliance.py
```diff
@@ -23,7 +24,7 @@ def get(self, mine_no):
             raise NotFound("No mine record in CORE.")
 
         result = cache.get(NRIS_COMPLIANCE_DATA(mine.mine_no))
-        if result is None:
+        if result is None and Config.ENVIRONMENT_NAME == 'prod':
```
matbusby-fw: Very nice!

taraepp: It makes tests fail unfortunately. 😭 
Not sure if it's worth keeping (although I wonder what those tests are actually testing because there's no way that they're getting real data in there)

matbusby-fw (approved):
👍 

I do think adding that NRIS check back in and updating the tests at some point would be worth it, but maybe not as part of this specific ticket.

PR #2452, opened 2023-03-02
[MDS-5113] - Incident triggers and templates

matbusby-fw:
## Objective 

[MDS-5113](https://bcmines.atlassian.net/browse/MDS-5113)

Made some updates to the incident trigger conditions and fixed the dynamic data on two of the templates


taraepp (approved):
👍 
LGTM

PR #2453, opened 2023-03-02
[MDS-5082] disabled Incident Mine manager recommendations field on Minespace

matbusby-fw:
## Objective 

[MDS-5082](https://bcmines.atlassian.net/browse/MDS-5082)


taraepp (approved):
👍

PR #2456, opened 2023-03-10
[MDS-5129] Change minespace-proponent role to mds_minespace_proponents

taraepp:
## Objective 
Getting rid of redundant roles, including minespace-proponent. Replace references with mds_minespace_proponents.

[MDS-5129](https://bcmines.atlassian.net/browse/MDS-5129)


matbusby-fw (approved):
Weee!

PR #2458, opened 2023-03-14
[MDS-5125] character limit on mine alerts

henryoforeh-dev:
## Objective 

[[MDS-5121](https://bcmines.atlassian.net/browse/MDS-5121)]

- Display the number of characters that are already input (as a fraction of the total number allowed) - to match the design.

 - The input shouldn’t allow more characters than what is allowed.
 
image

### DIFF
--- a/services/core-web/src/components/common/RenderAutoSizeField.js
+++ b/services/core-web/src/components/common/RenderAutoSizeField.js
@@ -61,12 +61,11 @@ const RenderAutoSizeField = (props) => {
         onChange={handleTextAreaChange}
         value={value}
       />
-      {props.maximumCharacters > 0 && (
-        <div className="flex between">
-          <span>{`Maximum ${props.maximumCharacters} characters`}</span>
-          <span className="flex-end">{`${remainingChars} / ${props.maximumCharacters}`}</span>
-        </div>
-      )}
+
+      <div className="flex between">
+        <span>{`Maximum ${props.maximumCharacters} characters`}</span>
+        <span className="flex-end">{`${remainingChars} / ${props.maximumCharacters}`}</span>
+      </div>
     </Form.Item>
   );
 };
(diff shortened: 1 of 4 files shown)


services/core-web/src/components/common/RenderAutoSizeField.js
```diff
@@ -16,37 +16,60 @@ const propTypes = {
...
+      }
+    >
+      <Input.TextArea
+        disabled={props.disabled}
+        id={props.id}
+        {...props.input}
+        autoSize={{ minRows: props.minRows }}
+        placeholder={props.placeholder}
+        onChange={handleTextAreaChange}
+        value={value}
```
matbusby-fw, line 63: Cool, this should definitely work.  I know AntD does have a [`showCount`](https://ant.design/components/input#components-input-demo-show-count) prop for the `Input.TextArea` that might make it a little bit simpler.  

I haven't played with it myself though so maybe that doesn't provide quite the behaviour we're looking for.  Did you look into that at all?

henryoforeh-dev: Great, I will have a look at it and refactor. Thanks.

services/core-web/src/components/common/RenderAutoSizeField.js
```diff
@@ -16,37 +16,60 @@ const propTypes = {
...
   minRows: PropTypes.number,
+  maximumCharacters: PropTypes.number,
 };
 
 const defaultProps = {
   placeholder: "",
   label: "",
   disabled: false,
   minRows: 3,
+  maximumCharacters: 300,
```
taraepp: This component is used in quite a few places, I'm not 100% sure we want to enable character limiting by default. Have we done any investigation into that?

henryoforeh-dev: Nice catch, I will refactor.

taraepp (approved):
LGTM

PR #2459, opened 2023-03-14
[CHORE] added lower() to username fetch from token

matbusby-fw:
## Objective 

Added .lower() function to parsing of user name from jwt


taraepp (approved):
👍

PR #2461, opened 2023-03-17
[MDS-4995] update memp document

henryoforeh-dev:
## Objective 

[MDS-4995](https://bcmines.atlassian.net/browse/MDS-4995)

Replaced MEMP PDF document in the home page with the one attached in the ticket.


matbusby-fw (approved):
Nice!

PR #2462, opened 2023-03-17
[MDS-5121] bug fix on on edit alert.

henryoforeh-dev:
## Objective 

[MDS-5121](https://bcmines.atlassian.net/browse/MDS-5121)

- There was a bug in the last work I did on this ticket. When you click on edit alert, the value previously entered in the message field was not showing. I have provided screen shots for the fix.

**When Edit alert is clicked**
edit_alert

**When create alert is clicked**
create_alert_with_typed_text


matbusby-fw (approved):
👍

PR #2463, opened 2023-03-17
[MDS-4868] Incidents Categorization

taraepp:
## Objective 
- Update FE with new incidents categories & subcategories.
- remove categories display from Minespace
- Legacy categories should still show for existing incidents that have those categories, but encourage them to be switched to new categories and don't allow them to be added to any incident that doesn't already have them.
- update emails because they reference the categories
- fix email datetimes by using the email datetime function that will translate to PST & include "PST"

[MDS-4868](https://bcmines.atlassian.net/browse/MDS-4868)
[MDS-5026](https://bcmines.atlassian.net/browse/MDS-5026)

### An incident with legacy categories ("Spill" is not available here because it was saved without it)
image
### Same but in view mode
image
### The incidents table
image
### Brand new incident or incident saved without any legacy categories
image
### Validation styling
image
### Removed from MineSpace table & form
image


services/core-web/src/components/Forms/incidents/IncidentCategorySelect.js
```diff
@@ -0,0 +1,129 @@
...
+  }).isRequired,
+};
+
+const defaultProps = {
+  data: [],
+  disabled: false,
+};
+
+// an item is considered a child of another item when its subType matches the parent's value
+const IncidentCategorySelect = (props) => {
```
matbusby-fw, line 28: Not major, but `IncidentCategorySelect` makes it sound like it's a dropdown select menu to me.  Maybe we could go with something more specific to checkboxes or more generic?

services/core-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -910,6 +909,10 @@ const renderEditSaveControls = (childProps, isEditMode, isNewIncident) => (
 
 export const IncidentForm = (props) => {
   const [uploadedFiles, setUploadedFiles] = useState([]);
+  // 2nd parameter in getDropdownIncidentCategoryCodeOptions controls inclusion of historic categories
+  const incidentCategoryCodeOptions = useSelector((state) =>
+    getDropdownIncidentCategoryCodeOptions(state, false)
```
matbusby-fw, line 914: Do we have a use-case where we do want to include the historic options?  Just wondering if the 2nd parameter is needed.

taraepp: I had to include them in order to display the legacy options!

This is what took me a ridiculous amount of time to wrap my head around- it already _does_ call for a second parameter, it just defaults to "true". That's done in createSelectorWrapper with the getVisibilityFilterOption, with the showActiveOnly parameter. I was waffling on the idea of including a link to something about selectors- on the one hand they are baffling but on the other hand they are a core functionality of Redux.

services/core-web/src/components/Forms/incidents/IncidentCategorySelect.js
taraepp: IncidentCheckboxGroup? I agree, this is not a very good name!

services/core-web/common/selectors/staticContentSelectors.js
```diff
@@ -274,8 +274,22 @@ export const getDropdownIncidentStatusCodeOptions = createSelectorWrapper(
...
+        subType: item.parent_mine_incident_category_code ?? null,
+      };
+    });
+  },
+  [
+    "description",
+    "mine_incident_category_code",
+    "is_historic",
+    "parent_mine_incident_category_code",
+  ]
```
henryoforeh-dev, line 292: Will "description" map to "item.mine_incident_category_code" ?

services/core-web/src/components/Forms/incidents/IncidentForm.js
matbusby-fw: Ah, okay, that makes sense to me!

services/core-web/common/selectors/staticContentSelectors.js
taraepp: Now that I look at it I'm thinking that while it works, this section is wonky from being changed a few times. 😅 
Initially I was using createDropdownList, which takes the label field, value field, is active field, and sub type, in that order, and this list of parameters makes sense in that context. But then I created a custom mapping function  because I had switched to a component that used key/title instead of value/label. And when I switched back I probably modified my mapping function instead of reverting when I could.

I'm going to try bringing it back to createDropdownList, I think it should work fine again!

taraepp: Oops, that wasn't quite right.
is_historic and is_active are basically opposite, so the custom mapping function was still necessary. However, custom mapping function doesn't require a parameter list like createDropdownList does, so the array can be removed (it wasn't doing anything).

henryoforeh-dev (approved):

PR #2464, opened 2023-03-17
[MDS-4936] Wording edit in NoD alert

henryoforeh-dev:
## Objective 

[MDS-4936](https://bcmines.atlassian.net/browse/MDS-4936)

Updated the wording on the Notice of Departure component to match content in the ticket.


taraepp (approved):
👍

PR #2466, opened 2023-03-20
[MDS-5139]  FNCS connection updates

matbusby-fw:
## Objective 

[MDS-5139](https://bcmines.atlassian.net/browse/MDS-5139)

- Updated Now list endpoint to properly return the Agent responsible for the NoW and the permit_no if they exist
- Added search parameters for permit_no and party


taraepp (approved):
LGTM

henryoforeh-dev (approved):

PR #2467, opened 2023-03-20
[MDS-5125] Fixing linting issues

isuru-aot:
## Objective 

[MDS-5125](https://bcmines.atlassian.net/browse/MDS-5125)

Fixing bad linting in .js files in code base (core-web, minespace-web and common) by running `npm run format:write-all`


matbusby-fw (dismissed):
Ooooh, tidy!

My only complaint really is the forcing of using double quotes.  I have an irrational dislike of double quotes.  Not really anything to do with your PR (and really just my preference), but I'd love to switch the linting rule to single quotes.  Thoughts @taraepp , @henryoforeh-aot?

taraepp:
> Ooooh, tidy!
> 
> My only complaint really is the forcing of using double quotes. I have an irrational dislike of double quotes. Not really anything to do with your PR (and really just my preference), but I'd love to switch the linting rule to single quotes. Thoughts @taraepp , @henryoforeh-aot?

One thing: removing the incident feature flags I think I did at least one no-verify push so we might want to hold off merging so we can catch those. 😅 

I like that it's one or the other at least for quotes (makes it easier to search when you're looking for a string). Javascript it doesn't technically matter if it's single or double, I'm not terribly opinionated on which we go with. 
**Single quote**: what I actually type in because it's less key presses. 
**Double quote**: can encapsulate a ' character within the string without having to escape it (but I think our contractions use a proper apostrophe character instead of a single quote).

henryoforeh-dev (dismissed):

isuru-aot:
[New PR](https://github.com/bcgov/mds/pull/2479) is raised for the same. Hence closing this PR without merging.

PR #2470, opened 2023-03-21
[MDS-5098] More Accurate Arch Diagram - v2

vyasworks:
## Objective 

[MDS-5098](https://bcmines.atlassian.net/browse/MDS-5098)


taraepp (commented):
I just have a question about what the "S" means on the deployments (ex: "S: mds")

vyasworks:
> I just have a question about what the "S" means on the deployments (ex: "S: mds")

S: is schema that this particular deployment uses. This is symbolized with the db connection icon.

taraepp:
> 

I think the icon didn't get included, I only see something like that for the redis cache:
image

vyasworks:
> > 
> 
> I think the icon didn't get included, I only see something like that for the redis cache: image

They are included, git's rich diff seems to be having troubles doing the rendering. 

Try downloading the svg or can refer this : 

https://apps.nrs.gov.bc.ca/int/confluence/download/attachments/42107578/MDS_Arch-Arch-V2.jpg?version=1&modificationDate=1679437013000&api=v2

simensma-fresh (approved):

taraepp (approved):

PR #2473, opened 2023-03-24
[MDS-5051] Adjusting links in no permit required IP letter

asinn134:
## Objective 

[MDS-5051](https://bcmines.atlassian.net/browse/MDS-5051)

- The reason for this change is _No Permit required IP_ letter's links were redirecting to non-existent pages. 
- This change updates the text displayed to the user before they preview or generate the letter
- This change updates the links found within the letter file

**This is the text user sees before generating or previewing the letter**:

Screenshot (27)

**This is the document before the change**:

Screenshot (29)

**This is the document after the change**:

Screenshot (25)


henryoforeh-dev (dismissed):

migrations/sql/V2023.03.22.01.06__update_npi_document_template.sql
```diff
@@ -0,0 +1,141 @@
...
+      "id": "rc_office_mailing_address_line_2",
+      "relative-data-path": "mine.region.regional_contact_office.mailing_address_line_2",
+      "read-only": true
+    },
+    {
+      "id": "application_type_code",
+      "relative-data-path": "now_application.application_type_code",
+      "read-only": true
+    }
+  ]' WHERE document_template_code ='NPI';
```
taraepp: If you perform the migration with flyway, does the data look how you expect it to? I'm thinking that a change made in afterMigrate__1__seed_data.sql may undo your changes. Just something to watch for.

asinn134: I did find that when I run the migration with flyway the data in the table sometimes changes to what I expect and sometimes stays the same, but I thought this might just be due to how I attempted to get the database to work on my local computer. 

For the afterMigrate__1__seed_data.sql, my understanding is that this only executes on the local environment? but maybe I misunderstood. Would you suggest adding the link change the afterMigrate__1__seed_data.sql file as well?

taraepp: I don't _think_ that afterMigrate only runs in the local environment, but it's possible that I'm missing something. Still, we would like our local dev environments to consistently show us the up-to-date information just like in other environments.

I'm personally not a fan of having `UPDATE` statements in the after migrate scripts, I think it's bad design (should be INSERT ... ON CONFLICT DO NOTHING or similar- UPDATE statements belong in migrations) so my gut is actually to remove the conflicting query from the file entirely.

asinn134: As discussed in the meeting I tried moving out all the **UPDATE** statements in the afterMigrate file, doing so I think there might be an issue with migration scripts. Is my thinking correct that the afterMigrate script since it seeds the data, it technically would run first? I think if that is the case if we remove the updates to now_application_document_type table like for example 

```
UPDATE now_application_document_type
SET document_template_code = 'NCL'
where now_application_document_type_code = 'CAL';
```

and place it in a migration script created in 2023, there may be some issues where past scripts are trying to make changes based on that data? I think that might be happening, because when I remove all updates in afterMigrate file and move them into a new migration script, there seems to be some errors occurring

For example:

```
mds_flyway_verify  | SQL State  : 23503
mds_flyway_verify  | Error Code : 0
mds_flyway_verify  | Message    : ERROR: insert or update on table "now_application_document_type" violates foreign key constraint "now_application_document_type_template_code"
mds_flyway_verify  |   Detail: Key (document_template_code)=(NCL) is not present in table "document_template".
mds_flyway_verify  | Location   : sql/V2023.03.24.12.04__update_queries_from_afterMigrade__1__seed_data.sql (/flyway/sql/V2023.03.24.12.04__update_queries_from_afterMigrade__1__seed_data.sql)
mds_flyway_verify  | Line       : 590
mds_flyway_verify  | Statement  : UPDATE now_application_document_type
mds_flyway_verify  | SET document_template_code = 'NCL'
mds_flyway_verify  | where now_application_document_type_code = 'CAL'
mds_flyway_verify  | 
mds_flyway_verify  | Caused by: Migration V2023.03.24.12.04__update_queries_from_afterMigrade__1__seed_data.sql failed
```

At the moment I changed it to just removing the conflicting update query in the afterMigrate file. 
I wonder though for something like the first example above with update for now_application_document_type table to set the document_template_code, maybe we can just change the inserts to include the document_template_code?

taraepp: Hmm, I do believe that the afterMigrate scripts run _after_ the migration, but that's a weird error you're getting in that case. Going to check this out locally to verify.

taraepp: According to their [documentation](https://flywaydb.org/documentation/concepts/callbacks) it should be run after each migrate run, unless it's somehow getting mixed up with afterEachMigrate. But I think this is enough of a rabbithole that it's not worth holding up the issue any longer.

taraepp (approved):

migrations/sql/V2023.03.22.01.06__update_npi_document_template.sql
asinn134: Ok, sounds good. Thanks Tara for checking this out on your end. I will go ahead and merge with just removing the conflicting query in afterMigrate seed file.

PR #2474, opened 2023-03-24
[MDS-4868] Story bugs: validation, category options

taraepp:
## Objective 
[MDS-4868](https://bcmines.atlassian.net/browse/MDS-4868)

- The checkbox group component isn't very good at being able to tell when the form is touched, or setting the state from that (doesn't register as a change?). Now the error message shows up better (ex: go to create an incident, click "save changes" without entering anything, or edit an incident originally created in MS from CORE- by default these will have no categories --> error message displays properly now)
- So many differences between the PDF and existing category options in DB that I just batch updated all of them. I made some very minor changes as discussed with Rebecca. 
- I thought I had renamed the incident category component but evidently not, fixed this

### DIFF
--- a/bin/setup_codespaces.sh
+++ b/bin/setup_codespaces.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 yes yes | make env
+nvm install
 yarn
 make be
 make seeddb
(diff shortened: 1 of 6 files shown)


matbusby-fw (approved):
Woooo!

PR #2475, opened 2023-03-24
[MDS-5144] date time validation issues

isuru-aot:
## Objective 

[MDS-5144](https://bcmines.atlassian.net/browse/MDS-5144)

- Removed the Incident Type form the Core Incident Template
- Fix the, Link to the "View in Minespace" not clickable issue on the "A final incident report"

Note: The other two points (1 & 3) mentioned in the JIRA issue, are addressed separately.


taraepp (approved):
👍

PR #2476, opened 2023-03-27
[MDS-5140] fix broken NOD link navigation

henryoforeh-dev:
## Objective 

[MDS-5140](https://bcmines.atlassian.net/browse/MDS-5140)

I am making this change to fix the broken navigation to Notice of Departments from the Mine Activity feed.

image

### DIFF
--- a/services/minespace-web/src/components/dashboard/mine/MineDashboard.js
+++ b/services/minespace-web/src/components/dashboard/mine/MineDashboard.js
@@ -48,7 +48,15 @@ const defaultProps = {
 const initialTab = "overview";
 
 export class MineDashboard extends Component {
-  state = { isLoaded: false, activeTab: initialTab, mineNotFound: false };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isLoaded: false,
+      activeTab: initialTab,
+      mineNotFound: false,
+    };
+  }
 
   componentDidMount() {
     const { id, activeTab } = this.props.match.params;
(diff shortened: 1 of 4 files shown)


services/minespace-web/src/components/dashboard/mine/noticeOfDeparture/NoticeOfDeparture.js
```diff
@@ -160,7 +160,7 @@ export const NoticeOfDeparture = (props) => {
   };
 
   useEffect(() => {
-    const nod = new URLSearchParams(location.search).get("nod");
+    const nod = new URLSearchParams(window.location.href.split("?")[1]).get("nod");
```
taraepp: I think you can use useParams here and it will be simpler, like in DamForm.js

services/minespace-web/src/components/dashboard/mine/MineDashboard.js
```diff
@@ -59,17 +67,25 @@ export class MineDashboard extends Component {
...
-      this.loadMine(id, activeTab);
+    if (
+      nextProps.match.params.id !== this.props.match.params.id ||
+      (this.extractActiveTab(activeTab) === "nods" && !this.state.isLoaded)
+    ) {
+      this.loadMine(id, this.extractActiveTab(activeTab));
     }
   }
 
+  extractActiveTab = (str) => {
```
taraepp: Can we use a more meaningful variable name than str?

Is the problem that is causing the bug essentially that instead of passing the activeTab like `?activeTab={tab}` that we're just passing it like `?{tab}`? (Or otherwise not in the format that the component expects) If so then we might want to fix that problem at the source to make it consistent, that is handled in the `NotificationDrawer` component.

services/minespace-web/src/components/dashboard/mine/MineDashboard.js
```diff
@@ -59,17 +67,25 @@ export class MineDashboard extends Component {
...
-      this.setState({ activeTab });
+      this.setState({ activeTab: this.extractActiveTab(activeTab) });
     }
     if (!nextProps.staticContentLoadingIsComplete) {
       this.loadStaticContent();
     }
 
-    if (nextProps.match.params.id !== this.props.match.params.id) {
-      this.loadMine(id, activeTab);
+    if (
```
taraepp: I find this condition a little difficult to read- if we were to translate this condition into English, what does it _mean_? Since we're referencing `this.extractActiveTab(activeTab)` in both the condition and the code block body, can we use the state variable or call the function once and store it in a variable?

services/minespace-web/src/components/layout/NotificationDrawer.js
```diff
@@ -168,7 +168,7 @@ const NotificationDrawer = (props) => {
   const activityClickHandler = async (notification) => {
     await handleMarkAsRead(notification.notification_guid);
     handleCollapse();
-    const routeSpecifics = await navigationHandler(notification);
+    const routeSpecifics = navigationHandler(notification);
```
taraepp, line 171: interesting, this is an async function on core-web but not on minespace! (although it doesn't look like it _needs_ to be async on core-web either)

taraepp (approved):
👍

PR #2478, opened 2023-03-27
[MDS-5144] Fixing minespace view incident url

isuru-aot:
## Objective 

[MDS-5144](https://bcmines.atlassian.net/browse/MDS-5144)

Fix the minspace url for view incident


taraepp (approved):
👍

PR #2484, opened 2023-04-04
[MDS-5158] added primary keys and index to permit conditions

matbusby-fw:
## Objective 

[MDS-5158](https://bcmines.atlassian.net/browse/MDS-5158)


taraepp (approved):
🙌

Mike-aot (approved):

PR #2490, opened 2023-04-11
[MDS-5149] Major Mine endpoint enhancement

isuru-aot:
## Objective 

[MDS-5149](https://bcmines.atlassian.net/browse/MDS-5149)

To enhance the response time of the major-mines endpoint (`/projects/dashboard?page=1&per_page=10`).

### DIFF
--- a/services/core-api/app/api/projects/project/resources/project.py
+++ b/services/core-api/app/api/projects/project/resources/project.py
@@ -173,7 +173,6 @@ class ProjectListDashboardResource(Resource, UserMixin):
                         update_timestamp_filter_pass = True
                 else:
                     update_timestamp_filter_pass = True
-
                 if status_code_filter_pass and update_timestamp_filter_pass:
                     record = {
                         'stage': stage,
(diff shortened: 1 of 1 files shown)


services/core-api/app/api/projects/project/resources/project.py
```diff
@@ -191,7 +188,9 @@ def get(self):
                         'project_lead_party_guid': project.project_lead_party_guid,
                         'project_lead_name': project.project_lead_name,
                         'update_timestamp': update_timestamp,
-                        'mine': project.mine
+                        'mine': {
+                            'mine_name': project.mine.mine_name
```
matbusby-fw: Is this the only field we're using here out of all the mine info :laughing:

matbusby-fw (dismissed):
LGTM!  I like seeing these returns get leaner.

services/core-api/app/api/projects/project/resources/project.py
isuru-aot: I've missed the `Commodity`  field, which is also retrieved from mine object. I just pushed the fix for that.

Table Headers => `Project Title`, `Project ID`, `Mine name`, `MRC, Stage`, `Review Status`, `EMLI Project Lead`, `Commodity`, `Updated Date`

Attached image shows the updated response, with the relevant fields for the table view.
Updated Response

henryoforeh-dev (approved):

PR #2491, opened 2023-04-12
[MDS-5193] - Implement Typescript

matbusby-fw:
### Had a little spare capacity so I took a stab at enabling Typescript

For now I only worked on the Minespace portion, but these concepts should work the same on Core.
Everything is working fine locally, and properly handling js, jsx, ts and tsx files
Updated the landing page to `LandingPage.tsx` and included some very light typescript
Updated `NoticeOfDeparture.js` to `NoticeOfDeparture.tsx`
updated jest to properly compile typescript pages
updated tests for these two components to .tsx

Wanted to throw a PR up at this point to get a couple of extra sets of eyes on this to see if anyone can see any issues or room for improvement.


services/minespace-web/package.json
```diff
@@ -29,6 +29,8 @@
     "@syncfusion/ej2-react-filemanager": "19.1.54",
     "@syncfusion/ej2-react-pdfviewer": "19.4.55",
     "@syncfusion/ej2-splitbuttons": "19.1.54",
+    "@types/jest": "^29.1.0",
```
vyasworks: fix versions?

vyasworks (commented):
Looks great! 

Would love to see this go to dev, are you able to make the changes on dockerfiles as well and see if it builds Mat?

services/minespace-web/package.json
matbusby-fw: Ah, good catch.  Yes, I started trying to update all of the jest stuff to 29, but couldn't get it to work, so rolled the others back to 24.  Will update this.

taraepp:
Exciting!

A couple of settings files that should be updated:
- .eslintrc.js should include ts/tsx files where we're configuring things for js files (the extensions array ~line 21)
- same with the root-level package.json for the lint-staged script ~line 34, 37 (lint-prepush is currently disabled, we could either remove or update that section too though)
- I think also the package.json files for core-web/minespace with the other scripts like format:check are just configured for js files
- some sonar-project.properties files (are we using this? I think so)
- I wonder about that exports.loadJS in services/core-web/webpack.parts.js (probably similar files in other directories)

matbusby-fw:
Good input @taraepp.  I'll follow up on these this week at some point when I have the capacity!

services/minespace-web/package.json
```diff
@@ -123,7 +132,6 @@
   },
   "optionalDependencies": {
     "enzyme-adapter-react-16": "1.14.0",
-    "jest": "24.8.0",
     "jest-canvas-mock": "2.1.0",
     "jest-environment-jsdom": "24.8.0",
     "jest-environment-jsdom-global": "1.2.0",
```
taraepp, line 137: I think the format scripts just below here (ln 150-152) and the equivalent bits in the core-web file are all that remains.

taraepp (approved):
🙌

PR #2494, opened 2023-04-17
[MDS-5020] Implement jwt manager with gold clients

vyasworks:
## Objective 

[MDS-5020](https://bcmines.atlassian.net/browse/MDS-5020)

Implement JWT Manager Switching


taraepp (approved):
LGTM

PR #2496, opened 2023-04-19
[MDS-5005] Unable to change mine name

asinn134:
## Objective

- It was found that mine#1650126 does still exist within production database, however it's deleted_ind was set to true, so it doesn't show up in the list of mines when searching in the mines page. When updating mine name we only retrieved results for comparison based on the name, so mine#1650126 will come back as a result even though deleted_ind has a value of true.

- The Change is so searching to see if name already exists only uses results with deleted_ind of value false for comparison 

[MDS-5005](https://bcmines.atlassian.net/browse/MDS-5005)


taraepp (approved):
LGTM

PR #2497, opened 2023-04-20
[MDS-5177] Permissions for Notice of Departure

taraepp:
## Objective 

[MDS-5177](https://bcmines.atlassian.net/browse/MDS-5177)
- ensure that the right people have the right access to the NoD information

### Summary of changes:
- BE permissions for NoD:
  - EDIT_DO --> EDIT_PERMITS
  - add restrictions to deleting NoD documents
  - allow VIEW_ALL to view the NoD data
- FE NoD modal form changes:
  - create a "view only" mode for the form when they don't have EDIT_PERMITS permission: 
    - don't show the file upload or delete
    - allow the form to be disabled
    - change buttons from Update/Cancel to Close
    - disable the form 
  - modal title is Edit or View NoD depending on what they're doing
    - there's no "Create" because it must be created in MS, didn't touch the MS side

#### Edit Mode
image
#### View Mode
image


henryoforeh-dev (approved):
Nice work!

PR #2505, opened 2023-04-26
[MDS-5036] - ts-loader version update

matbusby-fw:
## Objective 

Was seeing some errors using ts-loader version 8.2.0.  Upgraded to 8.2.4 (the most recent that is compatable with webpack v4)

[MDS-5036](https://bcmines.atlassian.net/browse/MDS-5036)


asinn134 (approved):

taraepp (approved):
👍

PR #2506, opened 2023-04-27
[MDS-5036] upgraded heap size for Core-Web build

matbusby-fw:
## Objective 

[MDS-5036](https://bcmines.atlassian.net/browse/MDS-5036)

Core-Web was running out of memory in the build function after the addition of Typescript.  Increased the allocation during build.


taraepp (dismissed):
👍

taraepp (approved):

henryoforeh-dev (approved):

PR #2507, opened 2023-04-27
[MDS-5224] TSF Update Error

taraepp:
### Some context
- adding new QP from the contacts page is very different from adding through the TSF page, where the contacts page was missing validation and fields that the TSF page had
- but the contacts page is the only way to view/edit historical data (and it does need editing) and I didn't want to close that by requiring that the appointment is the current one
- previous fix made it so that QP appt could be submitted through the contacts page but without the additional validations, so it was possible to create more than one active QP, which would cause an error the next time the form was submitted through the QP page
- the error it was triggering was not very readable

### Changes made
- fixed the error output by importing the correct BadRequest- which changed it from "HTTPStatusCode object isn't callable" to the intended message of "There is currently not exactly one active appointment"
- made the Start Date required when filling out the Add or Edit relationship form
- set end_current to true when the end date is undefined or in the future (today is considered the past)

[MDS-5224](https://bcmines.atlassian.net/browse/MDS-5224)

### DIFF
--- a/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
+++ b/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
@@ -1,6 +1,7 @@
 from datetime import datetime, timedelta
 from enum import Enum
-from http.client import BAD_REQUEST
+from werkzeug.exceptions import BadRequest
+
 
 from sqlalchemy import and_, nullsfirst, nullslast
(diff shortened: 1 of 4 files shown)


henryoforeh-dev (dismissed):
Nice work!

henryoforeh-dev (approved):

PR #2509, opened 2023-05-02
[MDS-5039] Adding gotchas.md

isuru-aot:
## Objective 

[MDS-5039](https://bcmines.atlassian.net/browse/MDS-5039)

As discussed in the retro, adding any useful findings (when fixing bugs) in a readme file.


services/core-api/sql/Gotchas.md
```diff
@@ -0,0 +1,21 @@
+# Gotchas
+
```
taraepp: So maybe I'm just not too keen on the gotchas file concept- I like info to go in "troubleshooting" sections of the related readme.
But I'm thinking anything specific to the database should go in the database readme (https://github.com/bcgov/mds/blob/develop/services/database/README.md) or migrations (https://github.com/bcgov/mds/blob/develop/migrations/README.md) or even the document manager (https://github.com/bcgov/mds/blob/develop/services/document-manager/backend/README.md), but this particular issue is a _very specific issue_, already captured in a bug ticket, and not really generalized to similar issues. I'm wondering what were all the "gotchas" that we wanted to capture, and is there a logical place to put all of them already- we did have an issue of documentation files not being very discoverable, so if it makes sense to create a gotchas file, it should be linked from the root README file.

isuru-aot: Thanks for the comment,

- If this is a _very specific issue_, the best thing would be adding this content to https://github.com/bcgov/mds/tree/develop/services/core-api#readme file under `troubleshooting` section. Because this issue is related to `core-api -> now-application API resource`.

- And root README seems so abstract to add this file as this is placed in core-api/sql 

What do you think?

taraepp: For root readme, if we're adding new documentation files, a link should be added in the root readme so that you can find it from the repo landing page. It functions kind of like a table of contents.

I'm just not sure about adding a bug to the readme- how many bugs should we add and then will there be a process to remove the bug from the repo after it's closed on Jira? It seems difficult to maintain. I think that's the purpose of Jira, moreso than project documentation.

isuru-aot: Well with these inputs, I'm having second thoughts, whether we want this gotha file or not. What others think @matbusby-fw @henryoforeh-aot ?

matbusby-fw: I think we can reformat this file a little without getting rid of it.  Maybe rather than call it Gotchas, call it something like "Datafix_SQLs.md"?

And then we can have a quick explanation on the use of each SQL file in that folder.  I think we can be fairly brief on the cause of the error, and just explain what the query is doing, how to run it, and what the result should be.

isuru-aot: Updated the PR with addressing the comments.

PR #2516, opened 2023-05-10
[MDS-5238] mines typescript conversion

henryoforeh-dev:
## Objective 

[MDS-5238](https://bcmines.atlassian.net/browse/MDS-5238)

- Converted MineAlert in core-web to typescript.
- Created necessary interface files.
- Converted MineAlert unit test file to typescript.
- Performed regression test (Create New Alert, Edit Active Alert, Remove Alert) to see if functionality is still working as expected
- Ran unit tests to ensure all is working as expected.


matbusby-fw (changes requested):
Looks awesome so far!  I was thinking the `mineReducer.js` file would be a good candidate to include in here as well for conversion.

Did this pass the check for the common folders matching?  Usually it insists that any changes within Core or Minespace's common folders be duplicated in the other frontend as well. **Ah, I see now that this file only exists on the Core side.  That makes sense**

services/core-web/src/actionCreators/mineAlertActionCreator.ts
```diff
@@ -0,0 +1,105 @@
...
+      })
+      .catch((err) => {
+        dispatch(error(reducerTypes.GET_MINE_ALERTS));
+        throw new Error(err);
+      })
+      .finally(() => dispatch(hideLoading()));
+  };
+
+export const createMineAlert =
+  (mineGuid: string, payload: IMineAlert): AppThunk<Promise<AxiosResponse<IMineAlert>>> =>
```
matbusby-fw: Was just chatting with Alex.  He had a good catch related to his PR.  I think the payload and the response will need their own interfaces as the data structure isn't the same.

henryoforeh-dev: Nice catch. I have updated my PR to fix this.

henryoforeh-dev:
> Looks awesome so far! I was thinking the `mineReducer.js` file would be a good candidate to include in here as well for conversion.
> 
> Did this pass the check for the common folders matching? Usually it insists that any changes within Core or Minespace's common folders be duplicated in the other frontend as well. **Ah, I see now that this file only exists on the Core side. That makes sense**

Oh, I just noticed this. currently fixing it.

services/common/src/interfaces/mineDocument.interface.ts
```diff
@@ -0,0 +1,11 @@
+export interface IMineDocument {
```
asinn134, line 1: @matbusby-fw @henryoforeh-aot @isuru-aot @taraepp @simensma-fresh  I was looking to setup the interface for MineDocument on my end as it is part of the response body of one of the functions in permit's actionCreator. I notice in the core-api response_models.py file the MINE_DOCUMENT_MODEL has a structure like this:

```
MINE_DOCUMENT_MODEL = api.model(
    'MineDocument', {
        'mine_document_guid': fields.String,
        'mine_guid': fields.String,
        'document_manager_guid': fields.String,
        'document_name': fields.String,
        'upload_date': fields.DateTime,
        'create_user': fields.String
    })
```

I am wondering now is there a difference in how response types are structured versus a type with the same name used in the reducer file? If so should we setup a folder called something like responses in the interface directory and place all the response types in there? or maybe I am over thinking this, please let me know.

henryoforeh-dev: When I created it initially, I added the extra attributes to test something but forgot to remove them. So the one you have there is correct. I will update my PR.

asinn134: Ah I see, ok thanks for clarifying I thought maybe there were two types with the same name in use but seperated.

asinn134 (dismissed):
Nice work~!

services/minespace-web/common/reducers/mineReducer.ts
```diff
@@ -1,13 +1,27 @@
...
 /**
  * @file mineReducer.js
  * all data associated with new mine/existing mine records is handled witnin this reducer.
  */
+interface MineState {
+  mines: IMine;
+  mineIds: string[];
+  mineNameList: IMine[];
+  minesPageData: IMine;
+  mineGuid: any;
```
matbusby-fw: Won't be able to avoid it entirely with these conversions, but it's probably best if we set as many `any` types to something more specific (ie. string)

services/minespace-web/common/reducers/mineReducer.ts
```diff
@@ -1,13 +1,27 @@
...
+  mineBasicInfoList: IMine[];
+  mineDocuments: IMineDocument[];
+  subscribedMines: IMine[];
+  mineComments: IMineComment[];
+  currentUserVerifiedMines: any[];
```
matbusby-fw: Do we know the shape of these?  Are they not an `IMine`?

matbusby-fw (changes requested):
Awesome!  This is looking great!  I just had a couple of nitpicky comments on the MineState interface.

matbusby-fw (approved):
Nice!

PR #2517, opened 2023-05-10
[MDS-5200] Fixed notifications not being displayed in core

simensma-fresh:
## Objective 

[MDS-5200](https://bcmines.atlassian.net/browse/MDS-5200)

Fixes an issue where notifications fail to load in core. The issue has been identified as IDIR usernames changing case from lower case to upper case during the SSO migration, and the activities endpoint being case-sensitive. 


taraepp (approved):
🎉

henryoforeh-dev (approved):
Nice!

PR #2521, opened 2023-05-15
[MDS-5264] Increased max node heap size on webpack workers

simensma-fresh:
## Objective 

[MDS-5264](https://bcmines.atlassian.net/browse/MDS-5264)

Increased max heap size of webpack workers to (hopefully) fix the out of memory errors that happen when building MineSpace


henryoforeh-dev (dismissed):
Solid!

matbusby-fw (approved):

henryoforeh-dev (approved):

PR #2522, opened 2023-05-15
[MDS-5249] Select Incident Timezone

taraepp:
## Objective 

[MDS-5249](https://bcmines.atlassian.net/browse/MDS-5249)

### How it works: 
- **props.displayFormat**: customize the input display (ie: show seconds, only show offset instead of abbrev, etc)
- **props.showTimezones**: default is BC timezones, customize the selection list (user's local timezone will be added to the list if it's within Canada)
- **props.timezone**: an override for the default timezone
- the component chooses the timezone in the order of
   - the saved form value
   - the timezone prop (intended for the mine's timezone)
   - the user's timezone, if it's within Canada (otherwise PT) 
- it expects there to be a corresponding field to save the timezone name, but this could be changed fairly easily in the future if we just want a correct UTC time
### How it looks like: 
(minor fix made after the gif- put form item in a Col, no change to functionality)
mds-5249-timezone


services/core-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -1009,6 +1016,6 @@ export default compose(
     form: FORM.ADD_EDIT_INCIDENT,
     enableReinitialize: true,
     touchOnBlur: true,
-    touchOnChange: false,
+    touchOnChange: true,
```
taraepp, line 1018: These forms were created with touchOnFalse: false and I have no idea why. The false setting created issues with validation messages showing up, and I did not notice any issues after I changed it.

services/core-web/src/tests/components/Forms/parties/__snapshots__/AddFullPartyForm.spec.js.snap
```diff
@@ -19,7 +19,7 @@ exports[`AddFullPartyForm renders properly 1`] = `
         <div
           className="center margin-large"
         >
-          <Memo(RadioGroup)
+          <Memo([object Object])
```
taraepp, line 22: The [object Object] looked more off than the other snapshot updates- I verified that this component still looks the same as it does on test.

services/minespace-web/common/constants/strings.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-key */
```
taraepp, line 1: Was getting "Missing key prop for element in array" on this file for MAP_UNAVAILABLE and CONTACT_ADMIN, which just didn't seem to make any sense in this context (lines 26, 35).

services/minespace-web/src/components/Forms/incidents/IncidentForm.js
```diff
@@ -310,15 +313,21 @@ const renderIncidentDetails = (childProps, formDisabled) => {
...
             id="incident_timestamp"
             name="incident_timestamp"
-            placeholder="Please select date and time"
-            component={renderConfig.DATE}
-            showTime
-            validate={[required, dateNotInFuture]}
             disabled={formDisabled}
+            validate={[dateNotInFutureTZ, required]}
+            component={(props) => (
+              <RenderDateTimeTz
```
taraepp: I did not put this component in config.js->renderConfig... controversial?

services/minespace-web/src/styles/overrides/antd-overrides.scss
```diff
@@ -457,11 +457,6 @@ th.ant-descriptions-item.vertical-description {
 // Radio
 .ant-radio .ant-radio-inner {
   border: 2px solid $primary-color;
-
-  &::after {
```
taraepp, line 461: antd fixed the centering of the radio inputs! 🎉

services/core-api/app/api/incidents/response_models.py
```diff
@@ -63,7 +63,8 @@
         'mine_name': fields.String,
         'mine_region': fields.String,
         'major_mine_ind': fields.Boolean,
-        'incident_timestamp': fields.DateTime,
+        'incident_timestamp': fields.DateTime(dt_format='iso8601'),
```
taraepp, line 66: Turns out this is all you need to do to use a timestamp with an offset (that and the parsing change). We're already using it with permits.

services/core-api/app/api/mines/incidents/resources/mine_incidents.py
```diff
@@ -125,6 +127,7 @@ def post(self, mine_guid):
         incident = MineIncident.create(
             mine,
             data['incident_timestamp'],
+            data['incident_timezone'],
```
simensma-fresh, line 130: @taraepp Will the timestamp be stored with the offset included, or will it be converted into UTC? (Ideally the latter 😅 )

taraepp: On the frontend I do a .toISOString() which converts it to UTC before sending it over (in the formatPayload function of MineIncident.js for Core).
Fun(?) thing I learned about the database:
If you go into the table for this field and manually change the offset, the value will adjust itself to have an offset of 0 so it always stays as UTC (so if I add an offset of 1hr, it will adjust the time by 1hr, and then save it with an offset of 0).

Short answer: converted to UTC

simensma-fresh: Ah missed that one!

Oo that's a fun fact! I had no idea that was a thing, pretty neat 📅

services/core-api/tests/helpers.py
```diff
@@ -0,0 +1,7 @@
+from datetime import datetime
+
+def get_datetime_iso8601_string(date: datetime):
```
taraepp, line 3: added some datetime testing helpers in a generic "helpers" file. Our version of python doesn't very easily generate the ISO 8601 string, but I thought this wasn't _too_ clunky.

services/core-web/src/styles/generic/layout.scss
```diff
@@ -115,9 +115,6 @@ html {
     height: 50%;
   }
 
-  & div {
```
matbusby-fw, line 118: This feels like it could have potential side-effects.  I've seen `inline-flex` used fairly frequently.  If we keep it, maybe give Jason a heads up to look out for odd spacing?

taraepp: I think I'll change it up a bit from what I have. I _believe_ the intention with this one is to space out the menu items in the top nav (selector compiles to `.inline-flex div`, the menu is the only change I notice).

image

The effect I wanted to remove was the "shadow" on the right side of this search tip, but I do see now that it also squishes the menu items together. I think I'll remove the rule and add a more specific one to target those items. For sure give a heads-up to Jason. 👍

matbusby-fw (dismissed):
Nice work!  Too many snaps!

henryoforeh-dev (dismissed):

simensma-fresh (dismissed):

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #2524, opened 2023-05-16
[MDS-5264] Skip waiting for rollout when promoting MineSpace with no changes

simensma-fresh:
## Objective 

[MDS-5264](https://bcmines.atlassian.net/browse/MDS-5264)
A good chunk of the "failed" Github action runs happened when trying to promote a service when no changes have been made to the service. This will cause `watch-deployment.sh ` to run until it times out after 15min as it will wait for a revision change that never happens.

Applied a fix for this to MS, will roll it out to the other services after doing some live tests

### DIFF
--- a/.github/workflows/minespace.deploy.test.yaml
+++ b/.github/workflows/minespace.deploy.test.yaml
@@ -42,8 +42,10 @@ jobs:
         with:
           oc: "4.7"
       - name: oc login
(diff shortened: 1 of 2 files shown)


matbusby-fw (approved):
🌮

taraepp (approved):
👍

asinn134 (approved):

PR #2530, opened 2023-05-23
[MDS-5255] Parties TypeScript conversion

matbusby-fw:
## Objective 

[MDS-5255](https://bcmines.atlassian.net/browse/MDS-5255)

Added interfaces to all parties action creators and reducers


asinn134 (dismissed):
Looks good, nice work!

asinn134 (approved):

PR #2532, opened 2023-05-25
[MDS-5251] Display Timezone

taraepp:
## Objective 
- update other parts of the UI to show the timezone properly (tables, emails)
- fixed a lot of UI bugs that I had previously created by upgrading antd
  - the icon one: several tables were using nearly identical functions to generate this icon, I integrated it into the CoreTable component
  - radio buttons: there was some prop overrides and type mismatches causing the selection to disappear, forced them all into strings
  -  dropdowns: the placement attribute was breaking it
- did some refactoring on the RenderDateTimeTZ component to make it nicer to use
   - made a normalize function to translate to ISO string
   - now it can optionally not show the time but still have a timezone (so instead of recording midnight UTC, will record midnight in the incident timezone)
   - set it up so that the timezone from the incident_timestamp is automatically used for other datetime inputs- this is the only one that's stored in the database so the only one I can count on for re-initializing the form.
   - had some state logic that wasn't working 100%, put in some useEffects for that
   - fixed it so that the field can be cleared

[MDS-5251](https://bcmines.atlassian.net/browse/MDS-5251)
[MDS-5291](https://bcmines.atlassian.net/browse/MDS-5291)
[MDS-5290](https://bcmines.atlassian.net/browse/MDS-5290)
[MDS-5287](https://bcmines.atlassian.net/browse/MDS-5287)


services/core-web/src/components/mine/AdministrativeAmendment/MineAdministrativeAmendmentTable.js
```diff
@@ -156,6 +134,7 @@ export class MineAdministrativeAmendmentTable extends Component {
         <div className="cap-col-height">
           {(trigger &&
             trigger.length > 0 &&
+            // eslint-disable-next-line react/jsx-key
```
taraepp, line 137: I maybe got a bit lazy here- I just have no idea how "item" is structured and what's appropriate to use as a key.

services/core-web/src/components/common/RenderCheckbox.js
```diff
@@ -18,7 +18,7 @@ const propTypes = {
 
 const RenderCheckbox = (props) => (
   <Form.Item validateStatus={props.meta.touched ? props.meta.error && "error" : ""}>
-    <Checkbox id={props.id} checked={props.input.value} {...props.input} disabled={props.disabled}>
+    <Checkbox id={props.id} {...props.input} disabled={props.disabled}>
```
asinn134, line 21: Is the `checked `property not valid anymore in the latest antd? Just wondering how removing it fixed the issue.

taraepp: I have some suspicions but I'm not 100% on the _why_.
- checked is still listed on the documentation, and we do have other checkboxes using that attribute that are working just fine
- the value was actually getting changed correctly, it just wasn't rendering it properly (console logging the form values showed that checking it did change the value)
- the props.input also passes a "value" that I think was in conflict with the checked attribute, I think the combination of checked/input.value/input.onChange was just not working.

asinn134: I see, good find!

services/core-web/src/components/common/RenderRadioButtons.js
```diff
@@ -22,42 +21,36 @@ const defaultProps = {
   customOptions: null,
 };
 
-class RenderRadioButtons extends Component {
-  state = { value: null };
+const RenderRadioButtons = (props) => {
```
matbusby-fw, line 24: 🙌

matbusby-fw (dismissed):
A lot to go through, but it looks great from what I can see!

simensma-fresh (dismissed):
👍

matbusby-fw (approved):

simensma-fresh (approved):

asinn134 (approved):
Nice work!

PR #2534, opened 2023-05-29
[MDS-5239] Converting rest of Mine Permit to Typescript

asinn134:
## Objective 

[MDS-5239](https://bcmines.atlassian.net/browse/MDS-5239)

- Converts MinePermitCard, PermitAmendmentFileUpload, PermitAmendmentUploadedFilesList, PermitConditionManagement, NOWActionWrapper, and AddCondition components to Typescript

- Converted PermitConditionManagement, NOWActionWrapper and AddCondition to functional components

- Sets the typing for the rest of Permit ActionCreators

- Converts the tests for the above components


services/core-web/src/components/mine/Permit/PermitAmendmentUploadedFilesList.tsx
```diff
@@ -28,6 +28,7 @@ export const PermitAmendmentUploadedFilesList = (props) => (
                 <p>Are you sure you want to remove this file?</p>,
                 <p>This cannot be undone.</p>,
               ]}
+              /* eslint react/jsx-key: 0 */
```
taraepp: Are these of type IPermitAmendmentMineDocument? Can we use that to assign a key?
If not, eslint-disable-next-line is a little easier to search

matbusby-fw: :100:

services/common/src/interfaces/dropDownListItem.interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface IDropDownListItem {
```
matbusby-fw: There is already an IOption interface here `services/common/src/interfaces/common/option.interface.ts`.  The only difference is that the value is `string`, but we can change it to `string | number` without any problem.

matbusby-fw (changes requested):
Looks awesome!  Just the one comment about a duplicate dropDownItem interface, and Tara's suggestion I think.  For the drop down, it doesn't matter which one we keep as long as we update wherever they're being used.

services/core-web/src/components/mine/Permit/PermitAmendmentUploadedFilesList.tsx
asinn134: This has been adjusted to include keys now. I honestly thought we only needed to add a key in the parent element in React mappings, so I was really confused about the error I was getting, but yeah adding keys to the paragraph elements solves this issue, thanks!

services/common/src/interfaces/dropDownListItem.interface.ts
asinn134: `IOption` has been adjusted now and `IDropDownListItem ` removed

matbusby-fw (approved):
👍

taraepp (approved):
👍

PR #2536, opened 2023-05-29
[MDS-5256] setup ts for activities

matbusby-fw:
## Objective 

Setup typescript for activities

[MDS-5256](https://bcmines.atlassian.net/browse/MDS-5256)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "husky": "8.0.2",
     "lint-prepush": "2.2.1",
     "lint-staged": "10.0.0",
-    "prettier": "2.8.0"
+    "prettier": "1.19.1"
   },
   "scripts": {
     "postinstall": "husky install"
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "husky": "8.0.2",
     "lint-prepush": "2.2.1",
     "lint-staged": "10.0.0",
-    "prettier": "2.8.0"
+    "prettier": "1.19.1"
```
taraepp, line 31: we had to downgrade prettier? 😮

matbusby-fw: The base package was set to 2.8.0, but Minespace, Core and the common package were all 1.19.1.  Making sure it's all on the same version.

taraepp (approved):

asinn134 (approved):
Nice!

PR #2541, opened 2023-05-31
[MDS-5309] Assorted UI bugs

taraepp:
## Objective 

[MDS-5309](https://bcmines.atlassian.net/browse/MDS-5309)

Dealing with these UI bugs:
image

image

image


matbusby-fw (approved):
OR, we could make the icons EVEN BIGGER!

asinn134 (approved):

PR #2545, opened 2023-06-02
[MDS-5318] Fixed sidebar scroll functionality after antd upgrade

simensma-fresh:
## Objective 

Fixed an issue where core web / Minespace crashes with a js error when using the scroll to hash functionality on pages that use the ScrollSideMenu/NowSideMenu components.

**Why?**
The `ref` property of the antd `Anchor` component was removed in antd `4.24.0` causing the `anchor` we try to scroll to to be undefined. 

**How?**
Switch to using plain old js query selector for scrolling as it currently  provides no suitable replacement for such functionality


henryoforeh-dev (approved):

taraepp (approved):
👍

PR #2546, opened 2023-06-02
[MDS-5315] Removed checking of timezone on legacy incidents

matbusby-fw:
## Objective 

updating a legacy incident was checking the validity of the timezone.  Added a check to only validate the timezone if the incident is not a legacy one.

[MDS-5315](https://bcmines.atlassian.net/browse/MDS-5315)


taraepp (approved):
Before I was thinking to force the updater to validate the timezone/timestamp, but I think I like this better- the updater does not necessarily know what the correct information is, so allowing them to skip it makes more sense.

(however, there _is_ still FE validation, so if they perform an action that "touches" the form --> triggers validation, they will still be forced to deal with it)

henryoforeh-dev (approved):

PR #2548, opened 2023-06-02
[MDS-5318] Upgraded Syncfusion to fix Doc Viewer error

simensma-fresh:
## Objective 

This should fix the `classLIst` error that popped up when using the document viewer in dev / test
image

**What caused it?**
Have no idea.
- Downgraded antd to previous version: it works!
- Put the document viewer outside the antd modal: It works!
- Upgraded to newest Syncfusion: It works!

So it looks like something in the new antd version made the current syncfusion document viewer version crash when running within the `Modal` component


matbusby-fw (dismissed):
Well huray for incidental upgrades!

henryoforeh-dev (dismissed):

isuru-aot (dismissed):

isuru-aot (approved):

PR #2550, opened 2023-06-02
[MDS-5318] Reverted syncfusion upgrade + fixed document viewer js error

simensma-fresh:
## Objective 

Reverted to the previous version of Syncfusion as it introduced this beautiful message

image

Managed to fix the js error in our current version by giving the PDF Viewer modal a container div (still don't fully understand why that fixed things 😓 )


matbusby-fw (approved):
Oh man.  That's unfortunate.  Don't we have a paid syncfusion licence though?

henryoforeh-dev (approved):

simensma-fresh:
@matbusby-fw  We do indeed, but will be for another time to upgrade i think

PR #2555, opened 2023-06-06
[MDS-5283] changing core upload area to match new mockup design

asinn134:
## Objective 

[MDS-5283](https://bcmines.atlassian.net/browse/MDS-5283)

- Updated Core upload area to match new mockup deisgn (https://www.figma.com/file/O7i6jpbKbju49HfL1OpozT/CORE-%26-Minespace-2021-2023?type=design&node-id=21066-63883)

image


matbusby-fw (dismissed):
Looks good!  Not strictly necessary, but one possible suggestion would be to make the fileTypes a prop for the fileUpload component so we don't have to keep rewriting the same text in the `labelIdle`

henryoforeh-dev (dismissed):

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #2557, opened 2023-06-07
[MDS-5329] Fixed error when opening the ESUP creation modal

simensma-fresh:
## Objective 

[MDS-5329](https://bcmines.atlassian.net/browse/MDS-5329)

Fixed an error when opening the ESUP creation modal.

**Cause**
Leaflet fails with an error complaining about a missing container - this seems to be an issue with the Modal component of the current version of antd, and is the same issue we experienced with Syncfusion


henryoforeh-dev (approved):

matbusby-fw (approved):
👍

PR #2558, opened 2023-06-08
[MDS-5286] Updating minespace uploader styling

isuru-aot:
## Objective 

[MDS-5286](https://bcmines.atlassian.net/browse/MDS-5286)

Update the styling for all the upload areas. And change the text from `Supporting Document Upload` to `Drag & drop your files or Browse`, the label in the other upload areas are kept as it is.

Screenshot 2023-06-08 100549

### DIFF
--- a/services/minespace-web/src/styles/overrides/filepond-overrides.scss
+++ b/services/minespace-web/src/styles/overrides/filepond-overrides.scss
@@ -5,7 +5,7 @@
 
 /* the text color of the drop label*/
 .filepond--drop-label {
-  color: $gov-grey-dark; //#555;
+  color: rgba(49, 49, 50, 1); //#555;
 }
 
 /* underline color for "Browse" button */
@@ -15,16 +15,6 @@
   &:focus {
     outline: none !important;
   }
-}
-
-/* the background color of the filepond drop area */
-.filepond--panel-root {
-  background-color: $white; //#eee;
-}
-
-/* the border radius of the drop area */
-.filepond--panel-root {
-  border-radius: $border-radius-base;
 }
 
 /* the border radius of the file item */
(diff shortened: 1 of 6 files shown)


services/minespace-web/src/styles/overrides/filepond-overrides.scss
```diff
@@ -5,7 +5,7 @@
 
 /* the text color of the drop label*/
 .filepond--drop-label {
-  color: $gov-grey-dark; //#555;
+  color: rgba(49, 49, 50, 1); //#555;
```
taraepp: I think this is the same as the variable $font-colour.

services/minespace-web/src/styles/overrides/filepond-overrides.scss
```diff
@@ -75,8 +65,8 @@
 
 /* bordered drop area */
 .filepond--panel-root {
-  background-color: transparent;
-  border: 2px solid $primary-color;
+  background-color: rgba(230, 235, 240, 1); /* the background color of the filepond drop area */
```
taraepp: Instead of a comment, would prefer to assign to variable. $filepond-drop-bg or something. 

I have a preference for hex format when we don't need the alpha channel, but that's probably just me. 😁

asinn134 (approved):

services/minespace-web/src/styles/overrides/filepond-overrides.scss
isuru-aot: Yep, you are correct. I corrected this with $font-colour.

services/minespace-web/src/styles/overrides/filepond-overrides.scss
isuru-aot: Updated the code accordingly.

taraepp (approved):
🎨

PR #2559, opened 2023-06-08
[MDS-5330] Adding spacing between input element and filepond element for ESUP

asinn134:
## Objective 

[MDS-5330](https://bcmines.atlassian.net/browse/MDS-5330)

- Adds space between input element and filepond element inside ESUP

previous:

image

new:

image

### DIFF
--- a/services/core-web/src/components/Forms/DocumentCategoryForm.js
+++ b/services/core-web/src/components/Forms/DocumentCategoryForm.js
@@ -78,8 +78,8 @@ export class DocumentCategoryForm extends Component {
                     <Popconfirm
                       placement="top"
                       title={[
-                        <p>Are you sure you want to remove this file?</p>,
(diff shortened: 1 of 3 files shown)


simensma-fresh:
Any way to get rid of the scrollbars around the document name/category section, or are they meant to be scrollable?

asinn134:
> 

@simensma-fresh that is a good question, let me check with Roop to make sure but from what I can see only this section uses the class that adds the scrollbars in the css. I think if we don't need it we can just remove it.

asinn134:
@simensma-fresh Confirmed with Roop that the scroll shouldn't be there. I replaced the spacing addition with just removing the css scroll.

henryoforeh-dev (approved):

simensma-fresh (approved):

PR #2560, opened 2023-06-08
[MDS-5331] Incidents 400 error

matbusby-fw:
## Objective 

[MDS-5331](https://bcmines.atlassian.net/browse/MDS-5331)

Updated the check for dangerous occurance status codes to only throw an error if they have been changed


asinn134 (approved):

simensma-fresh (approved):
👍

PR #2562, opened 2023-06-14
[MDS-5332] Fixing issue in updating mine status

isuru-aot:
## Objective 

[MDS-5332](https://bcmines.atlassian.net/browse/MDS-5332)

-  Fix the mine status selection
- Validate the status being updated correctly

Screenshot 2023-06-13 170057

### DIFF
--- a/services/core-web/src/components/common/RenderCascader.js
+++ b/services/core-web/src/components/common/RenderCascader.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { Form } from "@ant-design/compatible";
 import "@ant-design/compatible/assets/index.css";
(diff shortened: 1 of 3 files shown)


henryoforeh-dev (approved):

matbusby-fw (approved):
LGTM!

services/core-web/src/components/common/RenderCascader.js
```diff
@@ -23,28 +23,43 @@ const defaultProps = {
...
+    >
+      <Cascader
+        expandTrigger="hover"
+        id={props.id}
+        placeholder={props.placeholder}
+        options={props.options}
+        {...props.input}
+        changeOnSelect={props.changeOnSelect}
+        defaultValue={selectedValue}
+        value={handleCascaderChange(selectedValue)}
```
taraepp, line 57: I am curious why the change function is here for the value!

asinn134 (approved):

services/core-web/src/components/common/RenderCascader.js
isuru-aot: It handles the updates to the selected value. Without that the updated value doesn't get updated correctly.

taraepp: Interesting! I am surprised that it works for this attribute because it doesn't return a value. I have a feeling that if it were to be refactored to typescript that it would give an error.

PR #2564, opened 2023-06-15
[MDS-5345] Fix error while uploading document.

henryoforeh-dev:
## Objective 

[MDS-5345](https://bcmines.atlassian.net/browse/MDS-5345)

- Fixed a bug linked to uploading a document. This bug was caused by sending undefined/null **mines_guid**  as a query parameter to the project summary document upload end point.

- Updated swagger document

image


taraepp (approved):
LGTM

taraepp (changes requested):
Wait one thing- can we make sure the swagger is updated on the BE to reflect that the mine_guid is required?

simensma-fresh (approved):

taraepp (approved):

PR #2566, opened 2023-06-16
[MDS-5268] Document Versioning

matbusby-fw:
## Objective 

[MDS-5268](https://bcmines.atlassian.net/browse/MDS-5268)

- Created a custom `post-finish` hook in TUSD that returns the file version along with the creation timestamp
- added a migration to the docman database that adds a document_version table
- implemented functionality to save the document version when the `post-finish` hook is called

# Unfinished
- Started creating a post endpoint in docman that takes a document guid which will serve to update existing files
  - This is largely just a copy of the original POST function at this point (if anyone gets to it, feel free to use or get rid of it)

# Next Up
- Need to create a function in Core-Api that fetches the version information for files before their info is sent to the front end

### DIFF
--- a/services/document-manager/backend/app/docman/models/document.py
+++ b/services/document-manager/backend/app/docman/models/document.py
@@ -1,8 +1,6 @@
 import uuid
-from datetime import datetime
 
 from sqlalchemy.dialects.postgresql import UUID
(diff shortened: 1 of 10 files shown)


simensma-fresh:
Closing as a separate PR has been merged for this

PR #2567, opened 2023-06-17
[MDS-5048] Fixing EoR Status Change Issue

isuru-aot:
## Objective 

[MDS-5048](https://bcmines.atlassian.net/browse/MDS-5048)

- Fix the `Status` drop down so that users can select options under the status menu `Pending`, `Active`, `Inactive`.
- Fix the X button issue.

**Note**: Same issues were there for the `Ministry Acknowledged` drop down. So fix that as well except the default value.

Screenshot 2023-06-16 192331

### DIFF
--- a/services/core-web/common/components/PartyAppointmentTable.js
+++ b/services/core-web/common/components/PartyAppointmentTable.js
@@ -105,7 +105,7 @@ const PartyAppointmentTable = (props) => {
       dataIndex: "dates",
       render: (text, record) => (
         <div title="Dates">
-          {record.startDate} -{record.endDate}
+          {record.startDate} - {record.endDate}
         </div>
       ),
     },
--- a/services/core-web/common/components/PartyAppointmentTable.js
+++ b/services/core-web/common/components/PartyAppointmentTable.js
@@ -1,5 +1,5 @@
 import React, { useContext, useState } from "react";
-import { Col, Row, Table, Typography } from "antd";
+import { Col, Row, Table, Typography, Select } from "antd";
 import PropTypes from "prop-types";
 import { Field, FieldArray, change } from "redux-form";
 import { bindActionCreators } from "redux";
@@ -25,6 +25,8 @@ const defaultProps = {
   columns: [],
 };
 
+const { Option } = Select;
+
 const PartyAppointmentTable = (props) => {
   const { columns } = props;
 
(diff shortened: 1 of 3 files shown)


services/core-web/common/components/PartyAppointmentTable.js
```diff
@@ -85,14 +87,17 @@ const PartyAppointmentTable = (props) => {
       render: (status, record) => {
         if (isCore) {
           return (
-            <Field
-              id={`${record.rowName}.status`}
-              name={`${record.rowName}.status`}
-              component={renderConfig.SELECT}
-              data={statusColumns}
+            <Select
```
simensma-fresh: Curious as to why we're switching from `<Field>` to `<Select>`?
Seeing as we're using `redux-form` and `<Field component={renderConfig.SELECT}>` throughout the rest of the application, I would suggest that we stick to that if we can

simensma-fresh (changes requested):

services/core-web/common/components/PartyAppointmentTable.js
isuru-aot: Actually, that's something I wanted to discuss. I'll try harder to find a work around in the `Field`. Thanks.

isuru-aot: @simensma-fresh 
Could solve using the same `Field` component.

services/core-web/common/components/PartyAppointmentTable.js
```diff
@@ -104,10 +105,7 @@ const PartyAppointmentTable = (props) => {
       dataIndex: "dates",
       render: (text, record) => (
         <div title="Dates">
-          {record.startDate}
-          {' '}
--
-          {record.endDate}
+          {record.startDate} -{record.endDate}
```
taraepp: this is pretty minor but I think there's a space missing after the '-' before the end date

services/core-web/src/components/common/RenderSelect.js
```diff
@@ -53,7 +54,7 @@ const RenderSelect = (props) => {
       <Select
         virtual={false}
         disabled={props.disabled}
-        allowClear
+        allowClear={props.allowClear}
```
taraepp, line 57: nice improvement to the component

services/core-web/common/components/PartyAppointmentTable.js
isuru-aot: Nice catch. And I just fixed this.

taraepp (approved):
assuming checks pass LGTM 👍

henryoforeh-dev (approved):

simensma-fresh (approved):

PR #2568, opened 2023-06-19
[MDS-5270] Added functionality to archive documents in Core and MineSpace

simensma-fresh:
## Objective 

[MDS-5270](https://bcmines.atlassian.net/browse/MDS-5270)

Added functionality to archive project documents in Core and MineSpace.
- Added a bunch of project related filtering options to the Mine Document list endpoint a) to be used to look up archived files and b) (for a future story) to power the new "All documents" table
- Added `MineDocument.is_archived` to determine if a document has been archived
- Added Archive buttons to the document tables in "Project"
- Introduced a FeatureFlag utility to centralize feature flag definitions

**Note**
I've left the archive button as a button temporarily, as to keep the DocumentTable changes as minimal as I could. Will do a PR to move it into a dropdown once @taraepp's and @asinn134's DocumentTable changes are in

image
image
image
image


services/core-api/app/api/mines/documents/mine_document_search_util.py
```diff
@@ -0,0 +1,66 @@
...
+                    MajorMineApplication.project_guid == project_guid,
+                    ProjectSummary.project_guid == project_guid,
+                    ProjectDecisionPackage.project_guid == project_guid,
+                    InformationRequirementsTable.project_guid == project_guid,
+                ))
+
+        if major_mine_application_guid is not None:
+            qy = qy.join(MajorMineApplicationDocumentXref)\
+                .join(MajorMineApplication)\
+                .filter(MajorMineApplication.major_mine_application_guid == major_mine_application_guid)
```
henryoforeh-dev, line 54: I was wondering if there is need to also add deleted_ind == False as part of the filtering criteria.

henryoforeh-dev (dismissed):
I had only one comment but not sure if it makes much of a difference.

henryoforeh-dev (dismissed):

services/core-api/app/api/mines/documents/mine_document_search_util.py
taraepp: I have the same question! Perhaps defaulting the query to deleted_ind = false but allowing an override if we anticipate wanting to search them? But I remember that coming up in a conversation about search results- deleted results are showing up where the users don't want them to.

services/core-api/app/api/mines/documents/models/mine_document.py
```diff
@@ -41,11 +45,31 @@ def find_by_mine_document_guid(cls, mine_document_guid):
...
+        db.session.commit()
+
     def json(self):
         return {
             'mine_document_guid': str(self.mine_document_guid),
             'mine_guid': str(self.mine_guid),
             'document_manager_guid': str(self.document_manager_guid),
-            'document_name': self.document_name
+            'document_name': self.document_name,
+            'is_arhived': self.is_archived
```
taraepp: typo on is_archived!

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
```diff
@@ -1,26 +1,125 @@
...
+        if not mine:
+            raise NotFound('Mine not found.')
+
+        args = self.parser.parse_args()
+        mine_document_guids = args.get('mine_document_guids')
+
+        documents = MineDocument.find_by_mine_document_guid_many(mine_document_guids)
+
+        if len(documents) != len(mine_document_guids):
+            raise NotFound('Doucment not found')
```
taraepp: probably not very important but typo on Document here.

This seems like it could be difficult to troubleshoot- if you were to send 20 guids, 1 is missing, and trying to figure out which one it is. Maybe send back that guid or log it so we can find it in kibana?

services/core-api/app/api/mines/documents/mine_document_search_util.py
simensma-fresh: Good point! As far as I can tell we're actually not using the soft deletion functionality for the project related documents. When "deleting" them, we actually hard delete the Xref entry so they won't be included in the search at all. 
However definitely a good idea to add in the filter regardless for future proofing - will update the PR.

services/core-api/tests/mines/documents/test_mine_document_list_resource.py
```diff
@@ -0,0 +1,121 @@
...
+            headers=auth_headers['full_auth_header']
+        )
+
+        assert get_resp.status_code == 200
+        get_data = json.loads(get_resp.data.decode())
+
+        assert len(get_data['records']) == 0
+
+    def test_list_resource_one_document(self, test_client, db_session, auth_headers, setup_info):
+        """Should return no documents"""
```
taraepp: I think some of these descriptions got copy-pasted and not updated to match the test.

services/core-web/src/components/common/DocumentTable.js
```diff
@@ -96,7 +151,23 @@ export const DocumentTable = (props) => {
         </div>
       ),
     },
-  ];
+    !IN_PROD() && {
```
taraepp: can we put a name on this flag so it's easier to find?

services/core-api/tests/mines/documents/test_mine_document_list_resource.py
simensma-fresh: 😅

services/core-web/src/components/common/DocumentTable.js
simensma-fresh: @taraepp I got a bit carried away and introduced a `featureFlag` utility in the common package to centralize the flags and have them defined in one place 🏕️ 🔥 

Regardless of whether we end up defining them in ENV variables or using a Feature Flag tool, I think this will make that transition easier. Let me know your thoughts on this and I'll follow up with another PR to move the TSF flags over to this as well

taraepp: 🙌 
Oh I definitely meant just a comment but I do like this! Looks flexible enough to change the function in the featureFlags file to use a library or ENV variable down the road. 👍 

Heads-up: there is also a Projects feature flag still in effect. Before when I found that I just put a comment with FEATURE FLAG: PROJECTS on those instances.

taraepp (approved):
👍

henryoforeh-dev (approved):

services/core-web/src/components/common/DocumentTable.js
simensma-fresh: Started out as just a comment, buuuut couldn't help myself 😅

PR #2576, opened 2023-06-27
[MDS-5271] Adding function to view past versions of a file on front end

asinn134:
## Objective 

- Adds ability to view current and old versions of files in document table
- Adds icon to indicate to viewer how many past versions of a file exist
- Adds DocumentAction component for action drop down in document table, so far only has `Download file`, and `Open in Document Viewer`
- Only adjusted the tables found in Major Project's Final Application tab, and the `matchChildColumnsToParent ` prop is used to differentiate between the current structure of document tables and this PR's structure.

[MDS-5271](https://bcmines.atlassian.net/browse/MDS-5271)

image

image

image


matbusby-fw (dismissed):
Looks good to me!  It's not strictly necessary (especially on biggish PRs like this), but whenever possible, let's try to convert the .js files worked on within a PR to .ts/.tsx and class based components to functional.  I think that will be the best path forward to slowly updating everything.

henryoforeh-dev (dismissed):

services/core-web/src/components/common/DocumentTable.js
```diff
@@ -84,96 +139,178 @@ const withTag = (text, elem) => {
...
+              <span className="file-name-text" style={!record.numberOfVersions ? { marginLeft: "23px" } : {}}>{text}</span>
+              {record?.numberOfVersions > 0 ?
+                (<span className="file-history-container">
+                  <span>
+                    <Tooltip
+                      title={`This file has ${record.numberOfVersions} previous versions`}
+                      placement="top"
+                      mouseEnterDelay={1}
+                    >
+                      <span className="file-version-amount">
```
simensma-fresh: Would this one be a candidate for the [Tag](https://ant.design/components/tag) antd component? Used it for the Archived tag in the same table, and think it would be good to keep it consistent if possible 🙃

simensma-fresh (changes requested):
Just one little suggestion, otherwise looks good!

services/core-web/src/components/common/DocumentTable.js
asinn134: Hi Simen, this has been adjusted now, I changed `withTag` function a bit.

simensma-fresh (dismissed):
Awesome!

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -72,11 +74,31 @@ export const renderHighlightedTextColumn = (
   };
 };
 
-const withTag = (text: string, elem: ReactNode) => {
+const withTag = (record: any, elem: ReactNode, title: string) => {
   return (
-    <div className="inline-flex flex-between">
+    <div
+      className="inline-flex flex-between file-name-container"
+      style={!record.number_of_versions ? { marginLeft: "0" } : { marginLeft: "14px" }}
```
taraepp, line 81: I think this function needs a rename if it's going to be very specifically for document tags and not reusable for other use cases.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -111,18 +138,43 @@ export const renderDocumentLinkColumn = (
...
+          {text ?? placeHolder}
+        </div>
+      );
+      return withTag(record, content, title);
     },
     ...(sortable ? { sorter: nullableStringSorter(dataIndex) } : null),
   };
 };
+
+export const actionOperationsColumn = (title = "", dataIndex = "operations") => {
```
taraepp: This one is also just for document actions, not reusable for actions in general, needs rename. I would like to see a function where we can pass in an array like [{actionTitle, actionFunction, actionIcon}], but that may be out of scope for this ticket.

services/core-web/src/components/common/DocumentActions.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <div>Download File</div>
+        </button>
+      </Menu.Item>
+    </Menu>
+  );
+
+  return (
+    <div>
+      {/* @ts-ignore */}
+      <Dropdown overlay={menu} placement="bottomLeft">
```
taraepp: what are we ts-ignoring here?

services/core-web/src/components/common/CoreTable.tsx
```diff
@@ -9,6 +9,7 @@ interface CoreTableExpandConfig<T> extends ExpandableConfig<T> {
   rowKey?: string | ((record: any) => string);
   recordDescription?: string;
   subTableColumns?: ColumnType<any>[];
+  noSubTableExpandableRows?: boolean;
```
taraepp: Is there a reason not to just use rowExpandable = (record) => false? What is the goal here that this is serving?

services/core-web/src/components/common/DocumentTable.js
```diff
@@ -47,11 +57,55 @@ const defaultProps = {
...
+    case "PRM":
+      location = "Primary Document";
+      break;
+    case "SPT":
+      location = "Spatial Component";
+      break;
+    case "SPR":
+      location = "Supporting Document";
+      break;
+  }
```
taraepp: See src/constants/strings.js

We do have constants for MAJOR_MINES_APPLICATION_DOCUMENT_TYPE_CODE (this one might need pretty labels) and EMPTY_FIELD that we can refer to. Could probably just about index into it.

services/core-web/src/styles/components/Tables.scss
```diff
@@ -99,7 +99,7 @@ tr.ant-table-expanded-row, tr.ant-table-expanded-row:hover {
   }
 }
 
-.wrapped-text {
+.numberO .wrapped-text {
```
taraepp: There's a few different use cases for the .wrapped-text class, are you sure they're all within the .number0 class? Is the number0 class like the antd class for indicating the row is level 0 or what is its use case?

services/core-web/src/components/common/DocumentTable.js
```diff
@@ -47,11 +57,55 @@ const defaultProps = {
...
+const parseFiles = (versions, documentType) =>
+  versions.map((version, index) => ({
+    key: version.mine_document_version_guid,
+    file_location: renderFileLocation(documentType) || Strings.EMPTY_FIELD,
+    file_type: renderFileType(version.document_name) || Strings.EMPTY_FIELD,
+    number_of_versions: index === 0 ? versions.length - 1 : 0,
+    ...version,
+  }));
+
+const transformRowData = (document) => {
```
taraepp, line 81: I think I see the issue-
to use the table as designed, we want each record to look something like:
document_guid: abc,
upload_date: today,
...(other attributes about that document),
oldVersions: [{version number, guid, upload date, etc}]

where the old versions are an array attribute of the document. Then we can use the table normally, with getDataSource = record => record.oldVersions, and just pass in columns for the latest version, and another array of columns for the expanded rows. The table will handle the rest without any changes to it.

services/core-web/src/components/common/DocumentActions.tsx
taraepp: I just noticed that I'm getting (from a different component) console errors about this- `Warning: [antd: Dropdown] overlay is deprecated. Please use menu instead.` <-- can we use menu?

services/core-web/src/components/common/CoreTableCommonColumns.tsx
asinn134: This has been renamed now to `documentWithTag`

services/core-web/src/components/common/CoreTableCommonColumns.tsx
asinn134: This has been renamed now as well

services/core-web/src/components/common/CoreTable.tsx
asinn134: As discussed, this has been renamed to `matchChildColumnsToParent` and the purpose of this prop is because this PR's changes only applies to major project's final tab for the moment and having this prop is so we can keep using the current document table structure and slowly move to this PR's document table structure for other parts. So ultimately this prop is kind of temporary.

services/core-web/src/components/common/DocumentTable.js
asinn134: I ended up adding a new variable to the strings.js

```
export const MAJOR_MINES_APPLICATION_DOCUMENT_TYPE_CODE_LOCATION = {
  PRM: "Primary Document",
  SPT: "Spatial Component",
  SPR: "Supporting Document",
};
```

I think MAJOR_MINES_APPLICATION_DOCUMENT_TYPE_CODE is being used elsewhere and changing the keys in this object would affect the places it is being used, but if I didn't change the keys I think I would need to adjust text case (upper to lower) and add text like Document or Component to what we want. I think just adding another variable probably just simplifies things.

services/core-web/src/styles/components/Tables.scss
asinn134: `.number0` was removed from this, it is back to the way it is in developer branch

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -93,16 +125,21 @@ export const renderDocumentLinkColumn = (
     dataIndex,
     key: dataIndex,
     render: (text = "", record: any) => {
+      const recordWithArchiveIndicator = { ...record, showArchiveIndicator };
       const link = (
-        <div key={record.key ?? record[docManGuidIndex]} title={title}>
+        <div
```
taraepp, line 120: I think the title attribute may have gotten removed by accident. I think that the title attribute is for showing on narrow screens when it's too small for a table.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -93,16 +125,21 @@ export const renderDocumentLinkColumn = (
...
+          key={record.key ?? record[docManGuidIndex]}
+          className={record.number_of_versions !== undefined ? "file-name-text" : ""}
+          style={record?.number_of_versions === 0 ? { marginLeft: "38px" } : {}}
+        >
           <DocumentLink
             documentManagerGuid={record[docManGuidIndex]}
-            documentName={text}
+            documentName={record.document_name ?? text}
```
taraepp: This should be completely reliable as "text", should be no need for looking at document_name.
The contents of text in column render function is determined by antd internally by record[dataIndex], so just need to pass in the right attribute name to the function.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -111,18 +148,46 @@ export const renderDocumentLinkColumn = (
...
     ...(sortable ? { sorter: nullableStringSorter(dataIndex) } : null),
   };
 };
+
+export const documentActionOperationsColumn = (title = "", dataIndex = "operations") => {
+  return {
+    title,
+    dataIndex,
+    key: dataIndex,
+    render: (text: any, record: any) => {
```
taraepp: Does this need a dataIndex- are we expecting the records to have an "operations" attribute that we are using to index into the data?
If not, I'd recommend taking out dataIndex from the return, and change the render to `render: (record: any) => ...`

services/core-web/src/components/common/CoreTableCommonColumns.tsx
asinn134: The title attribute is added within `documentWithTag` function to a div element that wraps around the html content of the link variable. Originally I had left it there but noticed title text was showing up twice.

services/core-web/src/components/common/DocumentActions.tsx
asinn134: Honestly I used ts-ignore here mainly because I noticed other files in the project would have the ts-ignore above the dropdown component. I thought it might of been because overlay was deprecated, but I found even if I were to remove the ts-ignore it still works with overlay? might be missing some context but I am not exactly too sure why ts-ignore was used.

As suggested this has been changed to use menu instead of overlay now.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
asinn134: Ah I see, this has been reverted back to `text` now

services/core-web/src/components/common/CoreTableCommonColumns.tsx
asinn134: Thanks for catching this, your right I don't think we will need this, this has been removed now.

henryoforeh-dev (dismissed):

services/core-web/src/components/common/DocumentActions.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <div>Download File</div>
+        </button>
+      ),
+    },
+  ];
+
+  return (
+    <div>
+      <Dropdown menu={{ items }} placement="bottomLeft">
+        {/* @ts-ignore */}
```
taraepp: when we're putting in a ts-ignore, I really think it needs a comment saying which rule we're ignoring and why, if we aren't able to fix it.

henryoforeh-dev (approved):

services/core-web/src/components/common/DocumentActions.tsx
asinn134: Hi @taraepp this has been removed now, the ts-ignore was for `type="secondary"` but just tested and it works the same without it.

taraepp (approved):

PR #2577, opened 2023-06-27
[MDS-5340] implement cypress authentication

henryoforeh-dev:
## Objective 

[MDS-5340](https://bcmines.atlassian.net/browse/MDS-5340)

Implemented keycloak for cypress integration testing

### DIFF
--- a/cypress/realm-export.json
+++ b/cypress/realm-export.json
@@ -1,0 +1,2104 @@
+{
+  "id": "standard",
+  "realm": "standard",
+  "notBefore": 0,
+  "defaultSignatureAlgorithm": "RS256",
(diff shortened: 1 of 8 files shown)


cypress/realm-export.json
```diff
@@ -0,0 +1,2104 @@
+{
```
taraepp, line 1: I am curious how this was generated! But all looks good to me

taraepp (approved):

cypress/realm-export.json
henryoforeh-dev: I exported from the keycloak portal I installed locally.

matbusby-fw (approved):
👍

PR #2579, opened 2023-06-27
[MDS-5282] redesign table

taraepp:
## Objective 
Make tables consistent
- switched to typescript
   - used generics, extended base props so that functionality isn't limited but reasonable defaults are set 
- introduced a common column types file to generate common columns
- cleaned up document tables so that we're not, for example, changing every instance of "document_name" of a record to "name".
  - document columns call the common column types functions so that they're consistent with other columns and between document tables

[MDS-5282](https://bcmines.atlassian.net/browse/MDS-5282)
Bonus:
[MDS-5341](https://bcmines.atlassian.net/browse/MDS-5341)


services/core-web/src/components/common/CoreTable.tsx
```diff
@@ -0,0 +1,127 @@
+import React from "react";
```
taraepp, line 1: most important change!

Complete refactor of this file.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -0,0 +1,95 @@
+import React from "react";
```
taraepp: common file for generating columns.
Did not refactor _every_ table to use these functions, but some to test it out.

Some fun facts about columns!
The _dataIndex_ determines the first parameter that is passed to the render function. If there is no dataIndex, it will pass the entire record as the first parameter (so you can get rid of unused "text" parameters for things like action columns).
The _key_ is used for things like sorting and drag-dropping, it's good practice to give every column a key.

services/core-web/src/components/common/DocumentTable.js
```diff
@@ -32,76 +31,22 @@ const defaultProps = {
   additionalColumnProps: [],
   documentColumns: null,
   documentParent: null,
+  defaultSortKeys: ["upload_date", "dated"], // keys to sort by when page loads
```
taraepp, line 50: Not super visible here, but changed the keys that this table uses (now they're default parameters in the document columns file) to match the ones commonly returned by the API so we're not forced to do the same data transformation or overrides every time.

services/core-web/src/components/common/DocumentColumns.js
```diff
@@ -1,61 +1,52 @@
...
+  renderTextColumn,
+} from "./CoreTableCommonColumns";
 
-export const categoryColumn = (categoryDataIndex, documentCategoryOptionsHash) => {
-  return {
-    title: "Category",
-    dataIndex: categoryDataIndex,
-    render: (text) => <div title="Category">{documentCategoryOptionsHash[text]}</div>,
-  };
+export const documentNameColumn = (
```
taraepp, line 12: changed defaults here for data index to match data, and column headers to match Figma. Called other functions for more consistency. Kept them as separate functions here so the headers stay consistent.

services/core-web/common/utils/helpers.ts
```diff
@@ -141,8 +141,7 @@ export const dateSorter = (key) => (a, b) => {
   if (!b[key]) {
     return -1;
   }
-  // @ts-ignore
-  return moment(a[key]) - moment(b[key]);
+  return moment(a[key]).diff(moment(b[key]));
```
taraepp, line 144: changed this to use moment function

services/core-web/src/components/Forms/noticeOfWork/VerifyNoWContacts.js
```diff
@@ -108,10 +109,9 @@ const renderContacts = ({
               </p>
             </Col>
             {fields.map((field, index) => {
-              // eslint-disable-next-line no-unused-expressions
-              fields.get(index).id
-                ? (fields.get(index).id = fields.get(index).id)
-                : (fields.get(index).id = uuidv4());
+              if (!fields.get(index).id) {
```
taraepp, line 112: eslint was pretty angry about this bit of code and I also didn't like the use of ternary here.

services/core-web/src/components/Forms/noticeOfWork/VerifyNoWContacts.js
```diff
@@ -586,19 +586,14 @@ export class VerifyNoWContacts extends Component {
...
                   onSearch={(searchTerm) => this.handleSimpleSearch(searchTerm)}
                   onChange={(e) => this.setState({ searchTerm: e.target.value })}
-                  size="large"
                 />
                 <br />
                 <LoadingWrapper condition={!this.state.isLoading}>
-                  <Table
+                  <CoreTable
                     className="party-table"
-                    align="left"
```
taraepp, line 595: Table doesn't have align as a property anymore. _Columns_ do, but not tables.

services/core-web/src/components/mine/Compliance/ComplianceOrdersTable.js
```diff
@@ -67,70 +55,33 @@ const columns = [
...
-    render: (text, record) => <div title="Order No.">{record.order_no || "-"}</div>,
-    sorter: (a, b) => (a.order_no > b.order_no ? -1 : 1),
-  },
+  renderTextColumn("order_no", "Order No.", true, "-"),
   {
     title: "Violation",
     key: "violation",
     dataIndex: "violation",
-    render: (text, record) => <div title="Violation">{record.violation || "-"}</div>,
+    render: (text) => <div title="Violation">{text || "-"}</div>,
```
taraepp, line 71: this was something I wasn't sure about- in other places where there's no cell data, we use "N/A" or just blank. Here we are using "-" which we do nowhere else.

services/core-web/src/components/mine/NoticeOfWork/MineApplications.js
```diff
@@ -92,14 +91,6 @@ export class MineApplications extends Component {
     );
   };
 
-  onExpand = (expanded, record) =>
```
taraepp, line 95: Not necessary to write logic to handle row expansion, or have expandedRows state variables, _unless_ expansion is triggered by anything other than the user clicking the expand icon.

services/core-web/src/components/modalContent/NoticeOfDepartureModal.tsx
```diff
@@ -423,6 +381,15 @@ const NoticeOfDepartureModal: React.FC<
             </Button>
           ) : (
             <>
+              <Popconfirm
```
taraepp, line 384: this was a UI thing that bothered me- we had it like [Update][Cancel] which is the reverse of what we usually do. There was also a class on it that made the border size change when you hovered on it.

services/core-web/src/components/noticeOfWork/applications/permitGeneration/PermitAmendmentTable.js
```diff
@@ -61,23 +61,21 @@ export class PermitAmendmentTable extends Component {
...
+                  {renderDocumentLink(file)}
+                </li>
               ))}
             </ul>
           </div>
         ),
       },
     ];
-
+    console.log(this.transformRowData(this.props.permit.permit_amendments || []));
```
taraepp: whoops

services/core-web/src/styles/components/Tables.scss
```diff
@@ -7,8 +7,21 @@
   position: relative;
 }
 
-.ant-table-body {
+.ant-table-content {
   overflow-x: auto;
+  .ant-table-tbody {
```
taraepp, line 12: horizontal scrolling.

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkTable.js
```diff
@@ -126,7 +126,7 @@ export class NoticeOfWorkTable extends Component {
           <Input
             id={field}
             ref={(node) => {
-              this.searchInput = node && node.props.value;
+              this.searchInput = node && node?.input?.value;
```
taraepp, line 129: the search bug MDS-5341. I'm assuming antd changes went from props --> input.

.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   parser: "@typescript-eslint/parser",
   ignorePatterns: [
     "services/minespace-web/webpack.config.ts",
+    "services/core-web/webpack.config.ts",
```
taraepp, line 9: to resolve an eslint error I was getting

services/core-web/src/components/common/CoreTable.tsx
matbusby-fw: Very nice refactor!

services/core-web/src/components/mine/Compliance/ComplianceOrdersTable.js
matbusby-fw: That is odd.  Maybe a question for the design team, although I suspect they will want us to make it consistent with the rest of the site.

services/core-web/src/components/mine/Compliance/ComplianceOrdersTable.js
```diff
@@ -67,70 +55,33 @@ const columns = [
...
-    sorter: (a, b) => (a.inspector > b.inspector ? -1 : 1),
-  },
-  {
-    title: "Order Status",
-    key: "order_status",
-    dataIndex: "order_status",
-    render: (text, record) => <div title="Order Status">{record.order_status || "-"}</div>,
-    sorter: (a, b) => (a.order_status > b.order_status ? -1 : 1),
-  },
+  renderTextColumn("report_no", "Report No.", true, "-"),
```
matbusby-fw, line 74: Big fan of this `renderTextColumn` in general.  These columns are so bulky, this will make things much cleaner.

services/core-web/src/components/mine/Compliance/ComplianceOrdersTable.js
taraepp: I listed it on the Jira ticket as an issue I didn't deal with. I figure "make empty table cells consistent" can be its own ticket. Starting point would just be setting those "placeHolder" params on column functions = "N/A" (or whatever UX wants)

matbusby-fw (approved):
I think this is a new record for files changed for you.  Really nice work.  Hard to be completely thorough reviewing this many changes, but it all looks solid from what I can see!

henryoforeh-dev (approved):
Nice work Tara. A lot of effort went into this redesign.

PR #2582, opened 2023-06-30
[MDS-5268] Added backend support for document version management

simensma-fresh:
## Objective 

[MDS-5268](https://bcmines.atlassian.net/browse/MDS-5268)

Added backend support for document version management this includes

**Core API**
- `POST /<string:mine_guid>/documents/<string:mine_document_guid>/versions/upload` to initiate a new document version upload with docman and tustd. This creates a new `DocumentVersion` in docman.
- `POST /<string:mine_guid>/documents/<string:mine_document_guid>/versions` creates a `MineDocumentVersion` in core (should be used once the docman file upload has finished)
- `MineDocumentVersion` model has been added to track versions within core
- `MineDocument` now has a `versions` array returned that contains all new versions

**Docman**
- `POST /documents/<string:document_guid>/versions` creates a new DocumentVersion within docman and initiates the upload with tusd. Very similar to the `POST /documents` endpoint, so refactored that endpoint to share the common functionality
- Created DocumentVersion model to keep track of uploaded versions. Contains the `object_store_version_id` property which is the file version returned by S3
- Updated tusd hooks to record S3 version


services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -0,0 +1,148 @@
...
+    FILE_UPLOAD_PATH, FILE_UPLOAD_EXPIRY, TIMEOUT_24_HOURS, TUS_API_VERSION, TUS_API_SUPPORTED_VERSIONS, \
+    FORBIDDEN_FILETYPES
+from app.extensions import cache
+from flask import request, current_app, make_response, jsonify
+from werkzeug.exceptions import BadRequest, NotFound, RequestEntityTooLarge, InternalServerError, BadGateway
+
+CACHE_TIMEOUT = TIMEOUT_24_HOURS
+
+
+class DocumentUploadHelper:
```
matbusby-fw, line 19: Nice!  I was confused at first by the deletion of all the code from document.py.

matbusby-fw (dismissed):
Goooo Team!!  Looks great!  Thanks for taking over on this!

services/document-manager/backend/app/docman/resources/document_version_list_resource.py
```diff
@@ -0,0 +1,68 @@
...
+            version_guid=version_guid)
+
+        # Create document record
+        new_version = DocumentVersion(
+            id=version_guid,
+            document_guid=document.document_guid,
+            created_by='mds',
+            created_date=datetime.utcnow(),
+            file_display_name=filename,
+            upload_started_date=datetime.utcnow(),
```
taraepp, line 64: I'm probably missing something, but:
1. Will created_date and upload_started_date always be the same values? Do we _need_ both?
2. Is there a different spot where new versions uploaded by users are processed where their username gets put in the created_by field?
3. The bit with the id/version_guid: on ln 48 a version_guid is generated, but ln 58 it could just as easily be handled by the db, since it doesn't use the response from ln 50. Was the object_store_path meant to be passed in to the DocumentVersion?
4. Nitpick: I don't like the column names where "created_date" is a _timestamp_.

simensma-fresh: Answers to 1., and 4.: I do not like it particularly either. In practice created_date and upload_started_date will be the same (can't think of a scenario they wouldn't). There might be a legacy reason for keeping them both on the Document modal that I'm not aware of (maybe import of documents from MMS?). The only reason I kept both is to keep it as consistent with the `Document` model as I could. Happy to change it up though

simensma-fresh: 3: The reason for generating the version_guid "manually" is because the version is passed along to Tusd to initialize the document upload (document_upload_helper - initiate_document_upload line 22), that way we will avoid orphan `DocumentVersions` if it gets created by the DB and the Tusd initialization fails for any reason.

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):
Approved since it is only rebasing that was done.

taraepp (approved):

PR #2583, opened 2023-06-30
[MDS-5335] Editing Minespace User Mines

taraepp:
## Objective 
- in a modal, the multiselect component dropdown covers the input, preventing the users from searching. Make it visible.
- tons of issues came up in the console, dealt with a couple of them
  - not console issue, but sorted the list of mines for easier reading
  - UpdateMinsespaceUser: not previously flagged as an issue, but combining filterUserMines() to the list of mines was showing all of the user's mines in the dropdown more than once. 
  - helpers: antd complained about using props instead of children directly
  - AddPartyComponentWrapper: the componentWillMount function is deprecated, and not necessary anyway because the modal is destroyed on close now. Not sure why eslint was complaining about the render function return, maybe an issue with the Carousel?
  - UpdateMinespaceUser: both the parent and child of this component called it handleSubmit
  - PropTypes.minespaceUser: changed it to the field that's being used here.

[MDS-5335](https://bcmines.atlassian.net/browse/MDS-5335)

This one is more of a bandaid, if there is a better way to determine from within the multiselect that it's within a modal, I would like to know it. The multi select component really does need to know if it's within a modal or not in order to render correctly- because it puts calculated styles directly on the element, a CSS solution isn't very effective. I could have removed the getPopupContainer function entirely, but I know that this affects scroll outside of modals and there were many instances of the multi select component that would have to be checked.

### DIFF
--- a/services/core-web/common/utils/helpers.ts
+++ b/services/core-web/common/utils/helpers.ts
@@ -166,7 +166,7 @@ export const sortListObjectsByPropertyDate = (list, property) => list.sort(dateS
 
 // Case insensitive filter for a SELECT field by label string
 export const caseInsensitiveLabelFilter = (input, option) =>
-  option.props.children.toLowerCase().includes(input.toLowerCase());
+  option.children.toLowerCase().includes(input.toLowerCase());
 
 // function taken directly from redux-forms (https://redux-form.com/6.0.0-rc.1/examples/normalizing)
 // automatically adds dashes to phone number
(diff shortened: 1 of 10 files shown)


matbusby-fw (approved):
Looks good!  Had to look up the `getPopupContainer` property for the select component.  I can't think of a better way to detect a modal automatically, so I think you're solution fits!

henryoforeh-dev (approved):

PR #2584, opened 2023-07-05
[MDS-5365] converted document datetimes to strings

matbusby-fw:
## Objective 

The `upload_date` in `mine_document` was being returned as a datetime.  Changed this to a string.

[MDS-5365](https://bcmines.atlassian.net/browse/MDS-5365)

### DIFF
--- a/services/core-api/app/api/mines/response_models.py
+++ b/services/core-api/app/api/mines/response_models.py
@@ -99,7 +99,7 @@ MINE_DOCUMENT_VERSION_MODEL = api.model(
         'document_manager_guid': fields.String,
         'document_manager_version_guid': fields.String,
         'document_name': fields.String,
-        'upload_date': fields.DateTime(dt_format='iso8601'),
+        'upload_date': fields.String,
         'create_user': fields.String,
     })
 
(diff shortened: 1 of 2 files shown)


henryoforeh-dev (dismissed):

asinn134 (dismissed):

asinn134 (dismissed):

services/core-api/app/api/mines/response_models.py
```diff
@@ -109,7 +109,7 @@ def format(self, value):
         'mine_guid': fields.String,
         'document_manager_guid': fields.String,
         'document_name': fields.String,
-        'upload_date': fields.DateTime,
+        'upload_date': fields.String,
         'create_user': fields.String,
         'is_archived': fields.Boolean,
         'archived_date': fields.DateTime,
```
taraepp: is the archived date going to create the same problem?
I wonder if the format `fields.DateTime(dt_format='iso8601')` would also work.

taraepp (dismissed):

services/core-api/app/api/mines/response_models.py
matbusby: Good point.  Found another one in incident document model as well.  `fields.DateTime(dt_format='iso8601')` works too!

taraepp (dismissed):

henryoforeh-dev (dismissed):

services/core-api/app/api/mines/response_models.py
matbusby: Nevermind.  fields.DateTime(dt_format='iso8601') actually breaks the test, but str is returning it in ISO format at any rate.

henryoforeh-dev (approved):

PR #2586, opened 2023-07-07
[MDS-5367] Notice of Work Search

taraepp:
## Objective 

[MDS-5367](https://bcmines.atlassian.net/browse/MDS-5367)
- fixed the bug by making it distinguish between empty string (user cleared) and undefined (never entered)
- some TS-ification
   - I wanted a pageData interface and there was already a typed one, but I like generics so I changed it to generic (hence changes to parties)
   - staticContentSelectors was making TS compiler really angry- something to do with compiler picking the wrong function overload when parameters are optional and then complaining about wrong number of parameters. I find these functions hard to understand so I'm liking the clarity of adding the signature
   - throwing types in all over the place
   - I didn't even notice that helpers.ts had vars and lets in there that got changed to const, must have been prettier doing it for me
- little UI fix on the search box

before UI fix:
image

after UI fix:
image


services/core-web/common/utils/helpers.ts
```diff
@@ -46,18 +45,17 @@ export const createItemMap = <T>(array: T[], idField: string): ItemMap<T> => {
...
+  valueField: string,
+  isActiveField: string | false = false,
   subType = null,
   labelFormatter = null,
   orderByAlphabetically = true
 ) => {
   const options = array.map((item) => ({
     value: item[valueField],
     label: labelFormatter ? labelFormatter(item[labelField]) : item[labelField],
-    // @ts-ignore
```
taraepp, line 60: I don't like ts-ignores, just fixed the isActiveField type.

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.tsx
```diff
@@ -0,0 +1,167 @@
...
+} from "@common/selectors/noticeOfWorkSelectors";
+import * as routes from "@/constants/routes";
+import NoticeOfWorkTable from "@/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkTable";
+import NoticeOfWorkSearch from "@/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkSearch";
+import ResponsivePagination from "@/components/common/ResponsivePagination";
+import { PageTracker } from "@common/utils/trackers";
+import { INoticeOfWorkApplication, IPageData, IOption } from "@mds/common";
+import { RootState } from "@/App";
+
+export interface NoWSearchParams {
```
henryoforeh-dev, line 26: Will this interface be used in other pages ?

taraepp: I'm using it in both the HomePage and the Table file, but I don't _expect_ to use it anywhere else. Wasn't sure where the best place to put it was.

henryoforeh-dev: I was wondering if core-web/common would be a good place to put it.

taraepp: My thinking is that the interface is specifically for search parameters for this page, and so it belongs with the page. I think it would be clutter/obsfucation in core-web/common. ie: within the common directory, where would it make sense? Make a directory /pages/searchParams? Would we be moving it to the common directory just because it's an interface (and not for props) but not because it's needed by both core-web and minespace-web?

henryoforeh-dev: Sounds good. Maybe pages/searchParams can work, but I am also okay with leaving it as is since it is used by only 2 components.

henryoforeh-dev (approved):

services/common/src/interfaces/common/pageData.interface.ts
```diff
@@ -1,7 +1,5 @@
-import { IParty } from "@/index";
-
-export interface IPartiesPageData {
-  records: IParty[];
+export interface IPageData<T> {
```
matbusby-fw, line 1: Woo!  🙌

services/core-web/src/components/dashboard/noticeOfWorkHomePage/NoticeOfWorkHomePage.tsx
matbusby-fw: I think I prefer leaving it where it is for this case.  That, or maybe create it's own interface file within this same folder?

core-web/common will eventually be deleted hopefully, and this isn't an interface that is needed in Minespace, so I think placing it there wouldn't be the best fit.  Could maybe put it in src/interfaces for core-web.

matbusby-fw (approved):
Looks great to me.  Gave my two cents on the interface location as well.

PR #2590, opened 2023-07-14
[MDS-5361] Clean Up Console Logs

sjudzentis:
## Objective 

[MDS-5361](https://bcmines.atlassian.net/browse/MDS-5361)

Cleaning up local console logs as there are too many. Removing redux logger.

### DIFF
--- a/services/core-web/package.json
+++ b/services/core-web/package.json
@@ -81,7 +81,6 @@
     "redux": "4.0.5",
     "redux-form": "8.2.6",
     "redux-form-input-masks": "2.0.1",
-    "redux-logger": "3.0.6",
     "redux-mock-store": "1.5.4",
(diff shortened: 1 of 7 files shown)


services/core-web/package.json
```diff
@@ -81,7 +81,6 @@
     "redux": "4.0.5",
     "redux-form": "8.2.6",
     "redux-form-input-masks": "2.0.1",
-    "redux-logger": "3.0.6",
```
matbusby-fw, line 84: Yeah!  Get outta here redux-logger!  We don't need your spam! 🖕

matbusby-fw (approved):
Love it!  Thanks Scott!

taraepp (approved):
👍

PR #2591, opened 2023-07-17
[MDS-5337] Setup Cypress for MineSpace

henryoforeh-dev:
## Objective 

[MDS-5337](https://bcmines.atlassian.net/browse/MDS-5337)

- Setup Cypress for MineSpace.
- Documentation on setting up cypress.


.eslintrc.js
```diff
@@ -6,6 +6,8 @@ module.exports = {
   parser: "@typescript-eslint/parser",
   ignorePatterns: [
     "services/minespace-web/webpack.config.ts",
+    "services/minespace-web/cypress.config.ts",
+    "services/minespace-web/cypress/**",
```
taraepp, line 10: I am curious how many or what kind of eslint errors came up in the cypress/** directory! Would be nice to have it yell at us when we forget to remove an unused import or use a map function when we want a forEach.

taraepp (approved):
👍 
I'm curious about the eslint settings (I did notice some eslint-ignores in files that I thought were ignored entirely) but not a blocking issue. Great work getting cypress added in! 🎉

asinn134 (approved):

PR #2593, opened 2023-07-20
[MDS-5379] db error in final application page

asinn134:
- Database error in final application section of major projects was found to be caused by the optional query parameter major_mine_application_guid being passed a null value in the following endpoint: 
`/mines/<mine_guid>/documents?is_archived=true&major_mine_application_guid=`

- Changed it so the filters parameter passed to fetchMineDocuments function has properties that are conditionally included.

[MDS-5379](https://bcmines.atlassian.net/browse/MDS-5379)


henryoforeh-dev (approved):

matbusby-fw (approved):
Nice!

PR #2594, opened 2023-07-20
[MDS-5272] bulk file zipping

matbusby-fw:
## Objective 

- Allow the zipping of multiple files that exist in our storage,
- create a progress monitoring endpoint for the front end to poll
- upload the new zipped document to the object store and return a guid for it once the process is complete
- The endpoint for creating a zip accepts an array of document guids and an optional file name for the zipped file.  If there is no file name, it currently defaults to the mine number and a timestamp.
- This maintains the pretty path file structure for the documents within the zip file
- Zips are added to their own base directory - "zipped_docs"
  - Currently there is no exipiry set on this directory, but eventually we want to add one.

Zip documents endpoint -> `{{baseUrl}}/mines/{{mine_guid}}/documents/zip`
Zip progress endpoint -> `{{baseUrl}}/mines/documents/zip/{{task_id}}`

[MDS-5272](https://bcmines.atlassian.net/browse/MDS-5272)


henryoforeh-dev (approved):

services/document-manager/backend/app/docman/resources/document.py
```diff
@@ -70,6 +69,16 @@ def post(self):
...
+        # Create and send response
+        response = make_response(jsonify(document_manager_guid=document_guid), 201)
+        response.headers['Tus-Resumable'] = TUS_API_VERSION
+        response.headers['Tus-Version'] = TUS_API_SUPPORTED_VERSIONS
+        response.headers['Location'] = f'{Config.DOCUMENT_MANAGER_URL}/documents/{document_guid}'
+        response.headers['Upload-Offset'] = 0
+        response.headers['Upload-Expires'] = upload_expiry
+        response.headers[
+            'Access-Control-Expose-Headers'] = 'Tus-Resumable,Tus-Version,Location,Upload-Offset,Upload-Expires,Content-Type'
+        response.autocorrect_location_header = False
```
simensma-fresh, line 81: Think all of these are covered in the `initiate_document_upload` func, so probably not needed here

simensma-fresh (changes requested):
Overall looks good!

Are you able to add the new endpoints to tests/auth/test_expected_auth? Would be good to have as many endpoints as we can covered there.

Only concern would be that we're keeping the fully zipped file in memory before uploading it back to S3, which might cause some memory issues when running in Openshift when zipping many/large. However I'm happy to revisit later if it becomes an issue during testing

services/document-manager/backend/app/docman/resources/document.py
matbusby-fw: Hmmm, this is odd.  These lines came up during my rebase.  They weren't something I added.  I'm not sure why they are being flagged as new.  I'll double check the existing code to see if I can figure out where they came from.

asinn134 (approved):
Nice work!

simensma-fresh (approved):

PR #2595, opened 2023-07-20
[MDS-5385] document data missing from document table

asinn134:
- Document table data wasn't showing up in the document tables in the final application page. It was found the retrieval of current document data was setup to be taken from version array but current document data doesn't ever appear there. 

- Change made in `transformRowData` function inside `DoucmentTable.js` to just retrieve current document data from the document object

[MDS-5385](https://bcmines.atlassian.net/browse/MDS-5385)


simensma-fresh (approved):

matbusby-fw (approved):
Woot!

PR #2598, opened 2023-07-26
[MDS-5273] MS - Warn of similar file name as replacement

isuru-aot:
## Objective 

[MDS-5273](https://bcmines.atlassian.net/browse/MDS-5273)

Checking for existing files when uploading a file in MS project document upload section, 

- If the existing file is an archived one, alert the user with the following message
Screenshot 2023-07-25 210418

- If the existing file is non-archived one, alert the user with the following message and continue uploading the same file as a new version.
Screenshot 2023-07-25 210555

### DIFF
--- a/services/core-api/app/api/mines/documents/mine_document_search_util.py
+++ b/services/core-api/app/api/mines/documents/mine_document_search_util.py
@@ -8,7 +8,7 @@ from app.api.projects.project.models.project import Project
 from app.api.projects.information_requirements_table.models.information_requirements_table import InformationRequirementsTable
 from app.api.projects.information_requirements_table.models.information_requirements_table_document_xref import InformationRequirementsTableDocumentXref
 from app.api.mines.documents.models.mine_document import MineDocument
-from sqlalchemy import or_
+from sqlalchemy import desc, or_
 from app.extensions import db, api
 from sqlalchemy.orm import aliased
 
@@ -72,25 +72,16 @@ class MineDocumentSearchUtil():
         """
         Find Mine Documents by the document_name and the project_guid.
         """
-        query = db.session.query(MineDocument)
+        qy = db.session.query(MineDocument).filter_by(document_name=document_name, deleted_ind=False)
         
         if project_guid is not None:
-            # Create aliases for the joined tables
-            md_alias = aliased(MineDocument)
-            psdx_alias = aliased(ProjectSummaryDocumentXref)
-            ps_alias = aliased(ProjectSummary)
-            p_alias = aliased(Project)
+            qy = qy\
+                .join(ProjectSummaryDocumentXref)\
+                .join(ProjectSummary)\
+                .join(Project)\
+                .filter(Project.project_guid == project_guid)\
+                .order_by(desc(MineDocument.update_timestamp))
             
-            query = query\
-                .select_from(md_alias)\
-                .with_entities(md_alias.document_name, md_alias.mine_document_guid, p_alias.project_guid, md_alias.is_archived, \
-                    md_alias.mine_guid, md_alias.document_class, md_alias.update_timestamp, md_alias.update_user)\
-                .filter(md_alias.document_name == document_name, md_alias.deleted_ind == False)\
-                .join(psdx_alias, psdx_alias.mine_document_guid == md_alias.mine_document_guid)\
-                .join(ps_alias, psdx_alias.project_summary_id == ps_alias.project_summary_id)\
-                .join(p_alias, p_alias.project_guid == ps_alias.project_guid)\
-                .filter(p_alias.project_guid == project_guid)
-                
-            return query.first()
+            return qy.first()
         
         raise ValueError("Missing 'project_guid', This is required to continue the file upload process.")
--- a/services/core-api/app/api/mines/documents/mine_document_search_util.py
+++ b/services/core-api/app/api/mines/documents/mine_document_search_util.py
@@ -72,7 +72,6 @@ class MineDocumentSearchUtil():
         """
         Find Mine Documents by the document_name and the project_guid.
         """
-        project_guid=None
         query = db.session.query(MineDocument)
         
         if project_guid is not None:
--- a/services/core-api/app/api/mines/documents/mine_document_search_util.py
+++ b/services/core-api/app/api/mines/documents/mine_document_search_util.py
@@ -4,11 +4,13 @@ from app.api.projects.project_decision_package.models.project_decision_package i
 from app.api.projects.project_decision_package.models.project_decision_package_document_xref import ProjectDecisionPackageDocumentXref
 from app.api.projects.project_summary.models.project_summary_document_xref import ProjectSummaryDocumentXref
 from app.api.projects.project_summary.models.project_summary import ProjectSummary
+from app.api.projects.project.models.project import Project
 from app.api.projects.information_requirements_table.models.information_requirements_table import InformationRequirementsTable
 from app.api.projects.information_requirements_table.models.information_requirements_table_document_xref import InformationRequirementsTableDocumentXref
 from app.api.mines.documents.models.mine_document import MineDocument
 from sqlalchemy import or_
 from app.extensions import db, api
+from sqlalchemy.orm import aliased
 
 
 class MineDocumentSearchUtil():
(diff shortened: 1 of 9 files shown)


services/core-api/app/api/mines/documents/mine_document_search_util.py
```diff
@@ -64,3 +66,31 @@ def filter_by(cls, mine_guid, project_guid=None, major_mine_application_guid=Non
...
+            md_alias = aliased(MineDocument)
+            psdx_alias = aliased(ProjectSummaryDocumentXref)
+            ps_alias = aliased(ProjectSummary)
+            p_alias = aliased(Project)
+            
+            query = query\
+                .select_from(md_alias)\
+                .with_entities(md_alias.document_name, md_alias.mine_document_guid, p_alias.project_guid, md_alias.is_archived, \
+                    md_alias.mine_guid, md_alias.document_class, md_alias.update_timestamp, md_alias.update_user)\
+                .filter(md_alias.document_name == document_name, md_alias.deleted_ind == False)\
```
simensma-fresh: I think we should probably make this a case insensitive check so that `test.pdf` and `Test.pdf` are considered the same, as that's how most filesystems treat them.

Just a proposal to make this a little bit more readable would be to make use of SQLAlchemy's ability to recognize joining fields

So this could be something like:

```
query = db.session.query(MineDocument)
...
query = query.
    .join(ProjectSummaryDocumentXref)
    .join(ProjectSummary)
    .join(Project)
    .filter(MineDocument.document_name == document_name,...)
    .filter(ProjectSummary.project_guid=project_guid)
```

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -19,6 +20,28 @@
...
     document_manager_document_resource_url = f'{Config.DOCUMENT_MANAGER_URL}/documents'
+    
+    @classmethod
+    def validateFileNameAndInitializeFileUploadWithDocumentManager(cls, request, mine, project_guid, document_category):
+
+        metadata = cls._parse_request_metadata(request)
+        file_name = metadata.get('filename')
+        mine_document = MineDocumentSearchUtil.find_by_document_name_and_project_guid(file_name, project_guid)
+
+        if not mine_document: # No existing file found in this application hence continuing the file uploading
```
simensma-fresh, line 31: Can we add a "feature flag" to this so this won't block a release to prod until we're ready to go live with it?

Think elsewhere on the backend this has been done with:
`if Config.ENVIRONMENT_NAME != 'prod':`

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -19,6 +20,28 @@
...
+            content = f'{{"description" : "Archived file already exist with the given name: {file_name}", "status_code": 409, \
+                "status": "ARCHIVED_FILE_EXIST", "file_name": "{file_name}", "mine_guid": "{mine_document.mine_guid}", \
+                    "file_type": "{mine_document.document_class}", "update_timestamp": "{mine_document.update_timestamp}", "update_user": "{mine_document.update_user}", \
+                        "mine_document_guid": "{mine_document.mine_document_guid}"}}'
+            return Response(content, 409)
+        else: # The found file with the same name in this application is not archived.
+            content = f'{{"description" : "File already exist with the given name: {file_name}. Replace with the new version", \
+                "status_code": 409, "status": "REPLACEABLE_FILE_EXIST", "file_name": "{file_name}", "mine_guid": "{mine_document.mine_guid}", \
+                    "file_type": "{mine_document.document_class}", "update_timestamp": "{mine_document.update_timestamp}", "update_user": "{mine_document.update_user}", \
+                        "mine_document_guid": "{mine_document.mine_document_guid}"}}'
```
simensma-fresh: These are a bit difficult to read, would it be possible to maybe define them as a json obj and use `json.dumps` to convert them to a string?

simensma-fresh (changes requested):
Left a couple of suggestions to increase readability a bit. My main question would be that I see that the functionality has been implemented for the ProjectSummary, but it should probably also cover the other Major project files (IRT / Final Application / Decision Package) as well?

services/core-api/app/api/services/document_manager_service.py
matbusby-fw: Something like this:

content = {
    "description": f"File already exist with the given name: {file_name}. Replace with the new version",
    "status_code": 409,
    "status": "REPLACEABLE_FILE_EXIST",
    "file_name": file_name,
    "mine_guid": mine_document.mine_guid,
    "file_type": mine_document.document_class,
    "update_timestamp": mine_document.update_timestamp,
    "update_user": mine_document.update_user,
    "mine_document_guid": mine_document.mine_document_guid
}

return Response(json.dumps(content), 409, content_type='application/json')

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.js
```diff
@@ -11,19 +17,188 @@ const propTypes = {
...
+
+  const userInfo = useSelector(getUserInfo);
+
+  const handleCloseModal = () => {
+    setIsArchivedFileModalVisible(false);
+    setRelaplaceableFileModalVisible(false);
+    setShouldAbortUpload(true);
+  };
+
+  const childRef = useRef(null);
```
henryoforeh-dev: Is childRef used anywhere?

services/minespace-web/src/components/common/FileUpload.js
```diff
@@ -59,38 +64,78 @@ const defaultProps = {
...
+
     this.server = {
       process: (fieldName, file, metadata, load, error, progress, abort) => {
-        const upload = new tus.Upload(file, {
+        if (file) {
+          this.setState({ file: file, progress: progress, load: load });
+        }
+        let fileToUpload = this.state.file ? this.state.file : file;
+        let progressFn = this.state.progress ? this.state.progress : progress;
+        let loadFn = this.state.load ? this.state.load : load;
```
matbusby-fw:
```suggestion
        const fileToUpload = this.state.file ? this.state.file : file;
        const progressFn = this.state.progress ? this.state.progress : progress;
        const loadFn = this.state.load ? this.state.load : load;
```

services/minespace-web/src/components/common/FileUpload.js
```diff
@@ -59,38 +64,78 @@ const defaultProps = {
...
+                  message: `Failed to upload ${
+                    file && fileToUpload.name ? fileToUpload.name : ""
+                  }: ${err}`,
+                  duration: 10,
+                });
+              } else {
+                console.log(
+                  "notification disabled for status code(s):",
+                  this.props.notificationDisabledStatusCodes
+                );
```
henryoforeh-dev: Please check if this console.log is still required here.

matbusby-fw (changes requested):
Nice work Isuru!  Just a few things left to get it over the line, but it's looking really great.  Aside from Simen's comments I just notice a couple of let statements that could be const.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.js
```diff
@@ -11,19 +17,188 @@ const propTypes = {
...
-    allowRevert
-    allowMultiple
-  />
-);
+const notificationDisabledStatusCodes = [409]; // Define the notification disabled status codes
+
+export const ProjectSummaryFileUpload = (props) => {
+  const [archivedFileModalMessage, setArchivedFileModalMessage] = useState("");
+  const [isArchivedFileModalVisible, setIsArchivedFileModalVisible] = useState(false);
+  const [relaplaceableFileModalMessage, setRelaplaceableFileModalMessage] = useState("");
```
asinn134: Maybe it is intended but if its not I noticed something really small I think the spelling for these states should be `replaceableFileModalMessage`, and `isReplaceableFileModalVisible`

services/core-api/app/api/mines/documents/mine_document_search_util.py
isuru-aot: Updated accordingly.

services/core-api/app/api/services/document_manager_service.py
isuru-aot: Updated accordingly.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.js
isuru-aot: Nice catch. No it is not used so removed.

services/minespace-web/src/components/common/FileUpload.js
isuru-aot: Updated.

services/minespace-web/src/components/common/FileUpload.js
isuru-aot: Removed this.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.js
isuru-aot: Nice catch. Updated accordingly.

services/core-api/app/api/services/document_manager_service.py
isuru-aot: Added the flag to the [starting point](https://github.com/bcgov/mds/blob/mds-5273-minsespace-file-name-warning/services/core-api/app/api/projects/project_summary/resources/project_summary_document_upload.py#L29) of the new flow. And if its not PROD it will continue the same old flow.

isuru-aot:
@simensma-fresh 
Clarified the main point you highlighted ( `...it should probably also cover the other Major project files (IRT / Final Application / Decision Package) as well? `) with Rebecca, 
Response:
`It should definitely be applied to final applications, and should have functionality that can be transferred to other parts of Core's system. For this ticket I would not include adding this function to other parts of the user flow for it to be considered done. We can create another ticket for that.`

asinn134 (approved):
Nice work Isuru!

matbusby-fw (approved):

henryoforeh-dev (approved):

simensma-fresh (approved):

PR #2600, opened 2023-07-26
[MDS-5383] Core: Webpack cpu and build speed improvements

simensma-fresh:
## Objective 
Various improvements to reduce the sustained CPU usage when running core-web locally. The improvements fall into the following buckets:

- Upgraded webpack and other build tools to their latest versions
- Increased the poll interval for file change watchers
- Switched from babel to esbuild
- Switched to webpack filesystem cache that should reduce subsequent build times a bit

[MDS-5383](https://bcmines.atlassian.net/browse/MDS-5383)

**Before:**

image
image

**After:**

image
image

TODO:
- [ ] Apply the same changes for Minespace and common package
- [ ] Test on windows and in codespaces


.github/workflows/core-web.unit.yaml
```diff
@@ -25,25 +25,27 @@ jobs:
...
+      - name: Upgrade yarn
+        run: npm install -g yarn
+
       - name: Install Dependencies
-        run: yarn
+        run: yarn install
 
       - name: Build Common Package
-        run: cd services/common/ && yarn build
+        run: yarn workspace @mds/common build
```
taraepp, line 35: 🎨

asinn134 (approved):

henryoforeh-dev (approved):
Great work!

PR #2605, opened 2023-07-27
[MDS-5356] Actions Dropdown

taraepp:
## Objective 

[MDS-5356](https://bcmines.atlassian.net/browse/MDS-5356)


services/core-web/common/components/documents/DeleteDocumentModal.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { FC } from "react";
```
taraepp, line 1: this is copied from ArchiveDocumentModal, with some word changes.

services/core-web/common/models/documents/document.ts
```diff
@@ -0,0 +1,153 @@
...
+by the table should be *set* to the specific object, cannot expect to be able to consistently call its methods
+
+include "user_roles" property in the json used in the constructor to set allowed actions based on the user
+*/
+export class MineDocument {
+  public mine_document_guid: string;
+
+  public mine_guid: string;
+
+  public document_manager_guid: string;
```
taraepp, line 25: eslint added all these spaces on me and I was not feeling like changing eslint rules, even though it irritates me to look at it.

services/core-web/common/models/documents/document.ts
```diff
@@ -0,0 +1,153 @@
+import { USER_ROLES } from "@mds/common";
+
+export enum FileOperations {
+  View = "Open in document viewer",
+  Download = "Download file",
+  Replace = "Replace file",
+  Archive = "Archive file",
+  Delete = "Delete",
+}
```
taraepp, line 9: Made sense to me to put this here, but would make sense to put it in the documents directory instead

services/core-web/common/models/documents/document.ts
```diff
@@ -0,0 +1,153 @@
...
+    this.setAllowedActions(jsonObject.user_roles);
+  }
+
+  public getFileType() {
+    const index = this.document_name.lastIndexOf(".");
+    return index === -1 ? null : this.document_name.substring(index).toLocaleLowerCase();
+  }
+
+  public setAllowedActions(userRoles: string[] = []) {
+    this.allowed_actions = this.getAllowedActions(userRoles).filter(Boolean);
```
taraepp, line 105: https://bcmines.atlassian.net/browse/MDS-5356?focusedCommentId=22159 <-- some explanation of why things are set up this way (with functions setting a value instead of returning it)

services/core-web/src/components/common/CoreTable.tsx
```diff
@@ -1,20 +1,20 @@
 import React from "react";
 import { Table, TableProps, Tooltip } from "antd";
-import { ColumnType } from "antd/es/table";
+import { ColumnsType } from "antd/es/table";
```
taraepp, line 3: The change to ColumnsType was just because of my struggles with type (see above) and attempt to fix it by getting rid of differences between what we had and antd demos. It didn't seem to affect anything.

services/core-web/src/components/common/CoreTable.tsx
```diff
@@ -94,29 +94,26 @@ const CoreTable = <T,>(props: CoreTableProps<T>) => {
...
+            rowExpandable:
+              expandProps.rowExpandable ??
+              ((record) => expandProps.getDataSource(record).length > 0),
+            expandIcon: renderTableExpandIcon,
+            expandRowByClick: true,
+            expandedRowRender: expandProps.expandedRowRender ?? renderExpandedRow,
+            ...expandProps,
+          };
+    }
+    return { showExpandColumn: false };
```
taraepp, line 111: ran into an issue where if the dataSource had children, it would show the expand icon even if I didn't want it to. Forced it to be deliberate.

services/core-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -74,103 +72,57 @@ export const renderHighlightedTextColumn = (
...
-    ...(sortable ? { sorter: nullableStringSorter(dataIndex) } : null),
-  };
-};
-
-export const renderTaggedColumn = (
-  dataIndex: string,
-  title: string,
-  sortable = false,
-  placeHolder = ""
+export const renderActionsColumn = (
```
taraepp, line 82: I moved all the document-specific columns to DocumentColumns.

Put the interface here because the only reason to use it at all is this function (so importing both). Would be willing to move it though)

asinn134 (approved):

PR #2607, opened 2023-07-27
[MDS-5356] actions dropdown

taraepp:
## Objective 

[MDS-5356](https://bcmines.atlassian.net/browse/MDS-5356)

Original branch/PR:
https://github.com/bcgov/mds/pull/2605

* merge changes, basic switch to TS

* add update_timestamp to response model, changed CoreTable to use ColumnsType, made functin to calculate expandProps: allow more custom props and not allow 'children' to force expansion rendering unless on purpose. Moved some columns from common columns to document columns file, updated an import. Made document object model and an extension: cannot always guarantee method access so set more properties- testing added for proper usage. DocumentTable: changed matchChildColumns to showVersionHistory prop, used FileOperations interface to keep allowed actions structure consistent with docs themselves, parseDocuments run documents through constructor; get rid of archiveDocumentsArgs, instead let the document keep the mine guid, separated out old and new columns functions, it was getting confusing, found the table param 'childrenColumnName'

* filter allowed actions by the user's roles. Do the setting in the table because it's most DRY for now

* eslint determined to make my file not compile, switching to any

* brought changes over from Core to MS, and moved document models into common folder, updated imports

* delete document modal, extremely similar to archive document modal

* Minespace. Bring over styles to make the document table, document columns basically the same (minus a bug with userAccessData...). Brought some changes back to Core as well

* tests and snaps

* fix diff

* changed comment, it was a lie

* change function from public to protected, don't actually want it called externally

* diff

* removed some comments

* fix diff?


asinn134 (dismissed):

matbusby-fw (approved):
😅  👍

asinn134 (approved):

PR #2609, opened 2023-07-31
[MDS-5400] Fix bug on radio button

henryoforeh-dev:
## Objective 

[MDS-5400](https://bcmines.atlassian.net/browse/MDS-5400)

Fixed bug on radio button. The bug occurs when trying to save a strung data into a boolean column in the database.

### DIFF
--- a/services/core-web/src/components/common/RenderRadioButtons.js
+++ b/services/core-web/src/components/common/RenderRadioButtons.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { Form } from "@ant-design/compatible";
 import { Radio } from "antd";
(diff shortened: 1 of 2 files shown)


matbusby-fw (approved):
Nice!

services/core-web/src/components/common/RenderRadioButtons.js
```diff
@@ -23,13 +23,16 @@ const defaultProps = {
 
 const RenderRadioButtons = (props) => {
   const { meta, label, disabled, input, id, customOptions } = props;
-  const radioValue = typeof input.value !== "undefined" ? input.value.toString() : "";
 
   const options = customOptions ?? [
-    { label: "Yes", value: "true" },
-    { label: "No", value: "false" },
+    { label: "Yes", value: true },
+    { label: "No", value: false },
```
simensma-fresh, line 29: Will this affect any of our backend endpoints that expect a boolean as a string or would this purely be a frontend thing? Looks like the be checks for booleans as a string in a couple of places:
image

henryoforeh-dev: Good question, I will check.

henryoforeh-dev: I think it should be fine. The examples in the screenshot are for checks that happen during filtering. The respective columns in the tables in the backend store the data as boolean.
 
I also saw another ideal example in the explosives permit form. **closedOptions** is passed as an argument to customOptions parameter with boolean option values. 

**
image
**

image

simensma-fresh (approved):

PR #2610, opened 2023-07-31
[MDS-5394] Upgraded docman python version to 3.11

simensma-fresh:
## Objective 

[MDS-5394](https://bcmines.atlassian.net/browse/MDS-5394)

Upgraded docman to latest python version (3.11) to enable support for zip streaming. Surprisingly (?) few breaking changes found so far!

**Changes**
- Upgraded all python dependencies to their latest version
- Switched from flask_restplus to flask_restx (drop-in replacement) as flask_restplus is no longer supported and does not support python 3.11


henryoforeh-dev (approved):
Nice work!

matbusby-fw (approved):
Awesome!  love the upgrade to flask_restx.  It was bothering me that restplus was passed it's end of life.

PR #2611, opened 2023-07-31
[MDS-4153] Added missing VFCBC fields to NoW

matbusby-fw:
## Objective 

[MDS-4153](https://bcmines.atlassian.net/browse/MDS-4153)

Update core to accept and display new NoW fields from VFCBC

- added mine_purpose to fields that are "transmogrified" from VFCBC
- adjusted BE returns to include new data
- added fields in Core to show new fields
- refactored frontend components to TS
- Added interfaces for NoW

image
image


services/core-web/src/tests/components/noticeOfWork/applications/review/ReviewApplicationFeeContent.spec.js
```diff
@@ -31,292 +31,4 @@ describe("ReviewApplicationFeeContent", () => {
     const component = shallow(<ReviewApplicationFeeContent {...props} />);
     expect(component).toMatchSnapshot();
   });
-
-  it("MIN: isApplicationFeeValid === true if `adjustedTonnage` is set and exceeds proposed tonnage for MIN", () => {
```
matbusby-fw, line 35: After converting the `ReviewApplicationFeeContent` component into a functional component, these tests no longer work as you can't spy on functions within a functional component.  As this type of testing is generally avoided in favour of user flow testing, I made the decision to get rid of them for now.  We have upcoming tickets to implement Cypress tests for high value areas, and I think NoW is definitely one of those.

asinn134 (approved):

henryoforeh-dev (approved):

PR #2612, opened 2023-08-01
[MDS-5060] Switch to argocd cli to Force Sync Apps Upon Image Promotion

sjudzentis:
## Objective 

[MDS-5064](https://bcmines.atlassian.net/browse/MDS-5064)

When image promotion happens, ./gitops/commit.sh runs to push a new commit to tenant gitops, so that argo cd can redeploy. 

sometimes this commit is actioned on very fast before watch-deployment can start doing its job 

the revision number is updated and watch-deployment keeps waiting, hanging.. This is a false positive failure. The watch-deployment needs to be more robust in identifying a promotion and notify accurately. 

Solution is to use argocd cli to force sync of apps when image promotion happens and wait for the status to become healthy.

### DIFF
--- a/.github/workflows/core-api.build.deploy.dev.yaml
+++ b/.github/workflows/core-api.build.deploy.dev.yaml
@@ -66,10 +66,15 @@ jobs:
         uses: redhat-actions/openshift-tools-installer@v1
         with:
           oc: "4.7"
+      - name: Setup ArgoCD CLI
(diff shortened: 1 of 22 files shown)


matbusby-fw (approved):
Woo!  This is exciting.  The false fails are a real annoyance.  So many yamls.

henryoforeh-dev (approved):

PR #2614, opened 2023-08-02
[MDS-5363] Core: Fixed file upload 401 error for large files

simensma-fresh:
Replaces https://github.com/bcgov/mds/pull/2613 (so tests can successfully run)

**Objective**
[MDS-5363](https://bcmines.atlassian.net/browse/MDS-5363)

Wanted to test out how the current file zip functionality holds up for large files, but I kept hitting a 401 error while uploading a big file to core.

**Cause**
Currently, the auth token expires after 5min and the current implementation didn't account for that.

**Solution**
Make sure to use the current token for each tusd request

### DIFF
--- a/services/core-web/package.json
+++ b/services/core-web/package.json
@@ -91,7 +91,7 @@
     "styled-components": "5.2.0",
     "test-reducer-package": "1.0.14",
     "ts-loader": "8.4.0",
-    "tus-js-client": "1.8.0",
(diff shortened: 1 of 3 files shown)


matbusby-fw (approved):
👍

asinn134 (approved):

henryoforeh-dev (approved):

PR #2617, opened 2023-08-03
[MDS-5403] BE API to Get all Mine Alerts

isuru-aot:
## Objective 

[MDS-5403](https://bcmines.atlassian.net/browse/MDS-5403)

- List displays only active alerts
- Return the most recent alert at the top
- Include pagination: default page size=10

Screenshot 2023-08-03 162959

Updated API with more pagination details as per the PR comment:
Screenshot 2023-08-10 110936

### DIFF
--- a/services/core-api/app/api/mines/alerts/models/mine_alert.py
+++ b/services/core-api/app/api/mines/alerts/models/mine_alert.py
@@ -2,7 +2,6 @@ from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
 from app.api.utils.models_mixins import SoftDeleteMixin, AuditMixin, Base
 from app.extensions import db
-from sqlalchemy import func
 
 class MineAlert(SoftDeleteMixin, AuditMixin, Base):
     __tablename__ = "mine_alert"
--- a/services/core-api/app/api/mines/alerts/models/mine_alert.py
+++ b/services/core-api/app/api/mines/alerts/models/mine_alert.py
@@ -2,7 +2,7 @@ from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
 from app.api.utils.models_mixins import SoftDeleteMixin, AuditMixin, Base
 from app.extensions import db
-
+from sqlalchemy import func
 
 class MineAlert(SoftDeleteMixin, AuditMixin, Base):
     __tablename__ = "mine_alert"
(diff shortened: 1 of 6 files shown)


simensma-fresh (changes requested):
Can we add some integration tests to this endpoint?

You can see some examples of how to set them up and how they're generally set up in the `core-api/tests` folder. It would also be great if we could add the endpoint to the `tests/auth/test_expected_auth` list so we can verify that the role required has been applied correctly

isuru-aot:
> Can we add some integration tests to this endpoint?
> 
> You can see some examples of how to set them up and how they're generally set up in the `core-api/tests` folder. It would also be great if we could add the endpoint to the `tests/auth/test_expected_auth` list so we can verify that the role required has been applied correctly

Added tests. Please verify.

simensma-fresh (dismissed):

services/core-api/app/api/mines/alerts/models/mine_alert.py
```diff
@@ -56,3 +55,14 @@ def find_by_id(cls, _id):
...
     def find_by_mine_guid(cls, _id):
         return cls.query.filter_by(mine_guid=_id).filter_by(deleted_ind=False).all()
+    
+    @classmethod
+    def find_all_mine_alerts(cls, page, per_page):
+        query = cls.query.filter_by(deleted_ind=False).order_by(cls.create_timestamp.desc())
+
+        if (page):
+            result = query.paginate(page, per_page, error_out=False)
+            return dict([('total', result.total), ('records', result.items)])
```
taraepp: Can we send back more pagination data? current_page, total_pages, items_per_page, and total seem to be what we use, for consistency. I found something called PAGINATED_LIST that shows up in a few response_models.py files that could be worth taking a look at.

isuru-aot: Updated following the same PAGINATED_LIST way.

services/core-api/app/api/mines/alerts/resources/mine_alert.py
```diff
@@ -138,3 +138,31 @@ def delete(self, mine_guid, mine_alert_guid):
...
+        parser.add_argument(
+            'per_page', type=int, help='records per page', location='args', store_missing=False)
+        args = parser.parse_args()
+        page = args.get('page')
+        per_page = args.get('per_page') if args.get('per_page') else 10  # default per page is 10
+
+        records, pagination_details = MineAlert.find_all_mine_alerts(page, per_page)
+
+        if not records:
+            raise BadRequest('Unable to fetch global alerts.')
```
taraepp: Do we know it's a bad request though, if we don't get any results? I think BadRequest should probably be triggered if the page number they're requesting is greater than the total number of pages, negative number, etc, (validation issues) but otherwise I think we could just return [] for records. 

I'm thinking of a scenario where there's a local database with no mine alerts in the table, or if they're all deleted for some business reason, or if there were (hypothetical) filters that excluded all the results- what would we want to return then?

isuru-aot: As discussed, `filter_by` query will return an empty array in case there's no any records. Hence if checking on `not records` (`None`) is an unreachable code. So removed the `if` check.

taraepp (approved):

henryoforeh-dev (approved):

PR #2619, opened 2023-08-08
[MDS-4153] - Reinstate vfcbc fields ticket

matbusby-fw:
## Objective 

[MDS-4153](https://bcmines.atlassian.net/browse/MDS-4153)

This is just a reinstatement of the previously reverted PR (https://github.com/bcgov/mds/pull/2611).  Flyway was detecting an old migration from 2020 that had never been run on any of the databases and was named incorrectly.  It was trying to run this migration for some reason.

The only change from the initial PR is removing the broken migration after verifying it had not been run on any of the environments.


taraepp (approved):

henryoforeh-dev (approved):
Nice!

PR #2623, opened 2023-08-10
[MDS-5410] Minespace Management Button (#2622)

taraepp:
## Objective 
- make admin navigation (verified mines, minespace management) visible on admin dashboard again
   - added "exact" attribute back
- updated nvmrc with new node version, and added yarn install, nvm install to restart_codespaces script
- cleaned up some minor issues that generate a lot of console errors: 
   - added key for iterator, 
   - and updated AdminNavigation to functional+TS, and to use _items_ instead of _children_
      - the items vs children issue is to eventually be ready for new version of antd, and is for Menu components as well as I believe Tabs/TabPane, possibly a couple of others
      - changing this also got rid of a bunch of "warnKey"/"eventKey" messages- it seems that antd uses those internally when it catches that you're using children, but React flags it as an invalid attribute. 
      - changing from children to items also made antd generate different classnames or DOM structures and so new style file to deal with these changes. There were some conflicts with our existing rules, so notes added with the intention of us eventually removing the old styles

[MDS-5410](https://bcmines.atlassian.net/browse/MDS-5410)


simensma-fresh (dismissed):

services/core-web/src/components/admin/AdminNavigation.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Link } from "react-router-dom";
+import { Menu } from "antd";
+import { DownOutlined } from "@ant-design/icons";
+import { includes } from "lodash";
+import * as routes from "@/constants/routes";
+
+const propTypes = {
+  activeButton: PropTypes.string.isRequired,
+  openSubMenuKey: PropTypes.arrayOf(PropTypes.string).isRequired,
+};
```
henryoforeh-dev, line 11: Is it possible to rewrite this to be more "typescriptish"? 

const propTypes = {
  activeButton: string,
  openSubMenuKey: Array<string>,
};

henryoforeh-dev (dismissed):
Nice work! I just had one minor comment.

services/core-web/src/components/admin/AdminNavigation.tsx
matbusby-fw: I'm in favour of actually ditching propTypes in favour of interfaces

interface AdminNavigationProps {
activeButton: string,
openSubMenuKey: string[]
}

And then `export const AdminNavigation: FC<AdminNavigationProps>`

That's my preference, but either way works.

matbusby-fw (changes requested):
Yay!  Happy to have this back, and YAY Typescript conversions!

I like Henry's suggestion to make the one file more Typescripty by changing the PropTypes.  Not a hill to die on for me though if you prefer this pattern.

henryoforeh-dev (dismissed):

services/core-web/src/components/admin/AdminNavigation.tsx
```diff
@@ -1,17 +1,16 @@
...
-const propTypes = {
-  activeButton: PropTypes.string.isRequired,
-  openSubMenuKey: PropTypes.arrayOf(PropTypes.string).isRequired,
+interface AdminNavigationProps {
+  activeButton: string,
+  openSubMenuKey: string[],
 };
 
-export const AdminNavigation = (props) => {
+const AdminNavigation: FC<AdminNavigationProps> = (props: AdminNavigationProps) => {
```
matbusby-fw: FYI, once the FC<AdminNavigationProps> is there, it will infer the propType for props without explicitly typing it.

henryoforeh-dev (approved):

matbusby-fw (approved):

PR #2627, opened 2023-08-11
[MDS-5393] Convert file zipping to utilize file streaming

henryoforeh-dev:
## Objective 

[MDS-5393](https://bcmines.atlassian.net/browse/MDS-5393)

- Refactored the zipping functionality to utilize file streaming. This will help with zipping large files.
- Fixed a small bug that was causing SQLAlchemy dialect issue with Postgres.
- Added a small update to reflect the correct python version in the readme file.


services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+        total_size = sum(
+            self._client.head_object(Bucket=Config.OBJECT_STORE_BUCKET, Key=path)['ContentLength'] for path in
+            document_paths)
+
+        def local_files():
+            # Get the current timestamp
+            now = datetime.now()
+
+            def contents(path):
+                # Generator to read the file in chunks from S3
```
matbusby-fw: Even having written a lot of this, I find it to be difficult code to read through.  Have a couple thoughts on making the code comments a bit clearer for anyone who might deal with this down the road.

```suggestion
                # A generator function to slowly take data (in chunks) from S3 without loading the whole file in the memory.
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+                    yield chunk
+
+            # Return a generator with tuples of file parameters for each file
+            return (
+                (pretty_path, now, S_IFREG | 0o600, ZIP_32, contents(document_path))
+                for document_path, pretty_path in zip(document_paths, pretty_paths)
+            )
+
+        def to_file_like_obj(iterable):
+            # Converts an iterator to a file-like object
```
matbusby-fw:
```suggestion
            # This function takes a generator that continuously generates file data and produces a file-like object
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+
+            chunk = b''
+            offset = 0
+            it = iter(iterable)
+
+            def up_to_iter(size):
+                # Internal function to yield chunks from the iterable
```
matbusby-fw:
```suggestion
                # Internal function to yield chunks from the iterable.  Ensures we only read up to 'size' bytes from the file.
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+                            break
+                        else:
+                            offset = 0
+                    to_yield = min(size, len(chunk) - offset)
+                    offset = offset + to_yield
+                    size -= to_yield
+                    yield chunk[offset - to_yield:offset]
+
+            class FileLikeObj:
+                # File-like object to handle reading in chunks
```
matbusby-fw:
```suggestion
                # A class that creates an object which behaves like a file. Instead of reading from disk, we're reading from the generator.
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+
+                def read(self, size=-1):
+                    # Read a specific size from the iterator
+                    return b''.join(up_to_iter(float('inf') if size is None or size < 0 else size))
+
+            return FileLikeObj()
+
+        # Generator for zipped chunks including metadata
```
matbusby-fw:
```suggestion
        # Create an iterator to loop over the zipfiles and read chunks of the zip file data
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+        zipped_chunks = stream_zip(local_files())
+        # Convert the generator to a file-like object
```
matbusby-fw:
```suggestion
        # Convert the iterator to a file-like object
```

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -249,4 +253,86 @@ def upload_zip_file(self, file_data, key, file_size, progress_callback):
...
+        zipped_chunks_obj = to_file_like_obj(zipped_chunks)
+
+        # Get the total size of the zip file
+        zip_upload_progress_callback = ZipUploadProgressPercentage(total_size=total_size,
+                                                                   progress_callback=progress_callback)
+
+        # Upload the chunks to S3
```
matbusby-fw, line 331:
```suggestion
        #  Finally, the generated zip content is streamed to S3. The upload_fileobj function takes a file-like object as input, which in our scenario, is reading from the chunks generated by our zip streaming function. This approach ensures that our application only needs to load a small part of the file into memory at a time, allowing us to handle large files with minimal memory footprint.
```

matbusby-fw (changes requested):
Looks great!  I added some suggestions for code comments that could be added (you don't have to use exactly those words).  I think this would be helpful for any future devs that might work on it as it's a bit dense.

henryoforeh-dev:
@matbusby-fw I have updated the comments as suggested.

matbusby-fw (approved):

simensma-fresh (approved):

PR #2628, opened 2023-08-11
[MDS-5414] Fixed auth issue with some integrations in test env

simensma-fresh:
## Objective 

[MDS-5414](https://bcmines.atlassian.net/browse/MDS-5414)

Fixed an error that pops up in dev and test when using integration credentials. Cause of this is that the `JWT_OIDC_AUDIENCE_CYPRESS` env variable is not set in these environments causing the audience check to crash with an error. This was previously solved in production by adding a value for the env variable, but this was not done in dev/test

image


matbusby-fw (approved):
🕺

henryoforeh-dev (approved):

PR #2629, opened 2023-08-11
[MDS-5395] Add project document version in project summary

matbusby-fw:
## Objective 

[MDS-5395](https://bcmines.atlassian.net/browse/MDS-5395)

This ticket went a little bit sideways.  It was to allow viewing and managing of archived document versions, but the underlying uploading of versions wasn't working properly.  The component was uploading new documents instead of versions of the same document.  Was hoping to get this done and implement the ticket A/C, but it proved tricky to get everything right.

Moved the jira ticket back to open 😢 

Updated the Project Summary upload to create new versions
Modified previous logic to also check for existing documents before sending to the backend.
Updated the fileUpload component to handle new versions
Updated the DocumentUpload component to hit a different endpoint when a new version is uploaded
Created a new actionCreator in projectActionCreators for uploading versions
Small updates to the backend to return the created information.
Refactored some files to Typescript and Functional Components
Tried to refactor the FileUpload component to the same, but was struggling to get it to work properly, so I put that aside for now.

image


.eslintrc.js
```diff
@@ -77,5 +77,6 @@ module.exports = {
     "@typescript-eslint/default-param-last": 0,
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/no-children-prop": 0,
+    "@typescript-eslint/no-unused-expressions": 0,
```
taraepp: I think this rule I'd prefer to disable on a case-by-case basis, rather than globally. It catches a lot of the "map that should be a forEach" type code.

matbusby-fw: Oh!  Yeah, I had been playing with this setting because it was complaining about some legitimate code, but it dien't help in the end anyway.  We can just remove it.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.tsx
```diff
@@ -0,0 +1,318 @@
...
+  const [handleModalClose, setHandleModalClose] = useState<(() => void) | null>(null);
+  const [handleModalSubmit, setHandleModalSubmit] = useState<(() => void) | null>(null);
+  const [existingFiles, setExistingFiles] = useState([]);
+  const [version, setVersion] = useState();
+  const [uploadUrl, setUploadUrl] = useState(PROJECT_SUMMARY_DOCUMENTS(props.params));
+
+  const userInfo = useSelector(getUserInfo);
+
+  const handleCloseModal = () => {
+    handleModalClose && handleModalClose();
```
matbusby-fw, line 56: @taraepp these are the lines that eslint was complaining about.

.eslintrc.js
henryoforeh-dev: @taraepp I have removed it.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectSummaryFileUpload.tsx
taraepp: Silly eslint. This looks perfectly fine to me.

I do get this message when the hook runs, I wonder if it's related: `Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .`

taraepp (approved):

simensma-fresh (approved):

PR #2632, opened 2023-08-16
[MDS-5363] file submission frontend

asinn134:
## Objective 

- Changes made to DocumentTable component to allow for zipping and downloading bulk files on the front end
- CSS changes to fix the document table row alignment (https://bcmines.atlassian.net/browse/MDS-5420) and row hover issue introduced in one of my previous PRs
- Added a prop in DocumentTable component called `enableBulkActions`to help determine if multi select actions should be enabled
- Created a separate task/story for setting up the e2e cypress test (https://bcmines.atlassian.net/browse/MDS-5429)

[MDS-5363](https://bcmines.atlassian.net/browse/MDS-5363)

**User without mine project edit access is displayed download button instead of dropdown**:
Screenshot (150)

**No rows selected the action drop down to download zipped files is disabled**:
Screenshot (141)

**Row is selected, action drop down is enabled but individual action drop downs in each row are disabled**:
Screenshot (142)

**Compression begins with progress bar above notification**:
Screenshot (149)

**Compression is done and ready for download**:
Screenshot (144)

**Core view with document versions**:
Screenshot (151)

**Minespace table row alignment**:
Screenshot (140)

**What Minespace document table styling looks like with versions**:
Screenshot (153)


services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -49,6 +58,11 @@ interface DocumentTableProps {
   additionalColumnProps: { key: string; colProps: any }[];
   fileOperationPermissionMap: { operation: FileOperations; permission: string | boolean }[];
   userRoles: string[];
+  project?: IProject;
```
taraepp: Would suggest a few things for this file:
1. Refactor to not need the project passed in. Looks like we need the mine guid from it (stored in every document) and the text to show in the download, which will vary depending on where we're using the table (I think we had a prop called document parent that we used for something similar?). Could create a function in the document model like getParentEntityName too, which will be easy to extend for different file types and reusable everywhere.
2. Get the permission from the document itself instead of doing a role check in here. If we want to download the file, check if the file has permission to download, etc. Basically we want the document table to be as agnostic as possible when it comes to what type of files it's displaying.
3. This file is starting to get excessively long- can anything be extracted to its own file? Something like the progress bar or modal could be good candidates.

services/core-web/src/components/common/CoreTable.tsx
```diff
@@ -2,14 +2,14 @@ import React from "react";
...
-import { ExpandableConfig } from "antd/lib/table/interface";
+import { ExpandableConfig, TableRowSelection } from "antd/lib/table/interface";
 
 interface CoreTableExpandConfig<T> extends ExpandableConfig<T> {
   getDataSource?: (record: T) => any[];
   rowKey?: string | ((record: any) => string);
   recordDescription?: string;
   subTableColumns?: ColumnsType<any>;
-  matchChildColumnsToParent?: boolean;
+  showVersionHistory?: boolean;
```
taraepp: I don't think we should change the prop name here from matchChildColumnsToParent --> showVersionHistory, that should stay on DocumentTable.

With document table, there are other UI changes that are very specific to showing the version history (like the version tags, etc) so that is specifically what that prop is for. But the CoreTable doesn't just show documents, we want a prop name that makes sense here for any scenario that can have matching parent/child columns.

asinn134: That makes sense, the areas that had matchChildColumnsToParent changed are now reverted back to including it.

services/core-web/src/components/common/DocumentTable.tsx
asinn134: Great suggestions.

1. Project is no longer passed in as a prop now, and instead project title is setup inside the mine document type to be retrieved for use.

2. Added a multi select permission to the mine document type, and is checking each document in the table to make sure they all have multi select permission. No longer retrieving role for check here now.

3. The progress bar and modals are now extracted out.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -231,18 +256,204 @@ export const DocumentTable = ({
...
     ? { size: "small" as SizeType, rowClassName: "ant-table-row-minimal" }
     : null;
+
+  const handleRowSelectionChange = (value) => {
+    if (documentTypeCode === "") {
+      setDocumentTypeCode(value[0].major_mine_application_document_type_code);
+    }
+
+    const documentManagerGuids = value
+      .filter((row) => row.versions !== undefined)
```
henryoforeh-dev: Do we need to also check that row.versions is not null here too ?

asinn134: Oh I think this is changed now, it should look like this with the recent changes:

```
    const documentManagerGuids = rowSelection
      .filter((row) => row.is_latest_version)
```

and with `row.is_lastest_version`, it should always be either true or false.

henryoforeh-dev (dismissed):

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -70,6 +87,19 @@ export const DocumentTable = ({
...
 }: DocumentTableProps) => {
+  const [rowSelection, setRowSelection] = useState([]);
+  const [isBeginCompressionModalVisible, setBeginCompressionModalVisible] = useState(false);
+  const [isReadyForDownloadModalVisible, setReadyForDownloadModalVisible] = useState(false);
+  const [isCompressionProgressVisible, setCompressionProgressVisible] = useState(false);
+  const [compressionProgress, setCompressionProgress] = useState(0);
+  const [documentTypeCode, setDocumentTypeCode] = useState("");
+  const [notificationTopPosition, setNotificationTopPosition] = useState(0);
+  const [fileToDownload, setFileToDownload] = useState("");
+  const [projectTitle, setProjectTitle] = useState("");
```
taraepp: Instead of the project title, can we use "entity title" or something that can apply to any document?

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -231,20 +270,207 @@ export const DocumentTable = ({
...
+    },
+  ];
+
+  const rowSelectionObject = {
+    onChange: (selectedRowKeys: React.Key[], selectedRows: any) => {
+      handleRowSelectionChange(selectedRows);
+    },
+  };
+
+  return (
```
taraepp: There's a loooooot of if/ternary logic going on in the return statement here and I'm really not following it. Can we have a function or two to generate the elements for the table instead?

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -27,13 +30,22 @@ import {
 import { openDocument } from "../syncfusion/DocumentViewer";
 import { downloadFileFromDocumentManager } from "@common/utils/actionlessNetworkCalls";
 import { getUserAccessData } from "@common/selectors/authenticationSelectors";
+import { Dropdown, Button, MenuProps, notification, Progress } from "antd";
+import { DownOutlined } from "@ant-design/icons";
+import { USER_ROLES } from "@common/constants/environment";
```
taraepp: I think there's some unused imports in here.

services/core-web/src/components/modalContent/CompressedFilesDownloadModal.tsx
```diff
@@ -0,0 +1,38 @@
...
+    bodyStyle={{ minHeight: "150px" }}
+    style={{ padding: "40px" }}
+    zIndex={1500}
+  >
+    <Typography.Paragraph strong>
+      <CheckCircleOutlined style={{ color: "#45a766", fontSize: "20px", marginRight: "10px" }} />
+      Files ready for download
+    </Typography.Paragraph>
+    <Typography.Paragraph style={{ fontSize: "90%", marginLeft: "30px" }}>
+      {props.projectTitle} selected documents are ready for download.
```
taraepp: Again suggesting to change projectTitle to something generic.

services/core-web/src/tests/models/document.spec.ts
```diff
@@ -84,6 +84,7 @@ describe("MineDocument model", () => {
       FileOperations.Replace,
       FileOperations.Archive,
       FileOperations.Delete,
+      FileOperations.MultiSelect,
```
taraepp: Do we actually need a separate permission to multi select? Multi-selecting doesn't technically perform any actions to the file or give anyone view access that they didn't already have.

services/minespace-web/common/actionCreators/projectActionCreator.ts
```diff
@@ -505,6 +507,55 @@ export const removeDocumentFromMajorMineApplication = (
     .finally(() => dispatch(hideLoading()));
 };
 
+export const documentsCompression = (
+  mineGuid: string,
```
taraepp: This looks really generic! I'm wondering, is this a function that other documents can use or is it specific to projects?

services/core-web/src/tests/models/document.spec.ts
asinn134: Hmm... actually I think this is a good point. Technically according to the figma mockup, the permission check is to determine if the user will see a drop down or a download button for the bulk actions, so regardless of permissions the user should be able to bulk download, the only other action in the drop down is bulk archive so we can actually just check for archive permission here.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -49,6 +61,10 @@ interface DocumentTableProps {
   additionalColumnProps: { key: string; colProps: any }[];
   fileOperationPermissionMap: { operation: FileOperations; permission: string | boolean }[];
   userRoles: string[];
+  documentsCompression: ActionCreator<typeof documentsCompression>;
```
taraepp: Oooh actually if these _are_ generic functions that can be used by any document, we can just import them instead of passing them in as props. Way simpler to implement for the next one.

services/minespace-web/common/actionCreators/projectActionCreator.ts
asinn134: Ah I put it in projects because from what I know so far the compression only happens in projects. Is there a better location you would suggest for this?

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -93,8 +123,17 @@ export const DocumentTable = ({
       return doc;
     });
   };
+
   const documents = parseDocuments(props.documents ?? []);
 
+  useEffect(() => {
+    const isMultiSelect = documents.every((doc) =>
```
taraepp: Can we use the enableBulkActions prop? Can get rid of a state variable that way.

asinn134: Initially I tried setting up the condition check to see if every document in the table has permission within the MajorMineApplicationTab.js file, and the boolean result of that check would be what gets passed into enableBulkAcations prop.. however I ran into an issue where in MajorMineApplicationTab.js's  renderDocumentSection function you console.log the sectionDocuments, you would see documents having 6 allowed_actions, but if you were to access the same document's allowed_actions, for some reason it only shows 2 allowed_actions "Open in document viewer" and "Download File".
 Screenshot (155)
Not too sure what was going on there but I moved away from this because I think it makes sense to separate enableBulkAction prop from the check to see if all documents have permission, because technically even if the documents don't have the major mine application edit permission, the user should still be able to bulk download, just won't have permission to bulk archive.

services/minespace-web/common/actionCreators/projectActionCreator.ts
taraepp: Currently it only happens in projects but since there's nothing that makes it so that _only_ project documents can be compressed, I would put it in documentActionCreator. If we don't have it already I would make it.

services/core-web/src/components/common/DocumentTable.tsx
taraepp: Let me know if you're still having trouble with this and we can figure it out over VC. I had issues getting the permissions _as the result of a function_ (because the table will, in certain circumstances, turn an object proper into a plain JSON object with no methods) but there should not be any issues getting its attributes and we should sort that out. Ideally we'd want to be able to do something like "multiFileActions = allowedTableActions.filter(action => documents.every(document has permission for that action))"

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -231,20 +270,207 @@ export const DocumentTable = ({
...
+            });
+          }, 2000);
+        } else {
+          const documentTypeIdentifier = `.progressNotification-${documentTypeCode}`;
+          const notificationElement = document.querySelector(documentTypeIdentifier);
+          const notificationPosition = notificationElement.getBoundingClientRect();
+          setNotificationTopPosition(notificationPosition.top);
+          progressRef.current = true;
+          setCompressionProgressVisible(true);
+          const poll = async () => {
```
matbusby-fw: Nice implementation here!

services/minespace-web/src/components/common/CoreTable.tsx
```diff
@@ -64,15 +64,21 @@ const CoreTable = <T,>(props: CoreTableProps<T>) => {
...
           </Tooltip>
         ) : (
           <Tooltip
             title={`Click to show ${recordDescription}.`}
             placement="right"
             mouseEnterDelay={1}
           >
-            <PlusSquareFilled className="icon-lg--lightgrey" />
+            <PlusSquareFilled
+              style={{ fontSize: "24px", color: "#ded9d9" }}
```
matbusby-fw: I noticed you're using inline styles in a few areas.  For me, it's not a hill to die on, but I much prefer creating classes in the css files wherever possible.

matbusby-fw (dismissed):
Really nice work!

This one was a beast, but it's looking great.  Left a couple nitpicky comments.  I agree with Tara about the seperation of concerns to keep things as reusable as possible.

Thanks for taking the time to set out the PR itself with screenshots and descriptions!

🙌

henryoforeh-dev (dismissed):
Nice!

matbusby-fw (dismissed):

henryoforeh-dev (dismissed):

services/core-web/src/components/common/DocumentTable.tsx
asinn134: This has been adjusted to entityTitle now

services/core-web/src/components/common/DocumentTable.tsx
asinn134: Extracted the if/ternary logic out of the return statement here, and moved it into two separate functions. One for generating the bulk action html elements, and another function for generating the CoreTable

services/core-web/src/components/common/DocumentTable.tsx
asinn134: These have been removed now, thanks.

services/core-web/src/components/modalContent/CompressedFilesDownloadModal.tsx
asinn134: This has been adjusted to entityTitle now

services/core-web/src/tests/models/document.spec.ts
asinn134: MultiSelect has been removed now and is using document's archive permission instead to check.

services/minespace-web/common/actionCreators/projectActionCreator.ts
asinn134: The documentCompression and the pollDocument function have been moved to documentActionCreator now.

services/core-web/src/components/common/DocumentTable.tsx
asinn134: This should be using the `mapDispatchToProps` approach now.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -127,11 +126,11 @@ export const DocumentTable = ({
   const documents = parseDocuments(props.documents ?? []);
 
   useEffect(() => {
-    const isMultiSelect = documents.every((doc) =>
-      doc.allowed_actions.find((a) => a === "MultiSelect")
+    const isBulkArchive = documents.every((doc) =>
+      doc.allowed_actions.find((a) => a === "Archive file")
```
taraepp: nitpick: use the constant for "Archive file" rather than the string itself

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -305,7 +304,7 @@ export const DocumentTable = ({
       .filter((row) => row.is_latest_version)
       .map((filteredRows) => filteredRows.document_manager_guid);
 
-    setProjectTitle(rowSelection[0].project_title);
+    setEntityTitle(rowSelection[0].project_title);
```
taraepp: Can we set the _document's_ entity_title to equal project title so that we're not specific to project files in this component? Like in the model itself, this.entity_title = jsonObject.project_title.

services/core-web/src/components/common/DocumentTable.tsx
asinn134: This has been changed to use `FileOperations.Archive` now.

services/core-web/src/components/common/DocumentTable.tsx
asinn134: The document model file has been adjusted for this now.

services/core-web/src/components/common/DocumentTable.tsx
asinn134: As per our discussion, this issue doesn't interfere with any logic, so should be okay to leave as it is.

henryoforeh-dev (dismissed):

taraepp (dismissed):
LGTM
Great work on this PR, appreciate your patience with all my nitpicks! 🙌

matbusby-fw (dismissed):

asinn134:
@taraepp @henryoforeh-aot @matbusby-fw @simensma-fresh @isuru-aot @sjudzentis   I noticed yesterday after the demo, that I made a mistake, I left out the "Download All Application Files" button. 

Screenshot (158)

While setting up this button and allowing it to access the same modals as the table's bulk action dropdown or download button, I ended up having to extract a lot of the compression logic out of the `DocumentTable `and setting that logic up in its own file, `DocumentCompression`, so it can be used in MajorMineApplicationTab and also for each individual `DocumentTable`. With the current changes the "Download All Application Files" button and the individual table bulk download works fine, I was struggling a bit with finding a better way of doing this, if I can get this reviewed again that would be great.

asinn134:
> LGTM Great work on this PR, appreciate your patience with all my nitpicks! 🙌

Thanks, I appreciate your input, it is a good opportunity for me to learn! 🙌

services/core-web/common/models/documents/document.ts
```diff
@@ -69,6 +71,7 @@ export class MineDocument {
     this.archived_by = jsonObject.archived_by;
     this.archived_date = jsonObject.archived_date;
     this.is_latest_version = jsonObject.is_latest_version ?? true;
+    this.entity_title = jsonObject.project_title;
```
taraepp: So- the entity title should be on the MineDocument, but it should be the Type that _has a project title_ that sets it as such. Maybe = jsonObject.entity_title ?? "" on the base model so it's still defined but properly scoped.

asinn134: As discussed, this has been adjusted now.

taraepp (approved):

henryoforeh-dev (approved):

PR #2633, opened 2023-08-19
[MDS-5288] Notify changes in application in activity feed

isuru-aot:
## Objective 

[MDS-5288](https://bcmines.atlassian.net/browse/MDS-5288)

Completed:

1. Trigger an activity alert for a Core user who has subscribed to that mine when a mine does one or more actions: (adds a new file to an application, replaces a version of a file in an application, moves a file to archive
2. Notification will redirect to the documents page
3. The activity alert format  `File(s) in <project_title> has been updated for <mine_name>`
4. Send a notification in activity feed when one or multiple files have been changed in the submission.  
5. If more changes have been made on the same submission, no notification will be sent again if the previous notification of the submission in activity feed is unread.
6. If more changes have been made on the same submission, notification will be sent if the previous notification of the submission in activity feed is read.
7. New notification will be sent after 24 hours even if the previous notification wasn’t read and additional changes are made 

Todo:
Add time stamp for local timezone default hour:minute to last modified and state the timezone


services/core-api/app/api/activity/models/activity_notification.py
```diff
@@ -161,7 +166,18 @@ def create_many(cls, mine_guid, activity_type, document, idempotency_key=None, c
...
+        if(renotify_period_minutes > 0):
+          # Calculate the datetime that was renotify_period_minutes minutes ago
+          renotify_timestamp = datetime.utcnow() - timedelta(minutes=renotify_period_minutes)
+          # Filter to retrived unread records for renotifying
+          unread_renotify_query = cls.query.with_entities(cls.notification_recipient)\
+                .filter_by(notification_read=False)\
+                .filter(cls.create_timestamp < renotify_timestamp)\
+                .order_by(cls.create_timestamp)
+
+        unread_renotify_users = [res[0] for res in unread_renotify_query.all()]
```
simensma-fresh: I think this might produce an error if `renotify_period_minutes` is not set? As unread_renotify_query would be None

simensma-fresh (changes requested):

services/core-api/app/api/activity/models/activity_notification.py
isuru-aot: Address this by getting that line in to the if block and defining an empty list for `unread_renotify_users` before the if block. variable name is updated as `unread_renotify_and_read_notify_users` because same list is used to filter the read/unread notifications to decide whether to send the notifications or not.

simensma-fresh (dismissed):

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
```diff
@@ -105,6 +111,7 @@ class MineDocumentArchiveResource(Resource, UserMixin):
     @api.expect(ARCHIVE_MINE_DOCUMENT)
     @api.response(204, 'Successfully archived documents')
     def patch(self, mine_guid):
+        project_guid = request.args.get('project_guid', type=str)
```
taraepp: What happens when we send a document without a project?

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -94,6 +99,7 @@ export const DocumentTable = ({
     });
   };
   const documents = parseDocuments(props.documents ?? []);
+  const projectGuid = props.projectGuid;
```
taraepp: Document table is used for more than project documents. Cannot rely on project guid being defined. Once Alex's changes with entity_title and so on are merged in, would try following a similar pattern. No references to project should be in this file.

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
simensma-fresh: 100%, it would be great if we wouldn't have to pass in the `project_guid` here and instead look up the related project instead

simensma-fresh (changes requested):

services/core-api/app/api/mines/documents/resources/mine_document_version_resource.py
```diff
@@ -49,8 +57,16 @@ def post(self, mine_guid, mine_document_guid):
         if mine_document.is_archived:
             raise BadRequest('Cannot create new version of archived document')
 
-        return DocumentManagerService.initializeFileVersionUploadWithDocumentManager(
+        resp = DocumentManagerService.initializeFileVersionUploadWithDocumentManager(
             request, mine_document)
+        project_name = Project.find_by_project_guid(project_guid)
```
simensma-fresh: Same for the above. What happens if this is a file not associated with a project?
Would be great to look up the associated project here instead of having to pass in the `project_guid`

simensma-fresh (changes requested):

services/core-api/app/api/mines/documents/resources/mine_document_version_resource.py
```diff
@@ -49,8 +57,16 @@ def post(self, mine_guid, mine_document_guid):
...
+
+        if resp:
+            renotify_hours = 24
+            trigger_notification(f'File(s) in project {project_name} has been updated for mine {mine.mine_name}.',
```
simensma-fresh: I think this should be in the `MineDocumentVersionListResource` not in the `MineDocumentVersionUploadResource`. `MineDocumentVersionUploadResource` used when you first initialize an upload, where `MineDocumentVersionListResource` is used when the upload has completed - and we probably want to only send the notification then

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -79,7 +89,17 @@ def initializeFileUploadWithDocumentManager(cls, request, mine, document_categor
...
 
-        return Response(str(resp.json()), resp.status_code, resp.raw.headers.items())
+        resp = Response(str(resp.json()), resp.status_code, resp.raw.headers.items())
+
+        if Config.ENVIRONMENT_NAME != 'prod':
+            project = Project.find_by_project_guid(project_guid)
+
+            if resp:
+                renotify_hours = 24
+                trigger_notification(f'File(s) in project {project.project_title} has been updated for mine {mine.mine_name}.',
```
simensma-fresh: I think this is likely not the right place to put this as it's called when we first initialize the file upload. We don't want to send it until the file Upload has been finished and the related MineDocument has been created. I think you'll need to put this where the `MineDocument` for all the different categories (major mine application / project summary / project decision package / information requirements table)

simensma-fresh (changes requested):

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
```diff
@@ -127,6 +134,12 @@ def patch(self, mine_guid):
 
         MineDocument.mark_as_archived_many(mine_document_guids)
 
+        if len(mine_document_guids) > 0:
+            project = ProjectsSearchUtil.find_by_mine_document_guid(mine_document_guids[0])
```
taraepp: I think this is still too specific to projects.

Maybe pass in a parameter to represent the "activity type" and do something like a switch (ie: if parentEntityType == project then <do project related things>), or make a separate call to do the notification.

services/core-web/src/constants/routes.ts
```diff
@@ -297,6 +297,12 @@ export const VIEW_MINE_INCIDENT = {
   component: MineIncident,
 };
 
+export const DOCUMENT_MANAGEMENT = {
```
taraepp: I think it's fine that this one is specific to projects, but should rename to PROJECT_DOCUMENT_MANAGEMENT or similar.

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
isuru-aot: @taraepp 
This parameter `activityType` you suggest, it should be passed from the front end, right?

services/core-api/app/api/projects/project/projects_search_util.py
```diff
@@ -0,0 +1,29 @@
...
+        Find Project by the mine_document_guid.
+        """
+
+        qy = db.session.query(Project)
+
+        if mine_document_guid is not None:
+            prjquery = qy\
+                .filter(ProjectSummary.project_summary_id == ProjectSummaryDocumentXref.project_summary_id)\
+                .filter(ProjectSummary.project_guid == Project.project_guid)\
+                .filter(ProjectSummaryDocumentXref.mine_document_guid == mine_document_guid)
```
simensma-fresh: This also need to cover all the other project related documents I think?
- MajorMineApplicationDocumentXref
- ProjectDecisionPackageDocumentXref
- InformationRequirementsTableDocumentXref

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
simensma-fresh: @taraepp @isuru-aot I would actually suggest not to pass along other parameters here. In the end I think it's the backend that should decide what kind of document it is - we don't accidentally want to send out notifications for other types on documents

simensma-fresh (changes requested):

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
isuru-aot: Resolved by passing entity_type param from front end.

services/core-web/src/components/common/DocumentTable.tsx
isuru-aot: removed.

services/core-api/app/api/mines/documents/resources/mine_document_version_resource.py
isuru-aot: Updated accordingly

services/core-api/app/api/services/document_manager_service.py
isuru-aot: Updated accordingly.

services/core-web/src/constants/routes.ts
isuru-aot: Done

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
isuru-aot: Updated as discussed to decide the doc is valid to send notification.

services/core-api/app/api/projects/project/projects_search_util.py
isuru-aot: This is no more used in this flow

simensma-fresh: @isuru-aot From what I can tell, it's still referenced by major_mine_application and decision_package. If it's not used, we should probably remove it

simensma-fresh (dismissed):

services/core-api/app/api/projects/project/projects_search_util.py
isuru-aot: @simensma-fresh 
Oh, you are correct. Removed the project_search_util.py and all its references.

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -137,6 +158,14 @@ def update(self, irt_data, import_file=None, document_guid=None, add_to_session=
         if add_to_session:
             self.save()
 
+        if mine_doc and project:
+            if Config.ENVIRONMENT_NAME != 'prod':
```
taraepp: is it possible to squeeze in Simen's feature flag here?

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -137,6 +158,14 @@ def update(self, irt_data, import_file=None, document_guid=None, add_to_session=
...
+                renotify_hours = 24
+                mine = Mine.find_by_mine_guid(mine_doc.mine_guid)
+
+                trigger_notification(f'File(s) in project {project.project_title} has been updated for mine {mine_doc.mine_name}.',
```
taraepp: Are we calling trigger_notification twice? Once here and once in mine_document_resource.py? I'm guessing that the renotify param would prevent it from generating 2 notifications.

On the one hand, I like how the IRT is handling its own notifications (separation of concerns!) but on the other hand I don't like essentially the same method being copy-pasted into multiple classes (don't repeat yourself!) so I am on the fence about which is preferable.

services/core-web/src/components/navigation/NotificationDrawer.tsx
```diff
@@ -222,7 +227,7 @@ const NotificationDrawer: FC<INotificationDrawerProps> = (props) => {
                     )}
                     <Col>
                       <Typography.Text className="notification-info-text">
-                        {formatDateTime(activity.create_timestamp)}
+                        {formatDateTimeTz(activity.create_timestamp)}
```
taraepp: This is probably minor, but I don't think formatDateTimeTz is quite the right function here (it's intended for user inputted dates where we have the timezone as a separate data field so it gets passed in, and falls back on PT as a default). In this case we don't have a timezone to pass in but since it's a system-generated timestamp we would want to show it in the user's timezone.

I would suggest making yet another formatting function. I tested this one out and it worked on the incident table (same except it'll attempt to guess the user tz and default to PT if it fails. The guess function caches its guess). It would also be possible to use native JS datetime since moment.tz is only required to _convert_ between timezones, I just find it much harder to deal with the options for .toLocaleString() [MDS doc on options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#options)
```
export const formatDateTimeUserTz = (
  dateTime: Date | string,
): string =>
  {
    return moment(dateTime, true).isValid() && moment.tz(dateTime, moment.tz.guess() ?? DEFAULT_TIMEZONE).format(DATETIME_TZ_FORMAT)};
```

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
isuru-aot: Well, `information_requirement_table`, `major_mine_application`, `project_dicesion_package` and `project_summary` all these call `trigger_notification` function as those are different file uploads. File archive is captured in mine_document_resource, only.

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
isuru-aot: This is done.

services/core-web/src/components/navigation/NotificationDrawer.tsx
isuru-aot: Added a new function as suggested.

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
taraepp: Hmm, I'm wondering if there's a way we can reduce the repetition. Is the only difference between these functions the class name that's being queried? If so, I think that could be passed into a helper function (maybe in a project file?).

matbusby-fw: You could maybe add a universal function for this in `mine_document_resource`, but I don't know how much benefit there would be there.  

While these would all be doing the same function, there are enough unique elements in each, that creating the generic function is going require either a bunch of props passed in or do a bunch of checks.  Personally I think I'd be more in favour of the minor repetition.  A few more lines of code, but it's also easier to read.

matbusby-fw (dismissed):
I think this is looking really good.  I see Tara's point about making the document trigger notification functions more universal, but I'd be happy with it either way.

Nice work!  Documents are used in so many places that these tasks balloon pretty quickly, but it's looking good!

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
taraepp: I'd be satisfied with the really consistent parts being in a function.
- the renotify period
- the message text 
- optionally the feature flag check (not important since it'll go away)

It would be a _short_ function to have just have renotify_hours = 24 and the call to trigger_notification, but then if we want to change the time period, recipients, or the text we don't have to do it in 5 places.

isuru-aot: @taraepp I get your point. Actually `renotify_period` is already added as an optional param to `trigger_notification` function in `utils.py`. So only major repeating param is the text. Is there a way this text can be read from a file where we keep constants, so we can avoid the repetition?

taraepp (approved):
LGTM

matbusby-fw (approved):
🙌🏻

PR #2634, opened 2023-08-21
[MDS-5395] view archived doc versions

henryoforeh-dev:
## Objective 

[MDS-5395](https://bcmines.atlassian.net/browse/MDS-5395)

- Added functionality to enable viewing old versions of archived documents.
- Added functionality to display file location of archived documents.
- Fixed some css related issues on the document tables.

image

image


services/core-api/app/api/mines/documents/models/mine_document.py
```diff
@@ -37,6 +38,12 @@ class _ModelSchema(Base._ModelSchema):
     archived_by = db.Column(db.String(60))
 
     versions = db.relationship('MineDocumentVersion', lazy='joined')
+    major_mine_application_document_xref = db.relationship(
+        'MajorMineApplicationDocumentXref',
+        lazy='select',
+        uselist=False,
+        primaryjoin='and_(MajorMineApplicationDocumentXref.mine_document_guid == MineDocument.mine_document_guid, MineDocument.is_archived == True)'
+    )
```
simensma-fresh, line 41: I think there's a couple of more document types that would fall into this that can be archived (check out MineDocumentSearchUtil)
- ProjectSummaryDocumentXref
- ProjectDecisionPackageDocumentXref
- InformationRequirementsTableDocumentXref
- MajorMineApplicationDocumentXref

simensma-fresh (changes requested):

services/core-api/app/api/mines/documents/models/mine_document.py
henryoforeh-dev: Nice catch. I will implement them.

henryoforeh-dev: @simensma-fresh updated as advised.

simensma-fresh (dismissed):

services/core-web/common/components/documents/ArchivedDocumentsSection.tsx
```diff
@@ -3,10 +3,11 @@ import DocumentTable from "@/components/common/DocumentTable";
...
 import { DeleteOutlined } from "@ant-design/icons";
 import { Feature, isFeatureEnabled } from "@mds/common";
-import { MineDocument } from "@common/models/documents/document";
-
+import { MajorMineApplicationDocument } from "@common/models/documents/document";
+import { renderCategoryColumn } from "@/components/common/CoreTableCommonColumns";
+import * as Strings from "@common/constants/strings";
 interface ArchivedDocumentsSectionProps {
-  documents: MineDocument[];
+  documents: MajorMineApplicationDocument[];
```
taraepp: are we able to use the base model or does it not work that way? 
I am very hesitant to make this component only usable for one type of file. Prefer to change document model instead so that we are not relying on attributes specific to one extension of it.

services/core-web/common/components/documents/ArchivedDocumentsSection.tsx
```diff
@@ -16,6 +17,22 @@ const ArchivedDocumentsSection = (props: ArchivedDocumentsSectionProps) => {
     return <></>;
   }
 
+  const parseArchivedDocuments = () => {
+    return props.documents.map((obj) => ({
+      ...obj,
+      key: obj.mine_document_guid,
+      file_location_type_code: obj.major_mine_application_document_xref.major_mine_application_document_type_code
+        || obj.project_summary_document_xref.project_summary_document_type_code
```
taraepp: I'm thinking that when we set these attributes we can just set them as the file_location_type_code when we create the object and then not have to do a bunch of "OR"

services/core-web/common/components/documents/ArchivedDocumentsSection.tsx
henryoforeh-dev: I created the `file_location_type_code ` property in the child model. I saw something similar was done with the `major_mine_application_document_type_code` one.

taraepp: Can we instead pass in "file_location_type_code" as a prop like "categoryAttributeName", or use the category attribute in the document model? Currently we are not using the category attribute at all and I think this is a good use for it. Or pass in a prop like "extraColumns"? Otherwise, if that is not possible, and we are going to make this component be specific to MajorMineApplications we should rename it to reflect that.

henryoforeh-dev: I have updated it. The required attributes are now in the base model. I did it this way to mirror the model in core-api. The additionalColumns attribute is now a prop that will passed by the consuming component.

services/core-web/common/components/documents/ArchivedDocumentsSection.tsx
henryoforeh-dev: Updated.

services/core-web/common/models/documents/document.ts
```diff
@@ -69,9 +79,26 @@ export class MineDocument {
     this.archived_by = jsonObject.archived_by;
     this.archived_date = jsonObject.archived_date;
     this.is_latest_version = jsonObject.is_latest_version ?? true;
+    this.category_code = this.determineCategoryCode(jsonObject);
     this.setCalculatedProperties(jsonObject);
   }
 
+  private determineCategoryCode(jsonObject: any): string | undefined {
```
taraepp: Little nitpick:
- to make this method available to extended classes, change from private to protected
- the project_summary_document_xref, etc, I think those are only for MajorMineApplicationDocuments, right?
- then we can use the appropriate type in the archived documents section, and it'll have the appropriate permissions based on the type.

henryoforeh-dev: @taraepp Updated.

services/core-web/common/models/documents/document.ts
```diff
@@ -154,24 +123,74 @@ export class MineDocument {
...
+
+  protected makeChild(params: any, constructorArgs: any) {
     return new MajorMineApplicationDocument({
       ...params,
       major_mine_application_document_type_code:
         constructorArgs.major_mine_application_document_type_code,
     });
   }
 
+  protected setCalculatedProperties(jsonObject: any) {
```
taraepp: I don't think the content of this needs to be duplicated from the base model, since the base model passes the jsonObject to the makeChild method.

services/core-web/common/components/documents/ArchivedDocumentsSection.tsx
taraepp: Can we change this to use the MineDocument though?

matbusby-fw (dismissed):
Late to the party here, but it looks great!  Nice one!

taraepp (approved):
🎉 🎨

matbusby-fw (approved):

PR #2638, opened 2023-08-25
[MDS-5346]  - document replace functionality

matbusby-fw:
## Objective 

[MDS-5346](https://bcmines.atlassian.net/browse/MDS-5346)

- Implemented the document version replace functionality on MS and Core
- Updated backend versioning to properly track previous versions
- Updated document model to display updated versions properly

image
image
image


services/core-web/common/actionCreators/projectActionCreator.ts
```diff
@@ -615,7 +615,7 @@ export const postNewDocumentVersion = ({
     )
     .then((response: AxiosResponse<IProjectSummary>) => {
       notification.success({
-        message: "Successfully sent new document version",
+        message: "Successfully created new document version",
```
taraepp: I loooooove how generic this function looks. Is it generic enough to put it in a documentActionCreator file?

services/core-web/common/models/documents/document.ts
```diff
@@ -89,13 +89,12 @@ export class MineDocument {
 
     const versions = jsonObject.versions ?? [];
     if (this.is_latest_version && versions.length) {
-      this.number_prev_versions = versions.length - 1;
+      this.number_prev_versions = versions.length;
```
taraepp, line 92: 🎨

services/core-web/common/actionCreators/projectActionCreator.ts
matbusby-fw: Good idea!  I'll take a look at adding that on Monday!

services/core-web/src/components/Forms/projectSummaries/ProjectSummaryDocumentUpload.js
```diff
@@ -85,6 +85,7 @@ export class ProjectSummaryDocumentUpload extends Component {
             documents={this.props.initialValues?.documents?.reduce(
               (docs, doc) => [
                 {
+                  ...doc,
```
taraepp, line 88: I think that we've got some extraneous attributes here, lots of x: doc.x that should now be taken care of with ...doc. I think the key should already be in the doc, which would just leave the category to be specified.

services/core-web/common/components/documents/ReplaceDocumentModal.tsx
```diff
@@ -0,0 +1,167 @@
+import React, { FC, useState } from "react";
+
+import { Alert, Button, Col, Form, notification, Row, Typography } from "antd";
+import { MineDocument } from "@common/models/documents/document";
+import { formatDate } from "@common/utils/helpers";
+import FileUpload from "@/components/common/FileUpload";
+import { NEW_VERSION_PROJECT_SUMMARY_DOCUMENTS } from "@common/constants/API";
```
taraepp: I think that technically this file is actually usable by all documents! 🎉 
Just have a couple of files/consts named in a way that indicates they aren't.

services/minespace-web/cypress/support/commands.ts
```diff
@@ -26,7 +26,7 @@ Cypress.Commands.add("login", () => {
 
   cy.visit(url);
   cy.contains("Log in with BCeID").click();
-  cy.get("#username").type(Cypress.env("CYPRESS_TEST_USER"));
-  cy.get("#password").type(Cypress.env("CYPRESS_TEST_PASSWORD"));
-  cy.get("#kc-login").click();
+  // cy.get("#username").type(Cypress.env("CYPRESS_TEST_USER"));
```
matbusby-fw, line 29: These checks are failing due to something to do with the cypress keycloak.  Have disabled them for the time being and created a ticket to fix the issue.

asinn134 (approved):

simensma-fresh (approved):

PR #2640, opened 2023-08-30
[MDS-5450] Use selectin loader for Incident documents, and TSFs

simensma-fresh:
## Objective 

[MDS-5450](https://bcmines.atlassian.net/browse/MDS-5450)

Attempt to improve the performance of the Incident endpoints which in some circumstances causes the core api to error with an OOM error.

EXPLAIN ANALYZEing the filter query, I found that the document and TSF joins was a big reason for the amount of rows being returned by the query. 

Note:
This does not solve the overall issue around only returning the data we need - why do we need TSFs loaded when looking up an Incident? Turns out that's a bit of a bigger thing, but hopefully this PR should resolve the immediate issue.


asinn134 (approved):

matbusby-fw (approved):
🤞🏻

PR #2642, opened 2023-08-30
[MDS-5453] 404 error when trying to download all application files

asinn134:
## Objective 

[MDS-5453](https://bcmines.atlassian.net/browse/MDS-5453)

- When clicking download all application file button added a notification to inform user when files fail to retrieve for compression.

Screenshot (163)

### DIFF
--- a/services/document-manager/backend/app/services/object_store_storage_service.py
+++ b/services/document-manager/backend/app/services/object_store_storage_service.py
@@ -72,16 +72,19 @@ class ObjectStoreStorageService():
         def generate(result):
             for chunk in iter(lambda: result['Body'].read(1048576), b''):
                 yield chunk
-
-        s3_response = self._client.get_object(Bucket=Config.OBJECT_STORE_BUCKET, Key=path)
-        resp = Response(
+        try:
+            s3_response = self._client.get_object(Bucket=Config.OBJECT_STORE_BUCKET, Key=path)
+            resp = Response(
             generate(s3_response),
             mimetype='application/pdf' if '.pdf' in display_name.lower() else 'application/zip',
             headers={
                 'Content-Disposition':
                     ('attachment; ' if as_attachment else '') + ('filename=' + display_name)
             })
-        return resp
+            
+            return resp
+        except ClientError as e:
+            raise Exception(f'Failed to download the file: {e}')
 
     def upload_file(self, filename, progress=False):
         key = f'{Config.S3_PREFIX}{filename[1:]}'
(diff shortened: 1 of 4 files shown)


matbusby-fw (dismissed):
Looks good!  We still need to figure out why some of the files are failing, but this will give some good protection against error states.

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -72,16 +72,19 @@ def download_file(self, path, display_name, as_attachment):
...
             generate(s3_response),
             mimetype='application/pdf' if '.pdf' in display_name.lower() else 'application/zip',
             headers={
                 'Content-Disposition':
                     ('attachment; ' if as_attachment else '') + ('filename=' + display_name)
             })
-        return resp
+            
+            return resp
+        except ClientError as e:
```
taraepp, line 86: Is ClientError the only possible error type? I'm wondering if this should be a more generic exception.
Looks good though!

asinn134: I was wondering about that too, there are definitely more error types but I noticed most of the other functions in this file use ClientError and in the AWS Boto3 doc section they mention ClientError is the general exception. I think ClientError might be the most generic when it comes to Boto3.

Screenshot (167)

taraepp: The most generic exception will always just be "Exception" but if errors are handled on the FE I think it probably makes sense to let it produce a 500 if another type of error occurs (ex: if display_name gets passed in as undefined or not a string so can't do .lower() on it- errors resulting from the code or bad parameters).

Their language here is a little bit vague. They say it's "**a** general" exception, and "the most common" but they don't say it's "the gener**ic**" exception or that all errors from botocore will be ClientErrors. When I look at [this example here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#catching-botocore-exceptions) where they catch the ClientError and then catch a different botocore exception underneath it, that indicates to me that the botocore.exceptions.ParamValidationError does not get caught as a ClientError.

asinn134: Ah I see, then would you recommend setting it up like this:

```
        except Exception as e:
            raise Exception(f'Failed to download the file: {e}')
```

taraepp (approved):

matbusby-fw (approved):
👍🏻

PR #2644, opened 2023-08-31
[MDS-5467] - Download Document Versions

matbusby-fw:
[MDS-5467](https://bcmines.atlassian.net/browse/MDS-5467)
[MDS-5346](https://bcmines.atlassian.net/browse/MDS-5346)
[MDS-5471](https://bcmines.atlassian.net/browse/MDS-5471)

Included the fixes for 5346 and 5471 as they touched on the same elements modified in this PR

## Objective 

- Reworked the saving of versions (it looked like it was working with my previous changes, but wasn't actually saving the versionId)
- updated the actionless network call to download a file, to optionally include a versionId
- updated the documant manager api to accept the versionId on a download and utilize it if it exists.
- Added a MineDocumentVersion class to `document.ts`. Previously versions were using the same model as documents, but they do not adhere to the same structure.
- Made spatial files replacable in the `ReplaceDocumentModal`
- Removed actions from documents that do not yet have a mine_document_guid

### DIFF
--- a/services/core-web/common/models/documents/document.ts
+++ b/services/core-web/common/models/documents/document.ts
@@ -27,6 +27,8 @@ export class MineDocumentVersion {
 
   public update_timestamp: string;
 
+  public allowed_actions: FileOperations[];
+
   constructor(jsonObject: any) {
     this.mine_document_guid = jsonObject.mine_document_guid;
     this.mine_document_version_guid = jsonObject.mine_document_version_guid;
@@ -37,6 +39,7 @@ export class MineDocumentVersion {
     this.upload_date = jsonObject.upload_date;
     this.create_user = jsonObject.create_user;
     this.update_timestamp = jsonObject.update_timestamp;
+    this.allowed_actions = jsonObject.allowed_actions;
   }
 }
 
@@ -133,6 +136,10 @@ export class MineDocument {
               is_latest_version: false,
               mine_document_guid: this.mine_document_guid,
               document_manager_guid: this.document_manager_guid,
+              allowed_actions: [
+                this.file_type === ".pdf" && FileOperations.View,
+                FileOperations.Download,
+              ],
             },
             jsonObject
           )
--- a/services/core-web/common/models/documents/document.ts
+++ b/services/core-web/common/models/documents/document.ts
@@ -1,4 +1,4 @@
-import { USER_ROLES, Feature, isFeatureEnabled } from "@mds/common";
+import { Feature, USER_ROLES, isFeatureEnabled } from "@mds/common";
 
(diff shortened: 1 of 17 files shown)


services/core-web/common/models/documents/document.ts
```diff
@@ -150,25 +200,13 @@ export class MajorMineApplicationDocument extends MineDocument {
...
-      ?? information_requirements_table_document_xref?.information_requirements_table_document_type_code
-      ?? major_mine_application_document_xref?.major_mine_application_document_type_code
+      project_summary_document_xref?.project_summary_document_type_code ??
+      project_decision_package_document_xref?.project_decision_package_document_type_code ??
+      information_requirements_table_document_xref?.information_requirements_table_document_type_code ??
+      major_mine_application_document_xref?.major_mine_application_document_type_code
     );
   }
 
-  protected makeChild(params: any, constructorArgs: any) {
```
taraepp, line 160: I think there will be some difficulty in making versions a different type from the latest version. The file permissions are based on the type, and it keeps the properties consistent for document tables. If it works, that'd be a nice bit of clean-up, but I would be surprised if the document tables still have the right column data and allowed actions based on user roles.

matbusby-fw: The data in the columns is still displaying correctly for the versions actually 😄 

They have a handful of common fields between MineDocuments and MineDocumentVersions but passing the version info into a MineDocument was mostly just ending up with a bunch of undefined fields.  Since versions should always conform to the same structure, I think it makes sense to simplify them like this.

I did add the allowed actions for versions, since they should always just be download and or view.

taraepp: Ahh, that's right, we took out the File Location column that displays the major_mine_application_type_code, so it wouldn't produce an error with the columns we're using. If we're not planning on displaying versions with columns that are specific to the type then this shouldn't be a problem.

Might still suggest passing the view/download permission from the parent, so that we'd be able to restrict that based on permission (should be just as easy). I thought that the child data still wanted a unique key though- are you not getting a "Each record in table should have a unique key prop" message? No issues sorting? <-- that's when I ran into weird bugs

matbusby-fw: Hmmm, let me check the key issue.

Sure, we can pass the view/download permission from the parent.  Do we ever restrict this though?  I thought we only restricted the other functions.

asinn134 (dismissed):

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):

asinn134 (approved):

PR #2649, opened 2023-09-06
[MDS-5319] archive multiple files

henryoforeh-dev:
## Objective 

[MDS-5319](https://bcmines.atlassian.net/browse/MDS-5319)

- Implemented bulk archive in core-web and minespace-web.
- Fixed a bug that prevented re-rendering of documents in documents tab.

image

image


services/minespace-web/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -111,7 +112,7 @@ export const renderActionsColumn = (
       return (
         <div>
           {items.length > 0 && (
-            <Dropdown menu={{ items }} placement="bottomLeft">
+            <Dropdown menu={{ items }} placement="bottomLeft" disabled={isRowSelected}>
```
taraepp, line 115: We could probably extract a function from the bit above (the stuff that happens in filteredActions.map() to create the items) so that we can reduce some of the bulk in the DocumentTable, the bulkItems seem to be the same.

services/minespace-web/src/components/common/DocumentTable.tsx
```diff
@@ -266,20 +318,47 @@ export const DocumentTable = ({
...
-        recordDescription: "version history",
-        rowExpandable: (record) => record.number_prev_versions > 0,
-      }}
-    />
-  ) : (
-    <CoreTable columns={columns} dataSource={documents} {...minimalProps} />
+  const renderCoreTable = () => {
+    let element = (
+      <CoreTable columns={columns}  {...(enableBulkActions
+        ? {
```
taraepp: Can we do all the logic for the table props _before_ trying to render the table? Possibly like how we have minimalProps, have bulkActionsProps, versionProps. Would be nice to reduce the cognitive complexity and repetition here. I think that before your change, we had accidentally lost the condition=isLoaded prop from the second return statement.

services/minespace-web/src/components/common/DocumentTable.tsx
```diff
@@ -205,6 +239,24 @@ export const DocumentTable = ({
     },
   ].filter((action) => allowedTableActions[action.label]);
 
+  const bulkItems: MenuProps["items"] = [
+    {
+      key: "0",
+      icon: <DeleteOutlined />,
```
taraepp: Previously we had used this icon for Delete and `<InboxOutlined />` for Archive. Just want to double-check that it didn't get switched by accident.

services/minespace-web/src/components/common/DocumentTable.tsx
```diff
@@ -167,6 +171,36 @@ export const DocumentTable = ({
     });
   };
 
+  const handleRowSelectionChange = (value) => {
+    if (documentTypeCode === "") {
+      setDocumentTypeCode(value[0].major_mine_application_document_type_code);
```
taraepp: Not all documents have major_mine_application_document_type_code. I see where we're setting it but not where we're using it- can it be removed? 

I expanded the file looking for references to it and I think there's some unused prop types listed as well, I don't think fileOperationPermissionMap needs to be here, might be more.

services/minespace-web/src/components/common/CoreTableCommonColumns.tsx
henryoforeh-dev: I left this one as discussed in our call.

services/minespace-web/src/components/common/DocumentTable.tsx
henryoforeh-dev: Excellent call, updated in both core-web and minespace-web as discussed.

services/minespace-web/src/components/common/DocumentTable.tsx
henryoforeh-dev: Updated. I also sent a message to clarify same in ux-pod channel.

services/minespace-web/src/components/common/DocumentTable.tsx
henryoforeh-dev: Removed as discussed.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -348,42 +379,21 @@ export const DocumentTable = ({
...
-              : {})}
-            expandProps={{
-              childrenColumnName: "versions",
-              matchChildColumnsToParent: true,
-              recordDescription: "version history",
-              rowExpandable: (record) => record.number_prev_versions > 0,
-            }}
-          />
-        </div>
+        <CoreTable
```
taraepp: Much better, but I think we can just return a single instance.
like 
```
<CoreTable
condition
dataSource
columns
...bulkActions
...versions
...minimal
/>
```

The condition prop I don't think was removed on purpose, and minimalProps is going to be null anyway unless we explicitly pass in that we want it to be minimal. (If there's an issue with minimalProps in combination with versionProps, I think we can worry about it when we set both on the same table at the same time).

matbusby-fw: Ooh, tidy! 😁

taraepp (approved):
LGTM!

Thank you for tidying up the DocumentTable! 🎨

matbusby-fw (approved):
🚂 🚋 🚋 🚋 🚋

asinn134 (approved):

PR #2651, opened 2023-09-07
[MDS-5466]archive docs appearing randomly

asinn134:
## Objective 

[MDS-5466](https://bcmines.atlassian.net/browse/MDS-5466)

- fixed bug where projects were seeing all archived documents from a mine instead of the ones archived within that project


henryoforeh-dev (approved):
Clean!

taraepp (approved):

PR #2655, opened 2023-09-11
[CHORE] Fix cypress authentication in Minespace and Core

simensma-fresh:
## Objective 

Fixes a couple of issues with the Cypress auth in MS / Core.

**Roles**
The current `cypress` user used for testing has both the `mds_minespace_proponents` and roles required for core. A user with the `mds_minespace_proponents` role is not allowed to access core, so we need to split it into two separate users, one for core, and one for minespace.

I changed it to define users in `keycloak-users.json`, where you can specify the roles each of them should have, and define the roles themselves in realm-export.json. Note: these were exported from the dev gold KeyCloak instance, but none of the composite roles were ported over properly. Added in a `oidc-script-based-protocol-mapper` here to put the users roles into the `client_roles` token attribute, to match the token structure of the gold keycloak instance.

**Init of JWT cypress client**
Added in conditional initialization of the `jwt_cypress` client - it can be toggled on or off with the `ALLOW_CYPRESS_AUTH` env variable

**Cypress Studio**
Enabled [Cypress Studio](https://docs.cypress.io/guides/references/cypress-studio)


matbusby-fw (approved):
🚂 🚋 🚋 🚋 🚋

henryoforeh-dev (approved):

PR #2662, opened 2023-09-12
[MDS-5402] Core Homepage

taraepp:
## Objective 

[MDS-5402](https://bcmines.atlassian.net/browse/MDS-5402)
- update the core homepage

image
image


services/core-api/app/api/mines/alerts/models/mine_alert.py
```diff
@@ -17,6 +19,10 @@ class MineAlert(SoftDeleteMixin, AuditMixin, Base):
 
     end_date = db.Column(db.DateTime)
 
+    mine_table = db.relationship('MineMapViewLocation', lazy='joined', primaryjoin='and_(foreign(MineAlert.mine_guid) == remote(MineMapViewLocation.mine_guid))')
+    mine_name = association_proxy('mine_table', 'mine_name')
+    mine_no = association_proxy('mine_table', 'mine_no')
+
```
taraepp, line 25: I waffled over this _a lot_.
The reason I'm using MineMapViewLocation vs Mine is that joining _Mine_ results in several other tables being join, but MineMapViewLocation is a view of just a few columns from the mine table (so only one join). Did not add the mine name/no to the view, it was already there, but I did add those fields to the python model.

services/core-web/common/reducers/mineReducer.ts
```diff
@@ -15,7 +15,7 @@ interface MineState {
   mineGuid: string;
   mineBasicInfoList: IMine[];
   mineDocuments: IMineDocument[];
-  subscribedMines: IMine[];
+  subscribedMines: { records: IMine[]; loaded: boolean };
```
taraepp, line 18: The redux state logic doesn't update synchronously without tying this in, making my skeleton components "jump" as it loads. There's some room for discussion on how exactly to structure it though, basically when to separate it into 2 different functions or properties (search getGlobalMineAlerts to compare). Having the selector return an object with 2 properties (records & loaded) means we create fewer functions, but having 2 selectors like getXs and getXsLoaded is easier to implement retroactively without breaking anything.

services/core-web/src/App.tsx
```diff
@@ -12,16 +12,32 @@ export const store = configureStore();
...
-    <ScrollToTopWrapper>
-      <>
-        <Routes />
-        <ModalWrapper />
-        <DocumentViewer />
-      </>
-    </ScrollToTopWrapper>
-  </BrowserRouter>
-);
+export const useKey = (test: (e: KeyboardEvent) => boolean, ref) => {
```
taraepp, line 15: I haven't tested to see if we could use this to add other hotkeys to the page, but I don't see why not!

services/core-web/src/actionCreators/mineAlertActionCreator.ts
```diff
@@ -10,76 +10,100 @@ import * as reducerTypes from "@/constants/reducerTypes";
...
-        dispatch(error(reducerTypes.GET_MINE_ALERTS));
-        throw new Error(err);
-      })
-      .finally(() => dispatch(hideLoading()));
-  };
+export const fetchGlobalMineAlerts = (): AppThunk<Promise<AxiosResponse<IMineAlert>>> => (
+  dispatch
+): Promise<AxiosResponse<IMineAlert>> => {
+  dispatch(request(reducerTypes.GET_GLOBAL_MINE_ALERTS));
+  dispatch(showLoading());
```
taraepp, line 17: This function (fetchGlobalMineAlerts) was added, the rest in this file is formatting

services/core-web/src/components/homepage/HomeBanner.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Typography, Col, Row } from "antd";
+
+import SearchBar from "@/components/search/SearchBar";
+import { BACKGROUND } from "@/constants/assets";
+
+const HomeBanner = () => {
+  return (
+    <div
+      style={{ backgroundImage: `url(${BACKGROUND})`, padding: "64px 32px" }}
```
taraepp, line 10: I left some inline styles here! 🙈 
These are one-offs though, maybe doesn't matter.

services/core-web/src/components/search/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const IconFor = (item: ISearchResult) =>
+    ({
+      mine: <img className="icon-svg-filter" src={MINE} alt={item.value} height={25} />,
+      party: (
+        <img className="icon-svg-filter" src={PROFILE_NOCIRCLE} alt={item.value} height={25} />
+      ),
+      permit: <img className="icon-svg-filter" src={DOC} alt={item.value} height={25} />,
+    }[item.type]);
+
+  const getDropdownMenuItems = () => {
```
taraepp, line 40: The big change for the dropdown is here (I copied the 2 previous functions). Now this no longer generates the overlay is deprecated, use dropdown instead. It takes in menu items, which are easy to translate to a JSON format from `<Menu.Item>`, and now the element that triggers the dropdown (the search input) must be wrapped in the dropdown as a child. There are other places that need a similar update.

PR #2664, opened 2023-09-13
[MDS-5448] Refactor file upload process for better error handling

matbusby-fw:
Revised the file upload process to poll for upload status after the upload itself has succeeded (with the post-finish hook). 
Previously, no error was raised to the user if the upload failed at this point, resulting in documents that had failed without the user knowing about it.

This will cover the copying of file to it's new directory
Deleting the old file
and any other errors that occur in the post-finish hook

This is handled by adding a "Status" column to the document manager document table.  This column defaults to "In Progress".  It get's set to any error state that is raised, or "Success" if no errors occur.  The front end then polls the api for this status and doesn't mark the upload as complete unless it receives a "Success" status.

## Objective 

[MDS-5448](https://bcmines.atlassian.net/browse/MDS-5448)


services/document-manager/backend/migrations/versions/9be7dbc60ff8_add_status_message_to_document_table.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '9be7dbc60ff8'
+down_revision = 'c02786c20bfd'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("document") as batch_op:
+        batch_op.add_column(sa.Column('status', sa.String(length=255), server_default=text("'In Progress'"), nullable=True))
```
simensma-fresh, line 21: Will this mark all existing documents as "In Progress"?

Thinking it might be better to maybe lave it blank or set it to `Success` for existing ones to avoid any future confusion 😅

matbusby-fw: Yeah it will.  Good point that it might make it confusing.  I think I'll set existing ones as Success and keep the default for future ones.

simensma-fresh (approved):

asinn134 (approved):
Nice, thanks for putting this together.

henryoforeh-dev (approved):

PR #2670, opened 2023-09-15
[MDS-5500] Fixed error when downloading Notice of Work applications

simensma-fresh:
## Objective 

[MDS-5500](https://bcmines.atlassian.net/browse/MDS-5500)

Fixes error when downloading Notice of Work applications

image

**Cause**  
The `pretty_folder` is no longer passed along properly to Docman when generating these PDFs, for some reason when there's an underscore in the header name, Tusd does not pass the header along.

**Solution**
Updated header name to `Prettyfolder` and it looks like it fixed things.

Also updated the `send_file` params which also do not match.


matbusby-fw (dismissed):
Well that's weird.  Good catch!

isuru-aot (dismissed):

henryoforeh-dev (approved):

matbusby-fw (approved):

PR #2672, opened 2023-09-18
[MDS-5500] Poll for status when uploading generated file to docman

simensma-fresh:
## Objective 

[MDS-5500](https://bcmines.atlassian.net/browse/MDS-5500)

Includes a potential fix for the ESUP + NoW permit downloads. It looks like it's a similar issue to the one previously encountered for regular uploads / downloads, where the `post-finish` hook executes asynchronously, so the client is not made aware of any errors that happen, or in this case, we try to access the file before the generated PDFs have finished uploading - causing the 500.


matbusby-fw (approved):
Ah, yeah that makes sense!

henryoforeh-dev (approved):

asinn134 (approved):

PR #2675, opened 2023-09-19
[MDS-5474] create esup amendment

henryoforeh-dev:
## Objective 

[MDS-5474](https://bcmines.atlassian.net/browse/MDS-5474)

-Implemented CRUD for ESUP amendment.
-Moved common functionality between ESUP and ESUP amendment to PermitMixin.


services/core-api/app/api/utils/models_mixins.py
```diff
@@ -278,6 +282,40 @@ def deep_update_from_dict(self, data_dict, _edit_key=None):
                 raise (e)
         return
 
+class PermitMixin(object):
```
matbusby-fw, line 286: Is it possible to include the regular db columns in the mixin as well?  There are so many common fields here, that seems like it would be nice.  I haven't played with the mixins much though, so maybe it doesn't work that way?

matbusby-fw (dismissed):
Looks good to me!  Just one question about maybe including more columns in the permitMixin.

services/core-api/app/api/utils/models_mixins.py
henryoforeh-dev: Perfect recommendation. I will refactor and update my PR.

asinn134 (approved):
nice work!

matbusby-fw (approved):
Awesome!

PR #2677, opened 2023-09-19
[MDS-5413] TSF EOR UI fixes

matbusby-fw:
## Objective 

[MDS-5413](https://bcmines.atlassian.net/browse/MDS-5413)

- Changed "Company Name" to "Company Affiliation" in the create EOR modal to match what is used elsewhere.
- Added text under "Engineer of Record Term"
- removed popConfirm on creating a new EOR when one currently does not exist.
- Small patch to the file upload process for local development.  If not connected to the S3 storage, the backend doesn't ever run the post-finish hook on an upload, so it's status would never be updated from "In Progress".  Added code to change the status to success in this instance.

image
image


taraepp (approved):
LGTM

isuru-aot (approved):

PR #2678, opened 2023-09-19
[MDS-5402] Added fontawesome icon to homepage + added utility to fetch and configure fontawesome pro access key

simensma-fresh:
## Objective 

[MDS-5402](https://bcmines.atlassian.net/browse/MDS-5402)

Added fontawesome icons to homepage.

**env variable fetching**
I had some fun and expanded `make env` a little bit to look up and configure yarn to use the FontAwesome pro token that I added to the `local-dev-secrets` secret in the `4c2ba9-dev`  openshift namespace.

Why?  This is the first secret that is required to actually build and run the app locally, and it would be good to avoid sharing it manually as much as possible. I imagine we would be able to build on top of this for other secrets (hey S3 Access Key).

`make env` will now
1. Prompt to re-generate .env
2. Prompt to load secrets from OpenShift
3. Prompt to install `oc` binary (only in codespaces)
4. Prompt to log into Openshift
5. Configure yarn with FontAwesome secret

image
image
image
image
image


henryoforeh-dev (approved):
Cool!

asinn134 (approved):

PR #2683, opened 2023-09-20
[MDS-5507] Set Font awesome token when building core-web

simensma-fresh:
## Objective 

[MDS-5507](https://bcmines.atlassian.net/browse/MDS-5507)

Missed this one in the original PR. Sets the fontawesome auth token when building core, to allow for `yarn install` to succeed


matbusby-fw (approved):
Perfect!  Just came across this error! 🙌🏻

taraepp (approved):

PR #2685, opened 2023-09-20
[MDS-5495] Remove Actions Menu from IRT template upload (#2684)

taraepp:
## Objective 
- made the instance of the table view only, which will disallow non-viewing actions
- the actions menu is still there but only contains "Download" (for the excel files expected to be there)
- also cleaned up some unused/undefined variables
- fixed getUserAccessData by calling it at the same point we do on core, so that the data exists.
- bonus: CI/CD changes for font-awesome on MS

[MDS-5495](https://bcmines.atlassian.net/browse/MDS-5495)

### DIFF
--- a/.github/workflows/minespace.unit.yaml
+++ b/.github/workflows/minespace.unit.yaml
@@ -25,6 +25,12 @@ jobs:
         with:
           node-version: 14.x
 
+      - name: Set Fontawesome token
+        run: |
+          yarn config set 'npmScopes["fortawesome"].npmAuthToken' "${{ secrets.FONT_AWESOME_PACKAGE_TOKEN }}"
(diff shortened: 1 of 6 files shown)


asinn134 (approved):

henryoforeh-dev (approved):

.github/workflows/minespace.unit.yaml
```diff
@@ -25,6 +25,12 @@ jobs:
         with:
           node-version: 14.x
 
+      - name: Set Fontawesome token
```
matbusby-fw, line 28: 🙌🏻

matbusby-fw (approved):
Woot woot!

PR #2688, opened 2023-09-21
[FIX] - logging and nullable column on vc connnection table

Jsyro:
Want to test webhook configuration but made an error in original table definition.


matbusby-fw (approved):
👍🏻

taraepp (approved):

henryoforeh-dev (approved):

PR #2689, opened 2023-09-22
[CHORE] - Listen for connection webhooks from traction. 

Jsyro:
## Objective 

acapy sends webhooks when things happen. So far we are only listening for 'connections' but will need to listen for 'present_proof' in the future. 

When things happen to the connection, update the record in MDS for brevity and speed. 

We will need to key off of pending vs active connections to allow/enable actions.


taraepp (approved):

henryoforeh-dev (approved):

henryoforeh-dev (approved):

Jsyro:
I cannot merge this PR, someone else will need to do it. Thanks for the review!

PR #2693, opened 2023-09-25
[MDS-5476] ESUP party/select validation (#2692)

taraepp:
## Objective 
- make select inputs on the form validate against the list of options
- cleaned up a bunch of TS/proptype errors (proptype errors resolved mostly by switching to TS, new errors found in the switch)
- there's still a `can't update component from the function body of another component` error remaining, this seems to be originating from the loading bar component and likely within Lottie (library) itself.

[MDS-5476](https://bcmines.atlassian.net/browse/MDS-5476)

Added the validation on all 4 of the fields with "asdf" typed in here, which were passing the "required" that most had for validation methods:
image


services/core-web/src/styles/settings/themeOverride.scss
```diff
@@ -288,8 +289,7 @@
 }
 
 // POPUP CONFIRM
-.ant-popover-inner-content,
-.ant-popover-arrow {
```
taraepp, line 292: this was making an ugly white box behind the popover arrow.

henryoforeh-dev (dismissed):

services/core-web/common/utils/Validate.ts
```diff
@@ -237,13 +238,19 @@ export const validateIncidentDate = memoize((reportedDate) => (value) =>
...
+ * @param data: options to choose from
+ * @param allowEmptyData: perform validation with empty data array (free text on empty data set will be invalid)
+ */
+export const validateSelectOptions = memoize((data: IOption[], allowEmptyData = false) => (value):
+  | string
+  | undefined => {
   if (value && (data?.length > 0 || allowEmptyData)) {
     return data?.find((opt) => opt.value === value) !== undefined
       ? undefined
       : "Invalid. Select an option provided in the dropdown.";
```
matbusby-fw, line 251: I realize this is pre-existing, but this text bothers me.  "Invalid." Doesn't seem like the smoothest UI choice.  Even something like "Selection Invalid." would be better I think.

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
```diff
@@ -400,11 +411,10 @@ const mapDispatchToProps = (dispatch) =>
...
-  connect(mapStateToProps, mapDispatchToProps),
+  connect<StateProps, any, ExplosivesPermitFormProps>(mapStateToProps, mapDispatchToProps),
   reduxForm({
     form: FORM.EXPLOSIVES_PERMIT,
     touchOnBlur: true,
-    validate: validateBusinessRules,
     onSubmitSuccess: resetForm(FORM.EXPLOSIVES_PERMIT),
   })
-)(ExplosivesPermitForm);
+)(ExplosivesPermitForm) as FC<ExplosivesPermitFormProps>;
```
matbusby-fw, line 420: Was this the Typescript error you were getting that was hard to figure out?  Redux is dumb.

services/core-web/src/components/mine/ExplosivesPermit/ExplosivesPermit.tsx
```diff
@@ -193,13 +180,13 @@ export const ExplosivesPermit = (props) => {
   };
 
   const handleOpenExplosivesPermitDecisionModal = (event, record) => {
+    console.log("hi tara from handleOpenExplosivesPermitDecisionModal");
```
matbusby-fw: Love this addition.  Very practical.

services/core-web/src/tests/components/mine/ExplosivesPermit/__snapshots__/MineExplosivesPermitTable.spec.js.snap
```diff
@@ -1,323 +1,7 @@
...
-          "key": "upload_date",
-          "render": [Function],
-          "title": "Date",
-        },
-      ],
-    }
-  }
-  rowKey={[Function]}
-/>
+<ContextConsumer>
```
matbusby-fw: I don't know if it really matters since we're moving towards more cypress testing.  But to get nested components to generate the correct snap I've had better luck with `@testing-library/react` rather than enzyme.  

There's a good example here.
`services/core-web/src/tests/components/common/FileUpload.spec.tsx`

That said, maybe it's not worth bothering.  Up to you.

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
taraepp: It was the interaction between this component and its parent! What worked with one did not work with the other, and then I had mistakenly put one of the "state props" in the "ESUP props" interface which looked like the same error until I inspected it closer (it was no longer missing many props, just one). The injected form props gave me trouble here. 

I saw a recommendation on SO to separate the redux props from the regular ones and I did that. I like it that way, very clear what's expected to be passed in vs what the component handles itself.

matbusby-fw (changes requested):
Nice!  Just a couple small nitpicks and a console log.  Otherwise looks great!

matbusby-fw (dismissed):

services/core-web/src/tests/components/mine/ExplosivesPermit/__snapshots__/MineExplosivesPermitTable.spec.js.snap
taraepp: DONE.
I didn't like the especially-useless snap, but it took some fiddling because it's a withRouter component.

henryoforeh-dev (approved):
Neat!

matbusby-fw (approved):

PR #2696, opened 2023-09-26
[FIX] Add trailing slash to webhook endpoint (#2695)

taraepp:
* update state based on webhooks

* don't use enum inheritance

* trailing slash

* add trailing slash

## Objective 

[MDS-5486](https://bcmines.atlassian.net/browse/MDS-5486)


Jsyro (approved):
I can approve this, but it doesn't count. :(

henryoforeh-dev (approved):

matbusby-fw (approved):

PR #2700, opened 2023-09-27
[FIX] vc_connection is a list.

Jsyro:
## Objective 

vc_connection is a list, one for each invitation that's created. should add hybrid property to party in the future.


asinn134 (approved):

matbusby-fw (approved):
👍🏻

PR #2706, opened 2023-10-05
[MDS-5456] adding soft deletion of now_application_document_identity_xref

asinn134:
## Objective 

[MDS-5456](https://bcmines.atlassian.net/browse/MDS-5456)

- Changes for making imported submission documents in NoW soft deletable instead of giving error _None Type object has no attribute now_application_.

- The issue is because documents with import status Success do not have a record in the now_application_document_xref database table, instead they can be found in the now_application_document_identity_xref table.

- Database changes made to add deleted_ind column inside now_application_document_identity_xref and now_submissions.document tables.


henryoforeh-dev (dismissed):
Nice!

matbusby-fw (approved):
🙌🏻

henryoforeh-dev (approved):

PR #2707, opened 2023-10-06
[MDS-5536] Fix bug on division by zero

henryoforeh-dev:
## Objective 

[MDS-5536](https://bcmines.atlassian.net/browse/MDS-5536)

Set default values for image width and height to prevent division by zero bug.


asinn134 (approved):
nice!

isuru-aot (approved):

PR #2711, opened 2023-10-11
[MDS-5525] Remove Actions menu from Bond document table (#2710)

taraepp:
## Objective 
- get delete functionality working on the Securities/Bonds/Edit Bond form again: actions menu replace & delete functions don't work on this form
- on merge to develop, eslint had a ton of work to do in correcting whitespace and complaining about unused vars, cleaned those up
- trying to keep the DocumentTable in Core & MS as similar as possible, reordered things in both files so structure matches

[MDS-5525](https://bcmines.atlassian.net/browse/MDS-5525)

image
- I was fully intending to alter how the DocumentTable interprets rather than the parent elements to fix the problem
- and for getting rid of replace functionality where it's not wanted, that was what I did (I have decided that version history + replace are interconnected and if you aren't showing the resulting version history from a replace, don't allow replace)
- but as it turns out, 2 forms (BondForm, ReclamationInvoice) and _only 2 forms_ were outliers with delete functions that don't delete (they add files to a list that will not be included when the user submits the form to delete the association)
   - the only other components with delete document functionality are either project/application components where we implemented Actions menu, and Variance-related files that have a proper delete function that calls the BE 
- previously, this worked. But with the delete document modal, the user is not returned to the form when they click delete, so this delete function + delete modal can't work together
- in the future, we should fix the delete functions used on these forms. 
- refactoring the modal to not replace modal content on open new modal would also be nice (see modalReducer)

### DIFF
--- a/services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
+++ b/services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
@@ -119,7 +119,9 @@ export const ExplosivesPermitForm: FC<ExplosivesPermitFormProps &
     "now_application_guid"
   );
 
-  const [isHistoric, setIsHistoric] = useState<boolean>(!initialValues?.explosives_permit_id && props.isPermitTab);
+  const [isHistoric, setIsHistoric] = useState<boolean>(
+    !initialValues?.explosives_permit_id && props.isPermitTab
+  );
   const isESUP = props.userRoles.includes(USER_ROLES[Permission.EDIT_EXPLOSIVES_PERMITS]);
   // eslint-disable-next-line no-unused-vars
   const hasEditPermission = isESUP;
@@ -134,12 +136,12 @@ export const ExplosivesPermitForm: FC<ExplosivesPermitFormProps &
   const handleRadioChange = (e) => {
     setRadioSelection(e.target.value);
     setIsHistoric(e.target.value == 1);
-    setIsAmend(e.target.value==3);
+    setIsAmend(e.target.value == 3);
   };
 
   const handleOpenAddExplosivesPermitModal = () => {
-    setParentView(false)
-  }
+    setParentView(false);
+  };
 
   const descriptionListElement = (
     <div>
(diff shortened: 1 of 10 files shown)


services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
```diff
@@ -172,33 +177,44 @@ export const ExplosivesPermitForm: FC<ExplosivesPermitFormProps &
...
+          <Typography.Paragraph>
+            Let&apos;s get your permit started, in CORE you can...
+          </Typography.Paragraph>
           {descriptionListElement}
         </div>
-        <div  className="landing-list">
-          <h4 className="uppercase">DEFAULT TO "ADD EXISTING" FROM PERMIT PAGE / "CREATE NEW" FROM APPLICATION PAGE</h4><br/>
+        <div className="landing-list">
+          <h4 className="uppercase">
+            DEFAULT TO &quot;ADD EXISTING&quot; FROM PERMIT PAGE / &quot;CREATE NEW&quot; FROM
```
taraepp, line 204: escaped the quotes here because eslint wanted me to

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitForm.tsx
```diff
@@ -220,282 +236,289 @@ export const ExplosivesPermitForm: FC<ExplosivesPermitFormProps &
...
-            <Button className="full-mobile">
-              Cancel
-            </Button>
+            <Button className="full-mobile">Cancel</Button>
           </Popconfirm>
-          <Button disabled={isAmend} type="primary" onClick={(e) => handleOpenAddExplosivesPermitModal()}>
+          <Button
+            disabled={isAmend}
+            type="primary"
+            onClick={() => handleOpenAddExplosivesPermitModal()}
```
taraepp, line 244: and got rid of the unused function arg here, also because eslint wanted me to.

The rest was just eslint cleaning up the formatting on its own.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -41,20 +41,18 @@ interface DocumentTableProps {
...
   openDocument: any;
-  closeModal?: () => void;
+  closeModal: () => void;
   removeDocument: (event, doc_guid: string, mine_guid: string) => void;
-  archiveMineDocuments?: (mineGuid: string, mineDocumentGuids: string[]) => void;
+  archiveMineDocuments: (mineGuid: string, mineDocumentGuids: string[]) => void;
   onArchivedDocuments?: (docs?: MineDocument[]) => void;
-  onReplaceDocument?: (document: MineDocument) => void;
-  documentColumns: ColumnType<unknown>[];
+  documentColumns?: ColumnType<unknown>[];
```
taraepp, line 50: there were some props here that were completely unused that I removed (onReplaceDocument, handleRowSelectionChange) otherwise I made the reduxy ones required and some of the other ones not required.

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -83,13 +81,13 @@ export const DocumentTable: FC<DocumentTableProps> = ({
...
   const [documentsCanBulkDropDown, setDocumentsCanBulkDropDown] = useState(false);
-
   const { isFeatureEnabled } = useFeatureFlag();
 
   const allowedTableActions = {
     [FileOperations.View]: true,
     [FileOperations.Download]: true,
-    [FileOperations.Replace]: !isViewOnly,
+    // don't allow changes to version history where history is not shown
+    [FileOperations.Replace]: !isViewOnly && showVersionHistory,
```
taraepp, line 90: The actual change here. The rest below is rearranging.

henryoforeh-dev (approved):
Clean!

services/core-web/src/components/Forms/Securities/BondForm.js
```diff
@@ -71,6 +78,8 @@ export class BondForm extends Component {
     }));
   };
 
+  // TODO: this function will have to remove the file through a BE call
```
matbusby-fw, line 81: I know we don't want to balloon scope here, but with other TODOs I've found throughout the system, by the time we get around to looking at them, they're 2-3 years old and many of the processes surrounding them have changed enough that they don't really help.  Maybe this would make more sense as a ticket for the backlog?  Or maybe both?

matbusby-fw (approved):
Looks good!  Just left some thoughts on the use of TODOs.

asinn134 (approved):

PR #2712, opened 2023-10-11
[MDS-5337] Cache fontawesome using Artifactory

simensma-fresh:
## Objective 

[MDS-5337](https://bcmines.atlassian.net/browse/MDS-5337)
- Switched to downloading fontawesome packages from the [Artifactory BCGov service](https://developer.gov.bc.ca/Artifact-Repositories-\(Artifactory\)) in order to get around the Fontawesome npm package bandwith limits
- Added documentation on the setup, and how to upgrade Fontawesome to newer versions when required

This includes some type-related changes. Why? When re-generating `yarn.lock` in order to change the npm repository of Fontawesome, about 100 Typescript related react errors popped up. Why? Apparently some peer-dependencies requires `@types/react` of version `*` which resolves to the latest published react types, which caused things to throw errors. Solution: Force resolution of `@types/react`  to match React version we're using


services/core-web/common/components/SteppedForm.js
```diff
@@ -10,7 +10,7 @@ import { indexOf } from "lodash";
 const { Item } = Menu;
 
 const propTypes = {
-  children: PropTypes.arrayOf(PropTypes.node).isRequired,
+  children: PropTypes.arrayOf(PropTypes.any).isRequired,
```
simensma-fresh, line 13: For some reason I could not figure out, `PropTypes.node` would not recognize the `children` property type correctly, hence this unfortunate change

henryoforeh-dev (approved):

docs/feature_flags.md
```diff
@@ -88,9 +84,9 @@ const mapDispatchToProps = (dispatch) =>
     dispatch
   );
 
-export default compose(
-  connect(mapStateToProps, mapDispatchToProps),
-)(withRouter(FeatureFlagGuard(Feature.TSF_V2)(DamsPage)));
+export default compose(connect(mapStateToProps, mapDispatchToProps))(
```
matbusby-fw, line 87: Good catch on a needed documentation update!

services/core-web/common/components/Step.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from "react";
+
+interface StepProps {
+  disabled?: boolean;
+  key: string;
+}
+
+const Step: FC<StepProps> = (props) => {
```
matbusby-fw, line 8: Yay TS!

services/core-web/common/components/SteppedForm.js
matbusby-fw: Weird!  Well we can convert it to TS eventually and use an interface.

matbusby-fw (approved):
🚀 🐫

PR #2714, opened 2023-10-11
[MDS-5353] -Refactor Explosives Permit Form for better readability and functionality

matbusby-fw:
## Objective 

Was having an odd issue where when entering any value into a text input within a `<FieldArray/>` the form would trigger a re-render and lose focus.

This was causing the collapse panels on the esup form to close after every character entered.

It seems to be a weird interaction with Redux-Form (Damn you Redux!).

The solution that worked in the end was to rename the form being used (This doesn't make a lot of sense to me since the original form shouldn't have cause any problems since it wasn't being actively used in this modal, but it worked)

Cleaned up a couple of other things along the way as well.

[MDS-5353](https://bcmines.atlassian.net/browse/MDS-5353)


services/core-web/src/components/Forms/ExplosivesPermit/MagazineFormNew.tsx
```diff
@@ -233,11 +216,16 @@ export const MagazineForm = (props) => {
         <Row gutter={48}>
           <Col md={24}>
             {fields.map((field, index) => (
-              <Collapse key={index} className="magazine-collapse margin-large--bottom">
+              <Collapse
+                defaultActiveKey={activeKeys}
+                key={index}
```
taraepp, line 221: I am surprised that it let you use the index as the key without an eslint disable. Nothing wonky happens when you try to delete magazines[0] and magazines[1] is expanded or anything?

matbusby-fw: That's actually how it always was.  The problem is, the array being passed is created by redux-form fieldArray.  There is supposed to be a unique id passed by that, but trying it it didn't exist. 😠 

That did make me notice that adding the onChange caused the delete button to be unclickable.  Just added a stopPropagation to fix that.

matbusby-fw: But yeah, all the functions do work 😕

taraepp: That's wacky, I'm used to seeing // eslint-disable-next-line react/no-array-index-key where there's key={index} (why in some places but not in others? I was assuming they were all from --no-verify pushes)

Good that it works, I guess?

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
```diff
@@ -413,20 +413,12 @@ const mapStateToProps = (state) => ({
...
-  );
-
 export default compose(
-  connect(mapStateToProps, mapDispatchToProps),
+  connect(mapStateToProps, null),
   reduxForm({
-    form: FORM.EXPLOSIVES_PERMIT,
+    form: FORM.EXPLOSIVES_PERMIT_NEW,
     touchOnBlur: true,
     validate: validateBusinessRules,
```
taraepp: is validateBusinessRules defined? I don't see it anywhere

matbusby-fw: Yeah, it's within this same form at line 42

services/core-web/src/components/Forms/ExplosivesPermit/MagazineFormNew.tsx
matbusby-fw: TS doesn't complain about it here.  Not sure why.  Agreed that it's certainly not a pattern I'd prefer, but I struggled to find a decent alternative.

I actually did find unique id property for the fields, but whenever it triggered this re-render it would change that value as well.

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
taraepp: Ahhhhh so this must be based off an old version of this form. I had taken out that function before, because it doesn't work. It can test that there's a value, same as required, but it can't tell that it's valid, and so it doesn't catch anything that required doesn't.

matbusby-fw: Oh cool.  Didn't mess with that part at all.  I'll get rid of it.

taraepp: I am also not keen on this file removing type definitions and prop defaults. I just don't know why this looks like an even older version of the form than the file it's replacing so I'd probably want to do a diff with the other version for other changes.

taraepp (approved):

simensma-fresh (approved):

PR #2715, opened 2023-10-11
[MDS-5456] submission documents not filtering out deleted_ind == true records

asinn134:
## Objective 

[MDS-5456](https://bcmines.atlassian.net/browse/MDS-5456)

- Was found in test environment when deleting a now application document, a record gets left behind in the front end table with import status "In progress". It looks like this is because submission_documents property array in the get request to retrieve a now application still includes records that have `deleted_ind == True` from the now_submissions.document table


services/core-api/app/api/now_applications/resources/now_application_resource.py
```diff
@@ -49,6 +49,11 @@ def get(self, application_guid):
             application = transmogrify_now(now_application_identity, include_contacts=original)
             application.imported_to_core = False
 
+        for doc in application.submission_documents:
```
taraepp: Do we need the deleted documents showing up anywhere yet? Could also go to now_application.py model and modify submission_documents relationship and put deleted_ind=False right into the join.

asinn134: Yah that was my first though too, I tried setting it up in the now_application.py model by modifying the submission_documents relationship, but for some reason it still won't filter out the deleted_ind == True

 ```
   submission_documents = db.relationship(
        'Document',
        lazy='selectin',
        secondary=
        'join(NOWApplicationIdentity, Document, foreign(NOWApplicationIdentity.messageid)==remote(Document.messageid))',
        primaryjoin=
        'and_(NOWApplication.now_application_id==NOWApplicationIdentity.now_application_id, foreign(NOWApplicationIdentity.messageid)==remote(Document.messageid), remote(Document.deleted_ind)==False)',
        secondaryjoin='foreign(NOWApplicationIdentity.messageid)==remote(Document.messageid)',
        viewonly=True,
        order_by='asc(Document.id)')
```

, which lead me to doing it the way it is setup in the pr.

matbusby-fw: Hmmm, yeah, that seems like the cleanest place to filter these.  Where did you try adding the deleted_ind check?  Might need something in the secondary join?

like `secondaryjoin='and_(foreign(NOWApplicationIdentity.messageid)==remote(Document.messageid), remote(Document.deleted_ind)==False)'`

asinn134: The issue did end up being with the secondaryjoin, when I initially tested I added the deleted_ind check in the secondary join, but I didn't include `and_` which lead to it not working, nice catch.

taraepp (approved):

henryoforeh-dev (approved):

PR #2718, opened 2023-10-12
[MDS-5515] SchemaSpy updates

taraepp:
## Objective 
- restart the schemaspy deployment in openshift so that new pods are created with updated info
- have it be automatic so we don't have to do it manually
   - tied it into core-api deploys 
- it's not identical (doesn't run flyway first), but I setup a temporary workflow to run on PR to make sure it can run (will delete once verified) <-- DELETED, verified that reusable workflow is called successfully
- attempted to fix the run-if-failed in the core-api deployment workflows. Based on SO answer found here https://stackoverflow.com/questions/71430668/how-to-run-a-github-actions-job-on-workflow-failure by Yann Thomas-Gérard
   - if one of those executes sometime for us we can put it in other places 

[MDS-5515](https://bcmines.atlassian.net/browse/MDS-5515)

### DIFF
--- a/.github/workflows/core-api.build.deploy.dev.yaml
+++ b/.github/workflows/core-api.build.deploy.dev.yaml
@@ -52,6 +52,9 @@ jobs:
       - name: Push
         run: |
           docker push --all-tags ${{ secrets.CLUSTER_REGISTRY }}/${{ secrets.NS_TOOLS }}/${{ env.MIG }}
+      - name: SchemaSpy Rollout
+        uses: ./.github/workflows/schemaspy-rollout-dev.yaml
+        
 
   trigger-gitops:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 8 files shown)


.github/workflows/core-api.build.deploy.dev.yaml
```diff
@@ -77,8 +82,8 @@ jobs:
 
   run-if-failed:
     runs-on: ubuntu-20.04
-    needs: [trigger-gitops]
-    if: always() && (needs.trigger-gitops.result == 'failure')
+    needs: [build-backend, build-flyway, trigger-gitops]
```
taraepp, line 86: specifically _didn't_ put schemaspy-rollout in here because this notification will say that _core-api_ didn't deploy, and schemaspy has its own notification anyway

.github/workflows/schemaspy-rollout-dev.yaml
```diff
@@ -0,0 +1,12 @@
+name: SchemaSpy - DEV restart rollout
+
+on:
+  workflow_call:
+  workflow_dispatch:
```
taraepp, line 5: Wanted to be able to use workflow_call to have it be triggered from another workflow, and workflow_dispatch, so we can call it manually. Only needed separate files for each to allow manual calls.

matbusby-fw (approved):
Nice yaml-ing

simensma-fresh (approved):

PR #2724, opened 2023-10-13
[MDS-5550] Bug: Actions in menu don't do anything (#2723)

taraepp:
## Objective 
- make the actions do actions again
- little mix up here between our ActionMenu and the antd Dropdown menu and the ITableAction.
- ITableAction was made by us for the action menu and can't be passed directly to the Dropdown without processing it first
- extracted and exported the function that did the processing
- put it in the CoreTableCommonColumns too because why not (and removed a parameter that wasn't used anywhere)

[MDS-5550](https://bcmines.atlassian.net/browse/MDS-5550)

### DIFF
--- a/services/core-web/src/components/common/ActionMenu.tsx
+++ b/services/core-web/src/components/common/ActionMenu.tsx
@@ -8,8 +8,9 @@ interface ActionMenuProps {
   actionItems: ITableAction[];
   category: string;
 }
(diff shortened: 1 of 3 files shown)


matbusby-fw (approved):
Nice!  I like it!

asinn134 (approved):

PR #2728, opened 2023-10-17
[MDS-5502] Added support for React components in common package

simensma-fresh:
## Objective 

[MDS-5502](https://bcmines.atlassian.net/browse/MDS-5502)

**React components in common package**
Modified Webpack / typescript config to remove the need for building / watching the common package separately, and hot reloading of common component will work properly. This separate build step caused all sort of issues with getting React components working, and makes the dev experience a bit better for working with it.
Instead, we include the common package in the Webpack build process for core and MS.
- Updated webpack config to include common package in MS / core builds
- Removed separate common package build step from Dockerfiles and test runners
- Updated TSConfig files to recognize common package at `@mds/common`
- Moved the `DeleteDocumentModal` and Feature flag utilities into the common package as a test

**Type errors**
To the fun part! The above uncovered ~150 Type errors of various sorts, as it appears that the previous setup suppressed any errors related to the common package - including incorrectly referenced files, missing properties, and mismatch in types. The PR includes fixes for all of them to get the builds running with the new setup.


henryoforeh-dev (dismissed):

services/common/src/interfaces/activity/notificationDocumentMetadata.interface.ts
```diff
@@ -3,7 +3,7 @@ import {
   INotificationDocumentPermit,
   INotificationDocumentProject,
   INotificationDocumentMineTailingsStorageFacility,
-} from "@/index";
```
matbusby-fw, line 6: Oh good!  I actually just came across a few of these in my current work and thought it didn't look quite right (although wasn't showing any errors)

services/common/src/interfaces/index.ts
```diff
@@ -27,3 +28,9 @@ export * from "./user";
 export * from "./projects";
 export * from "./documentZipTask.interface";
 export * from "./documentZipProgress.interface";
+export * from "./common";
+export * from "./noticeOfWorkApplication.interface";
+export * from "./now";
+export * from "./noticeOfWork.interface";
+export * from "./noticeOfWorkDraftPermit.interface";
+export * from "./search/searchResult.interface";
```
matbusby-fw, line 36: Probably not something that needs to be dealt with here, but I just read an article discussing the performance implications of using many nested barrell files like this and how it ends up importing everything any time any of them are used.  Probably not an issue with interfaces and the like though.  With our previous setup being built we kind of needed to rely on this, but with new components and such are we able to use more direct imports?

services/common/src/interfaces/activity/notificationDocumentMetadata.interface.ts
simensma-fresh: In the previous setup `@/index` worked fine, but now it'll have to be `@mds/common` going forwards as to not conflict with those prefixes for files in minespace and core

services/common/src/interfaces/index.ts
simensma-fresh: Yeah, going forwards we should use direct imports, as you're correct that it's no longer needed. Added these as it solved a good chunk of errors that I didn't have to deal with 😅

services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -80,6 +57,7 @@ export const DocumentTable: FC<DocumentTableProps> = ({
   const [rowSelection, setRowSelection] = useState([]);
   const [isCompressionModal, setCompressionModal] = useState(false);
   const [isCompressionInProgress, setCompressionInProgress] = useState(false);
+  const [documentTypeCode, setDocumentTypeCode] = useState("");
```
taraepp: Are we calling setDocumentTypeCode anywhere? I don't think this will ever be anything other than "". I think it got half-implemented, and then I mentioned that the initial value was going to be undefined a lot of the time, and it didn't get fully removed.

services/filesystem-provider/.env-example
```diff
@@ -1,7 +1,7 @@
 OBJECT_STORE_HOST=nrs.objectstore.gov.bc.ca
-OBJECT_STORE_ACCESS_KEY_ID=
+OBJECT_STORE_ACCESS_KEY_ID=nr-core-tst
```
matbusby-fw, line 2: Nice!

services/core-web/src/components/mine/ExplosivesPermit/MineExplosivesPermitTable.tsx
```diff
@@ -41,6 +42,12 @@ interface MineExplosivesPermitTableProps {
   handleOpenViewExplosivesPermitModal: (event, record: IExplosivesPermit) => void;
 }
 
+type MineExplosivesColumnType = IExplosivesPermit & {
```
taraepp: I think this is the record type rather than the ColumnType? Little nitpicky, but it looks really odd to me to have `record: MineExplosivesColumnType` later on, surely we're dealing with Explosives Permits rather than listing out ColumnTypes in the table.

matbusby-fw (dismissed):
Amazing!  I'm very excited about this.  I regret not making the change to how the common package is used before doing the Typescript conversion.  Would have saved some big headaches for you!  Sorry about that!

services/core-web/src/components/mine/ExplosivesPermit/MineExplosivesPermitTable.tsx
```diff
@@ -471,7 +477,7 @@ const MineExplosivesPermitTable: FC<RouteComponentProps & MineExplosivesPermitTa
     <CoreTable
       condition={isLoaded}
       dataSource={transformRowData(data)}
-      rowKey={(record: IExplosivesPermit) => record.explosives_permit_guid}
+      rowKey={(record: MineExplosivesColumnType) => record.explosives_permit_guid}
```
taraepp: I see that we're actually setting the key up above- not necessary, but you could totally remove the rowKey attribute if you wanted to.

matbusby-fw (dismissed):
👯 👯 👯 👯

henryoforeh-dev (dismissed):

matbusby-fw (approved):
👯 👯 👯 👯  🐟  👯 👯 👯 👯

taraepp (approved):

PR #2729, opened 2023-10-17
[MDS-5505] Add cypress test for major projects. Part 2

henryoforeh-dev:
## Objective 

[MDS-5505](https://bcmines.atlassian.net/browse/MDS-5505)

- Implementation of Cypress test to upload/download documents.
- I added **data-cy** attribute to some of the elements them easier to test with cypress.


services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,86 @@
+describe("Major Projects", () => {
+    beforeEach(() => {
+        cy.login();
+        cy.wait(5000);
```
taraepp: I'm wondering if it's possible to set it to `wait for up to 5 seconds, until the Major Projects button appears`, instead of `wait for 5 seconds, and then look for the Major Projects button`.

Found some suggestions in their documentation: https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

matbusby-fw: You can set the universal timeout length in the the cypress config, or you can always add a timeout to any `cy.get()` as well.  ie. `cy.get('table[style="table-layout: auto;"]', {timeout: 5000})`

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+        // Get the button element by its text content
+        cy.contains('button', 'Major Projects')
+            // Check that it is visible and enabled
+            .should('be.visible')
+            .and('not.be.disabled')
+            // Click on it
+            .click({ force: true });
+
+        // Find the table and wait for it to be visible
+        cy.get('table[style="table-layout: auto;"]').should('be.visible').find('tbody > tr').eq(1) // get the second row
```
taraepp: Is there another selector we can use besides `table[style="table-layout: auto;"]`? I worry that selectors like that are quite brittle. I am okay with adding in IDs, data attributes for cypress to find things, but styles and classes defined by libraries are dicey.

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    it('should download a document successfully', () => {
+
+        // Find the row with 'dummy.pdf', scroll it into view, and hover over its 'Actions' button
+        cy.contains('tr', 'dummy.pdf')
+            .scrollIntoView()
+            .within(() => {
+                // Hover over the 'Actions' button within the row
+                cy.get('.ant-btn.ant-btn-default.ant-dropdown-trigger.permit-table-button')
```
taraepp: While we should definitely change permit-table-button className at some point to be actions-dropdown-button or something, I do think the inclusion of all the ant classes just makes it more likely to break.

services/core-web/cypress/e2e/majorprojects.cy.ts
matbusby-fw: Totally agree.  I think adding ids for the sake of testing will make them both more durable and more readable.

matbusby-fw: Not to mention making it easier to write the tests

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: I have incorporated the data-cy attribute into the Action menu to streamline its testing process with Cypress.

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: I have incorporated the data-cy attribute into the table to streamline its testing process with Cypress.

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: Updated as expected.

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,79 @@
...
+        cy.get("[data-cy=major-projects-table-open-button]", { timeout: 5000 })
+            .eq(1)
+            .find("button")
+            .click();
+
+
+        cy.get('a[data-cy="project-description-view-link"]', { timeout: 5000 }).click();
+
+        // Wait for the edit button to be visible and click it
+        cy.get('#project-summary-submit').then(($button) => {
```
matbusby-fw, line 18: Yeah, that feels a lot more readable!

matbusby-fw (dismissed):
👯‍♂️

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,79 @@
...
+            $button[0].click();
+        });
+    });
+
+    it("should upload a document successfully", () => {
+        // Access the file input element and attach a file from the fixtures directory.
+        cy.get('input[type="file"]').scrollIntoView().attachFile('dummy.pdf');
+
+        // This long time is required because it takes quite a while for the file (11kb) to upload.
+        cy.wait(25000);
```
taraepp: I think in the documentation it recommended putting an alias on the request and then you can wait for it to complete

services/core-web/src/components/homepage/HomeTopLinks.tsx
```diff
@@ -45,30 +45,35 @@ const HomeTopLinks = () => {
       route: routes.NOTICE_OF_WORK_APPLICATIONS.route,
       key: "now",
       icon: () => <FontAwesomeIcon icon={faTrianglePersonDigging} />,
+      datacy: "home-link-button-now",
```
taraepp: hmm, considering that these are all in the format of "home-link-button"-${key} I'm thinking we can simplify this by just having the logic performed directly to calculate that instead of passing as a prop.

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: Sure, that is ideal. I tried it and noticed random failures with no resolution.

services/core-web/src/components/homepage/HomeTopLinks.tsx
henryoforeh-dev: Updated. I removed datacy attributed and changed the replaced the key attribute with id. The reason for doing this is because React treats **key** as a special reserved property.

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: Updated it to now intercept, alias and wait. I still set the wait time to 25 seconds because of the time it takes to complete 3 service calls downstream.

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+        // Wait for file to download before continuing
+        cy.wait(5000);
+
+        /**
+         * Clean up by deleting file after downloading. This is to ensure that the 
+         * upload file runs multiple times without any issue
+         */
+
+        // Click the delete button in the dropdown
+        cy.contains('button', 'Delete')
```
taraepp: can we put a 5 second timeout here instead of the wait above?

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: Had to change the code again because it was passing 1or 2 times out of 8 runs or more on my machine but failing continuously in the pipeline.

services/core-web/cypress/e2e/majorprojects.cy.ts
henryoforeh-dev: Updated as expected.

taraepp (approved):
👍

matbusby-fw (approved):
🎺

PR #2731, opened 2023-10-18
[MDS-5551] ESUP open decision modal bug- form_spec of undefined

taraepp:
## Objective 
- the bug is because the `fetch` method populates the `documentContextTemplate` and this reduxy bit isn't quite synchronous so it wouldn't be populated in time. It would work on the second try but not the first.
   - I could not find a way to resolve the bug while maintaining this pattern, so I didn't, grabbing the data from the .then was my best solution
- in ApplicationTab.js- the `fetchNoticeOfWorkApplicationContextTemplate` in `handleGenerateDocument` actually calls/called the same reducer in a similar structure, so I was expecting it to fail as well, **but it didn't**
   - I don't know 100% why but I suspect that it's because ApplicationTab defines & uses the function directly whereas ExplosivesPermit passes it as a prop to the MineExplosivesPermitTable.
- I also noticed that our Actions menu was straying just a tiny bit from mockups (I used the original actions menu ticket as the source of truth: https://bcmines.atlassian.net/browse/MDS-5356)
  - moved the icon from the left to the right
  - consistent icon usage in feature-flagged (new) version of menu
  - put the delete in there too
  - made a wrapper function for the delete confirm modal
- checking user permissions is annoying, particularly when you can't just wrap the component, so I also added a selector function that does it more directly without having to introduce mapStateToProps.
[MDS-5551](https://bcmines.atlassian.net/browse/MDS-5551)

The delete confirmation modal and the current state of the action menu:
image

image


matbusby-fw (approved):
Looks good!  Nice clean up of the actions menu!

isuru-aot (approved):

PR #2732, opened 2023-10-19
[MDS-5553] (and 5542) Request a connection invitation for a permitee

Jsyro:
## Objective 

[MDS-5553](https://bcmines.atlassian.net/browse/MDS-5553)
[MDS-5542](https://bcmines.atlassian.net/browse/MDS-5542)

A Minespace proponent would like to establish a digital wallet connection between MDS (traction) and the mining companies digital wallet. 

This PR introduces functionality that allows a Minespace proponent to generate out-of-band with no payloads for the purposes of connecting two hyperledger aries agents. BC Digital Trust Services hosts `Traction`, a multi-tenanted service that is hosting the wallet for the 'Chief Permitting Officer of Mines' that needs to connection with the wallets of mining companies for the purpose of issuing Mines Act Permit as a verifiable credential. 

Remaining TODOs, I am not an expert in React, so please comment if i have done anything unexpected.

I need help on these: 
- [x] Fix bug where actionCreator is called continuously as long as modal is open, it should only activate on button
- [x] Loading animation while waiting for invitation response
- [x] Feature Flag the frontend

These I can do, but wanted to make the PR early:
- [x] Add restriction that 'Connection Info' button is only available for Major Mines
- [x] Add restriction that permit is open
- [x] If there is an active connection, indicate that and do not allow a new invitation to be created
- [x] "Issue Digital Credential" is only visible if there is an active connection,


Jsyro:
Some screenshots. 

image
image

Jsyro:
From discord message: 

Can help with some of this if you need:

- when creating a new javascript file, we are trying to create them as either .ts or .tsx files (.tsx for components, like CreateInvitation, or .ts for reducers, etc)
- I think we might be able to get rid of some of those <br> elements by using Typography component, I'll take a look once I'm out of meeting to see if I can find a better example to follow
- as far as FE unit tests: my really cheesy way to get these updated is to alter the package.json in services/core-web and minespace, add a -u to "test" under scripts to update the snapshots, then run yarn test (test:updateAll just doesn't work very well for me). Once it's finished running, all your snapshots will be updated. Just check them to make sure nothing weird happened before committing.

taraepp:
docs/feature_flags.md has really good instructions for using feature flags in the FE.
There's a flag already in flagsmith, `verifiable_credentials`, mostly need to add it to the Feature enum in featureFlag.ts- doc has examples of how to use and import feature flag utilities in different files.

Jsyro:
I think all the identified items are complete, Testing locally is tough due to webhook call backs from DEV traction, so there may be a small follow up. 

Unit tests cannot load dependencies, so i think this is good to go.

services/core-api/app/api/parties/party/models/party.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime
+from flask import current_app
```
taraepp: is this left in from doing debug logs?

taraepp (dismissed):
possibly an unused import in here in the BE but LGTM

services/core-web/common/reducers/verifiableCredentialReducer.js
henryoforeh-dev, line 1: Are we creating new reducer files as javascript ?

Jsyro: I updated to TS using tailingsReducer.ts as a template

henryoforeh-dev (approved):

PR #2755, opened 2023-10-26
[MDS-5509] Fixing Error: object of type NoneType has no len() on letter_date

isuru-aot:
## Objective 

[MDS-5509](https://bcmines.atlassian.net/browse/MDS-5509)

letter related arguments (`letter_date` and `letter_body`) are applicable in application flow only. Hence assigning default values (which was used earlier) to resolve the issue.

Related resources:
https://github.com/bcgov/mds/pull/2679/files


henryoforeh-dev (approved):

asinn134 (approved):
nice!

PR #2756, opened 2023-10-26
[FIX] - row level auth and invitations hybrid property

Jsyro:
connection_state not connection_status.

disable row level auth, no user context.


matbusby-fw (approved):
👯

asinn134 (approved):

PR #2760, opened 2023-10-27
[FIX] - Webhook attribute and set/list handling

Jsyro:
It looks like the sqlalchemy relationship was joining to all records, not limited by party guid, i'm hoping this fixes it.


taraepp (approved):

matbusby-fw (approved):
👍🏻

henryoforeh-dev (approved):

PR #2774, opened 2023-11-02
[MDS-5552]replacing now manage doc and admin with zip download

asinn134:
## Objective 

[MDS-5552](https://bcmines.atlassian.net/browse/MDS-5552)

- Replaces the existing download of NoW documents and Permit Package documents in NoW Manage Documents tab with download zip, similar to Major projects.

- Replaces the existing download of Permit Package documents in NoW Administrative tab with download zip, similar to Major projects.

image

image

### DIFF
--- a/services/core-web/src/components/common/DocumentTable.tsx
+++ b/services/core-web/src/components/common/DocumentTable.tsx
@@ -366,6 +366,7 @@ export const DocumentTable: FC<DocumentTableProps> = ({
         setCompressionModalVisible={setCompressionModal}
         isCompressionModalVisible={isCompressionModal}
         compressionInProgress={setCompressionInProgress}
+        showArchiveDownloadWarning={true}
       />
       {renderBulkActions()}
       {<CoreTable {...coreTableProps} />}
--- a/services/core-web/src/components/common/DocumentTable.tsx
+++ b/services/core-web/src/components/common/DocumentTable.tsx
@@ -366,7 +366,7 @@ export const DocumentTable: FC<DocumentTableProps> = ({
         setCompressionModalVisible={setCompressionModal}
         isCompressionModalVisible={isCompressionModal}
         compressionInProgress={setCompressionInProgress}
-        showArchiveDownloadWarning={true}
+        showDownloadWarning={showVersionHistory || canArchiveDocuments}
       />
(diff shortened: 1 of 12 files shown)


services/core-web/src/components/common/DocumentTable.tsx
```diff
@@ -366,6 +366,7 @@ export const DocumentTable: FC<DocumentTableProps> = ({
         setCompressionModalVisible={setCompressionModal}
         isCompressionModalVisible={isCompressionModal}
         compressionInProgress={setCompressionInProgress}
+        showArchiveDownloadWarning={true}
```
taraepp: would true be a reasonable default for this prop?

matbusby-fw (dismissed):
Nice.  Love the tweaks for reusability!

henryoforeh-dev (dismissed):

services/core-web/src/components/common/DocumentTable.tsx
asinn134: @taraepp that is a good point. I initially thought to set this prop as a way to temporarily disable the archive warning modal that pops up because NoW section hasn't had archive and versioning setup yet. At the same time, it is possible for other sections not just Major Projects or NoW to use document compression within the DocumentTable since DocumentTable is used in alot of different areas that also don't have archive or versioning files.

Would it make more sense to use DocumentTable's props `showVersionHistory` and `canArchiveDocuments`. So something like `showDownloadWarning={showVersionHistory || canArchiveDocuments}`

asinn134: Actually scratch what I mentioned above, the user is not able to select any archived documents or older versions of documents when they bulk download. I think just using DocumentTable's props `showVersionHistory `and `canArchiveDocuments` can work. So something like `showDownloadWarning={showVersionHistory || canArchiveDocuments}`

asinn134: I adjusted the DocumentTable's showDownloadWarning default value to be based on `showVersionHistory || canArchiveDocuments` in the latest commit.

henryoforeh-dev (approved):

matbusby-fw (approved):
👍🏻

PR #2789, opened 2023-11-07
 [MDS-5590] Adjusting esup creation flow for permit and application sections

asinn134:
## Objective 

[MDS-5590](https://bcmines.atlassian.net/browse/MDS-5590)

- Change so when creating a new permit, either in applications tab or permits tab, the modal opens to create a new permit
- Change so when creating a historic permit, either in applications tab or permits tab, the modal opens to create a historic permit record


henryoforeh-dev (approved):

matbusby-fw (approved):
Approvinated

PR #2791, opened 2023-11-08
[MDS-5601] - add security for traction webhook endpoint

Jsyro:
Traction is already providing them, but we will need new env vars.

## Objective 

[MDS-5601](https://bcmines.atlassian.net/browse/MDS-5601)

Only traction should be allowed to call this endpoint.

`TRACTION_WEBHOOK_X_API_KEY` is now a new environment variables for all environments.


henryoforeh-dev (approved):

taraepp (approved):

Jsyro:
Will not be merging until we have added the appropriate environment variables to each environment.

henryoforeh-dev (approved):

PR #2796, opened 2023-11-09
[MDS-5618] Shiny New Form

taraepp:
## Objective 
- make it do view and edit easily in the same component
- make it prettier
- no more legacy antd

[MDS-5618](https://bcmines.atlassian.net/browse/MDS-5618)

Decided not to do:
[MDS-5608](https://bcmines.atlassian.net/browse/MDS-5608)

hackathon-demo

### DIFF
--- a/services/common/src/components/FormWrapper.tsx
+++ b/services/common/src/components/FormWrapper.tsx
@@ -1,0 +1,93 @@
+import { Button, Row, Typography } from "antd";
+import React from "react";
+
+export const { Provider: FormProvider, Consumer: FormConsumer } = React.createContext({
+  view: false,
+});
+
+interface FormWrapperProps {
+  name: string;
+  onSubmit: any;
+  children: any;
+  showFooter?: boolean;
+  okText?: string;
+  onOk?: any;
+  cancelText?: string;
+  onCancel?: any;
+  loading?: boolean;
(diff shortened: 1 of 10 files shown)


services/common/src/components/FormWrapper.tsx
```diff
@@ -0,0 +1,93 @@
+import { Button, Row, Typography } from "antd";
+import React from "react";
+
+export const { Provider: FormProvider, Consumer: FormConsumer } = React.createContext({
+  view: false,
+});
```
taraepp, line 6: the Provider/Consumer is my favourite part! Before I put it in it was pretty annoying to have to add a parameter to each input.

services/core-web/src/components/Forms/ExplosivesPermit/MagazineFormNew.tsx
```diff
@@ -224,29 +235,31 @@ class MagazineForm extends Component<FormProps> {
           <Col md={24}>
             {fields.map((field, index) => (
               <Collapse
-                key={field.id}
+                key={field}
```
taraepp, line 238: totally unrelated to what I was doing: I kept getting duplicate key warnings in my console, and it turns out that **field does not have an id property**. So we should fix that. Field is a string.

PR #2797, opened 2023-11-09
[MDS-5562] add required data for revocation

Jsyro:
## Objective 

[MDS-5562](https://bcmines.atlassian.net/browse/MDS-5562)

This will not enable revocation, but it will ensure we have the right data to revoke any credential at a later date when we do have the functionality. Revocation could be done manually in the traction wallet, but this will be needed if we want to enable controls in minespace or MDS, and will work even if the credential exchange record in traction is deleted.

This may require a follow up, but this is the best way to test the webhook exchange between MDS and Traction.

### DIFF
--- a/migrations/sql/V2023.11.09.14.07__add_rev_reg_vc_columns.sql
+++ b/migrations/sql/V2023.11.09.14.07__add_rev_reg_vc_columns.sql
@@ -1,0 +1,5 @@
+ALTER TABLE party_verifiable_credential_mines_act_permit
+    ADD COLUMN IF NOT EXISTS rev_reg_id character varying(150); 
+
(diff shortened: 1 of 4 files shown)


taraepp (approved):
I think some comments on those db columns (rev_reg_id particularly is very cryptic) would be good, but that could come in a subsequent PR

henryoforeh-dev (approved):

Jsyro:
> I think some comments on those db columns (rev_reg_id particularly is very cryptic) would be good, but that could come in a subsequent PR

True, those are the fields that traction will return. Revocation Registry ID and Credential Revocation ID. but i should add some more context .

PR #2800, opened 2023-11-10
[MDS-5605] - migrate common folders into the common package

matbusby-fw:
## Objective

In order to drastically reduce the duplication of code, have refactored all shared redux resources to reside in the common package.

The vast majority of changes are just to how things are imported.  I'll comment where the bigger changes take place.

1. Created a global variable in the common package `REQUEST_HEADER` and update the webpack config in both Core and Minespace to inject this header as a module since this looks different on each system.  
2. Separated out the root reducer(s) into a shared reducer and unique reducers for each frontend
3. Updated jest config to also have the REQUEST_HEADER available
4. Moved constants to common package

[MDS-5605](https://bcmines.atlassian.net/browse/MDS-5605)


services/common/src/@Types/global.d.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
```
matbusby-fw, line 1: This type is to make the REQUEST_HEADER available globally in the common package (it's mapped to each FE's REQUEST_HEADER through their individual webpack configs.

services/common/src/constants/API.js
```diff
@@ -27,7 +27,9 @@ export const MINE_TSF = (mine_guid, mine_tailings_storage_facility_guid) =>
 export const DISTURBANCE_CODES = "/mines/disturbance-codes";
 export const COMMODITY_CODES = "/mines/commodity-codes";
 export const EDIT_TSF_REPORT = "";
-export const MINE_DOCUMENTS = (mine_guid) => `/mines/${mine_guid}/documents`;
+export const MINE_DOCUMENTS = (mine_guid, params = {}) =>
```
matbusby-fw, line 30: Checked the diff on the different api files and combined them

services/common/src/redux/rootState.ts
```diff
@@ -0,0 +1,29 @@
+import { sharedReducer } from "@mds/common/redux/reducers/rootReducerShared";
```
matbusby-fw, line 1: This file provides the typing for the redux stuff in the common package

services/common/src/redux/actionCreators/documentActionCreator.ts
```diff
@@ -1,12 +1,15 @@
...
 import { hideLoading, showLoading } from "react-redux-loading-bar";
-import CustomAxios from "@common/customAxios";
-import { createRequestHeader } from "@common/utils/RequestHeaders";
+
 import { notification } from "antd";
 import { ENVIRONMENT, IMineDocumentVersion } from "@mds/common";
+import { request, success, error } from "@mds/common/redux/actions/genericActions";
+import CustomAxios from "@mds/common/redux/customAxios";
+
+const createRequestHeader = REQUEST_HEADER.createRequestHeader;
```
matbusby-fw, line 12: This is the global `REQUEST_HEADER` being utilized (no import)

services/common/src/redux/reducers/rootReducerShared.ts
```diff
@@ -0,0 +1,67 @@
...
+  securitiesReducer,
+  staticContentReducer,
+  tailingsReducer,
+  userReducer,
+  varianceReducer,
+  verifiableCredentialReducer,
+  workInformationReducer,
+} from "../reducers";
+
+export const sharedReducer = {
```
matbusby-fw, line 35: All shared reducers now reside here in the common package

services/core-web/src/reducers/rootReducer.ts
```diff
@@ -0,0 +1,83 @@
+import { combineReducers } from "redux";
+import { createReducer } from "@common/utils/helpers";
+import mineAlertReducer from "./mineAlertReducer";
+import networkReducer from "./networkReducer";
+import { sharedReducer } from "@mds/common/redux/reducers/rootReducerShared";
+import * as reducerTypes from "@mds/common/constants/reducerTypes";
+import documentReducer from "./documentReducer";
+
+const coreReducer = {
```
matbusby-fw, line 9: Reducers that only existed in Core combined with the shared ones from the common package

services/minespace-web/src/reducers/rootReducer.ts
```diff
@@ -0,0 +1,18 @@
+import { combineReducers } from "redux";
+import authenticationReducer from "@/reducers/authenticationReducer";
+import userMineReducer from "@/reducers/userMineReducer";
+import networkReducer from "./networkReducer";
+import * as reducerTypes from "../constants/reducerTypes";
+import { createReducer } from "@common/utils/helpers";
+import { sharedReducer } from "@mds/common/redux/reducers/rootReducerShared";
+
+const minespaceReducer = {
+  ...sharedReducer,
```
matbusby-fw, line 10: Reducers unique to minespace combined with the shared reducers

services/common/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   // Change this to match your project
-  "include": ["src/**/*", "webpack.config.ts", "types"],
+  "include": ["src/**/*", "webpack.config.ts", "types", "src/@Types/global.d.ts"],
```
matbusby-fw, line 3: Including the new global type for Request Header

services/core-web/webpack.config.ts
```diff
@@ -73,6 +73,9 @@ const commonConfig = merge([
       new HtmlWebpackPlugin({
         template: PATHS.template,
       }),
+      new webpack.ProvidePlugin({
```
matbusby-fw, line 76: This exists for Core and Minespace.  It inserts the REQUEST_HEADER specific to each front end when utilizing it in the common workspace

taraepp (dismissed):
🚀🚀🚀🚀🚀

services/minespace-web/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     "esModuleInterop": true /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables 'allowSyntheticDefaultImports' for type compatibility. */,
     "forceConsistentCasingInFileNames": true /* Ensure that casing is correct in imports. */,
     "strict": false /* Enable all strict type-checking options. */,
+    "strictNullChecks": false,
```
taraepp: hmm, we do have significant bugs that come up because of missed null checks- was this to prevent TS errors about all the pre-existing cases? I think if/when we merge this we'll want to make a ticket to remove the line and fix all the errors.

henryoforeh-dev (dismissed):
Huge effort. Nice work!

services/minespace-web/tsconfig.json
matbusby-fw: Actually, good catch.  That had been added in at one point in the process to test something and I forgot to remove it.  I'll take it out.

henryoforeh-dev (approved):

taraepp (approved):

PR #2806, opened 2023-11-16
[MDS-5437] Implement KeyCloak logging

henryoforeh-dev:
## Objective 

[MDS-5437](https://bcmines.atlassian.net/browse/MDS-5437)

I have configured the **werkzeug** logger's logging level to **CRITICAL** and propagate to **False** in order to prevent the propagation of **INFO** logs. The **werkzeug** logger is the default logger responsible for displaying logs when the core-api is running in development, testing, and production environments. To replicate this logger locally, I added a new section to the LOGGING_DICT_CONFIG configuration setup and set the level attribute to **INFO** and the propagate attribute to True. 
I
This fix can only be tested after merging and testing it in the development environment.

### DIFF
--- a/services/core-api/app/config.py
+++ b/services/core-api/app/config.py
@@ -83,14 +83,14 @@ class Config(object):
                 'formatter': 'default',
             }
         },
+        'root': {
+            'level': FLASK_LOGGING_LEVEL,
+            'handlers': ['file', 'console']
+        },
         'loggers': {
-            '': {
-                'level': FLASK_LOGGING_LEVEL,
-                'handlers': ['file', 'console']
-            },
             'werkzeug': {
-                'level': 'ERROR',  # Set the level to ERROR to disable logging
-                'propagate': False  # Prevent the log messages from being propagated to the root logger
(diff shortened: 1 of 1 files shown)


services/core-api/app/config.py
```diff
@@ -86,6 +86,12 @@ class Config(object):
         'root': {
             'level': FLASK_LOGGING_LEVEL,
             'handlers': ['file', 'console']
+        },
+        'loggers': {
+            'werkzeug': {
+                'level': 'CRITICAL',
```
taraepp: above we have this set using an env variable (FLASK_LOGGING_LEVEL), I think that would be a good idea here, too. If we have a major production bug, then we could turn on more logging to get more info without doing a prod push.

taraepp (approved):

services/core-api/app/config.py
henryoforeh-dev: Updated.

asinn134 (approved):

simensma-fresh (approved):

PR #2808, opened 2023-11-16
[MDS-5606] Typescript conversions within Summary and Variance sections

asinn134:
## Objective 

[MDS-5606](https://bcmines.atlassian.net/browse/MDS-5606)

- Some conversions to Typescript within the Summary and Variance section of the Core codebase


henryoforeh-dev (dismissed):
Nice work!

services/common/src/interfaces/common/match.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface IMatch {
```
taraepp: I would say this is actually not necessary. In MineSummary, for example, you can replace
`const { id } = this.props.match.params`
with
`const { id } = useParams();`
and just get rid of the prop

services/core-web/src/components/mine/Summary/MineSummary.tsx
```diff
@@ -0,0 +1,289 @@
...
+  </Col>
+);
+
+// Since end date is stored at yyyy-mm-dd, comparing current Date() to
+// the the start of the next day ensures appointments ending today are displayed.
+const isActive = (pr) =>
+  (!pr.end_date ||
+    moment(pr.end_date)
+      .add(1, "days")
+      .toDate() > new Date()) &&
```
taraepp: we don't need to convert moment to a date.

moment(date).add(1 days).isBefore() <-- isBefore used without an argument will compare to the equivalent of new Date(). There's a few variations on that function like isAfter if I'm getting my < and > mixed up. Would do something similar below with the Date.parse.

services/core-web/src/components/mine/Summary/MineWorkInformation.tsx
```diff
@@ -0,0 +1,267 @@
...
+        </Col>
+      </Row>
+    </List.Item>
+  );
+
+  const dataSource = showAll
+    ? props.mineWorkInformations
+    : !isEmpty(props.mineWorkInformations)
+    ? [props.mineWorkInformations[0]]
+    : [];
```
taraepp, line 196: now I might be getting picky but did the linter run? I was wondering how this syntax could be real until I noticed that the indentation is wacky. 😂

taraepp (commented):
There's a few little things carried over that I think would be worth fixing but overall looks great!

services/core-web/src/components/mine/Summary/MineWorkInformation.tsx
asinn134: The linter definitely ran here. On my laptop the indentation looks like this

```
  const dataSource = showAll
    ? props.mineWorkInformations
    : !isEmpty(props.mineWorkInformations)
      ? [props.mineWorkInformations[0]]
      : [];
```
but the linter changes it to what we see in the pr.

services/core-web/src/components/mine/Summary/MineSummary.tsx
asinn134: Nice! I made the change it looks a lot cleaner now.

services/common/src/interfaces/common/match.interface.ts
asinn134: Yah looking at this again the IMatch really isn't necessary. It has been removed now and useParams() is used within MineSummary instead.

services/core-web/src/components/mine/Summary/MineWorkInformation.tsx
taraepp: _Gross_. But I accept this explanation. 👍

services/common/src/interfaces/common/match.interface.ts
taraepp: Nice!
I see the tests are giving you trouble- we've got a few examples of mocking this in MineIncident.spec, noticeOfDeparture.spec, and ProjectSummaryPage.spec

asinn134: Thanks for mentioning the examples, that helped a lot

henryoforeh-dev (dismissed):

services/common/src/interfaces/permits/permit.interface.ts
```diff
@@ -23,4 +23,5 @@ export interface IPermit {
   site_properties: IMineType;
   permit_prefix: string;
   mine_guid?: string;
+  amendments?: Partial<IPermitAmendment>[];
```
taraepp: Wait, we have `amendments` and `permit_amendments`.

asinn134: Oh whoops I forgot to remove this after some testing. Thanks for catching this.

taraepp (approved):

henryoforeh-dev (approved):

PR #2810, opened 2023-11-17
[MDS-5473] Fix esup amendment letter

henryoforeh-dev:
## Objective 

[MDS-5473](https://bcmines.atlassian.net/browse/MDS-5473)

- This pull request (PR) focuses on implementing the functionality for issuing ESUP amendment letters and certificates. In the previous PR, I had only implemented the functionality to preview the letter and certificate for ESUP amendments.

- I also added a missing action type constant for verifiable credentials that was causing build errors locally


services/core-api/app/api/mines/explosives_permit_amendment/resources/explosives_permit_amendment.py
```diff
@@ -14,6 +14,18 @@
 from app.api.mines.explosives_permit_amendment.models.explosives_permit_amendment import ExplosivesPermitAmendment
 class ExplosivesPermitAmendmentResource(Resource, UserMixin):
     parser = CustomReqparser()
+    parser.add_argument(
+        'amendment',
+        type=str,
+        store_missing=False,
+        required=False,
```
taraepp: I was wondering what this referred to! Can we add a help="some text to describe what's expected here"? I am not too keen on doing the string formatting on the FE for a value used on the BE, would prefer to send just the amendment number and do the string formatting where it's being used. 
The amendment_with_date though looks like it should be calculated on the BE if we're only using it on that letter template. Is it used anywhere on the FE?

services/common/src/constants/actionTypes.js
```diff
@@ -187,3 +187,6 @@ export const CLEAR_TAILINGS_STORAGE_FACILITY = "CLEAR_TAILINGS_STORAGE_FACILITY"
 
 // Dams
 export const STORE_DAM = "STORE_DAM";
+
+//Verifiable Credentials
+export const STORE_VC_WALLET_CONNECTION_INVITATION = "STORE_VC_WALLET_CONNECTION_INVITATION";
```
matbusby-fw: Good catch.  Added this in my current work as well.  I think that was my bad that this got lost in the shuffle of moving stuff around.

Jsyro: I'm hoping this can be merged and deployed to DEV/TEST before 2pm today. Minespace VC invitations won't show on the screen until this is in.

simensma-fresh (dismissed):

taraepp (approved):
I realize my comment earlier added extra work for a pre-existing issue! 😅 
Thanks for fixing it though! 🎉

isuru-aot (approved):

PR #2819, opened 2023-11-21
[MDS-5437] keycloak access log fix

henryoforeh-dev:
## Objective 

[MDS-5437](https://bcmines.atlassian.net/browse/MDS-5437)

- Improvised a way to display and format access logs similar to the way werkzeug logger works. This was done to be able to add traceid and client to the logs displayed. For some reason this is not possible (getting traceid and client information) with werkzeug logger.
- Set the werkzeug access logger level to CRITICAL.
- Removed the print error message that was getting displayed.

Sample screenshot of how the access logs now look.

image

### DIFF
--- a/services/core-api/app/config.py
+++ b/services/core-api/app/config.py
@@ -59,7 +59,7 @@ class Config(object):
     FLASK_LOGGING_LEVEL = os.environ.get('FLASK_LOGGING_LEVEL',
                                          'INFO')                # ['DEBUG','INFO','WARN','ERROR','CRITICAL']
     WERKZEUG_LOGGING_LEVEL = os.environ.get('WERKZEUG_LOGGING_LEVEL',
-                                         'INFO')  # ['DEBUG','INFO','WARN','ERROR','CRITICAL']
+                                         'CRITICAL')  # ['DEBUG','INFO','WARN','ERROR','CRITICAL']
     DISPLAY_WERKZEUG_LOG = os.environ.get('DISPLAY_WERKZEUG_LOG',
                                             True)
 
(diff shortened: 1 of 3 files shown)


services/core-api/app/config.py
```diff
@@ -25,8 +25,8 @@ def get_key_cloak_client_id():
                     if getJwtManager().audience:
                         return getJwtManager().audience
             except Exception as e:
-                # Handle the exception here (e.g., log it)
-                print(f"An error occurred: {e}")
+                # The exception is caught, but no error message is printed or logged
```
taraepp: I did find that I was able to log both the exception message and the stack trace here in the online editor.

But logging the exception message was just a tiny bit tricky. You can access it with `e.error` _but_ if the same method catches something like ValueError('text'), it'll have extra errors because .error doesn't exist, while directly logging the error does work in that case.

The stack trace was nice and simple!
Just import traceback and put in `print(traceback.format_exc())`.

services/core-api/app/date_time_helper.py
```diff
@@ -1,5 +1,8 @@
 from dateutil.relativedelta import relativedelta
+from datetime import datetime
 
+def get_formatted_current_time():
+    return datetime.now().strftime("%d/%b/%Y %H:%M:%S")
```
taraepp, line 5: Minor, but I remember writing some datetime helper functions... They're in services/core-apii/app/api/utils/helpers.py. There aren't any methods that I saw that this duplicates, but it looks like it could be combined maybe. This is a pre-existing thing so feel free to ignore it!

simensma-fresh (approved):

asinn134 (approved):

matbusby-fw (approved):
🙌🏻

PR #2820, opened 2023-11-22
[MDS-5640] ESUP Modal Bugs

taraepp:
## Objective 
- make all the actions do what they're expected to do
- make all the actions be what they're expected to be (for a particular record)
- pretty much iron out all the various user flows that a user can go through in ESUPs
- fixed some other bugs I noticed while I was at it

[MDS-5640](https://bcmines.atlassian.net/browse/MDS-5640)


services/common/src/redux/reducers/explosivesPermitReducer.ts
```diff
@@ -14,9 +14,27 @@ const initialState: IExplosivesPermitReducerState = {
...
   switch (action.type) {
     case actionTypes.STORE_EXPLOSIVES_PERMITS:
+      const records = action.payload?.records ?? [];
+      const explosivesPermits = records.map((permit) => {
+        const amendment_count = permit.explosives_permit_amendments.length;
+        const amendments = permit.explosives_permit_amendments.map((a, index) => {
+          const documents = a.documents.map((d) => ({
+            ...d,
+            explosives_permit_document_type_code: d.explosives_permit_amendment_document_type_code,
+          }));
```
taraepp, line 24: whyyyy did we give this a different name? It makes bugs. This is one of a few things that maybe should have been a backend fix, but I didn't want to expand scope to include a migration.

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
```diff
@@ -31,7 +31,6 @@ import {
   maxLength,
   number,
   required,
-  validateSelectOptions,
 } from "@common/utils/Validate";
```
taraepp, line 34: validateSelectOptions made app crash with too many updates so I made the decision to _get rid of it_ in this file. ...Fix the component instead? 😄

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
```diff
@@ -109,16 +112,36 @@ export const ExplosivesPermitFormNew: FC<ExplosivesPermitFormProps &
...
+  const mineManagers = partiesData.filter(
+    ({ mine_party_appt_type_code }) => mine_party_appt_type_code === "MMG"
+  );
+  const permittee = partiesData.filter(
+    ({ mine_party_appt_type_code, related_guid }) =>
+      mine_party_appt_type_code === "PMT" && related_guid === mines_permit_guid
+  );
+
+  const dropdown = (array = []) =>
+    array.map((item) => {
```
taraepp, line 138: removing ternaries anywhere I can!

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
```diff
@@ -577,17 +609,31 @@ export const ExplosivesPermitFormNew: FC<ExplosivesPermitFormProps &
         </Col>
       </Row>
       <Row className="flex-between form-button-container-row">
-        <Button onClick={() => setParentView(true)} className="full-mobile" type="ghost">
-          Back
-        </Button>
-        <Button type="primary" className="full-mobile" htmlType="submit" loading={props.submitting}>
-          {isHistoric ? "Submit" : "Finish And Generate Certificate"}
+        <Popconfirm
```
taraepp, line 612: I added a PopConfirm, and we show a Close button now where Back isn't appropriate.

services/common/src/redux/reducers/explosivesPermitReducer.ts
```diff
@@ -14,9 +14,27 @@ const initialState: IExplosivesPermitReducerState = {
...
+          return { ...a, documents, amendment_count, isAmendment: true, amendment_no: index + 1 };
```
asinn134, line 25: There probably is a change to how things work now but I think for the original permit we want the amendment_no to be zero, would having the index + 1 here cause the original permit to have amendment_no as one?

asinn134: o wait, this is just for the amendments and doesn't include the original I think? if so my mistake please ignore the comment above

taraepp: You got it. 👍 
Original permit gets assigned 0 below.
In some places we have the amendments including the original permit and some places we don't, which can get kind of confusing! 😅

services/common/src/redux/reducers/explosivesPermitReducer.ts
matbusby-fw: Yeah, I remember hitting this same point and making the same decision.  I think originally it was due to a lack of a full understanding of what needed to be differentiated here.  The amendments got created as pretty much a straight copy of the permits to start with.

asinn134 (approved):

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
matbusby-fw: I feel like we over use the PopConfirm and I've developed a distaste for using it when it doesn't need to be.  Can we apply that only when it's a close action.  Otherwise it's an unnecessary UI annoyance for users.

services/core-web/src/components/mine/ExplosivesPermit/MineExplosivesPermitTable.tsx
```diff
@@ -116,9 +114,45 @@ const MineExplosivesPermitTable: FC<RouteComponentProps & MineExplosivesPermitTa
...
-    type?: "permit" | "amendment"
-  ): ColumnType<MineExplosivesTableItem | amendmentsWithTotal> => {
+  const viewPermitAction: ITableAction = {
+    key: "view",
+    label: "View",
+    clickFunction: handleOpenViewExplosivesPermitModal,
+    icon: viewIcon,
+  };
+
+  const editPermitFunction = isFeatureEnabled(Feature.ESUP_PERMIT_AMENDMENT)
```
matbusby-fw, line 124: Nice cleanup!

matbusby-fw (approved):
Looks great.  Just one thought about making the popconfirm conditional *shakes fist at popConfirms

services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
taraepp: I would actually like to make it conditional on _the form being dirty_. I still want it on the "Back" because they could have entered data on the form.

PR #2823, opened 2023-11-24
[FIX] - Allow DB writes without JWT Authentication

Jsyro:
authorization should be handled well before this. this is to allow non-OIDC JWT's to write to the db if needed. Specifically verifiable credential webhook handler to receive calls from Traction

## Objective 

core-api needs to receive webhooks calls from traction and make database changes. 

If the request handling is reading for a username, but there is no `authorization` don't crash use the string `system` instead.

more feature flagging from namespace registration on app startup, to request time checks.


simensma-fresh (approved):

matbusby-fw (approved):
🙌🏻

PR #2824, opened 2023-11-24
[MDS-5607] - Exception handling + Email Notification

isuru-aot:
## Objective 

[MDS-5607](https://bcmines.atlassian.net/browse/MDS-5607)
[MDS-5617](https://bcmines.atlassian.net/browse/MDS-5617)

Hackathon work item for exception handling. 

Items covered:
- Added custom exceptions to mines namespace (resources and models)
- Added E-mail with `VIEW_ALL` rights
- Added notify admin feature when an error occurred (in FE)
- Added `ERROR_NOTIFY_RECIPIENTS` environment variable to configure email address for  error notification.


services/core-api/app/api/commons/resources/email_resource.py
```diff
@@ -0,0 +1,25 @@
+from app.api.services.email_service import EmailService
+from flask_restplus import Resource
+from flask import request
+
+from app.api.utils.access_decorators import (requires_any_of, VIEW_ALL)
+from app.api.utils.resources_mixins import UserMixin
+from app.api.exception.mds_core_api_exceptions import MDSCoreAPIException
+
+class EmailResource(Resource, UserMixin):
```
simensma-fresh: I would suggest that we maybe do not expose this as a generic way of sending emails accross the app. I've got some concerns with this in terms of potential abuse, as this would in essence allow anyone with access to send arbitrary emails on behalf of the mds team. The other concern I have would be that we allow plain html as input which would open up for html injection

As an alternative we could make this a specific "Report Error" endpoint that takes in the required fields so we can do some basic validation on it and create an email template that is specific for the error reporting in the templates folder which would escape what's passed in.

services/common/src/redux/customAxios.js
```diff
@@ -15,6 +19,61 @@ const formatErrorMessage = (errorMessage) => {
...
+  const reported_date = `${date} ${date.getHours()}:${date.getMinutes()}`;
+  const user = decodeJWT(createRequestHeader().headers.Authorization)
+  const environment_name = ENVIRONMENT.environment.toString().toUpperCase();
+
+  const email_title = "[MDS_INCIDENT_REPORT] [" + environment_name + "] - " + reported_date + " - " + business_message;
+  const email_body = `<p><b>Environment:</b> ${environment_name}</P>
+    <p><b>Business Error:</b> ${business_message}</P>
+    <p>
+      <b>Reporter's Name:</b> ${user.given_name} ${user.family_name}</br>
+      <b>Reporter's Email:</b> ${user.email}</br>
```
simensma-fresh: With the change suggested above to make the email endpoint specific to error reporting, I would suggest trying to get these details (user email / username etc) from the users token on the endpoint level so you would not be able to send these on behalf of another user

simensma-fresh (changes requested):

services/common/src/redux/customAxios.js
```diff
@@ -34,9 +93,22 @@ const CustomAxios = ({ errorToastMessage, suppressErrorNotification = false } =
         (errorToastMessage === "default" || errorToastMessage === undefined) &&
         !suppressErrorNotification
       ) {
+        console.error('Detailed Error: ', error?.response?.data?.detailed_error)
+        const notificationKey = 'errorNotification';
         notification.error({
+          key: notificationKey,
           message: formatErrorMessage(error?.response?.data?.message ?? String.ERROR),
+          description: <p style={{ color: 'grey' }}>If you think this is a system error please help us to improve by informing the system Admin</p>,
```
simensma-fresh: Could we put this behind a feature flag in case we decide on doing any iterations on this before releasing?

services/common/src/redux/customAxios.js
```diff
@@ -34,10 +62,33 @@ const CustomAxios = ({ errorToastMessage, suppressErrorNotification = false } =
         (errorToastMessage === "default" || errorToastMessage === undefined) &&
         !suppressErrorNotification
       ) {
-        notification.error({
-          message: formatErrorMessage(error?.response?.data?.message ?? String.ERROR),
-          duration: 10,
-        });
+        console.error('Detailed Error: ', error?.response?.data?.detailed_error)
```
matbusby-fw, line 65: Missed a console log 😜

services/core-api/app/api/mines/exceptions/mine_exceptions.py
```diff
@@ -0,0 +1,42 @@
...
+    """A Custom Exception for Errors in MINE Namespace"""
+
+    description = (
+        "Generic exeption for errors occurred in the Mine module"
+    )
+    def __init__(self, message = "Oops! Something went wrong", **kwargs):
+        super().__init__(message, **kwargs)
+        self.code = int(kwargs.get("status_code", 500))
+
+class ExplosivesPermitDocumentExeption(MineException):
```
matbusby-fw: Relatively minor, but there's a few instances here of this typo:

Exeption should be `Exception`

services/common/src/redux/customAxios.js
isuru-aot: Thanks @matbusby-fw,
Actually, I kept this purposely, as the detailed error will not be visible to FE anymore with this change. But it may not good practice. Shall I remove it?

services/core-api/app/api/commons/resources/email_resource.py
isuru-aot: Thanks Simen,
Added an endpoint for report-error, and only the error details are accepted in that. And email get sent with user info from BE. And the html template is used for the email content.

services/common/src/redux/customAxios.js
isuru-aot: This is addressed. Thanks.

services/common/src/redux/customAxios.js
isuru-aot: Added a feature flag from FE. Thanks

services/core-api/app/api/mines/exceptions/mine_exceptions.py
isuru-aot: Oh I "C"  :) 
Fixed. Thanks

services/common/src/redux/customAxios.js
matbusby: Well, I don't suppose it will hurt anything, but it might be of limited usefulness to us since we're not going to be seeing the user's console.

matbusby-fw (approved):
Looking good!  I'm excited about this feature!

henryoforeh-dev (approved):

simensma-fresh (approved):

PR #2826, opened 2023-11-27
[MDS-5604] Disable ESUP form submit while uploading documents

taraepp:
## Objective 
- disable the submit button on ESUP form while documents are uploading
- found & fixed bug where BE error would occur when using the "Edit Documents" action on an _amended_ ESUP (would occur on submit, due to document type field name mismatch)
- found another bug where all the necessary data for parties was being filtered out on the applications page- making previously selected parties show as their ID, and not showing expected options in new records- removed the filtering

[MDS-5604](https://bcmines.atlassian.net/browse/MDS-5604)

- I did put the "form is loading" tracking in with the modal actions: reason being they're already combined, consider refactor to separate file in future if we do more form upgrades
- had fun with redux hooks! Guided by docs https://react-redux.js.org/api/hooks
   - at first I implemented the useActions at the bottom of the page, but in the end it wasn't any better or less verbose than useDispatch so I cut it out. 
- had a real struggle with that test, adding redux logic forced me to essentially mock the redux form.
- bonus conversions to TS
- I took out a isMultiple param that wasn't doing anything (we were passing it as a prop to a component that didn't consume it)


services/core-web/src/components/Forms/ExplosivesPermit/ExplosivesPermitFormNew.tsx
```diff
@@ -168,7 +167,7 @@ export const ExplosivesPermitFormNew: FC<ExplosivesPermitFormProps &
   );
   const disabled = formMode === EsupFormMode.edit_document;
   const showBackButton = formMode === EsupFormMode.select_type_modal;
-  const isAmendment = formMode === EsupFormMode.amend;
+  const isAmendment = formMode === EsupFormMode.amend || initialValues.isAmendment;
```
taraepp, line 170: this was the change for bug with "Edit Documents" on amendment.

services/common/src/constants/actionTypes.ts
```diff
@@ -9,6 +9,7 @@ export const CLEAR = "CLEAR";
 export const OPEN_MODAL = "OPEN_MODAL";
 export const CLOSE_MODAL = "CLOSE_MODAL";
 export const UPDATE_MODAL_TITLE = "UPDATE_MODAL_TITLE";
+export const STORE_IS_LOADING = "STORE_IS_LOADING";
```
henryoforeh-dev: The name for this is a bit confusing. I see that it is used to check if form is loading.

taraepp: 👍 
Will change to STORE_FORM_IS_LOADING

henryoforeh-dev (approved):
Nice work!

asinn134 (approved):

PR #2828, opened 2023-11-29
[MDS-5634] link related projects part 1

henryoforeh-dev:
## Objective 

[MDS-5634](https://bcmines.atlassian.net/browse/MDS-5634)

This is the first part of this PR.

- Implemented end point to link 2 or more projects.
- Updated response payload of project to include linked projects (attribute name is called **project_links**).
- Updated TS interface/model to include **project_links** attribute.
- Implemented integration tests for new end point in core-api.


services/common/src/interfaces/projects/projectLink.interface.ts
```diff
@@ -0,0 +1,27 @@
+interface IProjectLinkContact {
+  name: string;
+}
+
+interface IProjectLinkSummary {
+  status_code: string;
```
taraepp: This one looks like it might want to be an enum instead of a string, which we probably should have done before. Do we have a finite list of project statuses?

services/core-api/app/api/projects/project_link/resources/project_link.py
```diff
@@ -0,0 +1,60 @@
...
+        'project_guid',
+        type=str,
+        store_missing=False,
+        required=False,
+    )
+    parser.add_argument(
+        'related_project_guid',
+        type=str,
+        store_missing=False,
+        required=False,
```
taraepp: Should these two guids be required arguments? And does the generated swagger show them in the params with mine_guid?

services/common/src/interfaces/projects/projectLink.interface.ts
matbusby-fw: Oh yeah!  Love me a good enum!

matbusby-fw (dismissed):
Love it!

asinn134 (approved):
nice work 👍

services/common/src/constants/enums.ts
```diff
@@ -134,3 +134,13 @@ export enum VC_CRED_ISSUE_STATES {
   credential_acked = "Active",
   deleted = "Active",
 }
+
+export enum PROJECT_SUMMARY_STATUS_CODES {
```
matbusby-fw, line 138: Beautiful!

matbusby-fw (approved):

taraepp (approved):
🎨

PR #2829, opened 2023-11-29
[MDS-5050] Adding active only status only for EOR and TQP

isuru-aot:
## Objective 

[MDS-5050](https://bcmines.atlassian.net/browse/MDS-5050)

Applied the filter only for TQP and EOR. So other mine_party_appt_types are returned without considering the status.

Screenshot 2023-11-29 122653
Screenshot 2023-11-29 122640

### DIFF
--- a/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
+++ b/services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
@@ -275,8 +275,14 @@ class MinePartyAppointment(SoftDeleteMixin, AuditMixin, Base):
                 party_guid=None,
                 mine_party_appt_type_codes=None,
                 include_permit_contacts=False,
-                active_only=True,
+                active_only=False,
                 mine_tailings_storage_facility_guid=None):
+
+        eor_and_tqp = (cls.mine_party_appt_type_code ==  "EOR"
+                                            or cls.mine_party_appt_type_code ==  "TQP" )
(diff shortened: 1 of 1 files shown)


henryoforeh-dev (approved):

services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -275,8 +275,14 @@ def find_by(cls,
...
                 mine_party_appt_type_codes=None,
                 include_permit_contacts=False,
-                active_only=True,
+                active_only=False,
                 mine_tailings_storage_facility_guid=None):
+
+        eor_and_tqp = (cls.mine_party_appt_type_code ==  "EOR"
+                                            or cls.mine_party_appt_type_code ==  "TQP" )
+        active_only_for_eor_and_tqp = ((cls.status == 'active' and eor_and_tqp)
+                                        or not eor_and_tqp)
```
asinn134, line 284: I am probably just missing something, but how come we need to check `or not eor_and_tqp`  for this part

taraepp (approved):

services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
isuru-aot: Good question.
Actually we want to consider status == active only for eor_and_tqp, anything other than eor_and_tqp has to be in the results. Because there are records in prod, with no status which needs to be in the results set.

PR #2830, opened 2023-11-29
[MDS-5049] TSF view and edit by roles

asinn134:
## Objective 

[MDS-5049](https://bcmines.atlassian.net/browse/MDS-5049)

- Adjusted within Core and Minespace the edit of TSF and Dam info to be available if user `has core_edit_tsf ` or `mds_minespace_proponents` roles
- User with the right roles assigned to them can edit or view TSF and Dam info
- Changed the TSF and Dam view and edit actions to be in a drop down menu instead of icons


services/core-web/src/components/mine/Tailings/MineTailingsTable.tsx
```diff
@@ -148,37 +206,31 @@ const MineTailingsTable: FC<RouteComponentProps & MineTailingsTableProps> = (pro
       title: "Actions",
```
taraepp: Would highly recommend replacing with renderActionsColumn.

services/core-web/src/components/mine/Tailings/MineTailingsTable.tsx
```diff
@@ -87,6 +85,66 @@ const MineTailingsTable: FC<RouteComponentProps & MineTailingsTableProps> = (pro
...
+            }}
+          >
+            <img src={EDIT_OUTLINE_VIOLET} alt="Edit Dam" />
+          </Button>
+        </AuthorizationWrapper>
+      </div>
+    );
+  };
+
+  const renderNewEditViewDropdown = (menu) => {
```
taraepp: renderActionsColumn!

asinn134:
I ended up not moving the tailing files into the common package for this ticket

services/core-web/src/components/mine/Tailings/MineTailingsTable.tsx
asinn134: This has been changed to use renderActionsColumn, thanks for the suggestion its a lot better with this.

services/core-web/src/components/mine/Tailings/MineTailingsTable.tsx
asinn134: This has been adjusted now as well.

services/minespace-web/common/components/tailings/AssociatedDams.tsx
```diff
@@ -46,6 +49,27 @@ const AssociatedDams: FC<AssociatedDamsProps> = (props) => {
     history.push(url);
   };
 
+  const editViewIcon = canEditTSF ? (
+    <img src={EDIT_OUTLINE} className="icon-sm padding-sm--right primary-colour" />
```
taraepp: Do people who have edit access not have the ability to just view the record?

services/core-web/common/components/tailings/AssociatedDams.tsx
```diff
@@ -12,22 +12,25 @@ import { getTsf } from "@mds/common/redux/reducers/tailingsReducer";
...
 import { useHistory } from "react-router-dom";
-import { EditIcon } from "@/assets/icons";
 import { ADD_DAM, EDIT_DAM } from "@/constants/routes";
-import { IDam, INoticeOfDeparture, ITailingsStorageFacility } from "@mds/common";
+import { IDam, ITailingsStorageFacility } from "@mds/common";
 import { RootState } from "@/App";
 import { ColumnsType } from "antd/lib/table";
 import CoreTable from "@/components/common/CoreTable";
+import { EDIT_OUTLINE } from "@mds/common/constants/assets";
+import { EyeOutlined } from "@ant-design/icons";
```
matbusby-fw: Smallish thing, but a big part of what sped up Minespace from Simen's work was actually importing antD icons individually so it doesn't grab everything at once.
```suggestion
import EyeOutlined from "@ant-design/icons/EyeOutlined";
```

services/minespace-web/common/components/tailings/AssociatedDams.tsx
```diff
@@ -12,22 +12,25 @@ import { getTsf } from "@mds/common/redux/reducers/tailingsReducer";
...
 import { useHistory } from "react-router-dom";
-import { EditIcon } from "@/assets/icons";
 import { ADD_DAM, EDIT_DAM } from "@/constants/routes";
-import { IDam, INoticeOfDeparture, ITailingsStorageFacility } from "@mds/common";
+import { IDam, ITailingsStorageFacility } from "@mds/common";
 import { RootState } from "@/App";
 import { ColumnsType } from "antd/lib/table";
 import CoreTable from "@/components/common/CoreTable";
+import { EDIT_OUTLINE } from "@mds/common/constants/assets";
+import { EyeOutlined } from "@ant-design/icons";
```
matbusby-fw:
```suggestion
import EyeOutlined from "@ant-design/icons/EyeOutlined";
```

services/minespace-web/src/components/dashboard/mine/tailings/TailingsTable.js
```diff
@@ -23,6 +23,8 @@ import AuthorizationWrapper from "@/components/common/wrappers/AuthorizationWrap
 import CoreTable from "@/components/common/CoreTable";
 import { Feature } from "@mds/common";
 import { useFeatureFlag } from "@mds/common/providers/featureFlags/useFeatureFlag";
+import { renderActionsColumn } from "@mds/common/components/common/CoreTableCommonColumns";
+import { EyeOutlined } from "@ant-design/icons";
```
matbusby-fw:
```suggestion
import EyeOutlined from "@ant-design/icons/EyeOutlined";
```

matbusby-fw (dismissed):
Looks great!  Just added a couple suggestions on importing Icons for better performance.  I'm guessing moving these components into the common package just became too much of a can of worms?

matbusby-fw (dismissed):
👍🏻

services/core-web/common/components/tailings/AssociatedDams.tsx
asinn134: Thanks, this should be resolved now

services/minespace-web/common/components/tailings/AssociatedDams.tsx
asinn134: Same here, this should be resolved now

services/minespace-web/src/components/dashboard/mine/tailings/TailingsTable.js
asinn134: Same here, should be resolved now too

services/minespace-web/common/components/tailings/AssociatedDams.tsx
asinn134: This has been adjusted now where people with edit access can choose to view or edit a TSF or dam

asinn134:
> Looks great! Just added a couple suggestions on importing Icons for better performance. I'm guessing moving these components into the common package just became too much of a can of worms?

Thanks! yah the moving of the components kind of complicated things, so I ended up just focusing on the story ac for this one.

henryoforeh-dev (dismissed):

henryoforeh-dev (dismissed):

services/common/src/constants/assets.ts
```diff
@@ -1 +1,2 @@
 export { default as SMALL_PIN_SELECTED } from "../assets/images/small-pin-selected.svg";
```
taraepp: We have _several_ instances of the edit icon! I don't think we need another.
We've got VioletEditIcon on Core which I just diffchecked against EditIcon (on both) and the difference is only the fill colour (which it could pick based on a boolean param, or be set via CSS.).

I am also curious about the extra classes on the icons, which I haven't had to use- are they not rendering correctly? Or do we have specific CSS rules in those other cases that should be more broadly-applied?

services/core-web/common/components/tailings/AssociatedDams.tsx
```diff
@@ -12,40 +12,76 @@ import { getTsf } from "@mds/common/redux/reducers/tailingsReducer";
...
+          handleNavigateToEdit(event, record, "edit");
+        },
+      },
+      {
+        key: "view",
+        label: "View Dam",
+        icon: <EyeOutlined className="icon-sm padding-sm--right" />,
+        clickFunction: (_event, record) => {
+          handleNavigateToEdit(event, record, (userAction === "edit" ? "editView" : "view"));
+        },
```
taraepp, line 70: What does editView mean? Seems like it would be simpler to just make it "edit".

services/core-web/common/components/tailings/TailingsSummaryPage.tsx
```diff
@@ -75,222 +77,246 @@ interface TailingsSummaryPageProps {
...
-    const [isReloading, setIsReloading] = useState(false);
-    const [uploadedFiles, setUploadedFiles] = useState([]);
+  const { mines, history, formErrors, formValues, mineGuid, tsfGuid, tab, userAction } = props;
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [isReloading, setIsReloading] = useState(false);
+  const [uploadedFiles, setUploadedFiles] = useState([]);
+  const [canEditTSF, setCanEditTSF] = useState(false);
+  const { renderConfig, components, routes, isCore } = useContext(TailingsContext);
+  const { Loading } = components;
+  const action = userAction ? userAction : "edit";
```
taraepp: set defaults when destructuring props like userAction="edit" on line 86.

services/core-web/common/components/tailings/AssociatedDams.tsx
asinn134: Ah editView was setup because we have a scenario where if a user can edit tsf and dams, they will be able to see the edit and view options in the action menu. If the user with edit access clicks on the dam view button, it will bring them to a dam page with all the user inputs disabled. When the user moves back out of the dam page and into the TailingsSummaryPage we still want them to have the ability to edit things. Having the editView helps let us know the user should still be able to edit. Let me know if it doesn't make sense we can talk about it further.

services/core-web/common/components/tailings/TailingsSummaryPage.tsx
asinn134: This has been adjusted now.

services/common/src/constants/assets.ts
asinn134: This has been removed, it now uses the same icons in the esup table. The extra classes used aren't needed when we keep the styling the same as esup table, its just if we were to stick to the styling of say the actions menu in the Permit table, without those extra classes the icon looks too small or it is missing the purple colour for Core and the blue for MineSpace.

services/core-web/common/components/tailings/AssociatedDams.tsx
taraepp: Looks like we are already tracking their permissions on the page that consumes it, so I'm not sure that we need this other variable to also indicate that they should have permission to edit. That userAction still really looks to me like it wants to be a boolean- it is always compared to a string literal. A string constant would be better than a string literal, but a boolean I really think would be best. You're either in view mode or edit mode,  and the page itself can track whether the user has permission to change that.

matbusby-fw (dismissed):

henryoforeh-dev (dismissed):

services/core-web/common/components/tailings/AssociatedDams.tsx
asinn134: I think I should have addressed the points in the latest commit. The string comparisons are converted to be boolean, the route will depend on a boolean to determine view or edit. I tried using `history.goBack()` but found the load time of using that versus `history.push` to go back to previous page to be noticeably longer, so ended up sticking with `history.push`.

henryoforeh-dev (dismissed):

services/core-web/common/components/tailings/AssociatedDams.tsx
asinn134: The boolean change has now been applied.

taraepp (approved):

henryoforeh-dev (approved):

PR #2831, opened 2023-11-29
[MDS-5648] Minespace homepage pagination on Mines

taraepp:
## Objective 
- added pagination to the MS homepage
   - RULES: 
      - disabled when loading
      - can switch between 10/20/50/100 per page, default 10
      - the page navigation is hidden if all the results fit on the page showing 10 results
         - I happened to switch it to 100 during testing and I couldn't go back to 10 (confusing) and that is why I did it that way.  
- also fancy list-style skeleton (sorry, code isn't very fancy, neither was antd's example code)
- some typescript fixes & conversions
- made `ReduxWrapper` component for testing- supplies the store to child components and can take an initial state
   - snapshot will show _loading_ state now instead of unauthenticated! Progress! 
- used redux hooks instead of connect(mapStateToProps, mapDispatchToProps). DO WE LIKE IT?
  - redux docs says we have more work to do: https://redux.js.org/usage/migrating-to-modern-redux

[MDS-5648](https://bcmines.atlassian.net/browse/MDS-5648)

image
image


henryoforeh-dev (approved):
Nice!

isuru-aot (approved):

services/minespace-web/src/components/pages/MinesPage.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const handlePaginationChange = (page, pageSize) => {
+    setDesiredPage(page);
+    setDesiredPageSize(pageSize);
+  };
+
+  const renderSkeleton = () => {
+    return (
+      <div className="link-card">
+        <ul>
+          {new Array(DEFAULT_MINES_PER_PAGE).fill({}).map((_, i) => (
```
matbusby, line 57: Nice!

matbusby (approved):
Love it!  Looks solid!

PR #2832, opened 2023-12-01
[MDS-5634] link related projects part 2

henryoforeh-dev:
## Objective 

[MDS-5634](https://bcmines.atlassian.net/browse/MDS-5634)

- Moved CoreTable components from MineSpace and Core to the common package.
- Updated reference to CoreTable in both MineSpace web and Core web.
- Implemented a functionality to display linked projects in both MineSpace web and Core web.
- Created unit tests.

image

image


taraepp (commented):
One easy thing: There's a slight difference in the CoreTable right now between core-web and minespace-web. The one on MS is the one that Simen made changes to icon imports to improve loading times. Would recommend carrying through those changes instead of using the file from core-web

taraepp (approved):
I would consider just replacing the ProjectLinkTable with CoreTable, since it just seems to be passing props to it anyway, but that is a minor thing. Looks good!

simensma-fresh (approved):

PR #2835, opened 2023-12-04
[MDS-5654] Updated crr table in MS with new columns/design

matbusby-fw:
## Objective 

[MDS-5654](https://bcmines.atlassian.net/browse/MDS-5654)

image


services/common/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -87,15 +87,24 @@ export interface ITableAction {
   icon?: ReactNode;
 }
 
-export const renderActionsColumn = (
-  actions: ITableAction[],
-  recordActionsFilter?: (record, actions) => ITableAction[],
+export const renderActionsColumn = ({
```
matbusby-fw, line 90: I was playing with implementing a fixed action column. I didn't end up enabling for the report table yet, but I ended up converting this to an object prop since it was getting a bit long to pass in a bunch of null values just to get to the last one.

services/common/src/redux/utils/helpers.ts
```diff
@@ -162,6 +162,9 @@ export const dateSorter = (key: string) => (a: any, b: any) => {
 export const nullableStringSorter = (path) => (a, b) => {
   const aObj = get(a, path, null);
   const bObj = get(b, path, null);
+  if (typeof aObj === "number" && typeof bObj === "number") {
```
matbusby-fw, line 165: This is to avoid an error if a number gets passed in here.

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
```diff
@@ -0,0 +1,202 @@
...
+const propTypes = {
+  mine: CustomPropTypes.mine.isRequired,
+  mineReports: PropTypes.arrayOf(CustomPropTypes.mineReport).isRequired, // eslint-disable-next-line react/no-unused-prop-types
+  mineReportDefinitionOptions: PropTypes.arrayOf(PropTypes.objectOf(PropTypes.any)).isRequired,
+  updateMineReport: PropTypes.func.isRequired,
+  createMineReport: PropTypes.func.isRequired,
+  fetchMineReports: PropTypes.func.isRequired,
+  openModal: PropTypes.func.isRequired,
+  closeModal: PropTypes.func.isRequired,
+};
```
henryoforeh-dev: Are these propTypes still required since there is already an interface with the same attributes defined below?

matbusby-fw: Oh, good catch.  Yes, they can go.

henryoforeh-dev (dismissed):

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
```diff
@@ -0,0 +1,191 @@
...
+  mineReports,
+  mineReportDefinitionOptions,
+  ...props
+}) => {
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [report, setReport] = useState(null);
+
+  useEffect(() => {
+    props.fetchMineReports(mine.mine_guid).then(() => {
+      setIsLoaded(true);
```
taraepp: would recommend the isMounted trick (or if you find something similar) to avoid the "can't do state update on unmounted component" error when setIsLoaded gets called.

services/common/src/components/common/CoreTableCommonColumns.tsx
taraepp: Good solution!

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
```diff
@@ -0,0 +1,191 @@
+import React, { FC, useEffect, useState } from "react";
+import { connect } from "react-redux";
+import { bindActionCreators } from "redux";
+import PlusCircleFilled from "@ant-design/icons/PlusCircleFilled";
+import { Button, Col, Row, Typography } from "antd";
+import moment from "moment";
+import PropTypes from "prop-types";
```
taraepp: wait, are you still using the PropTypes import? CustomPropTypes? I'm seeing a lot of unused imports lately... We should add an eslint rule and make it an error (instead of a warning).

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
```diff
@@ -0,0 +1,191 @@
...
+    });
+  };
+
+  return (
+    <Row>
+      <Col span={24}>
+        <Row>
+          <Col span={24}>
+            <div className="submit-report-button">
+              <AuthorizationWrapper>
```
taraepp, line 123: What's the authorization wrapper looking for? I thought it wanted a permission.

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
matbusby-fw: Oooh, good opportunity to switch to those redux hooks.  On it.

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
matbusby-fw: Good catch!  Using Intellij I usually just run `shift + cmd + l` and it runs a "clean up imports" function.  Forgot to run it on this one.  But I agree a rule to catch unused imports would be useful.  I'll look into it.

services/minespace-web/src/components/dashboard/mine/reports/Reports.tsx
matbusby-fw: 🤔 Didn't mess with the guts of this at all, so off the top of my head I'm not sure.  I'll take a peek.

matbusby-fw: Hmmm, looks like this wrapper is meant more to prevent use of certain functions if a user is signed in with an IDIR.  So it just makes sure they're using a BCEID.

henryoforeh-dev (approved):

taraepp (approved):

PR #2839, opened 2023-12-08
[MDS-5049]fix tsf mapping null error

asinn134:
## Objective 

[MDS-5049](https://bcmines.atlassian.net/browse/MDS-5049)

-There was an null error occurring when TSF page was opened in the test environment.

-The error was due to my recent merge not having all the up to date changes in the develop branch, so for example in my branch I had something like this renderActionsColumn(newTSFActions) when it should be renderActionsColumn({ actions: newTSFActions })

-Also did some small refactor in the TSF section

### DIFF
--- a/services/common/src/constants/strings.tsx
+++ b/services/common/src/constants/strings.tsx
@@ -173,6 +173,7 @@ export const NOTICE_OF_DEPARTURE_STATUS_VALUES = {
   determined_non_substantial: "determined_non_substantial",
   determined_substantial: "determined_substantial",
   withdrawn: "withdrawn",
+  self_authorized: "self_authorized",
 };
 
 export const NOD_TYPE_FIELD_VALUE = {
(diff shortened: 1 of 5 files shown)


services/common/src/constants/strings.tsx
```diff
@@ -173,6 +173,7 @@ export const NOTICE_OF_DEPARTURE_STATUS_VALUES = {
   determined_non_substantial: "determined_non_substantial",
   determined_substantial: "determined_substantial",
   withdrawn: "withdrawn",
+  self_authorized: "self_authorized",
```
asinn134, line 176: Noticed this was missing and causing an error in mds\services\core-web\src\components\modalContent\NoticeOfDepartureModal.tsx line 203. I just added this here not sure if anyone has anymore context or if this is the correct string value to set for this?

henryoforeh-dev (approved):

taraepp (approved):

PR #2841, opened 2023-12-08
[MDS-5655] - CRR getting started page

matbusby-fw:
## Objective 

1. Created a `ReportSteps` component as parent for the CRR steps
2. Created the `GettingStarted` child component for ReportSteps.
3. Common reports are stubbed in, but not functional
4. Report search input is likewise not functional
5. New Files live in the common package as per AC that this functionality will exist in Core in the future

[MDS-5655](https://bcmines.atlassian.net/browse/MDS-5655)

image


services/common/src/components/reports/ReportGetStarted.tsx
```diff
@@ -0,0 +1,105 @@
...
+        screen.
+      </Typography.Paragraph>
+      <Row>
+        <Col span={12}>
+          <Typography.Paragraph strong className="margin-large--top">
+            Report Code Requirement
+          </Typography.Paragraph>
+          <Form layout="vertical">
+            <Form.Item label="Enter Code Section">
+              <Input placeholder="10.4.1" />
```
taraepp, line 81: I thought this looked odd so I checked the mockup- looks like it's incomplete? "eg. 10.4.1 / Annual DSI"

henryoforeh-dev (approved):
Nice work!

services/common/src/components/reports/ReportGetStarted.tsx
matbusby-fw: image

https://www.figma.com/file/O7i6jpbKbju49HfL1OpozT/CORE-%26-Minespace-2021-2023?type=design&node-id=27411-6318&mode=design&t=DZ8ngmVT4lFY2Em2-0

This is the mockup I was working off of, where did you find that?

taraepp: To the right of this one! I still think it's not good placeholder text but if it's getting changed in another ticket then it'll do for now.

taraepp (approved):

PR #2842, opened 2023-12-08
[MDS-5637] Added volume mount for file uploads

simensma-fresh:
## Objective 

[MDS-5637](https://bcmines.atlassian.net/browse/MDS-5637)

Added separate volume for file uploads to permit service. This should prevent an "file not found" error when uploading a PDF through the rest api.

image


henryoforeh-dev (approved):

matbusby-fw (approved):
👍🏻

PR #2843, opened 2023-12-11
[MDS-5676] Updated webpack config to speed up loading of core

simensma-fresh:
## Objective 

[MDS-5676](https://bcmines.atlassian.net/browse/MDS-5676)

Copied over the webpack changes made to improve loading speeds of Minespace over to Core, with the hope that we'll see similar improvements there. This includes the following (low hanging fruits), but does not include things like lazy loading of the PDF viewer, and code splitting based on route, to keep application code changes to a minimum.

- Don’t bundle source maps
- Split PDF Viewer and leaflet into separate bundles
- Don’t bundle all Moment locales
- Use ES import compatible version of lodash
- Extract critical CSS + road the rest async
- Remove Keycloak redirect for checking if a user is logged in (perform this in background using the check-sso flag
- Update core web runner to enable better compression of files


taraepp (approved):
🙌

henryoforeh-dev (approved):

PR #2848, opened 2023-12-11
[MDS-5828] Run cypress tests against local DB

simensma-fresh:
## Objective 

[MDS-5828](https://bcmines.atlassian.net/browse/MDS-5828)

Run Cypress tests against local DB instead of dev environment. This will make them a bit more robust and prevent backend changes that break the tests to make its way into the codebase.

**Changes**
- Split Cypress tests into separate actions for Minespace and Core
- Added `make be-minimal` command that spins up a minimal version of the backend required for the cypress tests to run
- Upload container logs as artifact that you can download if tests fail
- Updated tests to run against minespace/core docker containers instead of manually running `yarn` to make it as close to the deployed version as possible + enables parallel building of the frontend and backend
- Moved full rebuilding of the backend container into `make be-rebuild` instead of having as part of `make be`, seeing as a full rebuild is rarely needed.
- `make seeddb` can now take a param specifying how many entities to seed the database with: `make ENTRIES=10 seeddb`
- Always seed the DB with a major mine as to make sure major project tests don't fail

### DIFF
--- a/.github/workflows/core-web.cypress.yaml
+++ b/.github/workflows/core-web.cypress.yaml
@@ -1,0 +1,87 @@
+name: CORE WEB - Cypress Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - services/common/**
+      - services/core-web/**
+      - services/core-api/**
+      - services/document-manager/**
+
+jobs:
+  run-cypress:
+    name: run-cypress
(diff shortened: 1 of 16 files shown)


matbusby-fw (approved):
Awesome!  Love that you seperated out the tests into their own workflows!

henryoforeh-dev (approved):

PR #2851, opened 2023-12-13
[MDS-5656] Add Report Details - CRR

isuru-aot:
## Objective 

[MDS-5656](https://bcmines.atlassian.net/browse/MDS-5656)

Screenshot 2023-12-13 110548

- The dependency files associated with the change are moved to the common location
- Added the existing vide from the modal.
- Added new fields according to the UI (some with dummy data)
- The file submission should be covered in the final step which is to be covered in MDS-5660.


services/common/src/components/common/FileUpload.js
```diff
@@ -0,0 +1,280 @@
...
+  itemInsertLocation: "before",
+  notificationDisabledStatusCodes: [],
+  shouldReplaceFile: false,
+  replaceFileUploadUrl: "",
+  file: null,
+  shouldAbortUpload: false,
+  onAfterResponse: () => {},
+};
+
+class FileUpload extends React.Component {
```
matbusby-fw: Interesting.  I really like the idea of having the fileUpload component moved to the common package, but I don't think they behave in exactly the same way.  

We don't need to remove this, but we'll need to consider rendering the differences seperately for each FE if we want to use this in Core.

services/common/src/components/common/RenderAutoComplete.js
```diff
@@ -0,0 +1,76 @@
+import React from "react";
```
matbusby-fw: Hmmm, this might get a bit sticky.  Tara has spent a bunch of time moving all of the form components into the common package and converting them to .tsx.  Going to have to coordinate merging these I think.

services/common/src/components/common/utils/actionlessNetworkCalls.js
```diff
@@ -0,0 +1,137 @@
+import { ENVIRONMENT } from "@mds/common";
```
matbusby-fw: This file already exists at: services/common/src/redux/utils/actionlessNetworkCalls.js

We can just use that one 😁 

image

services/common/src/components/modalContent/config.js
```diff
@@ -0,0 +1,45 @@
+// import AddReportModal from "@/components/modalContent/reports/AddReportModal";
+import EditReportModal from "@mds/common/components/modalContent/reports/EditReportModal";
+// import AddVarianceModal from "@/components/modalContent/variances/AddVarianceModal";
```
matbusby-fw: I like having a shared config for this, but I think we just delete everything not being used, and add them back in as needed. (Honestly would rather get rid of the config altogether, but I don't think that's happening right away)

services/common/src/components/reports/AddReportDetails.tsx
```diff
@@ -0,0 +1,476 @@
...
+  );
+  dropdownMineReportDefinitionOptionsFiltered = sortListObjectsByPropertyLocaleCompare(
+    dropdownMineReportDefinitionOptionsFiltered,
+    "label"
+  );
+};
+
+interface AddReportDetailsProps {
+  mineGuid: string;
+  updateMineReportSubmissions: any;
```
matbusby-fw: For anything coming from the `mapDispatchToProps` we can add a type like this:

`updateMineReportSubmissions: ActionCreator<typeof updateMineReportSubmissions>;`

This will give you all the correct typing when working with it throughout the file and make it easier to maintain.

services/common/src/components/reports/AddReportDetails.tsx
```diff
@@ -0,0 +1,476 @@
...
+  initialValues: any;
+  mineReportDefinitionOptions: any;
+  selectedMineReportCategory: any;
+  selectedMineReportDefinition: any;
+  updateDueDateWithDefaultDueDate: any;
+}
+
+const AddReportDetails: FC<AddReportDetailsProps> = (props) => {
+  const { mineGuid, dropdownMineReportCategoryOptions } = props;
+  const [state, setState] = useState({
```
matbusby-fw: Is this just an artifact of coming from a class based component at one point in time?

I think it's much cleaner to seperate all states into their own useStates as you've done lower down.  Makes it more consistent and readable.  And pretty.  It also makes it pretty.

services/common/src/components/common/utils/helpers.ts
```diff
@@ -0,0 +1,602 @@
+import {
```
matbusby-fw: This also exists in 

`services/common/src/redux/utils/helpers.ts` 💃🏻 🕺🏻

services/common/src/components/reports/AddReportDetails.tsx
```diff
@@ -0,0 +1,476 @@
...
+              mineGuid={mineGuid}
+              mineReportSubmissions={props.mineReportSubmissions}
+              updateMineReportSubmissions={updateMineReportSubmissions}
+            />
+          </Form>
+        </Col>
+
+        <Col span={24}>
+          <Form layout="vertical">
+            <ReportsTable
```
matbusby-fw: 🤔 I think this might not be the correct table.  I think the designs are just asking for a table of the documents, not the other reports.

I see you brought over the editReport modal and a few other things that are related to this table, so it would probably mean that we can leave those back where they were.

image

matbusby-fw (changes requested):
Lots of awesome work here!  I made a few comments, but the big one I think will need to be dealt with is the table at the bottom of your new page should just be a document table instead of the reports table.

Once that is swapped out, I think you'll find that a fair amount of the stuff moved to common doesn't need to be there (I think the reports table itself is fairly Minespace specific)

services/common/src/components/common/RenderAutoComplete.js
isuru-aot: Removed unnecessary props.

services/common/src/components/common/utils/actionlessNetworkCalls.js
isuru-aot: Removed.

services/common/src/components/modalContent/config.js
isuru-aot: Removed as these became unnecessary after addressing all the comments.

services/common/src/components/reports/AddReportDetails.tsx
isuru-aot: Updated.

services/common/src/components/reports/AddReportDetails.tsx
isuru-aot: Updated.

services/common/src/components/common/utils/helpers.ts
isuru-aot: Removed

services/common/src/components/reports/AddReportDetails.tsx
isuru-aot: As discussed added a new component.

services/common/src/assets/images/edit-pencil.svg
taraepp, line 1: Noooooo we don't need more copies of this file.
Use the antd one for actions menu.

services/common/src/components/common/FileUpload.js
```diff
@@ -0,0 +1,280 @@
+/* eslint-disable */
+import React from "react";
```
taraepp: now that I merged my PR, I think it's just the Callout and LinkButton that aren't already in common. This one's called RenderFileUpload now.

services/common/src/customPropTypes/common.js
```diff
@@ -0,0 +1,19 @@
+import { PropTypes, shape, arrayOf, oneOfType } from "prop-types";
```
taraepp: I'm not sure about bringing in customPropTypes to common when we're phasing it out. What can we do to remove it instead?

taraepp:
Generally looks good, but I do want to avoid pulling a bunch of stuff into common that's already slated for deletion, most notably customPropTypes which we use interfaces for now instead.
I guess, when I ran into that recently, when it was too much work to convert too many files to TS, I just commented out the customPropTypes and replaced it with "any", so that you're still able to see what the object should look like when it comes time to convert it.

services/common/src/constants/forms.js
```diff
@@ -0,0 +1,20 @@
+// Redux Forms
+export const ADD_REPORT = "ADD_REPORT";
```
taraepp: I think I also remade forms.js as a forms enum. Still works pretty much the same way but it only has one entry so far and doesn't trigger duplicate named import warnings/errors.

services/common/src/components/common/FileUpload.js
isuru-aot: This is updated accordingly

services/common/src/customPropTypes/common.js
isuru-aot: This is removed.

services/common/src/assets/images/edit-pencil.svg
asinn134: I might be wrong but it doesn't seem like you are using this image in your changes? Might be able to just remove this file.

asinn134 (approved):

matbusby-fw (approved):
Nice!  Thanks for all the hard work on this one.  Sorry the scope got a little out of control!

PR #2852, opened 2023-12-13
[MDS-5388] Add page size option

henryoforeh-dev:
## Objective 

[MDS-5388](https://bcmines.atlassian.net/browse/MDS-5388)

Added missing page size option.


matbusby-fw (approved):
Superstar!

isuru-aot (approved):

PR #2855, opened 2023-12-15
[MDS-5672] Fixed non re-rendering of documents section in incidents form

matbusby-fw:
## Objective 

The documents section of the incidentForm was holding on to old values if a user had navigated to a different incident previously.  This was because the render Documents section and all other sections were outside the form component and not re-rendering.  Moved the documents inside the form component and adjusted the handling logic slightly.

[MDS-5672](https://bcmines.atlassian.net/browse/MDS-5672)

### DIFF
--- a/services/core-web/src/components/Forms/incidents/IncidentForm.tsx
+++ b/services/core-web/src/components/Forms/incidents/IncidentForm.tsx
@@ -2,7 +2,7 @@ import React, { useState } from "react";
 import { withRouter } from "react-router-dom";
 import PropTypes from "prop-types";
 import { bindActionCreators, compose } from "redux";
-import { connect, useSelector } from "react-redux";
+import { connect, useSelector, useDispatch } from "react-redux";
 import { change, Field, FieldArray, formValueSelector, getFormValues, reduxForm } from "redux-form";
 import { LockOutlined, PlusOutlined } from "@ant-design/icons";
 import { Form } from "@ant-design/compatible";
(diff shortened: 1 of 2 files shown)


services/core-web/src/components/Forms/incidents/IncidentForm.tsx
```diff
@@ -142,794 +143,820 @@ const retrieveInitialReportDynamicValidation = (childProps) => {
   };
 };
 
-const renderInitialReport = (incidentCategoryCodeOptions, locationOptions, isEditMode) => {
```
matbusby-fw, line 145: This is mostly just moving all of these renders inside the component so they get updated fully if the user navigates between different incidents.

services/core-web/src/components/Forms/incidents/IncidentForm.tsx
```diff
@@ -142,794 +143,820 @@ const retrieveInitialReportDynamicValidation = (childProps) => {
...
+  );
+
+  const renderInternalDocumentsComments = (childProps, isEditMode, handlers, parentHandlers) => {
+    const incidentCreated = Boolean(childProps.formValues?.mine_incident_guid);
+    const internalMinistryDocuments = childProps.documents.filter(
+      (doc) => doc.mine_incident_document_type_code === "INM"
+    );
+    const dispatch = useDispatch();
+
+    const handleRemoveDocument = async (event, documentGuid: string) => {
```
matbusby-fw, line 741: This is the one change in the actual code.  Handling the deletion of documents here.  And then just calling the callback for cleanup.

asinn134 (approved):

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #2860, opened 2023-12-21
[MDS-5582] MS & Core - Allow more file types for report incident

isuru-aot:
## Objective 

[MDS-5582](https://bcmines.atlassian.net/browse/MDS-5582)

Screenshot 2023-12-21 145227
Screenshot 2023-12-21 145508


services/core-web/src/components/Forms/incidents/IncidentFormDocuments.tsx
```diff
@@ -70,7 +70,7 @@ const IncidentFormDocuments: FC<IncidentFormDocumentsProps> = ({
                 id={INITIAL_INCIDENT_DOCUMENTS_FORM_FIELD}
                 name={INITIAL_INCIDENT_DOCUMENTS_FORM_FIELD}
                 labelIdle='<strong>Drag & Drop your files or <span class="filepond--label-action">Browse</span></strong><br>
-              <div>Accepted filetypes: .kmz, .doc, .docx, .xlsx, .pdf</div>'
+              <div>Accepted filetypes: .kmz, .doc, .docx, .xlsx, .pdf, .jpeg, .tiff, .hiec</div>'
```
taraepp: Just noticed that .png isn't listed here! And should this include .msg or did that get taken out?

services/common/src/constants/fileTypes.js
```diff
@@ -30,6 +30,11 @@ export const IMAGE = {
   ".jpg": "image/jpg",
   ".jpeg": "image/jpeg",
   ".png": "image/png",
+  ".jp2": "image/jp2",
+  ".tif": "image/tiff",
+  ".tiff": "image/tiff",
+  ".heic": "image/heic",
+  ".heif": "image/heif",
 };
```
matbusby-fw, line 39: I think we'll want to create the message file type here as well to include in this update.  

Something like this:
```suggestion
export const MESSAGE = {
  ".msg": "application/vnd.ms-outlook",
};
```

services/core-web/src/components/Forms/incidents/IncidentFormDocuments.tsx
isuru-aot: .png and .msg was missed. Updated accordingly.

services/common/src/constants/fileTypes.js
isuru-aot: Thanks Done!

matbusby-fw (approved):
Nice!  🙌🏻

henryoforeh-dev (approved):
Nice!

PR #2861, opened 2023-12-22
[MDS-5673]adjustment to crr description comment

asinn134:
## Objective 

[MDS-5673](https://bcmines.atlassian.net/browse/MDS-5673)

- Adjusting crr general description input in Minespace

### DIFF
--- a/services/common/src/components/reports/AddReportDetails.tsx
+++ b/services/common/src/components/reports/AddReportDetails.tsx
@@ -269,7 +269,7 @@ const AddReportDetails: FC<AddReportDetailsProps> = (props) => {
 
         <Col span={12}>
           <Form layout="vertical">
-            <Form.Item label="* Report Compliance Year/Period">
+            <Form.Item label="Report Compliance Year/Period*">
               <Field
                 id="submission_year"
                 name="submission_year"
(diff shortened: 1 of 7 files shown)


simensma-fresh (dismissed):

services/common/src/components/reports/AddReportDetails.tsx
```diff
@@ -291,13 +297,17 @@ const AddReportDetails: FC<AddReportDetailsProps> = (props) => {
         </Col>
         <Col span={24}>
           <Form layout="vertical">
-            <Form.Item label="Report Title and Additional Comment">
-              <Input.TextArea
-                rows={3}
+            <Form.Item label="Report Title and Additional Comment*">
```
taraepp: Just to avoid some merge conflicts here- Form.Item is already within the Field element and doesn't need to be here too (I took them out on my branch, this is the only one that should make a conflict). Pass the label to Field to show it, might have to do that within a `props` object.

taraepp: Other than this one minor thing that's not much of a thing (I _can_ resolve it myself of course) everything looks good to me!

matbusby-fw (dismissed):

services/common/src/components/reports/AddReportDetails.tsx
asinn134: Sounds good, the Form.Item is removed now. I ended up not using label in the field or passing it a `props `object and set up the text in a `Typography.Text` due to some styling issue, hope this is okay?

matbusby-fw (dismissed):

taraepp (approved):

isuru-aot (approved):

PR #2865, opened 2024-01-03
[MDS-5675] Added support for direct S3 multipart file uploads

simensma-fresh:
## Objective 

[MDS-5675](https://bcmines.atlassian.net/browse/MDS-5675)

Added support for direct file uploads to S3. This is intended to speed up uploads to S3 as the current implementation using Tusd is not performing very well. Currently, the functionality has been feature flagged, so we can release it on our own time.

**Changes**
- Modified document creation logic to create a new S3 multipart upload instead of initializing the upload with Tusd. Parts are uploaded into the `/multipart` folder with a lifecycle rule that deletes uncompleted uploads after 1 day.
- Added `complete-upload` endpoint that completes the multipart upload and moves the file into its intended location
- Added configuration for running jest tests in the common folder
- Consolidated the minespace and core file upload components into `RenderFileUpload` and moved it into the common package
- A bit of cleanup of some of the file related files
- Added github action to run common package tests
- Enabled [desktop-lite](https://github.com/devcontainers/features/tree/main/src/desktop-lite) in codespaces so we can run Cypress there


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
+    "ghcr.io/devcontainers/features/desktop-lite:1": {}
+  },
+  "forwardPorts": [6080, 5901]
+}
```
simensma-fresh: This will in theory enable us to run Cypress in Codespaces https://github.com/devcontainers/features/tree/main/src/desktop-lite

henryoforeh-dev (dismissed):

services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -1,47 +1,69 @@
-import React from "react";
+/* eslint-disable */
```
taraepp: I always have to ask why when I see that eslint is disabled for the whole file!
Was it a specific rule? Is it something we think we can clean up in the future that's just too much scope for now?

services/common/src/components/forms/RenderFileUpload.spec.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import { render } from "enzyme";
+import RenderFileUpload from "./RenderFileUpload";
+import { Provider } from "react-redux";
+import { configureStore } from "@reduxjs/toolkit";
+import { sharedReducer } from "@mds/common/redux/reducers/rootReducerShared";
```
taraepp: I would suggest using the redux wrapper instead so that if we ever update the reducer setup or change the test to use an initial state, that we won't have to redo all the imports. Which I guess is only on core/MS right now.

services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -51,179 +73,332 @@ const defaultProps = {
...
+    file: props.file || null,
+    metadata: null,
+    load: null,
+    error: null,
+    progress: null,
+    abort: null,
+  });
+
+  let waterSound;
+  let flushSound;
```
taraepp, line 109: 😂

services/document-manager/backend/app/docman/resources/document_version_list_resource.py
```diff
@@ -58,14 +59,19 @@ def post(self, document_guid):
         new_version = DocumentVersion(
             id=version_guid,
             document_guid=document.document_guid,
-            created_by='mds',
+            created_by=User().get_user_username() or 'mds',
```
taraepp, line 62: oh wow. That was an oversight. Good catch!

services/core-web/src/store/configureStore.ts
```diff
@@ -8,6 +8,14 @@ const getStore = (preloadedState = {}) =>
       ...rootReducer,
       loadingBar: loadingBarReducer,
     },
+    middleware: (getDefaultMiddleware) =>
+      getDefaultMiddleware({
+        // Disable check for immutable state as this has caused a lot of errors with the redux-toolkit introduction
+        // TODO: Consider fixing these issues and re-enabling this check
```
taraepp, line 14: I think that normally this just pops up as a dismiss-able error on development mode. Unless that's not the case I might prefer to keep it in so at least it tells you when you _add_ code that breaks those rules. Environment variables maybe? For the immutable especially, the serializable I think is a lost cause. (This is already an enormous PR so a hypothetical switch to env variables would be another ticket)

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -87,36 +89,86 @@ def initiate_document_upload(cls, document_guid, file_path, folder, file_size, v
...
+        cache.set(
+                OBJECT_STORE_UPLOAD_RESOURCE(
+                    document_guid), object_store_upload_resource,
+                CACHE_TIMEOUT)
+        cache.set(OBJECT_STORE_PATH(document_guid),
+                      object_store_path, CACHE_TIMEOUT)
+                  
+        return object_store_path
 
     @ classmethod
```
taraepp: not anything you did here but I don't think there should be a space in `@classmethod`?

services/common/src/components/forms/RenderFileUpload.tsx
simensma-fresh: Definitely not intended, will remove. Copied it over from `FileUpload.js` and didn't notice 🙃 
image

services/core-web/src/store/configureStore.ts
simensma-fresh: Unfortunately it's shows up as a hard error here 😢 
image

services/common/src/utils/fileUploadHelper.ts
```diff
@@ -0,0 +1,191 @@
...
+      ENVIRONMENT.docManUrl + COMPLETE_MULTIPART_UPLOAD(uploadData.document_manager_guid),
+      payload,
+      createRequestHeader()
+    );
+  }
+
+  private _createMultipartUpload = async (): Promise<MultipartDocumentUpload> => {
+    const headers = createRequestHeader({
+      "Upload-Length": `${this.file.size}`,
+      // "Content-Type": this.file.type,
```
matbusby-fw:
```suggestion
```

Can this go?

matbusby-fw (dismissed):
Nice!  Love all the code comments you included.  Really helps the readability!

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -13,15 +13,41 @@
...
+    doc.status = str(status)
+    db.session.add(doc)
+    db.session.commit()
+
 
 class DocumentUploadHelper:
 
     @classmethod
     def initiate_document_upload(cls, document_guid, file_path, folder, file_size, version_guid=None):
+        folder = secure_filename(folder) if folder else folder
```
taraepp: Is this what's causing the the code scanning check failure below? The else path particularly? I think that's the first time I've seen that check raise an issue!

matbusby-fw (approved):
Nice!  Love all the code comments you included.  Really helps the readability!

services/document-manager/backend/app/docman/utils/document_upload_helper.py
simensma-fresh: Indeed. The `folder` parameter left without `secure_filename` would cause the check to fail - as it tries to create that folder on the filesystem (only for local development). Updated the fallback to `None` - check passes now.

simensma-fresh: That folder logic has been there for ages, so I guess the codescanning was enabled after this was last changed 😓

matbusby-fw (approved):

henryoforeh-dev (approved):

asinn134:
@simensma-fresh is Tusd still used after this change?

simensma-fresh:
@asinn134 Ideally no. I've left it in behind a feature flag in case issues are reported with the new solution, or we've missed a use case for it. Planning on following up with another ticket in a couple of months to verify that it's not receiving traffic since the upgrade, and then decommissioning it.

PR #2867, opened 2024-01-04
[MDS-5438] Implement logging information

henryoforeh-dev:
## Objective 

[MDS-5438](https://bcmines.atlassian.net/browse/MDS-5438)

. Implemented logging in document manager to display Timestamp, Level, Name, Keycloak Client ID, Filename: Line No, OTEL Trace ID and Message
. Removed unused imports.

### DIFF
--- a/services/document-manager/backend/app/__init__.py
+++ b/services/document-manager/backend/app/__init__.py
@@ -1,16 +1,8 @@
-import sys
-import json
 import os
-
 
 from flask import Flask, current_app, request
 from flask_cors import CORS
-from flask_restx import Resource
 from flask_restx.apidoc import apidoc
-from flask_migrate import MigrateCommand
-
-from flask_jwt_oidc.exceptions import AuthError
-from sqlalchemy.exc import SQLAlchemyError
 
 from app.docman.models import *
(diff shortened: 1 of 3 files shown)


simensma-fresh (dismissed):

services/document-manager/backend/app/__init__.py
```diff
@@ -24,12 +16,40 @@
...
-    config = config_object if config_object else Config
+    trace.set_tracer_provider(TracerProvider())
+
+    FlaskInstrumentor().instrument_app(app)
+
+    @app.after_request
+    def log_response_info(response):
+        from datetime import datetime
+        def get_formatted_current_time():
+            return datetime.now().strftime("%d/%b/%Y %H:%M:%S")
```
taraepp: I think that the datetime helper function imported from core-api should probably also go into a helper/utils file, would also prefer to avoid imports within functions.

The existing "helper.py" file is misnamed (it's a class with a specific purpose) but a new one can always be created.

taraepp (approved):

matbusby-fw (approved):
👍🏻

PR #2873, opened 2024-01-08
[MDS-5638] Permits service auth setup

matbusby-fw:
## Objective 

- Added middleware to intercept all routes on the permitting service
- Currently this will only allow access to the newly created service account for core-api
- Gitops PR to provide the necessary env variables following shortly

[MDS-5638](https://bcmines.atlassian.net/browse/MDS-5638)

### DIFF
--- a/bin/setenv.sh
+++ b/bin/setenv.sh
@@ -14,6 +14,7 @@ document-manager/backend
 filesystem-provider
 minespace-web
 tusd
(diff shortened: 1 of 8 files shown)


henryoforeh-dev (approved):
Excellent!

simensma-fresh (approved):

PR #2880, opened 2024-01-10
[MDS-5657] CRR Forms

taraepp:
## Objective 
- add view/edit forms
   - did up view/edit versions of input components that I was using, but didn't really do that for any that I wasn't using
   - but I did get rid of legacy antd even in the ones that aren't being used yet. And I didn't test them.
- clean up some styling issues
- fix issues with submit & validation not working
- added in a getMineById selector to simplify this very common action

[MDS-5657](https://bcmines.atlassian.net/browse/MDS-0000)


services/common/src/components/forms/FormWrapper.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { FC, useEffect } from "react";
+import { Form } from "antd";
+import { compose } from "@reduxjs/toolkit";
+import { connect, useDispatch, useSelector } from "react-redux";
+import { reduxForm, submit, getFormSubmitErrors, InjectedFormProps, ConfigProps } from "redux-form";
+
+export interface IFormContext {
```
taraepp, line 7: I decided to go with Context/Provider instead of redux after all. Why? Because redux-form already creates a form slice, and I thought that duplicating this functionality was more confusing, and I do like how the context/provider is scoped to the component.

services/common/src/components/forms/RenderDate.tsx
```diff
@@ -1,57 +1,85 @@
...
-const defaultProps = {
-  label: "",
-  placeholder: "",
-  disabled: false,
-  showTime: false,
-};
+interface DateInputProps extends BaseInputProps {
+  showTime?: boolean;
+  yearMode?: boolean;
+  disabledDate?: (currentDate) => boolean;
```
taraepp, line 14: merged the RenderYear component into this file and deleted it.

services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -361,44 +362,60 @@ export const FileUpload = (props: FileUploadProps) => {
...
-              const ext = exts && exts.length > 0 && `.${exts.pop()}`;
-              if (ext && ext in props.acceptedFileTypesMap) {
-                type = props.acceptedFileTypesMap[ext];
-              } else {
-                reject(type);
-              }
-            }
-            resolve(type);
-          })
+      <Form.Item
```
taraepp, line 365: Put this in a Form.Item because without giving a Form.Item with a name property, _it doesn't get passed in with values to onSubmit_. Put null checks on input & meta though because tests failed without them.

services/common/src/redux/reducers/mineReducer.ts
```diff
@@ -95,6 +95,8 @@ const mineReducerObject = {
 };
 
 export const getMines = (state: RootState): ItemMap<IMine> => state[MINES].mines as ItemMap<IMine>;
+export const getMineById = (state: RootState, mineGuid: string): IMine =>
+  state[MINES]?.mines[mineGuid] as IMine;
```
taraepp, line 99: I do wonder why we never did this before

services/common/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -113,8 +114,7 @@ export const renderActionsColumn = ({
         <div>
           {items.length > 0 && (
             <Dropdown menu={{ items }} placement="bottomLeft" disabled={isRowSelected}>
-              {/* // TODO: change button classname to something generic */}
-              <Button data-cy="menu-actions-button" className="permit-table-button">
+              <Button data-cy="menu-actions-button" className="actions-dropdown-button ">
```
matbusby-fw, line 117: I love seeing TODOs get TODONE

services/common/src/components/forms/FormWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  isEditMode: boolean;
+  isModal: boolean;
+}
+
+export const { Provider: FormProvider, Consumer: FormConsumer } = React.createContext<IFormContext>(
```
matbusby-fw, line 17: This is really great!

I'm a bit concerned though about making sure everyone knows how to use it properly.  
I much prefer this pattern of wrapping a form with this over using redux form on the export the way we have been doing, but maybe we could add a couple of code comments here to explicitly state that this is the intention and how it will work?
Hopefully that will help ensure that it gets used!

services/common/src/components/forms/RenderAutoSizeField.tsx
```diff
@@ -1,86 +1,84 @@
...
+interface AutoSizeProps extends BaseInputProps {
+  minRows?: number;
+  maximumCharacters?: number;
+}
 
-const RenderAutoSizeField = (props) => {
-  const [remainingChars, setRemainingChars] = useState(props.maximumCharacters);
-  const [value, setValue] = useState(props.input?.value ?? "");
+const RenderAutoSizeField: FC<AutoSizeProps> = ({
+  label = "",
```
matbusby-fw, line 16: 🙌🏻 I much prefer destructuring props like this as well!

services/common/src/redux/reducers/mineReducer.ts
matbusby-fw: Yeah.  I know I've wanted it before!  Nice!

services/common/src/components/forms/FormWrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  );
+};
+
+const mapStateToProps = (_state, ownProps) => ({
+  form: ownProps.name,
+  ...ownProps.reduxFormConfig,
+});
+export default compose(connect(mapStateToProps), reduxForm({}))(FormWrapper as any) as FC<
+  FormWrapperProps
+>;
```
henryoforeh-dev, line 114: I'm wondering, since this wrapper component is a functional one, would it have been possible to use a useSelector hook instead of mapping state to props here?

taraepp: There's some tricky redux logic here (I struggled with it!) that I put in to allow reduxForm to receive a dynamic form name and custom config (ie: you've probably seen touchOnBlur, onSubmitSuccess, etc. All that can be passed in as reduxFormConfig and it'll go into the reduxForm constructor)
ReduxForm actually creates a bunch of components (this screenshot is from a single instance):
image
The mapStateToProps is what I gotta do to pass those values "up" the tree, and so that reduxForm receives it like props.

henryoforeh-dev: Ah! I see why you had to do it this way. Cool cool.

services/common/src/components/reports/ReportFilesTable.tsx
```diff
@@ -26,17 +26,22 @@ export const ReportFilesTable: FC = () => {
   //Dummy data
   const files = [
     {
+      document_manager_guid: "1",
       file_name: "Report 1",
       "file_type ": "PDF",
       created_by: "admin",
     },
   ];
```
henryoforeh-dev, line 34: Is this dummy data a permanent thing ?

henryoforeh-dev (dismissed):
Nice!

services/common/src/components/reports/ReportFilesTable.tsx
taraepp: it shouldn't be. It just bothered me that the rowKey wasn't set to an attribute, so that's the only change I made here. I think we're going separate ticket.

henryoforeh-dev (approved):

matbusby-fw (approved):
Woot! 🙌🏻

PR #2882, opened 2024-01-12
[MDS-5660] - CRR contacts

matbusby-fw:
- Added fields for contacts in CRR report submission
- Added mine manager info in CRR report submission
- Added table to hold CRR contacts
- updated BE to handle new fields

## Objective 

[MDS-5660](https://bcmines.atlassian.net/browse/MDS-5660)

image
image

### DIFF
--- a/services/common/src/components/reports/ReportDetailsForm.tsx
+++ b/services/common/src/components/reports/ReportDetailsForm.tsx
@@ -108,7 +108,7 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
         (relationship) =>
           relationship.mine_party_appt_type_code === MinePartyAppointmentTypeCodeEnum.MMG
       );
-      if (mineManager) {
+      if (currentMineManager) {
         // Since the party relationships don't contain the required party data, fetch the party
         setMineManagerGuid(currentMineManager.party_guid);
         dispatch(fetchPartyById(currentMineManager.party_guid));
--- a/services/common/src/components/reports/ReportDetailsForm.tsx
+++ b/services/common/src/components/reports/ReportDetailsForm.tsx
@@ -1,7 +1,9 @@
-import { Alert, Col, Row, Typography } from "antd";
+import { Alert, Button, Col, Row, Typography } from "antd";
 import React, { FC, ReactNode, useEffect, useState } from "react";
-import { useSelector, useDispatch } from "react-redux";
-import { Field, getFormValues, change } from "redux-form";
+import { useDispatch, useSelector } from "react-redux";
+import { arrayPush, change, Field, FieldArray, getFormValues } from "redux-form";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faTrash } from "@fortawesome/pro-light-svg-icons";
 
 import {
   getDropdownMineReportCategoryOptions,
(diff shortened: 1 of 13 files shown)


services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -195,22 +275,18 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
               label="Report Title and Additional Comment"
               required
               props={{ maximumCharacters: 500, rows: 3 }}
-              placeholder="Include a precise and descriptive title to the report."
               component={RenderAutoSizeField}
               validate={[required, maxLength(500)]}
             />
+            <Typography.Text className="report-instructions">
```
taraepp, line 281: What happens when you click Review & Submit?

If this is the direction that we're wanting to go with the UI, I wonder if it would be worth adding as a prop for the input. I guess CSS would be the other route; the FormWrapper should give different classes for view/edit.

Other than that, LGTM.

matbusby-fw: I'm not sure how ubiquitous those instructions might be.  Might not be the worst idea to have them added to the fields as a prop, but I can't recall seeing designs for this kind of thing elsewhere much.

image
It looks the same on Review and Submit which I think is okay since it just reiterates the warning/info

taraepp: Ehhh they're not on the mockup, so I don't think they're supposed to be on the view mode. The one for the Report Title and Additional Comment seems pretty out of place.

I think it's worth checking with the ux team what their plans are.

matbusby-fw: Good point.  I'll get rid of them on the view mode

henryoforeh-dev (approved):

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -42,6 +88,40 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
     useSelector((state) => getFormValues(FORM.VIEW_EDIT_REPORT)(state)) ?? {};
   const { mine_report_category = "", mine_report_definition_guid = "" } = formValues;
+  const [mineManager, setMineManager] = useState<IParty>();
+  const [mineManagerGuid, setMineManagerGuid] = useState<string>("");
+
+  const partyRelationships: IPartyAppt[] = useSelector((state) => getPartyRelationships(state));
+  const parties = useSelector((state) => getParties(state));
+
+  useEffect(() => {
+    if (!partyRelationships.length) {
```
taraepp, line 98: I really dislike how we default this in the reducer to empty array! (instead of null or undefined)
I am wondering if there would be a scenario where the mine just _doesn't have_ any party relationships. Maybe it wouldn't be realistic for them to submit a report? Are all the party relationships in mine_party_appt? If so, `SELECT * FROM mine_report WHERE mine_guid IN
(SELECT mine_guid FROM mine WHERE mine_guid NOT IN
(SELECT DISTINCT mine_guid FROM mine_party_appt WHERE mine_guid IS NOT NULL));` on prod is suggesting to me that some mines may have issues submitting reports.

matbusby-fw: I don't disagree, but I think this would be better handled in it's own tech debt ticket.  partyRelationships are used in at least 13 different files and the clean up there will likely be somewhat involved.  Goes a bit out of the bounds of the camp site rule I think.

taraepp: Yeah mostly writing that down so that if such a bug surfaces in reports, we'll be expecting it. I'm fine with it being out of scope.

matbusby-fw: I'll write up a ticket for it as well

taraepp (approved):

services/common/src/components/reports/ReportDetailsForm.tsx
taraepp: I didn't get the excuse I was looking for to make a new redux slice, but hopefully that format is better at distinguishing "not loaded" and "no data"! And now that I look at it again, I don't think it's blocking loading, so it probably amounts to nothing. 👍

PR #2885, opened 2024-01-15
[MDS-5583] View Report

taraepp:
## Objective 
- brand new page for the report view since we're not on a modal anymore
- move document table stuff to common so we can use it without making it a prop (there are notes in the code where there were differences between Core & MS versions, but nothing I saw looked major)
- bring over dataMocks and give some mock data to the view form.
- adding in initialValues, dealing with any resulting bugs
- currently just console logging the values on submit

NOTE: report type & report name are a bit broken right now. This is expected and there's another ticket where this should be handled (5662 I think?)

[MDS-5583](https://bcmines.atlassian.net/browse/MDS-5583)


services/common/src/redux/reducers/reportReducer.ts
```diff
@@ -38,6 +46,12 @@ const reportReducerObject = {
 export const getReports = (state) => state[REPORTS].reports;
 export const getReportsPageData = (state) => state[REPORTS].reportsPageData;
 export const getMineReports = (state) => state[REPORTS].mineReports;
+export const getMineReportById = (state, reportGuid) => {
```
henryoforeh-dev, line 49: Is this function used to get the latest report or the first report ?

taraepp: it is meant to get the whole report, including all submissions.

services/minespace-web/src/components/dashboard/mine/reports/ReportPage.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const [isEditMode, setIsEditMode] = useState(false);
+
+  // get form data so we can submit it outside of the form
+  const formErrors = useSelector(getFormSubmitErrors(FORM.VIEW_EDIT_REPORT));
+  const formValues = useSelector(getFormValues(FORM.VIEW_EDIT_REPORT));
+
+  useEffect(() => {
+    if (!mine || mine.mine_guid !== mineGuid) {
+      dispatch(fetchMineRecordById(mineGuid));
+    }
```
henryoforeh-dev, line 35: Can the logic here can be moved to the other useEffect on line 38 since the useEffect gets triggered when either reportGuid or mineGuid changes ?

taraepp: hmm, yeah, I think I can. 👍

henryoforeh-dev (approved):
Nice!

simensma-fresh (approved):
Yayy for more things in the common folder! 🙌

PR #2891, opened 2024-01-17
[FIX] Enabled S3 uploads by default for local development

simensma-fresh:
## Objective 

The file upload speed upgrades broke the file system backed uploads that were enabled by default when developing locally. This PR makes use of Object Storage (S3) the default option, and adds functionality to `make env` to set the access key when run.

TODO at a later time: We can get rid of the old `OBJECT_STORE_ENABLED` logic, as it's no longer in use


henryoforeh-dev (approved):
Do we need to run any command locally to get this working ?

simensma-fresh:
If you already have the Object store env variables set in docman (OBJECT_STORE_ACCESS_KEY / OBJECT_STORE_ACCESS_KEY_ID / OBJECT_STORE_BUCKET) you shouldn't have to

taraepp (approved):
🙌

matbusby-fw (approved):
Oooh!

PR #2892, opened 2024-01-17
[MDS-5732] Report Contacts Bug

taraepp:
## Objective 
- the field name didn't match the data, so I made it consistent and now it works
- made an interface for the contacts while I was at it
  - didn't like that all the report interfaces were in a `mines` directory, so changed it to a `reports` directory
- bonus: navigate to "view" page on report submission, not the reports table

[MDS-5732](https://bcmines.atlassian.net/browse/MDS-5732)


services/common/src/interfaces/reports/mineReportContact.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface IMineReportContact {
+  mine_report_contact_id: string | number;
```
henryoforeh-dev, line 2: Is it possible the the id here could be a string or number ?

taraepp: it came back from the BE as a string of a number ("3"). So I was trying to play it safe as it's not actually implemented anywhere yet.

henryoforeh-dev (approved):

matbusby-fw (approved):
Approvinated!

PR #2893, opened 2024-01-17
[MDS-5662]Update report selection and submission flow

matbusby-fw:
- Added is_common field to mine_report_definition to facilitate the display of common reports
- included the assignment of this variable in the migration to the currently marked common reports
- Added select field to search reports by number or text
- Added functionality to auto-populate the selected report in the submission stage based on the selection in get started
- Removed category based functionality from the frontend (it actually wasn't even being used on the backend so it didn't need to be removed there

## Objective 

[MDS-5662](https://bcmines.atlassian.net/browse/MDS-5661)

image


services/common/src/components/reports/__snapshots__/ReportDetailsForm-view.spec.tsx.snap
```diff
@@ -78,10 +74,9 @@ exports[`ReportDetailsForm renders view mode properly 1`] = `
               <b />
                [TODO: plain language on what it is]
               <br />
-              <b />
-               [TODO: plain language on what it is]
-              <br />
-              <b />
+              <b>
+                undefined.undefined
```
taraepp: 🤨

services/common/src/interfaces/report/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./mineReportDefinition.interface";
```
taraepp: I guess- I called this directory _reports_. Doesn't matter to me, I am fine with either, but you'll probably want to pull now that it's merged and make it the same.

services/common/src/components/reports/ReportGetStarted.tsx
```diff
@@ -1,52 +1,59 @@
...
+    setSelectedReportDefinition(newReport);
+  };
+
+  useEffect(() => {
+    // Filter out common reports and sort alphabetically
+    const commonReportDefinitions = mineReportDefinitionOptions
+      .filter((report) => report.is_common)
+      .sort((a, b) => a.report_name.localeCompare(b.report_name));
+    setCommonReportDefinitionOptions(commonReportDefinitions);
+  }, [mineReportDefinitionOptions]);
```
henryoforeh-dev: The useEffect here and the one on line 26 have the same dependencies that trigger them. Is it possible to merge into one useEffect ?

matbusby-fw: Ah, good catch!

henryoforeh-dev (approved):
Excellent travail

taraepp (approved):

PR #2894, opened 2024-01-17
[MDS-5666] Fixing zipping failure due to duplicated files

isuru-aot:
## Objective 

[MDS-5666](https://bcmines.atlassian.net/browse/MDS-5666)

- Filtered the files that shown in the table view based on the corresponding guids availability
- Fix file guid retrieval in two different sets (core and app) of files to send for zipping.


henryoforeh-dev (approved):

matbusby-fw (approved):
Yeah, I can dig it!

PR #2895, opened 2024-01-17
[MDS-5737] Fixed opening of document viewer from action menu and other Document Viewer issues

simensma-fresh:
## Objective 

[MDS-5737](https://bcmines.atlassian.net/browse/MDS-5737)

Fixed a couple of issues around the document viewer
- The "Open in Document Viewer" action button downloads the document instead of opening in the document viewer in both minespace and core: Second param to `openDocument` is the document name, not guid
- The Document Viewer fails with an error in Minespace: PDF Viewer service url param name was wrong
- Document Viewer doesn't work locally: Newly uploaded files didn't have the S3_PREFIX appended for multipart uploads


matbusby-fw (approved):
📎

henryoforeh-dev (approved):
Nice work!

PR #2901, opened 2024-01-20
[MDS-5285] Adjustments to Core anchor side menu

asinn134:
## Objective 

[MDS-5285](https://bcmines.atlassian.net/browse/MDS-5285)

- Adjusted Core anchor side menu to match figma design
- Moved ScrollSideMenu component to common package
- Adjusted ScrollSideMenu component to accept icons

Screenshot (247)

Screenshot (249)


services/common/src/components/common/ScrollSideMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+ * @constant ScrollSideMenu renders react children with an active indicator if the id is in the url.
+ */
+
+interface ScrollSideMenuProps {
+  menuOptions: ISideMenuOption[];
+  featureUrlRoute: any;
+  featureUrlRouteArguments: (string | number)[];
+  history: any;
+  location: any;
+  match: any;
```
taraepp: Would highly recommend taking out history, location, and match as props, using hooks instead. (useHistory, useLocation, useParams). Easier to parse and work with, and less problems converting parent components to TS. You may be able to simplify a lot of the logic where you're doing null checks, etc, on those values.

asinn134: This has been adjusted to use hooks instead of props now.

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):

services/common/src/components/common/ScrollSideMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+    urlRoute = props.featureUrlRoute(...props.featureUrlRouteArguments, route);
+    if (route === history.location.hash) {
+      return;
+    }
+
+    history.push(urlRoute, { currentActiveLink: route });
+  };
+
+  useEffect(() => {
+    // If the user loads the page with a hash in the URL, start them off at the corresponding feature section.
```
matbusby-fw, line 35: Oh man 😓 .  Nice figuring!

matbusby-fw (approved):
Very nice!

services/common/src/components/common/ScrollSideMenu.tsx
asinn134: Yeah this part is a bit confusing, can't take the credit though! A previous developer had worked on this and figured it out :)

PR #2903, opened 2024-01-22
[MDS-5298] ministry follow up user flow

henryoforeh-dev:
## Objective 

[MDS-5298](https://bcmines.atlassian.net/browse/MDS-5298)

- Moved the current position of some fields.
- Changed header of one of the sections.
- Tested functionality to ensure nothing got broken as a result of moving the fields.

image


services/core-web/src/components/Forms/incidents/IncidentFormMinistryFollowup.tsx
```diff
@@ -150,10 +126,10 @@ const IncidentFormMinistryFollowup: FC<IncidentFormMinistryFollowupProps> = ({
               />
             </Col>
           )}
-          <Col span={24}>
+          {/* <Col span={24}>
```
matbusby-fw: Is there a need for these to be restored at some point?  I'm curious why it's just commented out.

henryoforeh-dev: Missed removing it. I will update.

matbusby-fw (approved):
Cool beans!

asinn134 (approved):

PR #2918, opened 2024-01-25
[MDS-5715] Project Agent

taraepp:
## Objective 
- Project Agent page
- fix bug where form navigation occurs regardless of form errors
- got some syntax errors popping up on other pages on the form, unused imports, when I went to look at them to see how to set up the page, did a little tidy
- at one point I thought I was going to use a checkbox, so I did up the RenderCheckbox. I ended up not using a checkbox, but I did verify that it works, so including that

[MDS-5715](https://bcmines.atlassian.net/browse/MDS-5715)
[MDS-5739](https://bcmines.atlassian.net/browse/MDS-5739)

image

### DIFF
--- a/services/common/src/redux/selectors/projectSelectors.ts
+++ b/services/common/src/redux/selectors/projectSelectors.ts
@@ -1,6 +1,7 @@
 import { createSelector } from "reselect";
 import { isEmpty } from "lodash";
 import * as projectReducer from "../reducers/projectReducer";
+import { IParty } from "../..";
 
 export const {
   getProjectSummary,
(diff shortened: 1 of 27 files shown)


services/common/src/redux/selectors/projectSelectors.ts
```diff
@@ -16,10 +17,18 @@ export const {
...
+    return agent;
+  }
+  return { ...agent, address: agent.address[0] };
+};
+
 export const getFormattedProjectSummary = createSelector(
   [getProjectSummary, getProject],
   (summary, project) => {
-    let formattedSummary = { ...summary, authorizationOptions: [] };
+    const agent = formatProjectSummaryAgent(summary.agent);
```
henryoforeh-dev, line 30: Will this throw an error if agent is null or undefined? I can't remember if strict null checks is enabled in our code base or not.

taraepp: agent doesn't have to be defined, but summary does. I think it must always be defined because it throws no errors on new entries.

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):

matbusby-fw (approved):
🙌🏻

PR #2923, opened 2024-01-26
[MDS-5734] Updating Core node version

asinn134:
## Objective 

[MDS-5734](https://bcmines.atlassian.net/browse/MDS-5734)

- Update Core node version to 20.11.0
- Fixed Core tests errors from the version update


taraepp (approved):

matbusby-fw (approved):
Nice work!  It's a pain having to update these tests, but it's nice to know they're in better shape now!

PR #2925, opened 2024-01-29
[MDS-5768] CORE Code Required Reports - View

taraepp:
## Objective 

[MDS-5768](https://bcmines.atlassian.net/browse/MDS-5768)

- made a wrapper for pages with scroll side menu, do some math for accurate alignment, etc, less duplicated code
- put in the view mode for CRR on Code
   - the edit mode is technically there too, issues with submitting, though
- made mines report table use new actions menu, added "View", but _did not_ update how existing functionality works- edit will still open in modal, and I merely copied over the "Download All" functions and made sure it behaved the same way after.

Form preview (that "success" alert should perhaps be dealt with on the ticket where Updating Status is handled)
image

The end result of scrolling math:
image


henryoforeh-dev (approved):

asinn134 (approved):

matbusby-fw (approved):
Looks great!

PR #2930, opened 2024-02-02
[MDS-5409] Upgraded Core API to Python 3.11

simensma-fresh:
## Objective 

[MDS-5409](https://bcmines.atlassian.net/browse/MDS-5409)

This PR upgrades the CoreAPI python version to 3.11 and dependencies to a version that supports the new python version.

**Changes**
- Upgraded Python version
- Switched from `flask-restplus` to [flask-restx](https://flask-restx.readthedocs.io/en/latest/) as flask-restplus is no longer maintained
- Upgraded SQLAlchemy to version 1.4.51, and Flask + Flask-SQLAlchemy to latest versions
- Switched to using PYJwt for auth validation, as `python-jose` is no longer maintained
- Use `mine_document` as a polymorphic identity for `MineDocument` as SQLAlchemy 1.4 does not support NULL values in the discriminator column.

Note: The SQLAlchemy version upgrade added about 100 warnings about missing `back_populate` or `overlaps` statements on our relationships which are resolved in this PR: https://github.com/bcgov/mds/pull/2931

image

PR #2931, opened 2024-02-02
[MDS-5409] Fixed SQLAlchemy relationship warnings in Core

simensma-fresh:
## Objective 

[MDS-5409](https://bcmines.atlassian.net/browse/MDS-5409)

The SQLAlchemy version upgrade performed in https://github.com/bcgov/mds/pull/2930 highlighted a couple of things with how we're mapping relationships between dependent models. In many cases we have conflicts between relationships so they will write data to the same columns, and we do not specify the intended behavior through `back_populates` or `viewonly` - the issue is explained in the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/14/errors.html#error-qzyx).

This is not a breaking change in SQLAlchemy 1.4, just a new warning highlighting the issue.
This PR resolves the warning (where possible), or silences it where it doesn't cause any issues

### DIFF
--- a/services/core-api/app/api/incidents/models/mine_incident.py
+++ b/services/core-api/app/api/incidents/models/mine_incident.py
@@ -134,7 +134,8 @@ class MineIncident(SoftDeleteMixin, AuditMixin, Base):
         'MineDocument',
         lazy='selectin',
         secondary='mine_incident_document_xref',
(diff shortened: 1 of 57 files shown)


matbusby-fw (approved):
Well that must have been a fun one...

henryoforeh-dev (approved):

taraepp (approved):

PR #2943, opened 2024-02-07
[MDS-5730] Made staging area and building detail polymorphic identity unique

simensma-fresh:
## Objective 

[MDS-5730](https://bcmines.atlassian.net/browse/MDS-5730)

🐉 🐉 🐉 🐉 

With the upgrade to SQLAlchemy 1.4, the Polymorphic identity warning with NoW building, camps, and Staging areas, described in https://bcmines.atlassian.net/browse/MDS-5494 has been upgraded to cause an error instead of just a warning, causing a 400 error when viewing a NoW with any of the entities associated 
image

The issue comes down to Staging Areas, Buildings and Camps all sharing the polymorphic identity "camp", as they have the same activity type.

This PR fixes the error by assigning a unique polymorphic identity to each of them, while still keeping the activity type the same. Whether Staging Areas, Buildings, and Camps all should have the same activity type is a question and rabbit hole for another day.

image

### DIFF
--- a/services/core-api/app/api/now_applications/models/activity_detail/activity_detail_base.py
+++ b/services/core-api/app/api/now_applications/models/activity_detail/activity_detail_base.py
@@ -1,6 +1,7 @@
 from marshmallow import fields
 from sqlalchemy import and_, func
 from sqlalchemy.schema import FetchedValue
+from sqlalchemy.sql import text, case
 from app.extensions import db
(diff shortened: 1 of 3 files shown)


matbusby-fw (approved):
Thanks for writing such a detailed description of the issue!  Nice work!

henryoforeh-dev (approved):

PR #2953, opened 2024-02-15
[MDS-5772] Adding internal comments

isuru-aot:
## Objective 

[MDS-5772](https://bcmines.atlassian.net/browse/MDS-5772)

NOTE: Added some tests but Quality Gate failed for 'bcgov-sonarcloud_mds_common'

Screenshot 2024-02-20 171254

- Migrate` CommentEditor.js, MinistryComment.js, MinistryCommentPanel.js, AuthorizationWrapper.js, styles.ts`  to core/common.
- Didn't convert all to TS


services/common/src/components/comments/CommentEditor.js
```diff
@@ -0,0 +1,103 @@
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import PropTypes from "prop-types";
+import { Form } from "@ant-design/compatible";
+import "@ant-design/compatible/assets/index.css";
```
taraepp: Would take out the old antd Form and legacy stylesheet if possible.

services/common/src/components/comments/__snapshots__/CommentEditor-add.spec.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CommentEditor renders properly 1`] = `null`;
```
taraepp: do we know why this test isn't working?

services/common/src/constants/permissions.ts
```diff
@@ -0,0 +1,26 @@
+export const VIEW_ALL = "role_view";
```
taraepp: I don't love bringing over this file which is kind of redundant, instead of using the USER_ROLES from environment.ts. The Authorization wrapper could be tweaked to use the userHasRole selector instead, which can take values from either.

services/common/src/components/comments/CommentEditor.js
```diff
@@ -0,0 +1,103 @@
...
+              placeholder="Enter your comment here"
+              showCount
+              maxLength={100}
+              onChange={this.handleChange}
+              value={this.state.comment}
+              name="comment"
+            />
+          </Form.Item>
+        )}
+        {// TODO: Hide until Minespace is updated to display comments.
```
taraepp: would just take this out.

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -98,6 +107,7 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
   formButtons,
   handleSubmit,
   currentReportDefinition,
+  createPermission = Permission.EDIT_INCIDENTS,
```
taraepp: Couple things:
- I don't think this needs to be a prop, could just be a constant.
- are we sure that the permission to comment on a report is the same permission to edit incidents?
- createPermission sounds like permission to create the report, maybe createCommentPermission?

services/common/src/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,115 @@
+/* eslint-disable */
```
taraepp: what kind of eslint errors do we get if we take this out?

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -34,6 +34,15 @@ import {
...
 import { getSystemFlag } from "@mds/common/redux/selectors/authenticationSelectors";
+import MinistryCommentPanel from "@mds/common/components/comments/MinistryCommentPanel";
+import { getMineReportComments } from "@mds/common/redux/selectors//reportSelectors";
+import {
+  createMineReportComment,
+  fetchMineReportComments,
+} from "@mds/common/redux/actionCreators/reportCommentActionCreator";
+import AuthorizationWrapper from "@mds/common/wrappers/AuthorizationWrapper";
+import * as Permission from "@mds/common/constants/permissions";
+import * as PermissionEnv from "@mds/common/constants/environment";
```
taraepp: Would suggest just importing USER_ROLES from here instead of the whole file as *. Something like the ProjectLinks file also does not use the permissions file, much simpler, and then that file can be deleted.

services/common/src/components/comments/__snapshots__/CommentEditor-add.spec.tsx.snap
isuru-aot: Updated.

services/common/src/constants/permissions.ts
isuru-aot: Removed this permission.ts and used environment.ts

services/common/src/components/comments/CommentEditor.js
isuru-aot: Removed.

services/common/src/components/reports/ReportDetailsForm.tsx
isuru-aot: Updated this accordingly.

services/common/src/wrappers/AuthorizationWrapper.js
isuru-aot: No errors 😄 Was just moving the file. (Learned the lesson, thanks for highlighting that)

services/common/src/components/reports/ReportDetailsForm.tsx
isuru-aot: Updated.

services/common/src/components/comments/CommentEditor.js
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from "react";
+import { connect } from "react-redux";
+import PropTypes from "prop-types";
+import { Input, Button } from "antd";
+import { getUserAccessData } from "@mds/common/redux/selectors/authenticationSelectors";
```
taraepp: I think you can remove a couple imports here- the getUserAccessData and the connect (mapStateToProps isn't being used, can be removed)

services/common/src/redux/reducers/authenticationReducer.ts
```diff
@@ -87,8 +87,11 @@ export const getSystemFlag = (state: RootState) => state[AUTHENTICATION].systemF
 export const isAuthenticated = (state: RootState) => state[AUTHENTICATION].isAuthenticated;
 export const getUserAccessData = (state: RootState) => state[AUTHENTICATION].userAccessData;
 export const getUserInfo = (state: RootState) => state[AUTHENTICATION].userInfo;
-export const userHasRole = (state: RootState, role: string) =>
-  state[AUTHENTICATION].userAccessData.includes(USER_ROLES[role] ?? role);
+export const userHasRole = (state: RootState, role: string) => {
+  const userAccessData = state[AUTHENTICATION]?.userAccessData;
+  if (!userAccessData) return false;
```
taraepp: I'm wondering if this is added in order to get the snapshot test to run? Check out MineReportTable.spec.tsx- there the userAccessData is populated via initialState and so we're able to return something other than null or empty div.

services/common/src/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,109 @@
...
+      >
+        {React.createElement("span", null, props.children)}
+      </Tooltip>
+    )
+  );
+};
+AuthorizationWrapper.propTypes = propTypes;
+AuthorizationWrapper.defaultProps = defaultProps;
+
+export default connect()(AuthorizationWrapper);
```
taraepp: I don't think that connect() without any parameters is necessary.

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -216,6 +227,42 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
     dispatch(change(FORM.VIEW_EDIT_REPORT, "documents", docs));
   };
 
+  const fetchComments = async () => {
+    setIsLoading(true);
+    await dispatch(fetchMineReportComments(mineGuid, formValues.mine_report_guid));
+    setIsLoading(false);
+  };
+
+  useEffect(() => {
```
taraepp, line 270: can take out this useEffect and default the isLoading status to false- inspect the formValues to grab the comments property- can `fetchLatestReportSubmission` on add comment to have formValues be updated properly.

isuru-aot: I tried to load comment section with formValues.comments initially and then to load it with fetchComments. However, once its assigned it always use the same source. Let's have a call on this please.

isuru-aot: As discussed, updated the backend to use mine report.

services/common/src/components/comments/CommentEditor.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { ReduxWrapper } from "@mds/common/tests/utils/ReduxWrapper";
+import CommentEditor from "./CommentEditor";
+
+describe("CommentEditor", () => {
```
taraepp, line 6: This one needs similar initialValues so that we get a useful snapshot out of it. So long as the addCommentPermission is what's in that userAccessData state it should work.

taraepp (dismissed):
🙌

services/common/src/wrappers/AuthorizationWrapper.js
```diff
@@ -0,0 +1,105 @@
...
+  isMajorMine: undefined,
+  inDevelopment: undefined,
+  inTesting: undefined,
+  permission: undefined,
+  showToolTip: true,
+  userRoles: [],
+};
+
+export const AuthorizationWrapper = (props) => {
+  const inDevCheck =
```
matbusby-fw, line 61: Certainly not part of your ticket, but we should make sure there's a backlog ticket to rework this AuthorizationWrapper.  It's not in line with our new feature flag system.

asinn134 (approved):
nice work!

matbusby-fw (approved):
👍🏻

PR #2955, opened 2024-02-20
[MDS-5771] Adding display of prr or crr report type

asinn134:
## Objective 

[MDS-5771](https://bcmines.atlassian.net/browse/MDS-5771)

-Adding display of prr or crr report type in report page
-Currently should only have Code Required Report selected because the user selection in MDS-5781 hasn't been worked on yet.

Screenshot (263)

### DIFF
--- a/services/common/src/components/reports/ReportDetailsForm.tsx
+++ b/services/common/src/components/reports/ReportDetailsForm.tsx
@@ -1,4 +1,4 @@
-import { Alert, Button, Col, Row, Typography, Radio } from "antd";
+import { Alert, Button, Col, Row, Typography } from "antd";
 import React, { FC, ReactNode, useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { arrayPush, change, Field, FieldArray, getFormValues } from "redux-form";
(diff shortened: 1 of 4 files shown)


services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -228,6 +229,19 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
         initialValues={initialValues}
       >
         <Row gutter={[16, 8]}>
+          {system === SystemFlagEnum.core && (
+            <Col span={24}>
+              <Typography.Paragraph>What is the type of the report?</Typography.Paragraph>
```
taraepp: Would recommend using the RenderRadio button instead of manually making it (and having the typography/required indicator not match...). Not keen on using 1/2 for values, hard to parse when you're debugging.

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -231,15 +242,22 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
         <Row gutter={[16, 8]}>
           {system === SystemFlagEnum.core && (
             <Col span={24}>
-              <Typography.Paragraph>What is the type of the report?</Typography.Paragraph>
-              <Radio.Group
+              <Field
+                name="report_type"
+                id="report_type"
+                required
                 disabled={true}
```
asinn134, line 249: I set disabled to true here because at the moment in Core we don't have the ability for the user to select a permit condition category code if they select PRR, my understanding is that will be worked on in MDS-5781. Should be able to adjust this to use `isEditMode` then

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -131,6 +137,11 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
     }
   }, []);
 
+  useEffect(() => {
+    const reportType = initialValues?.permit_condition_category_code ? "PRR" : "CRR";
+    dispatch(change(FORM.VIEW_EDIT_REPORT, "report_type", reportType));
+  }, [!formValues?.report_type]);
```
asinn134, line 143: I used `!formValues?.report_type` as the `useEffect` condition here because at the moment if a user discards the form changes and we are keeping `report_type` radio buttons disabled, the user cannot make a selection. This should be adjusted in the work done for MDS-5781

services/common/src/components/reports/ReportDetailsForm.tsx
asinn134: Thanks for the recommendation, this should be adjusted now.

henryoforeh-dev (approved):

taraepp (approved):

matbusby-fw (approved):
Nice work!

PR #2963, opened 2024-02-24
[MDS-5773] Adding regulatory authority radio selection in Core report page

asinn134:
## Objective 

[MDS-5773](https://bcmines.atlassian.net/browse/MDS-5773)

-Adding regulatory authority radio selection in Core report page
-Selection is based on cim_or_cpo column in the compliance_article database table

Screenshot (270)


services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -243,6 +249,39 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
+                disabled={true}
+                props={{
+                  isVertical: true,
+                }}
+                label="Who is the report for?"
+                component={RenderRadioButtons}
+                validate={[requiredRadioButton]}
+                customOptions={[
+                  { label: "Chief Permitting Officer", value: "CPO" },
+                  { label: "Chief Inspector of Mines", value: "CIM" },
```
taraepp: hindsight- we should have used CHI and CHP. 😅 
Backend: we can query by those values,
image
Frontend: they're already in Core Static Content (I was looking to see if there was an enum somewhere)
image

henryoforeh-dev: We can confirm with Rebecca. From the last work I did on the data side of this, there are only 3 possible values (CIM, CPO, Both) as specified in the spreadsheet of the a/c (MDS-5472). It now makes sense that the description of those codes are the same as some of the EMLI contact types. I only wonder what the code should be when it is Both.

henryoforeh-dev (dismissed):
Nice work, looks good.

services/common/src/components/reports/ReportDetailsForm.tsx
asinn134: Just leaving a note here, but it was decided by the team to leave it as it is for now.

taraepp (dismissed):

taraepp (approved):

henryoforeh-dev (approved):

PR #2965, opened 2024-02-26
[MDS-5781] PRR Submissions

taraepp:
## Objective 
- also move top "Save Changes" button in CORE into form so that it can submit.

[MDS-5781](https://bcmines.atlassian.net/browse/MDS-5781)

MINESPACE
(Getting Started)
image
(Add Report)
image
CORE
(View)
image
(Edit)
image


services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -62,17 +69,19 @@ const RenderContacts: FC<any> = ({ fields, isEditMode, mineSpaceEdit }) => {
...
-              name={`${contact}.name`}
-              component={RenderField}
-              label={`Contact Name`}
-              placeholder="Enter name"
-              disabled={!canEdit}
-              required
-              validate={[required]}
-            />
-          </Col>
+          {
```
henryoforeh-dev: Is this curly bracket required here ?

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -128,22 +140,37 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
+
   // minespace users are only allowed to add documents
   const mineSpaceEdit =
     system === SystemFlagEnum.ms && initialValues?.mine_report_guid && isEditMode;
 
+  useEffect(() => {
+    if (permit_guid && !permit) {
+      dispatch(fetchPermits(mineGuid));
+    }
+  });
```
henryoforeh-dev: Looks like the dependency array is missing in the useEffect here.

services/common/src/constants/strings.tsx
```diff
@@ -108,6 +109,11 @@ export const MINE_REPORTS_TYPE = {
   tailingsReports: "TAR",
 };
 
+export enum MINE_REPORTS_ENUM {
```
matbusby-fw: We do have the `services/common/src/constants/enums.ts` which might be a better place for this to live.

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -128,22 +140,37 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
+    getDropdownPermitConditionCategoryOptions
+  );
+  const selectedPermitCategory =
+    permit_condition_category_code &&
+    dropdownPermitConditionCategoryOptions.find(
+      (opt) => opt.value === permit_condition_category_code
+    );
+
+  const isCRR = report_type === "CRR";
+  const isPRR = report_type === "PRR";
```
henryoforeh-dev: I saw the Enum defined somewhere in the code. I think it can be used here too.

services/common/src/components/reports/ReportDetailsForm.tsx
taraepp: it's meant to run only when the page is loaded, not run again if the permit is changed (since it does fetchPermits)

matbusby-fw (dismissed):
Looks good!  Doesn't look like there should be too many conflicts with what I've been working with either.  Just one suggestion on moving where that enum is stored. 🙌🏻

services/common/src/components/reports/ReportDetailsForm.tsx
henryoforeh-dev: Sorry my comment should actually have been "Looks like an empty dependency array is missing". I think it will always run anytime the page/component is loaded.

matbusby-fw: Without the dependency array I think it would run on every render, but due to the if statement only fetch if there is a permit guid but no permit.  That should work I think, but could theoretically do it with `permit_guid` and `permit` in the dependency array as which might be the more recommended path.

services/common/src/constants/strings.tsx
taraepp: Yeaah I was debating the location because the MINE_REPORTS_TYPE is defined right above it, but enums is probably better. I think it's going to be a bit gross unless I overhaul the whole thing (not today...) I think I'll make an enum for MINE_REPORTS_TYPE too.

services/common/src/components/reports/ReportDetailsForm.tsx
taraepp: Now that I think of it, the partyRelationships one should probably be combined in this useEffect, it doesn't make sense to run that on every render either.

I open this link so often that I may as well bookmark it: https://stackoverflow.com/questions/58579426/in-useeffect-whats-the-difference-between-providing-no-dependency-array-and-an

taraepp: gasp, I did have it backwards, empty dep array it is!
https://react.dev/reference/react/useEffect#examples-dependencies

If I were to put something in there though, it would actually be the mineGuid. I am not sure it would be possible to change the mine guid without just reloading the page. But permit/permit_guid would cause extra fetches when the permit is changed in CORE.

services/common/src/constants/strings.tsx
matbusby-fw: Yeah, it actually stuck out because I'm working on an reverse of this enum:

```export enum MineReportType {
  "code-required-reports" = "CRR",
  "permit-required-reports" = "PRR",
  "tailings-reports" = "TAR",
}```

Used for routing purposes.

Comment (taraepp) [reply services/common/src/constants/strings.tsx]: I am literally about to commit it, but I went for the "use as few letters as possible" approach and came up with:
```
CRR = "CRR",
PRR = "PRR",
TAR = "TAR",
```

Debatable whether that's a good idea (that is, completely non-descript labels, that will make no sense if the value ever does change).... I suppose I'm putting it right under MINE_REPORT_SUBMISSION_CODES that does the same thing, and above MINE_INCIDENT_DOCUMENT_TYPE_CODE which.... also does the same thing. Hmm. I can see why you want that format for routing purposes.

It almost wants a whole object or something
```
REPORT_TYPES = {
  crr: {code: "CRR", "label: "Code Required Reports", route: "code-required-reports"},
  ...etc
}
```
But it felt like overkill. And then surely there'll be a mockup with "Reports required by the code" or some other non-matchy scenario just to make it useless and/or extremely convoluted. Maybe something to discuss at some point.

Comment (matbusby-fw) [reply services/common/src/constants/strings.tsx]: Hmmm, I'm kind of with you there.  It feels clunky to have 2-3 different strings/enums to handle this one thing.  I think the idea of the object might be the cleanest solution.

**Review by henryoforeh-dev — APPROVED** [review]
Nice work!

**Review by matbusby-fw — APPROVED** [review]

File: services/common/src/components/reports/ReportDetailsForm.tsx [hunk 5]
```diff
@@ -218,80 +241,125 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
+                  required
+                  disabled={true}
+                  props={{
+                    isVertical: true,
+                  }}
+                  label="What is the type of the report?"
+                  component={RenderRadioButtons}
+                  validate={[requiredRadioButton]}
+                  customOptions={[
+                    { label: MINE_REPORTS_ENUM.CRR, value: REPORT_TYPE_CODES.CRR },
```
asinn134, line 323: I should have made the string values in the options be from an enum, thanks for making the change here.

asinn134 (approved):

PR #2966, opened 2024-02-26
[MDS-5807] Fix CRR related issues

henryoforeh-dev:
## Objective 

[MDS-5807](https://bcmines.atlassian.net/browse/MDS-5807)

- Fixed a bug to show / hide help reference link on mine definition reports.
- Minor adjustment to show/hide title when a report either has a long description or not.
- Database migration script to update only compliance articles that have relationships with mine definition reports using the attached excel sheet.


taraepp:
I do think it would probably be worth it to make the data in 'cim_or_cpo' a foreign key to the emli_contact_type table so that we can use the data to contact people with specific roles using our existing data- perhaps verify what the eventual purpose is of the "Regulatory Authority" field, if it is to notify these people, should facilitate that.

henryoforeh-dev:
> I do think it would probably be worth it to make the data in 'cim_or_cpo' a foreign key to the emli_contact_type table so that we can use the data to contact people with specific roles using our existing data- perhaps verify what the eventual purpose is of the "Regulatory Authority" field, if it is to notify these people, should facilitate that.

Decided in tailgate to leave it as it is for now.

taraepp (approved):

asinn134 (approved):

PR #2967, opened 2024-02-26
[MDS-5790] - Core CRR/PRR

matbusby-fw:
- Utlized the new CRR/PRR component in Core (behind feature flag)
- Refactored MineReportInfo to handle both code required reports and permit required reports (The two components were almost identical, so it made more sense to reduce it to a single component.
- Refactored MineReportInfo.js to MineReportInfo.tsx and tranformed to class based component
- Added/adjusted routing in core to account for the use of the new create CRR component, and the amalgamated component for CRR/PRR.

## Objective 

[MDS-5790](https://bcmines.atlassian.net/browse/MDS-5790)

image
image


services/core-web/tsconfig.json
```diff
@@ -10,6 +10,7 @@
       "@mds/common/*": ["../common/src/*"],
       "@/*": ["src/*"]
     },
+    "module": "ESNext",
```
taraepp, line 13: nice. Lazy loading anywhere!

services/common/src/components/reports/ReportDetailsForm-edit.spec.tsx
```diff
@@ -19,6 +19,25 @@ const initialState = {
   },
 };
 
+function mockFunction() {
```
taraepp: does the ReportDetailsForm still need the mock function?

services/core-web/src/components/mine/Reports/MineReportInfo.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+  const handleReportFilterReset = () => {
+    setStateParams({
+      ...defaultParams,
+    });
+    history.replace(routes.MINE_REPORTS.dynamicRoute(mineGuid, reportType, defaultParams));
+  };
+
+  const renderTitle = () => {
+    switch (mine_reports_type) {
```
taraepp: I'm thinking that here, could just index into MINE_REPORTS_ENUM instead of a switch

matbusby-fw: Good call!

services/common/src/components/reports/ReportDetailsForm-edit.spec.tsx
matbusby-fw: Ah, you're right.  That's no longer needed.  With the implementation I had before rebasing it was required, but no longer!

services/core-web/tsconfig.json
matbusby-fw: Thought it might as well be there since it's set on Minespace.

services/common/src/components/reports/ReportDetailsForm-view.spec.tsx
```diff
@@ -23,6 +23,25 @@ const initialState = {
   },
 };
 
+function mockFunction() {
```
taraepp: haha sorry, same component, different test file

matbusby-fw: Yup, removed that as well

taraepp (approved):
👍

henryoforeh-dev (approved):

PR #2971, opened 2024-02-28
[MDS-5816] - CRR/PRR undefined in url

matbusby-fw:
- Fixed undefined value in URL when navigating back from reportSteps to Core CRR page
- Cleaned up navigating between CRR and PRR tabs so that it clears the mineReports before fetching

## Objective 

[MDS-5816](https://bcmines.atlassian.net/browse/MDS-5816)


henryoforeh-dev (approved):
Clean and cool!

isuru-aot (approved):

PR #2973, opened 2024-02-29
[MDS-5790] - Made getting started page for CRR/PRR reinitialize default values

matbusby-fw:
## Objective 

[MDS-5790](https://bcmines.atlassian.net/browse/MDS-5790)

The getting started page for CRRs/PRRs was holding onto the default value from the first way it was accessed and not updating if you navigated away and returned from the other flow.  Set enableReinitialize to true for the form, so this won't happen.


taraepp (approved):
Yaaaaay I'm glad it had a simple solution!

henryoforeh-dev (approved):

PR #2983, opened 2024-03-04
[MDS-5819] Unhandled Requested Report

taraepp:
## Objective 
- make things not blow up when a mine report without any submissions is loaded into new form.
- and be able to submit against it

## Strategy
- at the endpoint where the latest submission is fetched, if there _are_ no submissions, return the mine report instead
   - and give it a brand-new status of "NON" (sorry I was really running out of suitable keys to use, open to suggestions)
      - status shows up in UI as Report Requested 
   - but it is marshalled into the mine_report_submission model so it's got all the same properties, except with NULLs in places like mine_report_submission_guid 
   - a CORE user can save a report with the "NON" status and it will be treated like a requested report, even if there are already submissions against it
- MS user can fill out _most_ fields on report with NON. It is assumed that the CORE user will have put in the report name or permit & permit category, due date, submission year.
   - they can fill out: description comment, submitter name/email, add or remove report contacts 
- couple other things:
   - mine reports already had a calculated status code property, so I returned it instead of having to dig into the submissions each time

## Known Issues
- ~~filtering by status on the _mine_ reports table on CORE works, because it's a FE search, but on the reports page showing all mines, filtering does not work if NON is selected because it a report with no submissions doesn't get a status saved on it and I didn't adjust that query on the BE~~
  - fixed with change to mine_report.py-->@mine_report_status_code.expression
- to submit against a requested report, MS user does have to go to the report --> view --> edit report. Maybe not greatest UX. Reinstalled this button when status is NON and changed text to "Submit Report"

[MDS-5819](https://bcmines.atlassian.net/browse/MDS-5819)
My flow chart:
image
Editing a requested report in MS:
image


henryoforeh-dev (approved):

matbusby-fw (approved):
👍🏻

PR #2984, opened 2024-03-05
[MDS-5792] status codes and banners

henryoforeh-dev:
## Objective 

[MDS-5792](https://bcmines.atlassian.net/browse/MDS-5792)

- Created / Updated new banners to reflect status codes according to design.
- Implemented functionality to update status code for mine report submission.
- Commented code for saving Additional Comment.

image

image

image


taraepp (changes requested):
Updating the submission rather than creating a new one loses the audit trail. I do think that a new submission should be created in this case.

henryoforeh-dev:
> Updating the submission rather than creating a new one loses the audit trail. I do think that a new submission should be created in this case.

I thought creating a new status was out of the scope of this ticket.

taraepp:
> > Updating the submission rather than creating a new one loses the audit trail. I do think that a new submission should be created in this case.
> 
> I thought creating a new status was out of the scope of this ticket.

Can just use the the existing createReportSubmission functions to track the history. Much simpler to implement.

services/core-web/src/tests/components/modalContent/UpdateMineReportSubmissionStatusModal.spec.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { Provider } from "react-redux";
+import configureStore from "redux-mock-store";
+import UpdateMineReportSubmissionStatusModal from "@/components/modalContent/UpdateMineReportSubmissionStatusModal";
+import * as MOCK from "@/tests/mocks/dataMocks";
+import { reduxForm, reducer as formReducer } from "redux-form";
+
+const mockStore = configureStore({
```
taraepp: We do already have the ReduxWrapper component which will configure the store, wrap with provider, etc, just needs that `form` on ln 22 here passed in as initialState.

henryoforeh-dev: Updated

henryoforeh-dev:
> > > Updating the submission rather than creating a new one loses the audit trail. I do think that a new submission should be created in this case.
> > 
> > 
> > I thought creating a new status was out of the scope of this ticket.
> 
> Can just use the the existing createReportSubmission functions to track the history. Much simpler to implement.

Updated as advised.

taraepp (approved):

simensma-fresh (approved):

PR #2990, opened 2024-03-06
[MDS-5792] Update report status.

henryoforeh-dev:
## Objective 

[MDS-5792](https://bcmines.atlassian.net/browse/MDS-5792)

Added missing report status and description.

### DIFF
--- a/services/common/src/redux/utils/helpers.ts
+++ b/services/common/src/redux/utils/helpers.ts
@@ -624,6 +624,9 @@ export const getMineReportStatusDescription = (
 ) => {
   const currentDate = moment().format("YYYY-MM-DD");
   const MINE_REPORT_STATUS_DESCRIPTION_HASH = {
+    [MINE_REPORT_SUBMISSION_CODES.NON]: `The ministry has requested for the report from the proponent through MineSpace. Requested by ${
+      latestSubmission?.update_user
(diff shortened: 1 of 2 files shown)


taraepp (approved):

services/common/src/redux/utils/helpers.ts
```diff
@@ -624,6 +624,9 @@ export const getMineReportStatusDescription = (
 ) => {
   const currentDate = moment().format("YYYY-MM-DD");
   const MINE_REPORT_STATUS_DESCRIPTION_HASH = {
+    [MINE_REPORT_SUBMISSION_CODES.NON]: `The ministry has requested for the report from the proponent through MineSpace. Requested by ${
```
matbusby-fw, line 627: Is this text from the design?  The grammar feels awkward to me.  I would think it should be "The ministry has requested the report from the proponent through Minespace."

But if that's what the designers want, then 🤷🏻

matbusby-fw (approved):
Just left a thought on the grammar in the once sentance, but maybe that's what's wanted.  Otherwise looks great!

services/common/src/redux/utils/helpers.ts
henryoforeh-dev: Yes, I saw it in figma. 
image

PR #2998, opened 2024-03-12
[MDS-5713] Update Project Contacts to match AMS fields

asinn134:
## Objective 

[MDS-5713](https://bcmines.atlassian.net/browse/MDS-5713)

- Adjusted Project Contact model to include address, first_name, and last_name. Also took the data from the Project Contact table's name column and split it between first_name and last_name columns, then removed the name column.

- Adjusted Project Contact section of Minespace where primary and secondary contacts can now enter their address information

- Adjusted Minespace front end of Project Contact section to match figma design

Screenshot (281)

Screenshot (282)


services/common/src/redux/selectors/projectSelectors.ts
```diff
@@ -21,22 +21,38 @@ const formatProjectSummaryParty = (party): IParty => {
...
+const formatProjectContact = (contacts): IProjectContact[] => {
+  if (!contacts) {
+    return contacts;
+  }
+
+  const primaryContact = contacts.filter((contact) => contact.is_primary);
+  const secondaryContacts = contacts.filter((contact) => !contact.is_primary);
+  const formattedContacts = [...primaryContact, ...secondaryContacts].map((contact) => {
+    const address = contact?.address;
+    return { ...contact, address: address?.length > 0 ? address[0] : null };
```
taraepp: for lines 36 and 37, could just set `address = contact?.address[0]` and skip the ternary in 37.

I do wonder if you _need_ to separate out primary and secondary contacts, since they don't seem to be modified differently.

services/core-api/tests/factories.py
```diff
@@ -1266,14 +1266,26 @@ class Params:
...
+    first_name = factory.Faker('first_name')
+    last_name = factory.Faker('last_name')
     is_primary = True
     deleted_ind = False
 
     phone_extension = None
     job_title = None
     company_name = None
 
+    # @factory.post_generation
```
taraepp: got some commented out code left here and for name above!

services/core-web/src/components/mine/Projects/MineProjectTable.js
```diff
@@ -29,7 +31,7 @@ const transformRowData = (projects) => {
       project_id: project.project_id || Strings.EMPTY_FIELD,
       project_name: project.project_title,
       proponent_project_id: project.proponent_project_id || Strings.EMPTY_FIELD,
-      project_contact: contact?.name || Strings.EMPTY_FIELD,
+      project_contact: name || Strings.EMPTY_FIELD,
```
taraepp, line 32: you can see in the snap below- but when it's assigned "" where the field is undefined above, it'll always end up defined as "" and the `EMPTY_FIELD` will never take effect. Could probably skip the defaulting with `|| ""`, and have project be assigned like firstName && last_Name && ${firstName} ${lastName} || EMPTY FIELD. Sometimes those fields want to stay undefined.

services/minespace-web/src/components/Forms/projects/projectSummary/Agent.tsx
```diff
@@ -16,21 +16,20 @@ import {
...
   const isInternational = address_type_code === "INT";
-
-  const provinceOptions = useSelector(getDropdownProvinceOptions);
   // currently no endpoints, etc, for address_type_code
-  const countryOptions = [
-    { value: "CAN", label: "Canada" },
-    { value: "USA", label: "United States" },
-    { value: "INT", label: "International" },
-  ];
+  const { countryOptions } = props;
```
taraepp: hmm I'm thinking this would make sense to put in constants or enums instead of passing as props.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectContacts.tsx
```diff
@@ -0,0 +1,284 @@
...
+    <>
+      <Typography.Title level={3}>Project Contacts</Typography.Title>
+      <FieldArray name="contacts" countryOptions={props.countryOptions} component={contacts} />
+    </>
+  );
+};
+
+const selector = formValueSelector(FORM.ADD_EDIT_PROJECT_SUMMARY);
+const mapStateToProps = (state) => ({
+  contacts: selector(state, "contacts"),
```
taraepp: I see that above you've got contacts defined with a selector, so this one can go (and the selector on 271), and there's also dispatch defined from `useDispatch`.
So you can take out the contacts here because it's coming from the selector, and you could get rid of the mapDispatchToProps (and also take out arrayPush from props) by doing a dispatch(arrayPush(...)) instead. What the mapDispatchToProps does is have it automatically be dispatched as part of the function. Kind of confusing to do both though.

Then you'd be able to do export default ProjectContacts.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectContacts.tsx
```diff
@@ -0,0 +1,263 @@
...
+                      </Button>
+                    </Popconfirm>
+                  </Col>
+                </Row>
+              </>
+            )}
+            <Row gutter={16}>
+              <Col md={12} sm={24}>
+                <Field
+                  name={`${field}.first_name`}
```
asinn134, line 87: Other then the Typescript change, the other main changes in this file are:

- Adding first_name field (line 87)
- Adding last_name field (line 96)
- Adding address fields. (line 157 - 218)
- Adding the handleClearProvince function (line 28)
- Removal of mapDispatchToProps  and mapStateToProps

services/common/src/redux/selectors/projectSelectors.ts
asinn134: The ternary is changed now to the suggestion. 

Ah the reason I separated out the primary and secondary was in the project contacts section on the front end sometimes the primary contact data ends up being displayed as a secondary contact and a secondary contact ends up being displayed as the primary contact.

services/core-api/tests/factories.py
asinn134: This is removed now.

services/core-web/src/components/mine/Projects/MineProjectTable.js
asinn134: This has been adjusted now in several places to use a structure along the lines of `[primaryContact?.first_name, primaryContact?.last_name].join(" ").trim();`. Thanks for the suggestion, it is a lot cleaner now I think.

services/minespace-web/src/components/Forms/projects/projectSummary/Agent.tsx
asinn134: CountryOptions has been moved to the strings file as a constant now.

services/minespace-web/src/components/Forms/projects/projectSummary/ProjectContacts.tsx
asinn134: Thanks, I removed the mapDispatchToProps, the contacts selectors and went with dispatch(arrayPush(..)) now.

migrations/sql/V2024.03.11.10.00__add_project_contact_first_last_name_column_and_data.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE project_contact
+    ADD COLUMN IF NOT EXISTS first_name varchar,
+    ADD COLUMN IF NOT EXISTS last_name varchar;
```
henryoforeh-dev: first_name and last_name may require a character length.

services/common/src/constants/strings.tsx
```diff
@@ -414,3 +414,9 @@ export const MINE_REPORT_STATUS_HASH = {
 };
 
 export const VC_ACTIVE_CONNECTION_STATES = ["credential_acked", "credential-issued", "done"];
+
+export const CONTACTS_COUNTRY_OPTIONS = [
+  { value: "CAN", label: "Canada" },
+  { value: "USA", label: "United States" },
+  { value: "INT", label: "International" },
+];
```
henryoforeh-dev, line 422: Nice. work! I may just use this after merging it.

migrations/sql/V2024.03.11.10.00__add_project_contact_first_last_name_column_and_data.sql
asinn134: Sounds good, they have been adjusted to have character length 200 now.

henryoforeh-dev (approved):
Nice!

taraepp (approved):

PR #3006, opened 2024-03-15
[MDS-5834] - Fixed state updates after/during revocation

matbusby-fw:
## Objective 

Made sure they revocation modal showed a proper loading indicator while revoking and fixed state update after fulfilled.

[MDS-5834](https://bcmines.atlassian.net/browse/MDS-5834)

### DIFF
--- a/services/common/src/redux/slices/verifiableCredentialsSlice.ts
+++ b/services/common/src/redux/slices/verifiableCredentialsSlice.ts
@@ -29,82 +29,89 @@ const initialState: VerifiableCredentialsState = {
   verifiableCredentialConnections: [],
 };
 
-const reducers = (create) => ({
-  fetchCredentialConnections: create.asyncThunk(
-    async (payload: { partyGuid: string }, thunkAPI) => {
-      const headers = createRequestHeader();
-      thunkAPI.dispatch(showLoading());
-      const { partyGuid } = payload;
-
-      const response = await CustomAxios({
-        errorToastMessage: "default",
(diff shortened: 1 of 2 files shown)


services/common/src/redux/slices/verifiableCredentialsSlice.ts
```diff
@@ -29,82 +29,89 @@ const initialState: VerifiableCredentialsState = {
   verifiableCredentialConnections: [],
 };
 
-const reducers = (create) => ({
```
matbusby-fw, line 32: There were some typing errors here if reducers were defined outside of the createAppSlice and then passed in.  Defining them inside fixed that 🤷🏻 .

services/common/src/redux/slices/verifiableCredentialsSlice.ts
```diff
@@ -29,82 +29,89 @@ const initialState: VerifiableCredentialsState = {
...
-});
-
-const selectors = {
-  getCredentialConnections: (state): VerifiableCredentialsConnection[] => {
-    return state.verifiableCredentialConnections;
+      {
+        rejected: (state: VerifiableCredentialsState, action) => {
+          rejectHandler(action);
+        },
+        fulfilled: (state, action) => {
```
matbusby-fw, line 91: Other than moving things around, this is the only real change to this file

henryoforeh-dev (approved):
Nice!

asinn134 (approved):

PR #3008, opened 2024-03-15
[MDS-5831] Add nearest municipality to project application

matbusby-fw:
## Objective 

- Added municipality table to the backend and seeded it with data
- updated core_static_content to return municipalityoptions from the backend
- updated project summary to include a municipality guid
- added select to the LegalLandOwnerInformation tab of project summary in Minespace

[MDS-5831](https://bcmines.atlassian.net/browse/MDS-5831)

image


henryoforeh-dev (approved):
Nice work!

taraepp (approved):

PR #3011, opened 2024-03-18
[MDS-5863] Adjustment to AMS field verification

asinn134:
## Objective 

[MDS-5863](https://bcmines.atlassian.net/browse/MDS-5863)

- Adjustment to Minespace project description field verification so when feature flag is disabled sections that only appear with feature flag are not required.


matbusby-fw (approved):
🙌🏻

henryoforeh-dev (approved):

PR #3018, opened 2024-03-21
[FIX] Fixed capitalization of permit association proxy

simensma-fresh:
## Objective 

Fixes the capitalization for the `vc_locked association_proxy` as it causes an error when looking up a NoW.

image

image


henryoforeh-dev (approved):
Nice

asinn134 (approved):

PR #3019, opened 2024-03-21
[MDS-5857] Duplicate esup certificate magazine info

asinn134:
## Objective 

[MDS-5857](https://bcmines.atlassian.net/browse/MDS-5857)

- Was found that the magazine that just had it's deleted_ind set to True would show up inside the generated esup certificate.


henryoforeh-dev (approved):

matbusby-fw (approved):
Nice!

PR #3021, opened 2024-03-22
[MDS-5867] Update "Back to Reports" link in Core

matbusby-fw:
The "Back to Reports" link when viewing a report in Core was directing the user to the global reports list rather than back to the reports for that particular mine.  Updated the link.

## Objective 

[MDS-5867](https://bcmines.atlassian.net/browse/MDS-5867)


henryoforeh-dev (approved):
Nice!

asinn134 (approved):

PR #3024, opened 2024-03-25
[MDS-5658] Fixing compliance_article issue + changing log levels

isuru-aot:
## Objective 

Adding info level logs to see recipients for email notifications

### DIFF
--- a/services/core-api/app/api/mines/reports/models/mine_report.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report.py
@@ -129,9 +129,7 @@ class MineReport(SoftDeleteMixin, AuditMixin, Base):
     def __repr__(self):
         return '<MineReport %r>' % self.mine_report_guid
 
-    def send_report_update_email(self, is_edit, is_proponent, crr_or_prr):
-
-        report_code = "code" if crr_or_prr == "CRR" else "permit"
+    def send_crr_add_notification_email(self, is_edit, is_proponent):
         subject = f'Code Required Report Submitted for mine {self.mine.mine_name}'
         core_recipients = [MDS_EMAIL]
         ms_recipients = []
@@ -140,7 +138,7 @@ class MineReport(SoftDeleteMixin, AuditMixin, Base):
             core_recipients, ms_recipients = self.collectRecipients(is_proponent)
             core_recipients.extend(self.getReportSpecificEmailsByReportType())
 
-            core_report_page_link =  f'{Config.CORE_PRODUCTION_URL}/mine-dashboard/{self.mine.mine_guid}/required-reports/{report_code}-required-reports'
+            core_report_page_link =  f'{Config.CORE_PRODUCTION_URL}/mine-dashboard/{self.mine.mine_guid}/required-reports/code-required-reports'
             ms_report_page_link = f'{Config.MINESPACE_PRODUCTION_URL}/mines/{self.mine.mine_guid}/reports'
             c_article = self.mine_report_definition.compliance_articles[0]
             report_type = f'{c_article.section}.{c_article.sub_section}.{c_article.paragraph} - {self.mine.mine_name}'
@@ -175,7 +173,7 @@ class MineReport(SoftDeleteMixin, AuditMixin, Base):
             ms_email_body = open("app/templates/email/report/ms_new_report_submitted_email.html", "r").read()
             EmailService.send_template_email(subject, ms_recipients, ms_email_body, email_context, cc=None)
 
-        else:
+    def send_report_update_email(self, is_edit):
             recipients = [self.mine.region.regional_contact_office.email, MDS_EMAIL]
             if self.mine.major_mine_ind:
                 recipients = [MAJOR_MINES_OFFICE_EMAIL, MDS_EMAIL]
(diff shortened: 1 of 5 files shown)


matbusby-fw (dismissed):

services/core-api/app/api/services/email_service.py
```diff
@@ -192,6 +192,9 @@ def send_email(cls,
             current_app.logger.error(message)
             return
 
+        current_app.logger.info(
```
taraepp, line 195: Do we need _both_ debug and info?

matbusby-fw (dismissed):

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -129,66 +128,58 @@ def mine_report_status_description(cls):
...
+              "report_compliance_year": self.submission_year,
+              "report_due_date": due_date,
+              "report_recieved_date": (self.received_date).strftime("%b %d %Y at %I:%M %p"),
+          },
+          "minespace_login_link": Config.MINESPACE_PRODUCTION_URL,
+          "core_report_page_link": core_report_page_link,
+          "ms_report_page_link": ms_report_page_link
+        }
+
+        trigger_notification(f'Your ({self.mine_report_definition_report_name}) report has been recieved',
```
taraepp, line 174: I think this should be different text for CORE recipients

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -221,9 +211,9 @@ def collectRecipients(self, is_proponent):
 
         return core_recipients, ms_recipients
 
-    def getReportSpecificEmailsByReportType(self):
-        art = self.mine_report_definition.compliance_articles[0]
-        notificaiton_list = MineReportNotification.find_contact_by_compliance_article(art.section, art.sub_section, art.paragraph, art.sub_paragraph)
+    def getReportSpecificEmailsByReportType(self, compliance_details):
```
taraepp, line 218: not a big deal, just a minor style point- but just to keep in mind for the future, snake_case for python rather than pascalCase for function & variable names.

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -232,14 +222,29 @@ def getReportSpecificEmailsByReportType(self):
...
+            if ntf[1] and self.mine.major_mine_ind and PERM_RECL_EMAIL not in unique_recipients:
                 unique_recipients.add(PERM_RECL_EMAIL)
 
-            if ntf[2] and regional_email not in unique_recipients:
+            if ntf[2] and not self.mine.major_mine_ind and regional_email not in unique_recipients:
                 unique_recipients.add(regional_email)
 
         return list(unique_recipients)
 
+    def send_report_update_email(self, is_edit):
```
taraepp: would maybe rename this to send_crr_report_update_email for clarity

isuru-aot: This was an existing method. And this is being used in `..resources/mine_report.py`

services/core-api/app/api/services/email_service.py
isuru-aot: Changed from debug to info. And when we are using the templates its not possible to add the recipients list to the email body (for testing except prod). Hence just changed the existing debug log to in info.

services/core-api/app/api/mines/reports/models/mine_report.py
isuru-aot: updated

services/core-api/app/api/mines/reports/models/mine_report.py
taraepp: won't it get called for both CRRs and PRRs from the resource file even though it's only for CRRs? Should the resource file actually be calling `send_crr_and_prr_add_notification_email` instead? Perhaps `send_crr_and_prr_add_notification_email` should actually be changed back to `send_report_update_email` and then this function should be named to reflect its purpose. But if it's only for CRRs then the name should reflect that and references can be updated.

isuru-aot: Well, it was an exiting function, i think as name says its for report update. This has has been calling from ../resource/mine_report.py.

To make things clearer, this pr is for addition of crr or prr. So I didn't want to change an existing name to confuse things and separate logics inside with another if. Instead introduce this new function with a clearer name.

I think this way is better than putting every thing in to one function and separate things inside using another if.

taraepp: This function will break if a PRR report is passed to it and so it is specifically for CRRs. I think it's better to change the name of the function. Having an inaccurate name that does not reflect what the function does is confusing.

isuru-aot: @taraepp i think its easy to get into a call and discuss, waiting in discord.

`send_report_update_email` function is there from 1.5 tag (2021). And its been using in other places, which is not part of this ticket (5658). And also I used the same method thinking the ticket is only about CRR initially. Then after clarifications PRR was included as well, hence I introduced this separate method `send_crr_and_prr_add_notification_email` to make things clearer.
But if you think both these needs to merge, I suggest to do that in a separate ticket as `..resources/mine_reports.post()` is not part of this ticket, which needs to be tested if we merge these.

taraepp (approved):

henryoforeh-dev (approved):

PR #3025, opened 2024-03-25
[MDS-5833] allow reissuance after revocation

Jsyro:
## Objective 

[MDS-5833](https://bcmines.atlassian.net/browse/MDS-5833)

Active and valid
image

Immediately after revoked (not available for re-issuance)
image

Revoked but made available to second issuance
image

Other Changes: 
Objects modified by webhooks have a timestamp of the webhooks. this will mean if packets are received in the wrong order a newer state will not be overriden by an older state because the packets arrived in the reverse order.


henryoforeh-dev (dismissed):

henryoforeh-dev (dismissed):

services/common/src/redux/slices/verifiableCredentialsSlice.ts
```diff
@@ -4,6 +4,9 @@ import CustomAxios from "@mds/common/redux/customAxios";
 import { ENVIRONMENT, MINES_ACT_PERMITS_VC_LIST } from "@mds/common/constants";
 import * as API from "@mds/common/constants/API";
 import { RootState } from "@mds/common/redux/rootState";
+import { DataSourceItemType } from "antd/lib/auto-complete";
+import { debug } from "webpack";
+import { a } from "@mds/common/tests/mocks/dataMocks";
```
taraepp: What is 'a'? And is this + debug something that's no longer being used? I would guess it might cause issues with variables on `sort(a,b)`

henryoforeh-dev (approved):

services/common/src/redux/slices/verifiableCredentialsSlice.ts
Jsyro: removed

taraepp (approved):

PR #3027, opened 2024-03-26
[MDS-5658] PRR Issue Fixe

isuru-aot:
## Objective 

- Fix the notification link
- Fix the MS report name missing issue.

### DIFF
--- a/services/core-web/src/constants/routes.ts
+++ b/services/core-web/src/constants/routes.ts
@@ -355,13 +355,6 @@ export const REPORT_VIEW_EDIT = {
   component: ReportPage,
 };
 
-export const MINE_REPORT_VIEW_EDIT = {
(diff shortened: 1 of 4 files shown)


services/core-web/src/constants/routes.ts
```diff
@@ -355,6 +355,13 @@ export const REPORT_VIEW_EDIT = {
   component: ReportPage,
 };
 
+export const MINE_REPORT_VIEW_EDIT = {
```
taraepp: Is there a reason not to use the REPORT_VIEW_EDIT route above?

isuru-aot: oh reverted that. Can use the REPORT_VIEW_EDIT for it.

taraepp (approved):

asinn134 (approved):

PR #3028, opened 2024-03-27
[MDS-5658] Fixing previous mine manager issue

isuru-aot:
## Objective 

Below is a screenshot of PROD data (Second row is the valid manager), where mine-managers are exist without end_date + without status. Hence raising this PR to get the correct mine manager.
Screenshot 2024-03-26 220128

### DIFF
--- a/services/core-api/app/api/mines/reports/models/mine_report.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report.py
@@ -1,4 +1,5 @@
 import uuid
+from datetime import datetime, timezone
 
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.schema import FetchedValue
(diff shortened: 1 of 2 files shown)


services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -1,4 +1,5 @@
 import uuid
+from datetime import datetime, timezone
```
taraepp: minor, but are these new imports being used?

matbusby-fw: Ah, good catch.  I love cleaning up unused imports!  Otherwise this looks great Isuru!

isuru-aot: Removed unused imports :)

matbusby-fw (approved):

taraepp (approved):

PR #3032, opened 2024-04-01
[MDS-5860] Fix inconsistent toast error message from api calls

matbusby-fw:
## Objective 

Errors that were being thrown by the frontend were being overridden when an action creator contained `throw new Error()`, but it's invocation didn't contain a `catch` block.

Did a search throughout the frontends and only found a very limited number of catch blocks, so removed the throw new Error from all actionCreators where this didn't exist.  

[MDS-5860](https://bcmines.atlassian.net/browse/MDS-5860)

image

### DIFF
--- a/services/core-api/app/api/notice_of_departure/resources/notice_of_departure_list.py
+++ b/services/core-api/app/api/notice_of_departure/resources/notice_of_departure_list.py
@@ -1,6 +1,6 @@
 from app.api.activity.models.activity_notification import ActivityType
 from flask_restx import Resource, reqparse, inputs
-from werkzeug.exceptions import NotFound
+from werkzeug.exceptions import NotFound, BadRequest
 from app.extensions import api
 from app.api.utils.resources_mixins import UserMixin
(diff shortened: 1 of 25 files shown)


henryoforeh-dev (dismissed):
Clean clean!

asinn134 (dismissed):
Nice catch

services/core-api/app/api/notice_of_departure/resources/notice_of_departure_list.py
```diff
@@ -69,6 +69,7 @@ def get(self):
     @api.expect(CREATE_NOD_MODEL)
     @api.marshal_with(NOD_MODEL, code=201)
     def post(self):
+        raise BadRequest('TEST this has been a bad request')
```
taraepp: I have a feeling that this wasn't intended to be merged?

taraepp: Other than that LGTM

henryoforeh-dev (approved):

asinn134 (approved):

PR #3036, opened 2024-04-04
[MDS-5712] AMS Application Purpose

taraepp:
## Objective 
- Authorizations Involved page on MS
- change data shape- make necessary updates on CORE

[MDS-5712](https://bcmines.atlassian.net/browse/MDS-5712)

This is what it looks like when one of the required fields is missing on the BE:
image

Tech debt:
- did not update BE tests
- styling is a bit off
- CORE has an edit and a view mode- there is no view mode implemented for authorizations involved


services/core-api/app/api/projects/project_summary/models/project_summary_authorization.py
```diff
@@ -18,26 +22,191 @@ class ProjectSummaryAuthorization(SoftDeleteMixin, AuditMixin, Base):
...
+            if permit_type[0] == 'AMENDMENT':
+                v = Validator(ams_amendment_schema, purge_unknown=True)
+            elif permit_type[0] == 'NEW':
+                v = Validator(ams_new_schema, purge_unknown=True)
+            else:
+                v = Validator(common_ams_schema, purge_unknown=True)        
+        elif authorization.get('project_summary_authorization_type') == 'OTHER':
+            v = Validator(other_schema, purge_unknown=True)
+
+        if not v.validate(authorization):
```
taraepp, line 164: The errors aren't terribly _pretty_ (by default...) but they get the job done. I found it easier to use this than to do all the `if`'s and `len`'s and `not in`'s I'd be doing otherwise. It's not specific about which authorization is causing the issue, for the reason that I didn't want to cause a 500 trying to get an attribute that may be causing the 400, so just logged the whole record on the BE if it's invalid.

Documentation: https://docs.python-cerberus.org/validation-rules.html

services/common/src/components/forms/RenderHiddenField.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { FC, useState, useEffect } from "react";
+import { Input, Form } from "antd";
+import { BaseInputProps, getFormItemLabel } from "./BaseInput";
+
+/**
+ * Just can't conform the UI to play nice with redux-form?
```
matbusby-fw, line 6: 😆 Love it.  Reminds me of an infomercial. "Has this ever happened to you!?"

henryoforeh-dev (approved):
lol the PR needs sizing for reviewing too. Nice work.

matbusby-fw (approved):
Thanks some fancy dancy SQL.  Nice work!

migrations/sql/V2024.03.23.1.46__ams_authorizations.sql
```diff
@@ -0,0 +1,87 @@
...
+    authorization record;
+    auth_no text;
+    auth_array text[];
+    i integer;
+    n integer;
+
+    other_types text[];
+    type_count integer;
+    delete_guids uuid[];
+BEGIN
```
asinn134, line 28: Nice work with the sql scrip here 👍

asinn134 (approved):
A lot of work done, great job!

PR #3037, opened 2024-04-04
[MDS-5860] added detailed message to network error toasts

matbusby-fw:
## Objective 

[MDS-5860](https://bcmines.atlassian.net/browse/MDS-5860)

Add the detailed error message to the toast message that pops up on errors returned from the API

### DIFF
--- a/services/common/src/redux/customAxios.js
+++ b/services/common/src/redux/customAxios.js
@@ -3,9 +3,9 @@ import { notification, Button } from "antd";
 import * as String from "@mds/common/constants/strings";
 import React from "react";
 import * as API from "@mds/common/constants/API";
-import { ENVIRONMENT } from "@mds/common";
+import { ENVIRONMENT } from "@mds/common/constants";
 import { createRequestHeader } from "./utils/RequestHeaders";
-import { Feature, isFeatureEnabled } from "@mds/common";
+import { Feature, isFeatureEnabled } from "@mds/common/utils";
 
 // https://stackoverflow.com/questions/39696007/axios-with-promise-prototype-finally-doesnt-work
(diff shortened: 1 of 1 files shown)


services/common/src/redux/customAxios.js
```diff
@@ -3,9 +3,9 @@ import { notification, Button } from "antd";
 import * as String from "@mds/common/constants/strings";
 import React from "react";
 import * as API from "@mds/common/constants/API";
-import { ENVIRONMENT } from "@mds/common";
+import { ENVIRONMENT } from "@mds/common/constants";
```
taraepp, line 6: it's a bit funny to me to see your one-man mission to make imports better. 😂 
Keep fighting the good fight!

taraepp (approved):

henryoforeh-dev (approved):

services/common/src/redux/customAxios.js
matbusby-fw: Haha, my IDE complains about it, and I hate seeing red lines

isuru-aot (approved):

services/common/src/redux/customAxios.js
taraepp: ahhhh mine doesn't, so I assumed it was preference. But mine did start complaining about unused imports in python recently, which I have been enjoying. 🪓

matbusby-fw: It is satisfying to do these little cleanups.  With the common repo there's also the fact that there are many variables with identical names in different folders (legacy from before the common repo).  It all seems to mostly work, but it has the potential to break without more specific imports.

PR #3038, opened 2024-04-04
[MDS-5895] Setting report email links with right env

asinn134:
## Objective 

[MDS-5895](https://bcmines.atlassian.net/browse/MDS-5895)

-Report email links were always pointing to production regardless of environment, just adding adjustment
to point to the url depending on environment.

### DIFF
--- a/services/core-api/app/api/incidents/models/mine_incident.py
+++ b/services/core-api/app/api/incidents/models/mine_incident.py
@@ -299,7 +299,7 @@ class MineIncident(SoftDeleteMixin, AuditMixin, Base):
                 "mine_name": self.mine_table.mine_name,
                 "mine_no": self.mine_table.mine_no,
             },
-            "incident_link": f'{Config.CORE_PRODUCTION_URL}/mines/{self.mine.mine_guid}/incidents/{self.mine_incident_guid}',
+            "incident_link": f'{Config.CORE_PROD_URL}/mines/{self.mine.mine_guid}/incidents/{self.mine_incident_guid}',
         }
 
         minespace_context = {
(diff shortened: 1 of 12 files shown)


matbusby-fw (dismissed):

isuru-aot (dismissed):

services/core-api/app/config.py
```diff
@@ -119,8 +119,22 @@ class Config(object):
...
+                                         'http://localhost:3000')
+    
     MINESPACE_PRODUCTION_URL = os.environ.get('MINESPACE_PRODUCTION_URL',
                                               'https://minespace.gov.bc.ca')
+    MINESPACE_TEST_URL = os.environ.get('MINESPACE_TEST_URL',
+                                              'https://minespace-test.apps.silver.devops.gov.bc.ca')
+    MINESPACE_DEV_URL = os.environ.get('MINESPACE_DEV_URL',
+                                              'https://minespace-dev.apps.silver.devops.gov.bc.ca')
+    MINESPACE_LOCAL_URL = os.environ.get('MINESPACE_LOCAL_URL',
+                                              'http://localhost:3020')
```
taraepp, line 136: would prefer a function or two in here (or some utils file, whatever) like "get_current_env_ms/core_url" that returns the correct URL over doing if-statements where it's consumed. More reusable, or setting a variable accordingly.

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -310,7 +310,7 @@ def send_incidents_email(self):
                 "mine_name": self.mine_table.mine_name,
                 "mine_no": self.mine_table.mine_no,
             },
-            "minespace_incident_link": f'{Config.MINESPACE_PRODUCTION_URL}/mines/{self.mine.mine_guid}/incidents/{self.mine_incident_guid}',
```
asinn134, line 313: There were a bunch of other places that had links setup with production urls, I wasn't too sure if they need to be adjusted so I left them alone for now.

services/core-api/app/config.py
asinn134: Thanks, it makes more sense to do it this way. I added a new function in the helpers file to retrieve the env url now.

isuru-aot (approved):

henryoforeh-dev (approved):

taraepp (approved):

PR #3039, opened 2024-04-05
[MDS-5872] - get/view cred exch data from traction

matbusby-fw:
## Objective 

- Fetch the credential exchange data from Traction
- Allow it to be viewed in a modal while inspecting a digital credential in Core

[MDS-5872](https://bcmines.atlassian.net/browse/MDS-5872)

image

### DIFF
--- a/services/core-web/src/components/mine/DigitalPermitCredential/CredentialContentModal.tsx
+++ b/services/core-web/src/components/mine/DigitalPermitCredential/CredentialContentModal.tsx
@@ -1,16 +1,29 @@
-import { Col, Row, Typography } from "antd";
-import React, { useEffect } from "react";
+import React, { FC, useEffect, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import {
   fetchCredentialExchangeDetails,
   getCredentialExchangeDetails,
 } from "@mds/common/redux/slices/verifiableCredentialsSlice";
+import { ICredentialExchange, IMine } from "@mds/common";
+import DigitalCredentialDetails from "@/components/mine/DigitalPermitCredential/DigitalCredentialDetails";
 
-const CredentialContentModal = ({ partyGuid, credExchId }) => {
-  const { Paragraph } = Typography;
+interface CredentialContentModalProps {
+  partyGuid: string;
+  credExchId: string;
+  mine: IMine;
+}
+
+const CredentialContentModal: FC<CredentialContentModalProps> = ({
+  partyGuid,
+  credExchId,
+  mine,
+}) => {
+  const [details, setDetails] = useState();
   const dispatch = useDispatch();
-  const credentialExchangeDetails: any[] = useSelector(getCredentialExchangeDetails);
-  const credentialExchangeDetail: any = credentialExchangeDetails.find(
+  const credentialExchangeDetails: ICredentialExchange[] = useSelector(
+    getCredentialExchangeDetails
+  );
+  const credentialExchangeDetail: ICredentialExchange = credentialExchangeDetails.find(
     (ced) => ced.credential_exchange_id === credExchId
   );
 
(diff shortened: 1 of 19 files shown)


services/core-web/src/components/mine/DigitalPermitCredential/ViewDigitalPermitCredential.tsx
```diff
@@ -60,72 +38,51 @@ export const ViewDigitalPermitCredential: FC = () => {
...
   const permitHistoryColumns = [
-    renderDateColumn("issue_date", "Issued"),
-    renderDateColumn("authorization_end_date", "Expiry"),
     {
-      key: "current_permittee_digital_wallet_connection_state",
+      title: "Credential Type",
+      render: () => "AnonCreds",
+    },
+    renderDateColumn("last_webhook_timestamp", "Last Updated"),
+    // renderTextColumn("cred_exch_state", "Status"),
```
henryoforeh-dev: Will this commented line of code be used in future?

services/core-web/src/components/mine/DigitalPermitCredential/CredentialContentModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    dispatch(fetchCredentialExchangeDetails({ partyGuid, credentialExchangeGuid: credExchId }));
+  }, []);
+
+  const formatTractionDate = (dateString: string) => {
+    const year = dateString.slice(0, 4);
+    const month = dateString.slice(4, 6);
+    const day = dateString.slice(6, 8);
+
+    return `${year}-${month}-${day}T00:00:00`;
+  };
```
henryoforeh-dev: Is it possible to move this to the helpers file in common/utils ?

matbusby-fw: I suppose we might as well.  I left it here because I don't think it's likely to be needed anywhere else, but I guess it might as be available.

services/core-web/src/components/mine/DigitalPermitCredential/ViewDigitalPermitCredential.tsx
matbusby-fw: Nope!  Good catch!

services/core-web/src/components/mine/DigitalPermitCredential/DigitalCredentialDetails.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { IMine, IMineCommodityOption, IMineDisturbanceOption, IPermit } from "@mds/common";
+import {
+  getMineCommodityOptions,
+  getMineDisturbanceOptions,
+} from "@mds/common/redux/selectors/staticContentSelectors";
+import { useSelector } from "react-redux";
+
+const { Title, Paragraph } = Typography;
+
+interface DigitalCredentialPermit extends IPermit {
```
taraepp: this looks like it might want to be in the interfaces directory?

taraepp (dismissed):

henryoforeh-dev (dismissed):

Jsyro:
I'm not sure how the group feels about this, but I would suggest making the credential details modal data look different than the component on the page itself. 

The data that is in the credential should look more technical and immutable as it's not native data anymore.
The heading all needs to be very clear that we are looking at the exact values that are in the digital credential we have given to the company, not an internal record). 

One possible presentation would align with OCA, but that's probably more than we want to chew off at this point. 

https://bcgov.github.io/aries-oca-explorer/identifier/A2UZSmrL9N5FDZGPu68wy%3A2%3Abc-mines-act-permit%3A1.1.1

Even just a raw table of key, value pairs could achieve the same result. 

I don't want to make it ugly, but i do want to make it visually different. Would like to hear Roop's thoughts.

henryoforeh-dev (dismissed):

taraepp (approved):

henryoforeh-dev (approved):

PR #3040, opened 2024-04-06
[MDS-5686] Remove crash details in error responses in CORE-API

isuru-aot:
## Objective 

[MDS-5686](https://bcmines.atlassian.net/browse/MDS-5686)

- Added trace id to error responses in core-api
- Added the trace id to notification message in MS and Core
- Added the trace id to the error email 
   - Added a link to Kibana logs which with a filter to log the corresponding error messages
   - Introduced two separate email template for MS and Core

Note: In addition to the task, removed the unnecessary Custom Exception in `mines/exceptions/mine_exceptions.py` and update the refences with appropriate `messages` and `status codes`


services/core-api/app/api/mines/explosives_permit/resources/explosives_permit.py
```diff
@@ -221,8 +220,7 @@ def put(self, mine_guid, explosives_permit_guid):
             explosives_permit.save()
         except Exception as e:
             current_app.logger.error(e)
-            raise MineException("Oops!, Something went wrong while updaitng the explosive permit information",
-                                detailed_error = e)
+            raise MineException("Something went wrong while updaitng the explosive permit information")
```
taraepp: typo on updating.
I do wonder a bit about using the "mine exception" type for all these- I do see that we also have `ExplosivesPermitDocumentException`, `ExplosivesPermitExeption` (some spelling errors there 😓 ) but also not sure that we need a different exception type for every object type when we're setting a custom error message in every instance.

isuru-aot: Removed unnecessary exceptions and used only `MineException` with appropriate message and status codes.

services/core-api/app/config.py
```diff
@@ -207,6 +207,9 @@ def JWT_ROLE_CALLBACK(jwt_dict):
     # Enable flag caching and evalutation. If set to True, FLAGSMITH_KEY must be set to a server side FLAGSMITH_KEY
     FLAGSMITH_ENABLE_LOCAL_EVALUTION=os.environ.get('FLAGSMITH_ENABLE_LOCAL_EVALUTION', 'false') == 'true'
 
+    # Kibana
+    KIBANA_BASE_URL = os.environ.get('KIBANA_BASE_URL', 'https://kibana-openshift-logging.apps.silver.devops.gov.bc.ca')
```
henryoforeh-dev, line 211: For this environment variable a gitops PR need to be created too.

henryoforeh-dev (approved):

services/core-api/app/api/mines/explosives_permit/resources/explosives_permit.py
taraepp: I know the typo is pre-existing, but should still change 'updaitng' to 'updating'

isuru-aot: @taraepp Well, that's updated already. Please see all the changes.

taraepp: So it is! Must have been looking at it wrong. 🤷‍♀️

taraepp (approved):

PR #3041, opened 2024-04-08
[MDS-5724] AMS field mapping

henryoforeh-dev:
## Objective 

[MDS-5724](https://bcmines.atlassian.net/browse/MDS-5724)

Implemented field mapping and API integration between AMS and core-api.

Implementation also covers [MDS-5849](https://bcmines.atlassian.net/browse/MDS-5849)


services/core-api/app/api/services/ams_api_service.py
```diff
@@ -0,0 +1,169 @@
+import requests
+import json
+
+from app.api.municipalities.models.municipality import Municipality
+from app.config import Config
+from app.date_time_helper import get_date_iso8601_string
+
+
+class AMSApiService():
```
simensma-fresh, line 11: Would be awesome if we could add a test or two for this service. Even if it's just something like "given this example authorization - here's the payload in plain json of what gets posted to AMS"

simensma-fresh (changes requested):

services/core-api/app/api/services/ams_api_service.py
```diff
@@ -0,0 +1,169 @@
...
+                                     is_legal_land_owner,
+                                     is_crown_land_federal_or_provincial,
+                                     is_landowner_aware_of_discharge_application,
+                                     has_landowner_received_copy_of_application
+                                     ):
+        """Creates a new AMS authorization application"""
+
+        ams_results = []
+        headers = {
+            'bearer': Config.AMS_BEARER_TOKEN,
```
taraepp, line 89: Just wondering what happens if the bearer token is expired, if it can catch and refresh, or how the ams_result will be saved if it comes back as an error (will it be populated with None? trigger a 500?).

services/core-api/app/api/services/ams_api_service.py
henryoforeh-dev: Nice suggestion. I will update.

taraepp: ooo yeah and I think I'd want a test ensuring that the authorization still gets saved in our db if AMS errors out for whatever reason (bad request, expired token)

services/core-api/app/api/services/ams_api_service.py
henryoforeh-dev: Updated the code to include try catch. During the call to AMS service, If there is any error that is not bad request (400) related, the results returned will be an empty array. If an empty array is returned, the lines of code for populating the attributes in project_summary_authorizations table is skipped. 

image

services/core-api/app/api/services/ams_api_service.py
henryoforeh-dev: Updated with 2 unit tests for successful outcome and unsuccessful outcome.

matbusby-fw (dismissed):
💪🏻 Super duper!

simensma-fresh (dismissed):

simensma-fresh (approved):

matbusby-fw (approved):

PR #3042, opened 2024-04-08
[MDS-5885] - Report submit button set loading indicator

matbusby-fw:
Added a loading status to the submit button so the user will be unable to click submit multiple times

## Objective 

[MDS-5885](https://bcmines.atlassian.net/browse/MDS-5885)


taraepp (approved):

henryoforeh-dev (approved):
Nice!

PR #3043, opened 2024-04-08
[MDS-5899] - now by now numbers

matbusby-fw:
## Objective 

- Created a new post endpoint for fetching a list of NoW.
- A `post` endpoint is being utilized due to the expected high number of passed numbers (200+)
- Refactored current now_application list resource to have a base class that could be shared with the new resource (since the filtering logic is quite complex and is used in the new resource as well)

POST `{{baseUrl}}/now-applications/now-numbers`

Payload:

{
    "now_numbers": ["1","2","3"]
}

[MDS-5899](https://bcmines.atlassian.net/browse/MDS-5899)


henryoforeh-dev (approved):
Nice work!

simensma-fresh (approved):

PR #3049, opened 2024-04-12
[MDS-5923] Add Missing Author field to edit NoW document form

taraepp:
## Objective 
- add missing field
- note that author, title, and date are all conditional on "part of permit package" being checked

[MDS-5923](https://bcmines.atlassian.net/browse/MDS-5923)

image

### DIFF
--- a/services/core-web/src/components/mine/Projects/ProjectSummary.tsx
+++ b/services/core-web/src/components/mine/Projects/ProjectSummary.tsx
@@ -8,7 +8,6 @@ import EnvironmentOutlined from "@ant-design/icons/EnvironmentOutlined";
 import ArrowLeftOutlined from "@ant-design/icons/ArrowLeftOutlined";
 
 import NullScreen from "@/components/common/NullScreen";
-import { getMines } from "@mds/common/redux/selectors/mineSelectors";
 import {
   getFormattedProjectSummary,
   getProject,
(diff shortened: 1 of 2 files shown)


services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -109,12 +107,11 @@ export const ProjectSummary: FC = () => {
           setIsValid(false);
         });
     }
-    return dispatch(fetchMineRecordById(projectMineGuid)).then(() => {
-      const mine = mines[projectMineGuid];
+    return dispatch(fetchMineRecordById(projectMineGuid)).then((res) => {
```
taraepp, line 110: this made cypress fail, so I tested it out locally and it failed locally too. 🤷‍♀️ 
Race condition- it's not (always) finished populating the redux state immediately in the .then so you get undefined errors.

taraepp: technically it probably shouldn't be fetching the mine record unless it doesn't already exist in the state (which it will if you navigated to the page in most ways aside from directly loading the page). But this whole page needs an overhaul and this is not the ticket.

asinn134 (approved):

henryoforeh-dev (approved):

matbusby-fw (approved):

PR #3054, opened 2024-04-17
[MDS-5561] revoke and reoffer newest amendment

Jsyro:
## Objective 

[MDS-5561 ](https://bcmines.atlassian.net/browse/MDS-5561)

This adds a flask CLI command that revoke a credential by credential_exchange_id and offers the newest permit_amendment on the same mine as the old credential as a vc to the same connection.

image


services/core-api/app/api/services/traction_service.py
```diff
@@ -69,8 +69,8 @@ def create_oob_connection_invitation(self,party: Party):
         
         return response
     
-    def offer_mines_act_permit(self, connection_id, attributes):
-
+    def offer_mines_act_permit_111(self, connection_id, attributes):
+        # https://github.com/bcgov/bc-vcpedia/blob/main/credentials/bc-mines-act-permit/1.1.1/governance.md#261-schema-definition
```
henryoforeh-dev, line 79: Is this still in use ?

henryoforeh-dev (dismissed):
Nice work! just a cosmetic comment.

services/core-api/app/api/services/traction_service.py
Jsyro: Yes, that's the governance document that has to match this implementation.

Jsyro:
After testing multiple questions have arisen. 

1) Should this revoke all previous issued credentials and offer the newest (the code shouldn't the code assume every previous credential has been properly processed)
2) This needs to cancel any pending offers of any old credentials too.
3) It's sending the replacement on the wrong connection, this is fatal error i need to rectify and ensure cannot happen.

Jsyro:
Ok, new approach that leverage existing assumptions and code. 

when a new amendment is create, all VC's for that permit ON THAT MINE, are revoked, of if 'in-progress' cancelled with a problem report. 

Second task is to offer the permit_amentment to the permits current permittee (which infers the current_permittee using existing logic). 

The two jobs are run separately, so if revocation fails, there will be two active credentiasl that are slightly different, but retrying revocation is easy. 

If offering fails, then then user should be able to use minespace as normal to get it again.

henryoforeh-dev (dismissed):

matbusby-fw (dismissed):
🙌🏻

henryoforeh-dev (dismissed):

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -0,0 +1,121 @@
...
+        if ce.cred_exch_state in PartyVerifiableCredentialMinesActPermit._active_credential_states:
+            traction_svc.revoke_credential(connection.connection_id, ce.rev_reg_id, ce.cred_rev_id, reason)
+            
+            attempts = 0
+            while not ce.cred_rev_id:
+                sleep(1)
+                db.session.refresh(cred_exch)
+                attempts += 1
+                if attempts > 60:
+                    raise Exception("Never received webhook confirming revokation credential")
```
matbusby-fw, line 39: Maybe for the purposes of tracking potential failures it would be simplest to make an update to the cred_exch record to indicate the failure.  Either adding a column to that table, or using the `cred_exch_state`?

Jsyro: hmm. true. because if the call to traction returns success but traction fails to successfully revoke, that will probably need different consideration than if that call to traction fails.... hmm. 

Let me take another look. thanks.

henryoforeh-dev (dismissed):

services/core-api/app/api/verifiable_credentials/manager.py
matbusby-fw: `error_description` looks good.  Allows for some versatility.

matbusby-fw (dismissed):

matbusby-fw (approved):

henryoforeh-dev (approved):

Jsyro:
Gif for future testing in DEV. 
Auto_revoke_and_offer

PR #3063, opened 2024-04-22
[MDS-5933] HSRC Edit FE

taraepp:
### What is outstanding:
- [x] there are no tests made yet
- [ ] it doesn't just show HSRCM article_act_code, it shows everything in that table
- [x] BUG: when there's a new item added from the "Add" form, it doesn't get added to form data (this should probably happen in the useEffect that watches complianceCodes)
- [ ] BUG: you have to click on the date on the calendar for expiry date to change it, can't just change the text (bug on form component, I thought it was annoying)
- [ ] there is no viewing that a record is_expired. 
- [ ] there is no filtering by is_expired. 
- [ ] there is no assigning or viewing of associated Mine Reports
- [ ] Sometimes performance is not very good!
   - I took out the table sorting because it was crashing. BE pagination would handle this
   - validating against existing article numbers on the FE takes a lot of memory to search a 1000+ item array. 
- [x] a snapshot test is failing, probably because it's a shallow enzyme test. 
- [ ] I did not verify that the user role filtering works. I do not have the role myself unless someone gave it to me.

[MDS-5933](https://bcmines.atlassian.net/browse/MDS-5933)

image
image
image
image


services/core-web/src/components/admin/complianceCodes/ComplianceCodeManagement.tsx
```diff
@@ -0,0 +1,315 @@
...
+    const editedValues = editedIds.map((id) => {
+      return values[`${formPrefix}${id}`];
+    });
+    // array of compliance articles that have been changed
+    dispatch(updateComplianceCodes(editedValues)).then((resp) => {
+      if (resp.payload) {
+        setIsEditMode(false);
+        setEditedIds([]);
+      }
+      console.log(resp);
```
henryoforeh-dev: Forgot to remove this.

henryoforeh-dev (approved):
Cool stuff!

simensma-fresh (approved):

PR #3064, opened 2024-04-23
[MDS-5933] added auth guard and styling to expired codes

matbusby-fw:
## Objective 

[MDS-5933](https://bcmines.atlassian.net/browse/MDS-5933)

image


henryoforeh-dev (dismissed):

services/core-web/src/constants/permissions.js
```diff
@@ -24,3 +24,4 @@ export const EDIT_PROJECT_SUMMARIES = "role_edit_project_summaries";
 export const EDIT_INCIDENTS = "role_edit_incidents";
 export const EDIT_TSF = "role_edit_tsf";
 export const ABANDONED_MINES = "role_abandoned_mines";
+export const EDIT_COMPLIANCE_CODES = "core_edit_code";
```
taraepp: I see that this one is not quite like the ones above it... It may not work as expected.

So it kind of bothers me that we use this file solely to index into the roles defined in environment.ts and so I made userHasRole to accept either or, like so:
image

Could either make this one be role_edit_compliance_codes, or modify AuthorizationGuard to get userHasRole in its mapStateToProps instead.

services/core-web/src/components/admin/complianceCodes/ComplianceCodeManagement.tsx
```diff
@@ -243,6 +246,13 @@ const ComplianceCodeManagement: FC = () => {
     isEditMode && editModeResetColumn,
   ].filter(Boolean);
 
+  const setExpiredRowBackground = (record: IComplianceArticle) => {
+    if (isAfter(new Date(), new Date(record.expiry_date))) {
```
taraepp, line 250: I think that each of the compliance articles should have an is_expired data index

services/core-web/src/constants/permissions.js
matbusby-fw: Yeah, noticed that too, should be better now.

services/core-web/src/components/admin/complianceCodes/ComplianceCodeManagement.tsx
matbusby-fw: You mean to the end of adding that expired column?  I like it, but I think we should leave that for MVP 2.0 personally.

henryoforeh-dev (approved):

simensma-fresh (approved):

services/core-web/src/components/admin/complianceCodes/ComplianceCodeManagement.tsx
taraepp: Sorry I know this is closed now but I meant that it doesn't have to be calculated again here- each of the entries already has an "is_expired" property.

PR #3066, opened 2024-04-24
[MDS-5949] New Code Edit error

taraepp:
## Objective 
- adding a new code to the list was registering as a change, which was getting discarded when you went to edit dates and then clicked cancel, and then it would still be on the table and you could still edit the non-existent value and things would go wrong on save. Made it re-initialize before that point so it doesn't get discarded
- I had already made a RenderSubmit button component which didn't make things _worse_, and changed some functions to async, so that got included
- there was also spacing issues on View, gave it a classname so you can control adding that space
- but then also found that formClassName was performing boolean math instead of string concatenation, giving me the wrong classname (it was always 'edit', because it was always true). Fixed that- which made for a number of snapshot changes.

[MDS-5949](https://bcmines.atlassian.net/browse/MDS-5949)

### DIFF
--- a/services/common/src/components/forms/RenderSubmitButton.tsx
+++ b/services/common/src/components/forms/RenderSubmitButton.tsx
@@ -1,0 +1,38 @@
+import React, { FC, ReactNode, useContext } from "react";
+import { useSelector } from "react-redux";
+import { isDirty, isSubmitting } from "redux-form";
+import { FormContext } from "./FormWrapper";
+import { Button } from "antd";
+import { BaseButtonProps } from "antd/lib/button/button";
+
+interface RenderSubmitButtonProps {
+  buttonText?: string | ReactNode;
+  buttonProps?: BaseButtonProps;
+}
+
+const RenderSubmitButton: FC<RenderSubmitButtonProps> = ({
+  buttonText = "Save Changes",
+  buttonProps,
+}) => {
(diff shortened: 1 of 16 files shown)


simensma-fresh (approved):

henryoforeh-dev (approved):

services/common/src/components/forms/RenderSubmitButton.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { FC, ReactNode, useContext } from "react";
+import { useSelector } from "react-redux";
+import { isDirty, isSubmitting } from "redux-form";
+import { FormContext } from "./FormWrapper";
+import { Button } from "antd";
+import { BaseButtonProps } from "antd/lib/button/button";
+
+interface RenderSubmitButtonProps {
```
taraepp, line 8: Did not fix bug but did not make it any worse. 🤷‍♀️

services/core-web/src/components/admin/complianceCodes/ComplianceCodeManagement.tsx
```diff
@@ -271,6 +272,7 @@ const ComplianceCodeManagement: FC = () => {
           <Button
             onClick={() => openAddModal()}
             disabled={isEditMode}
+            loading={isLoading}
```
taraepp, line 275: Disabled the buttons while the data is loading. Why not.

PR #3070, opened 2024-04-25
[CHORE] Clean up relationship def warnings + others

Jsyro:
Resolving sqlalchemy warnings and a few other housekeeping things. 

`is` and `is not` are reference comparisons in python, not equivalence. 
`overlays' tells sqlalchemy we know that these two definitions collide, so it's up to us to ensure they are compatible. 

~~`nearest_municipality` has no references ( i searched for .nearest_municpality and only found this usage which doesn't make much sense to me, the parameter is not required and it appears to be assigning it to the wrong attribute 
https://github.com/bcgov/mds/blob/607d19476c931b1f89547691bad38d48a8dcd191/services/core-api/app/api/projects/project_summary/models/project_summary.py#L525~~

relationship definitions that use backref are creating the relationships is both directions in one spot, which can be good or bad, the team should decide on if you want to use back ref everywhere (less lines), or no where (more possible errors, but clarity on what relationships exist).


services/core-api/app/api/incidents/resources/mine_incident_notes.py
```diff
@@ -96,7 +96,7 @@ def post(self, mine_incident_guid):
 
         data = self.parser.parse_args()
         note_content = data.get('content')
-        if note_content is None or note_content is '':
+        if note_content is None or note_content == '':
```
taraepp, line 99: ...I did not know this. Learn something new every day!

taraepp:
I think the nearest_municipality thing is just a bit of a misnomer. I took a look on the FE and on the form, the field is named `nearest_municipality` but the value is 100% a guid. Looks like it's getting renamed from the DB attribute (nearest_municipality_guid) in the response model (nearest_municipality). Which is indeed confusing.

matbusby-fw (approved):
Nice clean up.  I also didn't realize that difference between `is/is not` and `==/!=`

henryoforeh-dev (approved):

isuru-aot (approved):

PR #3074, opened 2024-04-26
[MDS-5809] - updated error handling flow for incorrect IRT upload

matbusby-fw:
## Objective 

Updating to the import component caused the error reporting modal to not function properly.

This section of the code needs some love (and got a fair bit of it in this ticket, but didn't want to baloon scope too much).  Finding lots of odd choices here that are hard to follow and fix.

[MDS-5809](https://bcmines.atlassian.net/browse/MDS-5809)

image

### DIFF
--- a/services/minespace-web/src/components/Forms/projects/informationRequirementsTable/IRTFileImport.tsx
+++ b/services/minespace-web/src/components/Forms/projects/informationRequirementsTable/IRTFileImport.tsx
@@ -77,21 +77,21 @@ export const IRTFileImport: FC<IRTFileImportProps> = ({
     renderTextColumn("create_user", "Imported By"),
   ];
 
-  const handleCreateInformationRequirementsTable = (
+  const handleCreateInformationRequirementsTable = async (
     projectGuid: string,
     file: IFileInfo,
     documentGuid: string
   ) => {
-    dispatch(createInformationRequirementsTable(projectGuid, file, documentGuid));
+    return dispatch(createInformationRequirementsTable(projectGuid, file, documentGuid));
   };
 
-  const handleUpdateInformationRequirementsTableByFile = (
+  const handleUpdateInformationRequirementsTableByFile = async (
     projectGuid: string,
     informationRequirementsTableGuid: string,
     file: IFileInfo,
     documentGuid: string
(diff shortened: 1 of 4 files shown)


services/minespace-web/src/components/Forms/projects/informationRequirementsTable/IRTFileImport.tsx
```diff
@@ -77,21 +77,21 @@ export const IRTFileImport: FC<IRTFileImportProps> = ({
...
   ];
 
-  const handleCreateInformationRequirementsTable = (
+  const handleCreateInformationRequirementsTable = async (
     projectGuid: string,
     file: IFileInfo,
     documentGuid: string
   ) => {
-    dispatch(createInformationRequirementsTable(projectGuid, file, documentGuid));
+    return dispatch(createInformationRequirementsTable(projectGuid, file, documentGuid));
```
taraepp, line 85: dispatch returns an action, doesn't it? It looks like it maybe wants to be await, but hard to tell in JS files.

henryoforeh-dev (approved):

services/minespace-web/src/components/Forms/projects/informationRequirementsTable/IRTFileImport.tsx
matbusby-fw: I thought so too, but I guess not.  ESLINT threw an error if I added an await there. (and it works without)

taraepp: I guess the function consumer probably just wants something that _returns_? I find it odd but 🤷‍♀️ 
(I had to double-check, it does return an action)

taraepp (approved):

services/minespace-web/src/components/Forms/projects/informationRequirementsTable/IRTFileImport.tsx
matbusby-fw: Yeah, I did the same thing.  🤷🏻 

Honestly the setup here isn't great, but I think there are a lot of threads to pull here 😬

PR #3075, opened 2024-04-26
[MDS-5707] Added history view to Tailings Storage Facilities

simensma-fresh:
## Objective 

[MDS-5707](https://bcmines.atlassian.net/browse/MDS-5707)

*History View*
Added a TSF History view to the basic info page in core/minespace.

image
image
image

*SQLAlchemy Continuum*
Introduced [SQLAlchemy Continuum](https://sqlalchemy-continuum.readthedocs.io/) for "automatic" handling of version tracking. Basically, just add a `__versioned__ = {}` property to your model and it'll create a Version model behind the scene that gets automatically added to on insert, updates, and deletes.

*Migrations*
Added a utility command `make generate_version_table_migration TABLE=<table_name>` that you can use to automatically generate a database migration for the `_version` table you want to create, and a DB migration to backfill existing data.


services/common/src/components/tailings/BasicInformation.tsx
```diff
@@ -36,9 +36,51 @@ interface BasicInformationProps {
   isEditMode: boolean;
 }
 
+// Provide a mapping of the field names to the title and data for the field
+// This is used to display the field names and values in a more user-friendly way in the history modal
+const historyDiffValueMapper = {
```
matbusby-fw, line 41: For the purposes of these value mappers, will we need to always provide an exhaustive list of all fields we want tracked?  I'm just thinking about what will happen if the data changes and updates to these get missed.

docs/audit_log.md
```diff
@@ -0,0 +1,40 @@
...
+                mine_no = db.Column(db.String(10))
+                mine_name = db.Column(db.String(60), nullable=False)
+
+                __versioned__ = {} # <--- This statement is required to enable versioning
+        ```
+
+2.  Create a version table for the model by running the following command:
+
+    ```bash
+    make generate_version_table_migration TABLE=mine
```
matbusby-fw, line 31: This is awesome.  For future updates to the base table, I'm assuming we just have to make sure the versioned table gets updated as well?  Will it break things if that step gets missed?

matbusby-fw (approved):
Looks awesome!  Left a couple of questions regarding usage of the diff table elements and continuum, but I love it!

services/common/src/components/tailings/BasicInformation.tsx
simensma-fresh: I set it up to default to using the values as is if missing from the mapper, and just transforming the title from snake_case to Title Case - I'll update the comments to make that more clear

docs/audit_log.md
simensma-fresh: Exactly. The same changes will have to be performed on the `_version` table if the `base` table is updated. Things will break if mised

services/common/src/components/comments/CommentEditor.spec.tsx
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import { render } from "@testing-library/react";
 import { ReduxWrapper } from "@mds/common/tests/utils/ReduxWrapper";
 import CommentEditor from "./CommentEditor";
+window.scrollTo = jest.fn();
```
simensma-fresh, line 5: Just added these a couple of places to resolve some `jest` warnings

henryoforeh-dev (approved):

PR #3088, opened 2024-05-08
[MDS-5802] Major Projects Summary validation errors

asinn134:
## Objective 

[MDS-5802](https://bcmines.atlassian.net/browse/MDS-5802)

- Removed validate_phone_no helper function call for project contacts and project party. The cereberus schema validation will do this now.

- Adjusted `ams_terms_agreed` to not only accept True. This was causing a validation error when a user tries to update the Authorization Involved section by selecting something like `Air emissions discharge permit` input. By selecting `Air emissions discharge permit` it requires the user to agree to the ams terms and condition in the declaration section and because this application was submitted and now has `"SUB"` as it's `status_code`, a full validation is done on all sections and we will get the validation error for `ams_terms_agreed` not being true.

### DIFF
--- a/services/core-api/app/api/projects/project/models/project.py
+++ b/services/core-api/app/api/projects/project/models/project.py
@@ -258,10 +258,6 @@ class Project(AuditMixin, Base):
         for contact in contacts:
             updated_contact_guid = contact.get('project_contact_guid')
             new_address_data = contact.get('address', None)
-            if new_address_data:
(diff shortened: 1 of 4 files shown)


matbusby-fw (dismissed):
⚡

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):

isuru-aot (approved):

PR #3092, opened 2024-05-09
[MDS-5973] Setting up featureflag checks for project summary backend validation

asinn134:
## Objective 

[MDS-5973](https://bcmines.atlassian.net/browse/MDS-5973)

- Added condition to check for ams_agent feature flag to determine if we include validation on agent, applicant, facility, legal land, and authorization


simensma-fresh (approved):

matbusby-fw (approved):
Nice!

PR #3105, opened 2024-05-21
[MDS-5754] - Update check for test jwt secrets in core-api to not crash on NONETYPE

matbusby-fw:
## Objective 

[MDS-5754](https://bcmines.atlassian.net/browse/MDS-5754)

### DIFF
--- a/services/core-api/app/extensions.py
+++ b/services/core-api/app/extensions.py
@@ -76,7 +76,7 @@ def getJwtManager():
         raise AuthError({'code': 'auth_fail',
                     'description': 'Token issuer oidc.gov.bc.ca is no longer supported. Please contact the mds team.'}, 401)
 
-    if test_config.JWT_OIDC_TEST_ISSUER is not None:
+    if test_config.JWT_OIDC_TEST_ISSUER is not None and test_config.JWT_OIDC_TEST_ISSUER in iss:
         jwt_result = jwt
     else:
         jwt_result = get_jwt_by_audience(aud)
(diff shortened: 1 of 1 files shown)


services/core-api/app/extensions.py
```diff
@@ -76,7 +76,7 @@ def getJwtManager():
         raise AuthError({'code': 'auth_fail',
                     'description': 'Token issuer oidc.gov.bc.ca is no longer supported. Please contact the mds team.'}, 401)
 
-    if iss in test_config.JWT_OIDC_TEST_ISSUER:
+    if test_config.JWT_OIDC_TEST_ISSUER is not None:
```
simensma-fresh: Think we also might need the `iss in test_config.JWT_OIDC_TEST_ISSUER` check here😅
Think it's the case locally that `JWT_OIDC_TEST_ISSUER` is set but the users token is actually issued by the `mds-public` issuer

simensma-fresh (changes requested):

Jsyro:
Why is the development deployment using test_config? or am i misunderstanding the naming? 

I'm not sure what this component does.

matbusby-fw:
> iss in test_config.JWT_OIDC_TEST_ISSUER

This was just added, but it seems to be just used for the integration tests.  So this update is to make sure it doesn't run unless those integration test variables are present.

simensma-fresh (approved):

isuru-aot (approved):

PR #3110, opened 2024-05-23
[MDS-5984] Feature/5984 poc public w3c credentials

Jsyro:
## Objective 

[MDS-5984](https://bcmines.atlassian.net/browse/MDS-5984)

The feature is feature_flagged with a new key that will need to be configured @simensma-fresh 

Endpoint that takes a permit_amendment and returns a w3c signed credential for it. the format of the credential is underreview, currently just re-using the anoncreds data shape.

Keys are the other issue, acapy has deprecated an old endpoint that has a critical function that allows us to re-use a verkey from a public did, I may have to make that change myself. 

The modern endpoint cannot use did:sov: because of aca-py quirks, so can only use a did:key: which does not have a public verkey, so cannot be verified. 

I've also hard coded the did:indy:bcovrin:test: prefix, but the prod key is on did:indy:candy, and we want to use `did:web:` in the production use case (as UNTP requires did:web, or maybe the even newer did:tdw)


services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -43,81 +46,130 @@ def revoke_all_credentials_for_permit(permit_guid: str, mine_guid:str, reason:st
...
+        credential = {
+            "@context":
+            ["https://www.w3.org/2018/credentials/v1", {
+                "@vocab": "urn:bcgov:attributes#"
+            }],
+            "type": ["VerifiableCredential"],
+            "issuer": {
+                "id": did,
+            },
+            "issuanceDate": "2010-01-01T19:23:24Z",
```
simensma-fresh: Just curious if the issuanceDate is supposed to be hardcoded?

services/core-api/app/api/verifiable_credentials/models/w3c_map_credentials.py
```diff
@@ -0,0 +1,28 @@
...
+from app.api.mines.permits.permit.models.permit import Permit
+from app.api.mines.permits.permit_amendment.models.permit_amendment import PermitAmendment
+
+class UNTPCCredential():
+    pass
+
+
+class PartyVerifiableCredentialW3CCredential(AuditMixin, Base):
+    """Verificable Credential"""
+    __tablename__ = "party_verifiable_credential_w3c_credential"
```
simensma-fresh: We'll need to add a Flyway migration to match this new table otherwise this will cause some issues when deployed to dev

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -0,0 +1,101 @@
...
+        help='GUID of the permit amendment.',
+        location='json',
+        store_missing=False)
+
+    @api.expect(parser)
+    @api.doc(
+        description=
+        "returns a signed w3c credential for a specific permit_amendment, using new aca-py endpoints, but cannot use public did."
+    )
+    def post(self):
```
simensma-fresh, line 50: Could we add in a role check for these endpoints? 
I assume they should only be available for Minespace proponents and certain core users. Just so we don't accidentally forget and leave them open when we eventually turn on the feature flag

simensma-fresh (changes requested):

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: It is not. it should be the issue_date (and should actually be removed from the contents).

services/core-api/app/api/verifiable_credentials/models/w3c_map_credentials.py
Jsyro: I'll remove this. it's currently unused, good catch.

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: Good catch. not sure what it will be. but i'll make it MINE_EDIT for now. 

Minespace likely won't be interacting with this directly. most likely it will be celery tbh.

simensma-fresh: Makes sense, sounds good to me

simensma-fresh (approved):

matbusby-fw (approved):
I haven't done any work with dids before now so I don't have much insight there.  Aside from Simen's comments though I'm not seeing anything out of place.

PR #3111, opened 2024-05-23
[FIX] update vscode settings

Jsyro:
vscode no longer respects the existing format of the settings.json file..

I restored the existing yapf formatting needs, but not sure if others are using a different tool for formatting or not.

Ideally we locally enforce the same formatting/linting that the pipeline expects.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,12 +4,11 @@
   "files.trimFinalNewlines": true,
   "editor.renderWhitespace": "none",
   "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
I'm not generally using vs code, but looks solid from what I can tell.

simensma-fresh (approved):
👍 All for this. Eslint is automatically run (in fix mode) as an on-commit hook, but python formatting is not, so does indeed make sense to enable that on save

PR #3112, opened 2024-05-23
[MDS-5803] Adjustments for Minespace project description UI cleanup

asinn134:
## Objective 

[MDS-5803](https://bcmines.atlassian.net/browse/MDS-5803)

- Adjusted Minespace project summary sections based on figma mockups. This includes changes to styling, text, and moving some inputs that were previously in the Facility Operator section into the Legal Land section.

- This isn't part of the ticket but I did some adjustment to the Project Summary Update's back end validation. There are a bunch of properties in the payload from facility and legal land section that are on the surface level of the payload instead of inside a section specific property, so for example `facility_type` isn't inside something like `facility_operator`. Previously the validation of these properties were done in the `validate_project_summary_surface_level_data` method but I moved the validation of them to the methods that specifically validate those page sections. 

  The reasoning for this was because if for example a validation error occurred for the `facility_type` property I think it makes more sense to show this error in the array that contains errors for the section `facility_type` input is entered in, Mine Components and Offsite Infrastructure (previously was  Facility Location and Operator), instead of in the surface level error array. 

```
{
            'surface_level_data': [],
            'basic_info': [],
            'project_contacts': [],
            'authorizations': [],
            'applicant_info': [],
            'agent': [],
            'location_access_land_use': [],
            'mine_component_and_offsite': [],
            'declaration': [],
}
```


services/common/src/components/projectSummary/FacilityOperator.tsx
```diff
@@ -1,56 +1,26 @@
...
 import { FORM } from "../..";
 import { getDropdownProvinceOptions } from "@mds/common/redux/selectors/staticContentSelectors";
 import RenderRadioButtons from "../forms/RenderRadioButtons";
 import RenderAutoSizeField from "../forms/RenderAutoSizeField";
-import CoreMap from "../common/Map";
 import { normalizePhone } from "@mds/common/redux/utils/helpers";
 
 export const FacilityOperator: FC = () => {
   const formValues = useSelector(getFormValues(FORM.ADD_EDIT_PROJECT_SUMMARY));
-  const formErrors = useSelector(getFormSyncErrors(FORM.ADD_EDIT_PROJECT_SUMMARY));
```
asinn134, line 28: The figma mockup had some fields removed out of the facility section (now called Mine Components and Offsite Infrastructure) and setup within the legal land section (now called Location, Access and Land Use). So the work here is just removing them.

services/common/src/components/projectSummary/LegalLandOwnerInformation.tsx
```diff
@@ -1,31 +1,63 @@
...
+    facility_latitude,
+    facility_longitude,
+    legal_land_desc,
+    facility_pid_pin_crown_file_no,
+  } = formValues;
 
+  const [pin, setPin] = useState<Array<string>>([]);
   const municipalityOptions = useSelector(getDropdownMunicipalities);
 
+  const dataSourceOptions = [
```
asinn134, line 41: A lot of the work in this file is just adding in the code that supports the fields that we moved in from the Facility section.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -457,6 +457,66 @@ def validate_project_party(cls, party, section):
         if not v.validate(party):
             return json.dumps(v.errors)
         return True
+    
+    @classmethod
+    def validate_mine_component_offsite_infrastructure(self, data):
```
asinn134, line 462: All the adjustments in this file was to move the validation of section specific properties out of the method that validates surface level properties and instead into the methods that actually validate the section these specific properties exist in.

So for example `facility_pid_pin_crown_file_no exists` on the surface level of the payload but on the front end it is in the Location, Access and Land Use section (formerly labelled as Legal Land Owner Information)

henryoforeh-dev (approved):
Nice

isuru-aot (approved):

PR #3117, opened 2024-05-24
[FIX] W3C Credential needs datetime, not date. 

Jsyro:
using utc to match the rest, could become time-zone aware, but given it's all dates and not times, it's not that important.


henryoforeh-dev (approved):

matbusby-fw (approved):
👍🏻

PR #3107, opened 2024-05-22
[MDS-5964] IRT File Upload 415 Issue

isuru-aot:
## Objective 

[MDS-5964](https://bcmines.atlassian.net/browse/MDS-5964)

_Why are you making this change? Provide a short explanation and/or screenshots_

PR #3123, opened 2024-05-27
[MDS-5990] IRT Submit Button Issue

isuru-aot:
## Objective 

[MDS-5990](https://bcmines.atlassian.net/browse/MDS-5990)

1. Initially in [MDS-5964](https://github.com/bcgov/mds/pull/3107/files) removed the json data readings from the request. But when submitting the IRT at the last step, BE update the `current_status_code`. Hence reverted the change back and added a validation for request.json to address the 415 issue existed before.
I left the email notification part as it is.

2. FE removed the AuthorizationWrapper in order to address the submit button not visible issue for users in MS

### DIFF
--- a/services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table.py
+++ b/services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table.py
@@ -55,9 +55,20 @@ class InformationRequirementsTableResource(Resource, UserMixin):
                     import_file)
                 irt_updated = irt.update(sanitized_irt_requirements, import_file, document_guid)
                 return irt_updated
-            else:
-                current_app.logger.error("Error occurred while retrieving file | document_guid")
-                raise BadRequest('Missing file information')
+
+            if request.json:
+                data = request.json
+                current_status_code = irt.status_code
+                irt_updated = irt.update(data)
+                if current_status_code != 'APV' and data['status_code'] == 'APV':
+                    irt.send_irt_approval_email()
+                elif current_status_code != 'SUB' and data['status_code'] == 'SUB':
+                    irt.send_irt_submit_email()
+                    # Trigger notification for newly submitted IRT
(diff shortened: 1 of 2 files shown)


services/minespace-web/src/components/pages/Project/InformationRequirementsTableStepForm.tsx
```diff
@@ -158,26 +158,24 @@ const StepForms = ({
                   project?.information_requirements_table?.irt_guid
                 )}
               >
```
taraepp, line 160: SonarCloud is making 2 complaints about this file:
1. remove the import for AuthorizationWrapper
2. rename stepFormsProps --> StepFormProps

May as well fix them!

services/core-api/app/api/projects/information_requirements_table/resources/information_requirements_table.py
```diff
@@ -55,9 +55,20 @@ def put(self, project_guid, irt_guid):
                     import_file)
                 irt_updated = irt.update(sanitized_irt_requirements, import_file, document_guid)
                 return irt_updated
-            else:
-                current_app.logger.error("Error occurred while retrieving file | document_guid")
-                raise BadRequest('Missing file information')
+
+            if request.json:
```
matbusby-fw, line 59: I think if we're going to add this back in we should actually be creating a new endpoint that does this.  While the if check should prevent the failure we had before, I don't think it's best practice to have a single endpoint accepting both json data and form data since they're mutually exclusive.  

Can we move this to it's own endpoint (maybe 'status')?

isuru-aot: Raised a separate [PR for this](https://github.com/bcgov/mds/pull/3128). Hence closing this PR.

isuru-aot:
Created a separate [PR](https://github.com/bcgov/mds/pull/3128) with the new endpoint change

isuru-aot:
https://github.com/bcgov/mds/pull/3128

PR #3126, opened 2024-05-28
[MDS-5972] Update CORE homepage links

taraepp:
## Objective 
- match new links in ticket
- ticket had option of linking to the HSRC pdf or its webpage- went with the PDF because a) seems more consistent, and b) sometimes external links break
- there wasn't yet a snapshot test for the CORE homepage so I made one

[MDS-5972](https://bcmines.atlassian.net/browse/MDS-5972)


henryoforeh-dev (approved):

matbusby-fw (approved):
🥇

PR #3130, opened 2024-05-29
[MDS-5989] Major Project Document uploads not saving all uploaded files

asinn134:
## Objective 

[MDS-5989](https://bcmines.atlassian.net/browse/MDS-5989)

-The issue was because support and spatial uploads happening at the same time could override each other and prevent some files from being saved.

-Adjustment made to support and spatial file uploads to have their own separate fields.

### DIFF
--- a/services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
+++ b/services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
@@ -1,17 +1,9 @@
-import React, { FC, useState } from "react";
-import { change, Field, formValueSelector } from "redux-form";
-import { connect } from "react-redux";
+import React, { FC, useEffect } from "react";
+import { change, Field, getFormValues } from "redux-form";
+import { connect, useSelector } from "react-redux";
 import { Form, Typography } from "antd";
 import { bindActionCreators } from "redux";
-import {
-  CSV,
-  DOCUMENT,
-  EXCEL,
-  IMAGE,
-  OTHER_SPATIAL,
-  SPATIAL,
-  XML,
-} from "@mds/common/constants/fileTypes";
+import { CSV, DOCUMENT, EXCEL, IMAGE, OTHER_SPATIAL, XML } from "@mds/common/constants/fileTypes";
 import DocumentTable from "@/components/common/DocumentTable";
 import {
   documentNameColumn,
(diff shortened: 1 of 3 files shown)


services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
```diff
@@ -72,12 +62,25 @@ export const DocumentUpload: FC<DocumentUploadProps> = (props) => {
     ...IMAGE,
   };
 
+  useEffect(() => {
+    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "spatial_documents", []);
+    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "support_documents", []);
+  }, []);
```
taraepp, line 68: just checking- do the fields populate properly when you view an existing record (with existing spatial/support documents)?

services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
```diff
@@ -72,12 +62,25 @@ export const DocumentUpload: FC<DocumentUploadProps> = (props) => {
...
+
+  useEffect(() => {
+    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "documents", [
```
taraepp, line 71: I don't think there's a field on the page anymore named documents, so I am wondering where this value is used

services/minespace-web/src/tests/components/Forms/projects/projectSummary/__snapshots__/DocumentUpload.spec.js.snap
```diff
@@ -1,155 +1,42 @@
...
+          "getState": [Function],
+          "replaceReducer": [Function],
+          "subscribe": [Function],
+        },
+        "unsubscribe": null,
+      },
+    }
+  }
+>
+  <DocumentUpload
```
taraepp, line 33: I don't _love_ that it doesn't show the render of DocumentUpload anymore- does it behave differently if you look at component.firstChild in the expect?

services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
asinn134: Yup the fields show up when someone has uploaded a spatial or support document in the file uploader.

services/minespace-web/src/components/Forms/projects/projectSummary/DocumentUpload.tsx
asinn134: The project summary form has a documents field that gets submitted as part of the form to the back end.

services/minespace-web/src/tests/components/Forms/projects/projectSummary/__snapshots__/DocumentUpload.spec.js.snap
asinn134: Just tested it out and component.firstChild behaves the same in the expect

simensma-fresh (approved):

services/minespace-web/src/tests/components/Forms/projects/projectSummary/__snapshots__/DocumentUpload.spec.js.snap
taraepp: bummer! It _might_ work if you used the react-testing-library render instead of the enzyme shallow, but it might also be a giant can of worms.

taraepp (approved):
Don't want to hold up too long over a snapshot test. 👍

services/minespace-web/src/tests/components/Forms/projects/projectSummary/DocumentUpload.spec.js
```diff
@@ -24,7 +26,12 @@ beforeEach(() => {
 
 describe("DocumentUpload", () => {
   it("renders properly", () => {
-    const component = shallow(<DocumentUpload {...dispatchProps} {...props} />);
+    const component = shallow(
```
matbusby-fw, line 29: Tara's correct that this might open up a whole can of worms, but it might be worth trying render from react-testing library.

The syntax is something like this:

```
const { container } = render(...)

expect(container.firstChild).toMatchSnapshot();
```

The can of worms part comes from the fact that it will try to fully render the component including child components, but is more reliable to use as well.

PR #3131, opened 2024-05-30
[MDS-6005] Notify MMO on report submission for 10.4.4(g)

taraepp:
## Objective 
[MDS-6005](https://bcmines.atlassian.net/browse/MDS-6005)
- Just to note something I thought was strange: perm recl does _not_ get an email sent without the entry in mine_report_notification which is added by the migration (and this is the only record that references that contact) but with it, it sends _twice_
- once to `PermRecl@gov.bc.ca` and again to `permrecl@gov.bc.ca`
- I decided that this was probably a can of worms though- I tested emailing _myself_ with different upper/lower and only got it once so I concluded that this is fine. Couple things that might be causing this that I didn't want to change the logic for:
   - we have two defined constants for this same email, MAJOR_MINES_OFFICE_EMAIL, and PERM_RECL_EMAIL (and it's also in emli_contacts table)
   - mine_report.py-->getReportSpecificEmailsByReportType: looks like it adds PERM_RECL for major mines but only if there are contacts associated with the compliance article
   - and also there is no lower-casing in the comparison to see if an email is already in the list


henryoforeh-dev (approved):

henryoforeh-dev (approved):

matbusby-fw (approved):
👍🏻

asinn134 (approved):

isuru-aot (approved):

PR #3134, opened 2024-05-30
[MDS-6004] Data handling on project summary form

taraepp:
## Objective 
There were a few issues:
- the conditions for testing if the page is loaded are different for new & existing records. For existing records, we want to wait for the formattedProjectSummary, for new, only the mine
- data was reverting when Saving a page and then clicking Back!
   - this is because the form wasn't picking up changes to formattedProjectSummary, so there was a whole series of state changes between clicking save & getting the new data back where it would use the old data to initialize the form (component was re-mounted but data hadn't been fetched yet). Setting enableReinitialize = true fixed that (see reduxForm docs for further details: https://redux-form.com/6.2.1/docs/api/reduxform.md/)
- and then I broke the test because I changed the loading conditions (making it return "Loading" component)
   - gave the isLoaded and isEditMode state values on mount that were appropriate.
   - updated mock data to match, because default values check that data matches requested guids
- When removing a category of authorizations, it wasn't changing it properly in the form values, so it wasn't getting deleted
- removed some returns that weren't being used and redundant setStates in troubleshooting process - this didn't fix anything but it didn't break anything either, and it's cleaner.

[MDS-6004](https://bcmines.atlassian.net/browse/MDS-6004)

### DIFF
--- a/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
+++ b/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
@@ -43,9 +43,9 @@ import { IProjectSummaryDocument } from "@mds/common/interfaces";
 import {
   renderTextColumn,
   renderDateColumn,
+  renderCategoryColumn,
 } from "@mds/common/components/common/CoreTableCommonColumns";
 import DocumentTable from "@mds/common/components/documents/DocumentTable";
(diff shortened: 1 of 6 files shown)


simensma-fresh (approved):

matbusby-fw (approved):
Good catches and cleanup!

PR #3138, opened 2024-06-04
[MDS-5397] update python

henryoforeh-dev:
## Objective 

[MDS-5397](https://bcmines.atlassian.net/browse/MDS-5397)


matbusby-fw (approved):
Love that we're moving to slim image!

simensma-fresh (approved):

PR #3140, opened 2024-06-05
[MDS-6006] Allow ability to edit and remove the orgbook associations

isuru-aot:
## Objective 

[MDS-6006](https://bcmines.atlassian.net/browse/MDS-6006)

- Users can change an orgbook party connection to a party through the party edit modal 
- User with core_admin can delete an existing orgbook party connection through the party edit modal
   - This is a hard delete from `party_orgbook_entity` table.

Screenshot 2024-06-05 at 5 21 43 PM
Screenshot 2024-06-05 at 5 22 03 PM

### DIFF
--- a/services/common/src/components/parties/OrgBookSearch.tsx
+++ b/services/common/src/components/parties/OrgBookSearch.tsx
@@ -60,11 +60,9 @@ const OrgBookSearch: FC<OrgBookSearchProps> = ({
     setIsSearching(false);
   };
 
-  const placeHolder = "Start typing to search OrgBook...";
-
   useEffect(() => {
     if (selectedParty !== current_party) {
-      setSelectedParty(placeHolder);
+      setSelectedParty(undefined);
     }
   }, [current_party]);
 
@@ -102,7 +100,7 @@ const OrgBookSearch: FC<OrgBookSearchProps> = ({
         showSearch
         showArrow
         labelInValue
-        placeholder={placeHolder}
+        placeholder="Start typing to search OrgBook..."
         notFoundContent={isSearching ? <Spin size="small" indicator={<LoadingOutlined />} /> : null}
         filterOption={false}
         onSearch={handleSearchDebounced}
--- a/services/common/src/components/parties/OrgBookSearch.tsx
+++ b/services/common/src/components/parties/OrgBookSearch.tsx
@@ -34,7 +34,6 @@ const OrgBookSearch: FC<OrgBookSearchProps> = ({
 
   const [options, setOptions] = useState([]);
   const [isSearching, setIsSearching] = useState(false);
-  const [isAssociated, setIsAssociated] = useState(current_party !== undefined);
   const [selectedParty, setSelectedParty] = useState(current_party);
 
   const handleChange = () => {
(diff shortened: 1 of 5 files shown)


services/common/src/components/parties/OrgBookSearch.tsx
```diff
@@ -54,6 +60,14 @@ const OrgBookSearch: FC<OrgBookSearchProps> = ({ isDisabled = false, setCredenti
     setIsSearching(false);
   };
 
+  const placeHolder = "Start typing to search OrgBook...";
+
+  useEffect(() => {
+    if (selectedParty !== current_party) {
+      setSelectedParty(placeHolder);
```
henryoforeh-dev: Any particular reason for setting the selected party state to be the constant placeholder value "Start typing to search OrgBook..." ?

services/core-api/app/api/parties/party/resources/party_orgbook_entity_list_resource.py
```diff
@@ -59,3 +59,15 @@ def post(self, party_guid):
...
+    @api.doc(description='Delete a Party OrgBook Entity.')
+    @requires_role_mine_admin
+    @api.marshal_with(PARTY_ORGBOOK_ENTITY, code=201)
+    def delete(self, party_guid):
+        party_orgbook_entity = PartyOrgBookEntity.find_by_party_guid(party_guid)
+        if party_orgbook_entity is None:
+            raise NotFound('OrgBook entity not found.')
+
+        party_orgbook_entity.delete()
+        return None, 204
```
henryoforeh-dev, line 72: 1. The @api.marshal_with decorator specifies a status code of 201, which is used for "Created" responses. However, for a successful deletion, the status code should be 204 (No Content).

2. The @api.expect(parser) decorator suggests that some input validation or parsing is expected, but it is not being utilized within the function.

services/common/src/components/parties/OrgBookSearch.tsx
isuru-aot: In the flow, 
1. User sees the place holder (initial step, when no party is associated)
2. Then user types and select one party and click on `Associate` button. This is will disabled the search box
3. Then (if) user want to change the selected party, user clicks on `Disassociate` button (screen shots in the PR)
4. In step (3), if assigned selected party as 
    - setSelectedParty(NONE) -> it will show None in the search box
    - setSelectedParty("") -> it will show the empty in the search box.

So I thought assigning the place holder to the party, makes things easier for the user.

If there's better options to handle this, please let me know.

isuru-aot: @henryoforeh-dev actually, setSelectedParty(undefined) resolve the issue. I'll commit this

services/core-api/app/api/parties/party/resources/party_orgbook_entity_list_resource.py
isuru-aot: This is addressed.

henryoforeh-dev (approved):

Jsyro (approved):
Delete and remake makes sense to me in this case. 
Only admins can see button and disassociate, good call! assuming Rebecca agrees.

I wonder if we should start showing the Party NAME and Orgbook names together (the original feature overrode the party name with the orgbook name). But that is future work.

PR #3145, opened 2024-06-07
[MDS-5927] Bring Projects Form to CORE

taraepp:
## Objective 
Bring Projects form to CORE to make it match MS

Bonus:
- streamlined a lot of props so components handled things themselves
- the ProjectSummaryForm is the only instance using the `common` SteppedForm, so made it handle the form itself, check for errors, etc.
- some bugs in authorizations documents (going away when you navigate to a different page, showing up in both new/amendment)

Tech Debt:
- there's no "cancel confirmation" when you try to leave the page on CORE
- CORE had the "edit/view mode" toggle, and I did keep that, but some of the components don't _have_ a view mode yet. What's a checkbox supposed to look like, for example? The Add button still shows up on the Related Projects page
- previously CORE users could remove and archive documents, if they had either admin role or edit project summaries. Did not bring this in yet
- the status doesn't display on the header for every page yet, and if you want to update it you have to do that under "Ministry Contact"
   - I'm also not really sure if either of those fields should be required, so they're not
- if we decide to fix the view mode, should probably also stop the tabs disabling while it's in View mode

[MDS-5927](https://bcmines.atlassian.net/browse/MDS-5927)

image


services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -134,8 +134,11 @@ export const FileUpload = (props: FileUploadProps) => {
     const fileTypeList = listedFileTypes ?? Object.keys(acceptedFileTypesMap);
     const fileTypeDisplayString =
       fileTypeList.slice(0, -1).join(", ") + ", and " + fileTypeList.slice(-1);
+    const fileSize = props.maxFileSize
+      ? ` with max individual file size of ${props.maxFileSize}`
+      : "";
     const secondLine = abbrevLabel
-      ? `<div>We accept most common ${fileTypeDisplayString} files</div>`
+      ? `<div>We accept most common ${fileTypeDisplayString} files${fileSize}.</div>`
```
taraepp, line 141: just for Denise- making file-size show up by default.

services/common/src/components/forms/Step.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from "react";
```
taraepp, line 1: One of many, many components to get moved or copied to common

services/common/src/components/forms/SteppedForm.tsx
```diff
@@ -1,54 +1,57 @@
-import { Button, Col, Form, Menu, Popconfirm, Row } from "antd";
+import React, { FC, ReactElement, useEffect, useState } from "react";
+import { useSelector, useDispatch } from "react-redux";
+import { getFormSyncErrors, getFormValues, submit } from "redux-form";
+import { Button, Col, Menu, Popconfirm, Row, StepProps } from "antd";
```
taraepp, line 4: Basically, SteppedForm in this state:
- makes its own FormWrapper with the settings that are ideal for it
- checks for errors
- passes down isEditMode to FormWrapper to enable a view mode
- transforms the values according to a function prop

services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
```diff
@@ -540,14 +459,15 @@ const RenderAuthCodeFormSection = ({
   );
 };
 
-export const AuthorizationsInvolved = (props) => {
+export const AuthorizationsInvolved = () => {
```
taraepp, line 461: initialValues does not need to be passed as a prop when we're using reduxForm. Redux has all that in the state already (both initial & current values, and data about errors, etc) so we can use a selector to get that context _right where we need it_ instead of prop drilling.
Most of the edits here are just taking out props that duplicate redux state.

services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
```diff
@@ -388,55 +325,52 @@ const RenderEMAAuthCodeFormSection = ({
       <Field
         name={`${code}.types`}
         component={RenderGroupCheckbox}
-        label="What type of authorization is involved in your application?"
         required
         validate={[requiredList]}
         normalize={normalizeGroupCheckBox}
+        onDrop={(event) => {
```
taraepp, line 330: This onDrop though was for a bug. I noticed that when I had "amendment" checked, and dropped a file, that it unchecked the box on me! Can be replicated in test, currently.

services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
```diff
@@ -49,46 +49,43 @@ import DocumentTable from "@mds/common/components/documents/DocumentTable";
...
-  const updateAmendmentDocuments = (doc: IProjectSummaryDocument, code: string) => {
-    const index = 0;
-    const response = dispatch(
+  const updateAmendmentDocuments = (doc: IProjectSummaryDocument) => {
+    const type = isAmendment ? "AMENDMENT" : "NEW";
+    dispatch(
       arrayPush(
         FORM.ADD_EDIT_PROJECT_SUMMARY,
-        `authorizations.[${code}].AMENDMENT.[${index}].amendment_documents`,
+        `authorizations.[${code}].${type}[${index}].amendment_documents`,
```
taraepp, line 81: this here was for the documents duplicating. The change from AMENDMENT to ${type} is to stop them duplicating across new/amendment, and the index now being a prop set by the parent is what stops them from being duplicated across multiple amendments.

services/common/src/components/projectSummary/DocumentUpload.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+  },
+  [STATIC_CONTENT]: {
+    projectSummaryPermitTypes: MOCK.BULK_STATIC_CONTENT_RESPONSE.projectSummaryPermitTypes,
+    projectSummaryAuthorizationTypes:
+      MOCK.BULK_STATIC_CONTENT_RESPONSE.projectSummaryAuthorizationTypes,
+  },
+};
+
+describe("DocumentUpload", () => {
+  it.skip("renders properly", () => {
```
taraepp: Oops. I _knew_ I committed at least one skipped test in there. Try to get it working.

services/common/src/components/projectSummary/DocumentUpload.tsx
```diff
@@ -144,13 +129,13 @@ export const DocumentUpload: FC<DocumentUploadProps> = (props) => {
...
-    projectGuid: initialValues.project_guid,
-    projectSummaryGuid: initialValues.project_summary_guid,
+    mineGuid: mine_guid,
+    projectGuid: project_guid,
+    projectSummaryGuid: project_summary_guid,
   };
 
   const tableDocuments =
-    initialValues?.documents?.map(
+    documents?.map(
```
taraepp, line 132: I think that having initialValues here was also making bugs (not showing files that weren't in initialValues)

services/common/src/setupTests.ts
```diff
@@ -23,7 +31,7 @@ jest.mock("@mds/common/providers/featureFlags/useFeatureFlag", () => ({
     isFeatureEnabled: () => true,
   }),
 }));
-
+window.scrollTo = jest.fn();
```
taraepp, line 34: I was getting so many console errors from window.scrollTo and useLayoutEffect that I ran out of terminal lines and couldn't see the tests that were actually failing! Mocking made that better.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -186,23 +186,17 @@ def proponent_project_id(self):
         return None
 
     @classmethod
-    def find_by_project_summary_guid(cls, project_summary_guid, is_minespace_user=False):
-        if is_minespace_user:
-            return cls.query.filter_by(
-                project_summary_guid=project_summary_guid, deleted_ind=False).one_or_none()
-        return cls.query.filter(ProjectSummary.status_code.is_distinct_from("DFT")).filter_by(
+    def find_by_project_summary_guid(cls, project_summary_guid):
```
taraepp, line 189: no longer filtering out DRAFT on CORE.

services/core-web/src/constants/routes.ts
```diff
@@ -195,21 +195,20 @@ export const MINE_PRE_APPLICATIONS = {
 };
 
 export const ADD_PROJECT_SUMMARY = {
-  route: "/mines/:mineGuid/project-description/new",
-  dynamicRoute: (mineGuid) => `/mines/${mineGuid}/project-description/new`,
+  route: "/mines/:mineGuid/project-description/new/:tab",
```
taraepp, line 198: changed both these routes to match the name on MS so I could use GLOBAL_ROUTES

services/minespace-web/src/tests/components/project/projectSummaryPage/ProjectSummaryPage.spec.tsx
```diff
@@ -8,7 +8,8 @@ import {
   REGIONS,
 } from "@mds/common/tests/mocks/dataMocks";
 import { PROJECTS, STATIC_CONTENT } from "@mds/common/constants/reducerTypes";
-import { ReduxWrapper } from "@/tests/utils/ReduxWrapper";
+import { ReduxWrapper as CommonWrapper } from "@mds/common/tests/utils/ReduxWrapper";
+import { ReduxWrapper as MSWrapper } from "@/tests/utils/ReduxWrapper";
```
taraepp, line 12: 🧀

services/common/src/components/projectSummary/DocumentUpload.tsx
```diff
@@ -63,15 +43,17 @@ export const DocumentUpload: FC<DocumentUploadProps> = (props) => {
   };
 
   useEffect(() => {
-    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "spatial_documents", []);
-    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "support_documents", []);
+    dispatch(change(FORM.ADD_EDIT_PROJECT_SUMMARY, "spatial_documents", []));
```
asinn134: Thanks for setting up the documents table to show the uploaded files without needing to go back to see them

services/common/src/components/projectSummary/AuthorizationSupportDocumentUpload.tsx
```diff
@@ -34,9 +31,7 @@ export const AuthorizationSupportDocumentUpload: FC<AuthorizationSupportDocument
     const newDocuments = documents.filter(
       (file) => fileToRemove.serverId !== file.document_manager_guid
     );
-    newDocuments.forEach((newDoc) => {
-      updateAmendmentDocuments(newDoc, code);
-    });
+    updateAmendmentDocuments(newDocuments);
```
taraepp, line 34: the changes to updateAmendmentDocuments is a fix for the bug Alex posted in the bugs channel today. Before it was updating a single document, not really having the power of _removing_ documents. Now it takes as its parameter the desired value of the field, allowing us to remove documents here.

services/common/src/components/projectSummary/DocumentUpload.tsx
```diff
@@ -63,15 +43,12 @@ export const DocumentUpload: FC<DocumentUploadProps> = (props) => {
...
-    props.change(FORM.ADD_EDIT_PROJECT_SUMMARY, "documents", [
-      ...spatial_documents,
-      ...support_documents,
-    ]);
+    dispatch(
+      change(FORM.ADD_EDIT_PROJECT_SUMMARY, "documents", [
+        ...spatial_documents,
+        ...support_documents,
+      ])
+    );
```
taraepp, line 51: Making sure that documents persist between user navigation events and are added to the table when uploaded. spatial & support variables are populated right in formatProjectSummary

services/common/src/models/documents/document.ts
```diff
@@ -163,10 +163,10 @@ export class MineDocument implements IMineDocument {
     _userRoles: string[] = [],
     is_latest_version: boolean = this.is_latest_version
   ) {
-    const canModify = is_latest_version && !this.is_archived;
-    if (!this.mine_document_guid) return [];
+    const canModify = is_latest_version && !this.is_archived && this.mine_document_guid;
+    const canView = this.file_type === ".pdf" && this.document_manager_guid;
```
taraepp, line 167: For cypress tests. So that the actions column can be there before we save the file (download/view actions require document_manager_guid, not mine_document_guid, but I kept the other actions from being allowed)

services/common/src/components/projectSummary/LegalLandOwnerInformation.spec.tsx
```diff
@@ -13,10 +14,12 @@ const mockStore = configureStore({
...
-
-const WrappedLegalLandOwnerInformation = reduxForm({
-  form: "ADD_EDIT_PROJECT_SUMMARY",
-})(LegalLandOwnerInformation);
+const formName = "ADD_EDIT_PROJECT_SUMMARY";
+const WrappedLegalLandOwnerInformation = (
+  <FormWrapper name={formName} onSubmit={() => {}}>
+    <LegalLandOwnerInformation />
+  </FormWrapper>
+);
```
henryoforeh-dev, line 22: Wondering if the ReduxWrapper can be used here for consistency.

services/common/src/components/forms/Step.tsx
matbusby-fw: More more more!

services/common/src/components/forms/SteppedForm.tsx
```diff
@@ -104,12 +133,29 @@ const SteppedForm = (props) => {
...
           <div className="stepped-form-form-container">
-            <Form layout="vertical">{children.find((child) => child.key === tabs[tabIndex])}</Form>
+            <FormWrapper
+              name={formName}
+              onSubmit={() => {}}
+              initialValues={initialValues}
+              isEditMode={isEditMode}
+              reduxFormConfig={{
+                touchOnBlur: true,
+                touchOnChange: false,
```
matbusby-fw: I'm wondering if it might be worth passing this value as a prop with a default of false.  I can see some potential scenarios where we would want touchOnChange to be true (even though we don't need it now)

taraepp: I'm thinking that when we get to the point of wanting variable configuration, to have it take in FormWrapperProps (or extend the interface) to allow you to override _everything_. But at this point it feels like over-engineering.

taraepp: I decided to do it.

services/common/src/components/projectSummary/LegalLandOwnerInformation.spec.tsx
taraepp: With the .getActions used in the test, can't use ReduxWrapper. This one was only updated to make it stop console logging errors about missing form context.

henryoforeh-dev: Yes, this is correct. getActions() is a method of the store object which is encapsulated in ReduxWrapper.

henryoforeh-dev (approved):

services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
```diff
@@ -49,46 +49,43 @@ import DocumentTable from "@mds/common/components/documents/DocumentTable";
...
-  status_code: string;
-  submission_date: string;
-  project_summary_description: string;
-  project_guid: string;
-  documents: IProjectSummaryDocument[];
-}
-
-const RenderEMAPermitCommonSections = ({ props }) => {
-  const { code } = props;
+const RenderEMAPermitCommonSections = ({ code, isAmendment, index }) => {
```
matbusby-fw, line 52: Love this change.  Much clearer

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -125,202 +105,103 @@ export const ProjectSummary: FC = () => {
...
-            { mrc_review_required, contacts, project_lead_party_guid },
+            { projectGuid },
+            { mrc_review_required: payload.mrc_review_required, contacts: payload.contacts },
             "Successfully updated project.",
             false
           )
         );
       })
-      .then(() => {
+      .then(async () => {
```
matbusby-fw, line 151: Do we need this async here?

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -125,202 +105,103 @@ export const ProjectSummary: FC = () => {
...
-      ...updatedAuthorizations,
-    };
-
-    delete payloadValues.authorizationTypes;
-    return payloadValues;
-  };
-
-  const handleUpdate = (message) => {
-    const { project_guid, mrc_review_required, contacts, project_lead_party_guid } = formValues;
+  const handleUpdateProjectSummary = async (payload, message) => {
```
matbusby-fw, line 129: The combo of .then and async await here feels a bit clunky.  Is it possible to make it all async/await?

matbusby-fw (approved):
Looking great.  Left one comment on maybe making one of the functions fully async/await rather than a mix of .then and async.

Some of the refactors are a bit tricky to track in a PR, but it all looks really solid!

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
taraepp: That's probably a good thing to do. I don't want to expand the scope _too_ much though, and want to unblock AMS work so I'll leave it as something to keep in mind for the future.

PR #3162, opened 2024-06-19
[CHORE] - Draft endpoints for W3C/UNTP Mines Act Permit Creds

Jsyro:
## Objective 

Separated UNTP models into their own package: https://github.com/Jsyro/untp-models

Three new endpoints for testing. 
1) AnonCred Equivalent W3C Credential using new acapy endpoints 
- can't use public did, so have to use private did:key, not usable in prod at the moment. 
2) AnonCred Equivalent W3C Credential using deprecated acapy endpoint
- Does what want, but marked for deprecation
3) New command to pre-process any untp credentials for publication. 

I've added a new environment variable: 

CHIEF_PERMITTING_OFFICER_DID_WEB="did:web:untp.traceability.site:parties:regulators:CHIEF-PERMITTING-OFFICER"

for now, the did:web will change in the near future, but this will work for POc


services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -87,6 +97,87 @@ def offer_newest_amendment_to_current_permittee(permit_amendment_guid: str,
...
+
+
+def publish_all_pending_vc_to_orgbook():
+    records_to_publish = PermitAmendmentOrgBookPublish.find_all_unpublished(unsafe=True)
+
+    for record in records_to_publish:
+        current_app.logger.warning("NOT sending cred to orgbook")
+        # resp = requests.post(ORGBOOK_W3C_CRED_POST, record.signed_credential)
+        # assert resp.status_code == 200, f"resp={resp.json()}"
+        current_app.logger.warning(record.signed_credential)
```
taraepp: ??
This looks maybe like it was a testing method? The commented out lines seem to be doing the work.

services/core-api/app/api/verifiable_credentials/models/orgbook_publish_status.py
```diff
@@ -0,0 +1,38 @@
...
+    orgbook_credential_id = db.Column(
+        db.String, nullable=True)                                  # not sure this will be able to be populated
+
+    def __repr__(self):
+        return f'<PermitAmendmentOrgBookPublishStatus unsigned_payload_hash={self.unsigned_payload_hash}, permit_amendment_guid={self.permit_amendment_guid}, sign_date={self.sign_date}, publish_state={self.publish_state}>'
+
+    @classmethod
+    def find_by_unsigned_payload_hash(cls,
+                                      unsigned_payload_hash,
+                                      *,
```
taraepp, line 30: What's up with the *?

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -103,3 +103,40 @@ def post(self):
...
+    def post(self):
+        if not is_feature_enabled(Feature.JSONLD_MINES_ACT_PERMIT):
+            raise NotImplementedError("This feature is not enabled.")
+        current_app.logger.warning("untp endpoint")
+
+        data = self.parser.parse_args()
+        permit_amendment = PermitAmendment.find_by_permit_amendment_guid(
+            data["permit_amendment_guid"])
+        if not permit_amendment:
+            raise BadRequest("Permit amendment not found")
```
taraepp, line 134: should this be a BadRequest or a NotFound? I'd lean towards NotFound. 🤷‍♀️

services/core-api/app/api/verifiable_credentials/models/orgbook_publish_status.py
Jsyro: https://www.vegardstikbakke.com/python-keyword-only/

It allows you force keyword arguments. It helps ensure no one ever accidentally uses 'unsafe' as a positional argument.

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: NotFound / 404 is for "Resource Not Found". 

If the permit_amendment_guid was in the URL, it would be 404, but given it's a payload processing thing, i believe 400 is more correct. 

Should permit_amdendment_guid be a URL parameter... maybe.

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: There is no API to publish to yet. that functionality needs to be added to Orgbook, so it's a stub for now.

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -87,6 +97,87 @@ def offer_newest_amendment_to_current_permittee(permit_amendment_guid: str,
...
 
 
+def process_all_untp_map_for_orgbook():
+    """Revoke the existing credential and offer a new one with the newest amendment."""
+    # find all parties connected to orgbook
+    # find all most recent permit_amendments for each permit they are permittee's on.
+    # for each permit_amendment, produce the untp_cc_map payload
+
+    # https://metabase-4c2ba9-prod.apps.silver.devops.gov.bc.ca/question/2937-permit-amendments-for-each-party-orgbook-entity
+    permit_amendment_query_results = db.session.execute("""
```
Jsyro, line 116: It was way easier for me to construct this from Metabase and bring this in. 

I can look at making this object relationships if that is preferred (this SQL query will be much faster though), but performance is not a huge concern.

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +288,116 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
         }
         return credential
+
+    @classmethod
+    def produce_untp_cc_map_payload(cls, did: str, permit_amendment: PermitAmendment):
+        """Produce payload for Mines Act Permit UNTP Conformity Credential from permit amendment and did."""
+        ANONCRED_SCHEME = "https://hyperledger.github.io/anoncreds-spec/"
+
+        curr_appt = permit_amendment.permittee_appointments[0]
+        for pa in permit_amendment.permittee_appointments:
```
simensma-fresh: Any reason to do this comparison again? Just seeing as `order_by start_date desc` was already added to the new  permittee_appointments relationship

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +288,116 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+                        ]),
+                    effectiveDate=datetime(2024, 5, 14, tzinfo=pytz.timezone("UTC")).isoformat()),
+                subjectFacilities=[facility],
+                subjectProducts=products,
+                sustainabilityTopic=codes.SustainabilityTopic.Governance_Compliance)
+        ]
+        issue_date = permit_amendment.issue_date
+        issuance_date_str = datetime(
+            issue_date.year, issue_date.month, issue_date.day, 0, 0, 0,
+            tzinfo=pytz.timezone("UTC")).isoformat()
```
simensma-fresh: According to Sonarcloud, it's preferred to use `datetime.tzinfo` instead of `pytz.timezone` when using `datetime` as there's a slight difference in how timezones are handled between them, hard to say if it'll actually have an impact here. https://sonarcloud.io/project/issues?resolved=false&pullRequest=3162&id=bcgov-sonarcloud_mds_core-api&open=AZAyy_vxY_vcUsn0Ofd6&tab=why

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +288,116 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+            cc.ConformityAssessment(
+                referenceRegulation=cc.Regulation(
+                    id="https://www.bclaws.gov.bc.ca/civix/document/id/complete/statreg/96293_01",
+                    name="BC Mines Act",
+                    issuingBody=base.Party(
+                        name="BC Government",
+                        identifiers=[
+                            base.Identifier(identifierValue="did:indy:candy:LTNyw5R14J66CrF7tmV3i8")
+                        ]),
+                    effectiveDate=datetime(2024, 5, 14, tzinfo=pytz.timezone("UTC")).isoformat()),
```
simensma-fresh: Is the effectiveDate supposed to be hardcoded?

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +288,116 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+
+        class W3CCred(BaseModel):
+            model_config = ConfigDict(
+                populate_by_name=True, json_encoders={datetime: lambda v: v.isoformat()})
+
+            context: List[Union[str, dict]] = Field(alias="@context")
+            type: List[str]
+            issuer: dict[str, str]
+            issuanceDate: str
+            credentialSubject: cc.ConformityAttestation
```
simensma-fresh: Any reason for keeping the class inside the function? Thinking it might be good to move it out of there in case we want to reference it elsewhere

simensma-fresh (changes requested):

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: that warning recommends using zoneinfo package if python version 3.9 or greater. we are using python 3.8 right?

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: yes. that date is date the Mine Act was Passed. however now the date shown in the webpage has changed again..... so maybe we need a different date. 

https://www.bclaws.gov.bc.ca/civix/document/id/complete/statreg/96293_01

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: no, but i should probably move that to the library (or find a library that already implements the w3c vc data model). 

https://www.w3.org/TR/vc-data-model-2.0/

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: That was so bad that was confusing and wrong at the same time. See change and comment.

services/core-api/app/api/verifiable_credentials/manager.py
simensma-fresh: We're actually on python 3.11 😄

services/core-api/app/api/verifiable_credentials/manager.py
simensma-fresh: That makes sense!

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: ok, well i rebuild my venv and am using zoneinfo now.

simensma-fresh (dismissed):

simensma-fresh (dismissed):

services/core-api/app/api/services/traction_service.py
```diff
@@ -147,22 +148,19 @@ def sign_jsonld_credential_deprecated(
         self,
         did,
         verkey,
-        credential,
-    ):
+        credential: BaseModel,
+    ) -> dict:
         options = {"verificationMethod": did + "#verkey", "proofPurpose": "assertionMethod"}
```
PatStLouis, line 154: We need to be careful with hard coding the `verficationMethod` id, might want to provide the whole verification method instead of the did + hard coded id

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -11,18 +18,30 @@
...
 
 
+#this should probably be imported from somewhere.
+class W3CCred(BaseModel):
+    model_config = ConfigDict(
+        populate_by_name=True, json_encoders={datetime: lambda v: v.isoformat()})
+
+    context: List[Union[str, dict]] = Field(alias="@context")
+    type: List[str]
+    issuer: dict[str, str]
```
PatStLouis: issuer can be a string or a dict

PatStLouis:
```suggestion
    issuer: Union[str, Dict[str, str]]
```

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -11,18 +18,30 @@
...
+    issuanceDate: str
```
PatStLouis, line 42:
```suggestion
    # TODO: update to `validFrom` for vcdm 2.0 once available in aca-py/traction, which is an optional property
    issuanceDate: str
```

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +303,104 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+                base.Identifier(
+                    scheme=ANONCRED_SCHEME,
+                    identifierValue=str(orgbook_entity.registration_id),
+                    identifierURI=orgbook_cred_url)
+            ])
+
+        facility = cc.Facility(
+            name=permit_amendment.mine.mine_name,
+            geolocation=
+            f'https://plus.codes/{plus_code_encode(permit_amendment.mine.latitude, permit_amendment.mine.longitude)}',
```
PatStLouis, line 354: nice

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +303,104 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+                "https://github.com/bcgov/bc-vcpedia/blob/main/credentials/bc-mines-act-permit/1.1.1/governance.md",
+                name="BC Mines Act Permit Credential (1.1.1) Governance Documentation"),
+            issuedBy=untp_party_cpo,
+            issuedTo=untp_party_business,
+            validFrom=issuance_date_str,                                                                                                                                   #shouldn't this just be in the w3c wrapper
+            assessments=untp_assessments)
+
+        w3c_cred = W3CCred(
+            context=["https://www.w3.org/2018/credentials/v1", {
+                "@vocab": "urn:bcgov:attributes#"
```
PatStLouis, line 400: Lets start thinking about a context file/vocab for the mines stuff

services/core-api/app/commands.py
```diff
@@ -72,17 +66,13 @@ def _create_cypress_data():
...
+            mine = MineFactory(major_mine_ind=True, mine_name='Evergreen Cypress Mine')
 
-            ## Create a minespace user with data corresponding to 
+            ## Create a minespace user with data corresponding to
             ## the Cypress test user (cypress/keycloak-users.json)
             minespace_user = MinespaceUserFactory(
                 email_or_username='cypress@bceid',
-                keycloak_guid='a28dfc3a-5e5c-4501-ab2f-399d8e64f2c8'
-            )
+                keycloak_guid='a28dfc3a-5e5c-4501-ab2f-399d8e64f2c8')
```
PatStLouis, line 75: could this be a variable instead of a hard coded value?

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: true... next PR. 

I'll put that in the backlog.

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -195,3 +304,104 @@ def produce_map_01_credential_payload(cls, did: str, permit_amendment: PermitAme
...
+                subjectProducts=products,
+                sustainabilityTopic=codes.SustainabilityTopic.Governance_Compliance)
+        ]
+        issue_date = permit_amendment.issue_date
+        issuance_date_str = datetime(
+            issue_date.year, issue_date.month, issue_date.day, 0, 0, 0,
+            tzinfo=ZoneInfo("UTC")).isoformat()
+
+        cred = cc.ConformityAttestation(
+            id="http://example.com/attestation/123",
```
taraepp, line 384: what's with the example.com URL? Mostly asking because SonarCloud is upset that it's not https

Jsyro: That will the URL where the credential will be hosted, but we have no web accessible endpoint for these credentials, so it's a placeholder value.

simensma-fresh (approved):

taraepp (approved):

PR #3170, opened 2024-06-26
[MDS-5883] retry ams failure

henryoforeh-dev:
## Objective 

[MDS-5883](https://bcmines.atlassian.net/browse/MDS-5883)

Implemented a feature that allows users to view and resend failed AMS submissions.

image


services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const transformedProjectSummaryAuthorizationTypes = useSelector(
+    getTransformedProjectSummaryAuthorizationTypes
+  );
+  const dropdownProjectSummaryPermitTypes = useSelector(getDropdownProjectSummaryPermitTypes);
+  const notApplicableText = "N/A";
+
+  let processedOtherActPermitResult: any[] = [];
+
+  const createStatusColumn = (text: string) => ({
+    title: "Status",
```
taraepp: column should have a key attribute. If it's going to be for submitted or failed, probably good to use that for the key. (Either a key or a dataIndex but the key makes more sense here)

services/common/src/interfaces/projects/authorizationSummary.interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface IAuthorizationSummaryColumn {
```
taraepp: I think it would make more sense to have the interface just be for "IAuthorizationSummary" (as it's the record type that the column is displaying and could be reused in other places with authorization summaries).

services/minespace-web/src/components/dashboard/mine/projects/Projects.tsx
```diff
@@ -18,6 +19,7 @@ export const Projects: FC = () => {
   const projects = useSelector(getProjects);
 
   const handleFetchData = () => {
+    dispatch(fetchPermits(mine.mine_guid));
     dispatch(fetchProjectsByMine({ mineGuid: mine.mine_guid })).then(() => {
```
taraepp: These two fetches look like they want to be in a Promise.all so that it gets set as loaded after both are completed.

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.spec.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { shallow } from "enzyme";
+import { ProjectDescriptionTab } from "./ProjectDescriptionTab";
+import FeatureFlagContext from "@mds/common/providers/featureFlags/featureFlag.context";
+
+describe("ProjectDescriptionTab", () => {
+  it("renders properly", () => {
+    const component = shallow(
+      <FeatureFlagContext.Provider
+        value={{
+          isFeatureEnabled: () => true,
```
taraepp: We do already have in MS setupTests.js this function mocked. Not a big deal but it shouldn't be necessary.

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
+import React, { useEffect, useState } from "react";
+import { Row, Col, Typography, Button, Alert, Badge } from "antd";
+import Callout from "@/components/common/Callout";
```
taraepp: the Callout did get moved to common if you wanted to use that one

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
...
+import { IAuthorizationSummaryColumn } from "@mds/common/interfaces";
+import { Link } from "react-router-dom";
+
+import {
+  getDropdownProjectSummaryPermitTypes,
+  getTransformedProjectSummaryAuthorizationTypes,
+} from "@mds/common/redux/selectors/staticContentSelectors";
+import { ColumnsType } from "antd/es/table";
+import * as routes from "@/constants/routes";
+import { useHistory } from "react-router-dom";
```
taraepp: I'm seeing multiple imports from the same source- there's another one from react-router-dom that can be combined

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.spec.tsx
taraepp: I think it would be worth switching out the shallow for the react-testing-library `render`. This one doesn't seem to be generating much of a snapshot.

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
...
+              style={{ marginBottom: "12px" }}
+            />
+          )}
+          <CoreTable dataSource={environmentalManagementActData} columns={otherActColumns} />
+          <br />
+          <Typography.Title level={5}>Water Sustainability Act</Typography.Title>
+          <CoreTable dataSource={waterSustainabilityActData} columns={otherActColumns} />
+          <br />
+          <Typography.Title level={5}>Forestry Act</Typography.Title>
+          <CoreTable dataSource={forestryActData} columns={otherActColumns} />
```
taraepp: It's not a required attribute for CoreTable because it has a default of "key", but it is good to set the rowKey attribute if the data doesn't have a `key` attribute. You may be getting React key errors in the console.

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
+import React, { useEffect, useState } from "react";
```
asinn134: For the new project description overview tab I think it is meant to be for both Core and Minespace? I am noticing the ticket doesn't specify, but I know the new submission page I added was for both Core and Minespace and is meant to redirect to the project description overview tab.

May have to check with Rebecca or Denise I think?

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,279 @@
...
+            <Alert
+              message={"Submission Failed Please Retry"}
+              showIcon
+              type="error"
+              description={
+                "One or more of your environment authorization application has not been submitted successfully. Please retry the submission."
+              }
+              action={
+                <Link to={routes.ADD_PROJECT_SUMMARY.dynamicRoute(project.mine_guid)}>
+                  <Button>Retry Failed Submission</Button>
```
simensma-fresh: Just to confirm - I thought this was supposed to actually retry the failed jobs, not take you back to the detail page (AC 7, 8, 9 of the ticket)

simensma-fresh (changes requested):

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+            <br />
+            <Typography.Title level={5}>Water Sustainability Act</Typography.Title>
+            <CoreTable
+              rowKey="project_summary_authorization_guid"
+              dataSource={waterSustainabilityActData}
+              columns={otherActColumns}
+            />
+            <br />
+            <Typography.Title level={5}>Forestry Act</Typography.Title>
+            <CoreTable dataSource={forestryActData} columns={otherActColumns} />
```
taraepp: just this instance of the table here is missing the rowKey. 😅 
I'm assuming it's also `project_summary_authorization_guid`

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+import {
+  updateProjectSummary,
+  fetchProjectById,
+} from "@mds/common/redux/actionCreators/projectActionCreator";
+import { isArray } from "lodash";
+import { removeNullValuesRecursive } from "@mds/common/constants/utils";
+import Loading from "@mds/common/components/common/Loading";
+
+export const ProjectDescriptionTab = () => {
```
taraepp: Just noticed this- there's `export const ProjectDescriptionTab` up here, and then `export default ProjectDescriptionTab` at the bottom. Don't need both!
I'd go with the export at the bottom so you can import it like `import ProjectDescriptionTab from <path>`

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+  const getPermitNumber = (permit_guid: string): string => {
+    const permit = permits.find(({ permit_guid: id }) => id === permit_guid);
+    return permit?.permit_no ?? notApplicableText;
+  };
+
+  const loadMinesActPermitData = (authorizations) => {
+    const result = authorizations
+      .filter(
+        (authorization) => authorization.project_summary_authorization_type === "MINES_ACT_PERMIT"
```
simensma-fresh: Would be good to put the project summary authorization type into an enum if we don't already have one

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+      )
+      .map((authorization) => {
+        const dateSubmitted = formatDateTimeTz(authorization.ams_submission_timestamp);
+        const projectType = parseProjectTypeLabel(authorization?.project_summary_permit_type[0]);
+        const permitNo =
+          authorization?.project_summary_permit_type[0] === "AMENDMENT"
```
simensma-fresh: Would be good to put the project summary permit type into an enum if we don't already have one

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+    processedOtherActPermitResult = [];
+  };
+
+  const loadOtherActPermitData = (authorizations) => {
+    const environmentalManagementActTypes = [
+      "AIR_EMISSIONS_DISCHARGE_PERMIT",
+      "EFFLUENT_DISCHARGE_PERMIT",
+      "REFUSE_DISCHARGE_PERMIT",
+      "MUNICIPAL_WASTEWATER_REGULATION",
+    ];
```
simensma-fresh: Could we put these and the constants below into enums so they can be shared for future use?

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+    if (output.MINES_ACT_PERMIT) {
+      output.MINES_ACT_PERMIT = input.filter(
+        (auth) => auth.project_summary_authorization_type === "MINES_ACT_PERMIT"
+      );
+    }
+
+    return output;
+  };
+
+  const handleRetryAMSSubmissionClicked = async () => {
```
simensma-fresh, line 323: There seems to be a lot of logic here just to retry the submission to AMS. Any reason not to set up an endpoint in the API and re-trigger the jobs with the data we already have?

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+      // Normalize facility operator's address
+      if (Array.isArray(payload.facility_operator.address)) {
+        payload.facility_operator.address =
+          payload.facility_operator.address.length === 0
+            ? null
+            : payload.facility_operator.address[0];
+      }
+      await dispatch(
+        updateProjectSummary(
```
simensma-fresh, line 357: I believe this endpoint also sends notifications - they should probably not be sent when retrying the jobs as we're not really making any other changes to the project summary

simensma-fresh (changes requested):

services/minespace-web/src/components/pages/Project/ProjectDescriptionTab.tsx
asinn134: The retry should be working as intended now

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: the rowKey has been added now

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: The export at the top has been removed now.

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: Enum for this has been created now

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: Same here, enum has been created

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: enum has been created for this as well

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: the notification has been turned off now when a retry is attempted

henryoforeh-dev (commented):
Reviewed the last 3 commits. Looks good to me. I am not able to approve since I also worked on this.

matbusby-fw (dismissed):

services/common/src/components/common/Loading.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const isCore = systemFlag === SystemFlagEnum.core;
+  const defaultOptions = {
+    loop: true,
+    autoplay: true,
+    animationData: isCore ? coreLoader : mineSpaceLoader,
+  };
+  return (
+    <div id="loading-screen">
+      <div id="loader">
+        <Lottie options={defaultOptions} />
```
taraepp: This looks a little strange to me!
It looks like the coreLoader/mineSpaceLoader are SVGs pretending to be JSON files.
Lottie also gives us a million deprecation warnings.
What if we just use the antd `<Spin />` like we have in `LoadingWrapper.js`? Currently it gets passed the colour of the spinner, but that seems a lot simpler as the change variable between MS/CORE (assuming that we can't just CSS it)

simensma-fresh: ☝️ Also vote for not introducing lottie again, it's a big library and way overkill for a loading-spinner 😅

simensma-fresh (changes requested):

services/minespace-web/src/components/pages/Project/ProjectPage.js
```diff
@@ -227,13 +242,18 @@ export class ProjectPage extends Component {
...
+                defaultActiveKey={this.props.match.params?.tab || tabs[0]}
                 onChange={(activeTab) => this.handleTabChange(activeTab, irtStatus)}
                 type="card"
               >
                 <Tabs.TabPane tab="Overview" key="overview">
                   <ProjectOverviewTab navigateForward={this.navigateFromProjectStagesTable} />
                 </Tabs.TabPane>
+                {this.props.isFeatureEnabled(Feature.AMS_AGENT) && (
+                  <Tabs.TabPane tab="Project Description" key="project-description">
+                    <ProjectDescriptionTab />
```
taraepp, line 254: Sorry, only noticed this now that I'm working in the same area. 😅 
BUT- this component, and the CORE equivalent, are the parent of these Tabs, right? And this parent always has the project loaded if `isLoaded`. 
I think you can just pass the project as a prop then and take the fetch out from the child component. One less fetch!

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -0,0 +1,521 @@
...
+                project_summary_permit_type: [AMS_AUTHORIZATION_TYPES.AMENDMENT],
+              })
+            )
+          );
+        }
+      }
+    });
+    return {
+      authorizations: updatedAuthorizations,
+      ams_authorizations: { amendments: amendAmsAuthorizations, new: newAmsAuthorizations },
```
taraepp: How similar is this function to the one in ProjectSummaryForm? If we're copy/pasting, perhaps it's time to make it a helper function.

services/common/src/components/common/Loading.tsx
asinn134: Sounds good, lottie is taken out now and uses the antd `<Spin />`

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: This is changed to a helper function now. Thanks for the help with this.

services/minespace-web/src/components/pages/Project/ProjectPage.js
asinn134: Wait maybe my understanding is incorrect but I think these parent components only fetch on `componentDidMount`? Also I think we still need the project fetch inside ProjectDescriptionTab when a retry is attempted so we can also dispatch the changes.

simensma-fresh (dismissed):

simensma-fresh:
Oh just noticed there's a failed cypress test 😅

services/core-web/cypress/e2e/createproject.cy.ts
```diff
@@ -38,7 +38,6 @@ describe("Major Projects", () => {
 
     // SAVE & CONTINUE - skip to Purpose & Authorization
     cy.contains("Save & Continue").click({ force: true });
-    cy.contains("Project Lead", { timeout: 10000 });
```
asinn134, line 41: I couldn't find in Core an element with Project Lead in Major Project Summary section so I ended up removing this test that was causing an error.

henryoforeh-dev (commented):
Looks good to me. I am not able to approve because I also worked on it.

taraepp (approved):

simensma-fresh (approved):

simensma-fresh (approved):

PR #3174, opened 2024-07-03
[MDS-6013] BE spatial validation

matbusby-fw:
## Objective 

- Created migrations for core and docman to add document bundles
- Created endpoint in docman to `complete-bundle`
  - This will run validation on a set of document_guids (from the docman db) and pass them through geomark.  It will return the geomark id and the document_bundle_guid
- The endpoint will save a document_bundle on the backend and add a reference to this bundle to each of the documents referenced.
- Currently created geomarks will expire after a set amount of time (90 days I think)
  - Rebecca and I met with Samson Lee to discuss options here.  We're following up on these potentialities including the following:
    - Not saving the geomark id, and instead for future usage making an api call to MPSD to retrieve the mapping information required to view items
    - Getting a key/permissions to create non-expiring geomarks.  This may be preferable as it would have less reliance on the GIS team.
- Updated accepted file types on the frontend to accomodate all mime-types required for spatial bundles.

[MDS-6013](https://bcmines.atlassian.net/browse/MDS-6013)


services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +292,145 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+            doc.document_bundle = bundle
+            if geomark_response.get('url'):
+                doc.upload_completed_date = datetime.utcnow()
+            db.session.add(doc)
+
+        db.session.add(bundle)
+        db.session.commit()
+
+        if bundle.error:
+            raise BadRequest(bundle.error)
```
taraepp, line 435: do we know that this is a bad request?

taraepp:
should probably add BE tests

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -0,0 +1,15 @@
...
+from app.api.utils.models_mixins import SoftDeleteMixin, AuditMixin, Base
+
+
+class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
+    __tablename__ = 'mine_document_bundle'
+
+    bundle_id = db.Column(UUID(as_uuid=True), primary_key=True)
+    bundle_guid = db.Column(UUID(as_uuid=True), nullable=False)
+    bundle_name = db.Column(db.String(100), nullable=False)
+    name = db.Column(db.String(100), nullable=False)
```
simensma-fresh: Think the max length is defined as 300 in the migration for name and geomark_id

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -0,0 +1,15 @@
...
+from app.extensions import db
+from app.api.utils.models_mixins import SoftDeleteMixin, AuditMixin, Base
+
+
+class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
+    __tablename__ = 'mine_document_bundle'
+
+    bundle_id = db.Column(UUID(as_uuid=True), primary_key=True)
+    bundle_guid = db.Column(UUID(as_uuid=True), nullable=False)
+    bundle_name = db.Column(db.String(100), nullable=False)
```
simensma-fresh: Do we need both a `name` and a `bundle_name`?

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +292,145 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+        cls.validate_bundle(bundle_documents)
+
+        if len(bundle_documents) > 1:
+            file_path = f'{name}.shpz'
+            cls.zip_spatial_files(bundle_documents, name)
+        else:
+            file_path = bundle_documents[0].file_display_name
+            cls.download_kml_kmz_files(bundle_documents[0])
+
+        geomark_response = cls.send_spatial_file_to_geomark(file_path)
```
simensma-fresh, line 410: I see there's a cleanup call to remove the downloaded file at the end of the `send_spatial_file_to_geomark` method - however I think there's still a chance that the file is not removed if the call to geomark fails. I'm wondering if we could wrap this in a `try / finally` block to make sure it gets cleaned up regardless of the outcome.

```python
try:
    geomark_response = cls.send_spatial_file_to_geomark(file_path)
finally:
    os.remove(file_path)
```

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +292,145 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+        return shpz_path
+    @classmethod
+    def download_kml_kmz_files(cls, doc):
+        # Download the file from object store
+        response = ObjectStoreStorageService().download_file(doc.object_store_path, doc.file_display_name, False)
+
+        # Check the response status
+        if response.status_code == 200:
+            file_data = response.get_data()  # obtain file data from response body
+            with open(doc.file_display_name, 'wb') as file:
```
simensma-fresh: I'm wondering if we could sanitize this file path (e.g using `secure_filename`) here and for the `zip` file writing above Just seeing as `name` is "user provided" https://werkzeug.palletsprojects.com/en/3.0.x/utils/#werkzeug.utils.secure_filename

simensma-fresh: And could we maybe put these in a separate folder, just to avoid overwriting any other existing files. Maybe putting them in something like `/tmp/spacial`?

simensma-fresh (changes requested):

services/document-manager/backend/app/docman/utils/document_upload_helper.py
matbusby-fw: It's about the best we can do I think.  Unless there's a server error on their side they actually return a 200 status code with a response formatted like this:

```
{
    "resultFormat": [
        "json"
    ],
    "format": [
        "shpz"
    ],
    "body_Error": "Geometry is outside the valid area for coordinate system Unknown (0). Check the geometry and coordinate system.",
    "error": "Geometry is outside the valid area for coordinate system Unknown (0). Check the geometry and coordinate system."
}
```

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +293,153 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+                }
+                # Send the request
+                response = requests.post(
+                    url,
+                    data=multipart_data,
+                    headers=headers
+                )
+        finally:
+            # Delete the file
+            os.remove(file_path)
```
simensma-fresh, line 388: I think we'll have the same issue here with a storage leak that the downloaded file will still be present if the `assert` above fails.

Possible to wrap this whole method to be safe?

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +293,153 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+
+            extra_extensions = document_extensions - allowed_extensions
+            if extra_extensions:
+                raise ValueError(
+                    f'Found non spatial bundle file types in spatial bundle: {", ".join(extra_extensions)}')
+
+    @classmethod
+    def zip_spatial_files(cls, bundle_documents, name):
+        oss_service = ObjectStoreStorageService()
+        shpz_path = secure_filename(f'{name}.shpz')
```
simensma-fresh: Could we also put this into the `/tmp` folder?

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -285,3 +293,153 @@ def complete_upload(cls, key, new_key, doc_guid, versions=None, version_guid=Non
...
+        # Download the file from object store
+        response = ObjectStoreStorageService().download_file(doc.object_store_path, doc.file_display_name, False)
+
+        # Check the response status
+        if response.status_code == 200:
+            file_data = response.get_data()  # obtain file data from response body
+
+            dir_path = 'tmp/spatial'
+            os.makedirs(dir_path, exist_ok=True)
+            file_path = os.path.join(dir_path, doc.file_display_name)
```
simensma-fresh: Could we also do the `secure_filename` on this one? 
Just good practice to do with any user provided names of files to avoid any chance of path traversal attacks

simensma-fresh (changes requested):

migrations/sql/V2024.06.17.12.00__add_mine_document_bundle_table.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS mine_document_bundle (
```
asinn134: I just merged a .sql file with the date V2024.07.04.13.26, might need to adjust the date for this sql file I think.

simensma-fresh (dismissed):

migrations/sql/V2024.06.17.12.00__add_mine_document_bundle_table.sql
matbusby-fw: Oh, good catch!  Will update the date.

simensma-fresh (approved):

asinn134 (approved):

PR #3176, opened 2024-07-04
[MDS-6009] Altering unique constraint of permit_permit_no_unique in permit table

asinn134:
## Objective 

[MDS-6009](https://bcmines.atlassian.net/browse/MDS-6009)

- Adjusting the unique constraint of permit_no column in the permit table so that a soft deleted permit_no will not prevent us from creating a new permit with the same permit_no. (kudos to Mat for the idea)


matbusby-fw (approved):
🙌🏻

taraepp (approved):

PR #3177, opened 2024-07-08
[MDS-6064] Removed changing default cipher for NRIS download

simensma-fresh:
## Objective 

[MDS-6064](https://bcmines.atlassian.net/browse/MDS-6064)

DEFAULT_CIPHERS is no longer a property on .ssl_ causing the following error:
Tested out downloads locally without it and they work fine without any ssl related errors.

image


matbusby-fw (approved):
🌈

PR #3181, opened 2024-07-10
[MDS-6014] Spatial File FE

taraepp:
## Objective 
- show spatial files in a table
- upload spatial files in a modal
   - validate on FE
- show the geomark on the map in a view modal

## Misc. changes:
- change to map component to show geojson _features_
   - there were some issues with the timing of the call to `fitBounds` and with no minimum height set _in_ the component, issues rendering without  a container (see in action: look at the legal land owner page on the form on test, and delete the other elements in the same row. With no sibling elements, it will disappear)
- modal wrapper/add party component form: upgraded to tsx and props tidied up (there were unused props)
   - extracted the main modal logic to a common component, used to wrap around separate CORE/MS implementations 
- RenderFileUpload: dealt with an extension having the potential for multiple MIME-types. Cleared up variable ambiguity (lots of eslint warnings about reusing names). Destructured props, deleted unused, and attached the prop type to the component. Fixed bug where uppercase extensions weren't interpreted as valid. Disallowed empty files (everywhere that the common component is used). But a major takeaway here is that **RenderFileUpload will now yell at you if you give it bad props**
- I initially considered using the DocumentTable. That didn't go very well, but I did take out the bit where DocumentTable took DocumentTable as a prop.

[MDS-6014](https://bcmines.atlassian.net/browse/MDS-6014)


services/common/src/redux/createAppSlice.ts
```diff
@@ -4,3 +4,8 @@ import { buildCreateSlice, asyncThunkCreator } from "@reduxjs/toolkit";
 export const createAppSlice = buildCreateSlice({
   creators: { asyncThunk: asyncThunkCreator },
 });
+
+export const rejectHandler = (action) => {
+  console.log(action.error);
+  console.log(action.error.stack);
```
simensma-fresh, line 10: Intentional to leave the console logs here?

services/document-manager/backend/Dockerfile
```diff
@@ -17,4 +17,6 @@ EXPOSE 5001
 
 RUN mkdir -p /tmp/celery
 
+RUN mkdir -p /tmp/spatial
```
simensma-fresh: I see that this doesn't match exactly the folder the assets are downloaded to. We should probably add the missing `/` to the docman paths

image

simensma-fresh (dismissed):
Loving all the cleanup! 🧹 

Two small comments, but looks great!
(and a merge conflict)

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+    },
+  ];
+
+  const isFinalStep = currentStep === stepContent.length - 1;
+
+  return (
+    <FormWrapper
+      isModal
+      name={modalFormName}
+      onSubmit={async (values) => {
```
matbusby-fw: Not a huge deal, but I this function is getting a bit large to live inline.  Might be nice as it's own function.

services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -67,44 +64,54 @@ interface FileUploadProps extends BaseInputProps {
...
 
   beforeAddFile?: (file?: any) => any;
   beforeDropFile?: (file?: any) => any;
   onProcessFiles?: () => void;
-  onRemoveFile?: () => void;
+  onRemoveFile?: (error, file) => void;
   addFileStart?: () => void;
 }
 
-const defaultProps = {
```
matbusby-fw, line 80: 🙌🏻

services/common/src/components/forms/RenderFileUpload.tsx
```diff
@@ -585,12 +592,4 @@ export const FileUpload = (props: FileUploadProps) => {
...
-const mapDispatchToProps = (dispatch) =>
-  bindActionCreators(
-    {
-      pollDocumentUploadStatus,
-    },
-    dispatch
-  );
-
-export default withFeatureFlag(connect(null, mapDispatchToProps)(FileUpload));
+export default FileUpload;
```
matbusby-fw, line 595: Nice solution on this file as a whole!  Happy to see some of the clean up too!

services/common/src/redux/createAppSlice.ts
taraepp: This one yeah. We should probably have a _better_ handler for errors, but it's what we're currently doing in slices and I thought it better to at least make it a _shared_ function.

services/document-manager/backend/Dockerfile
matbusby-fw: I was working with this in my PR too, but from what I can tell the folder doesn't get persisted to the resulting container after this command (/tmp/celery also doesn't exist if you exec into the container after creation)

matbusby-fw: In my PR I added the creation in the init file for docman.

matbusby-fw (dismissed):
Looks great.  A couple thoughts/suggestions, but it's looking nice and clean!

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
taraepp: Agreed. Extracted a function!

services/document-manager/backend/Dockerfile
taraepp: I just took this stuff out from both Dockerfiles. 👍

matbusby-fw (approved):

services/common/src/components/documents/DocumentCompression.tsx
```diff
@@ -13,18 +11,23 @@ import DocumentCompressionWarningModal from "./DocumentCompressionWarningModal";
 import DocumentCompressedDownloadModal from "./DocumentCompressedDownloadModal";
 
 interface DocumentCompressionProps {
-  documentType: string;
```
asinn134, line 16: Thanks for cleaning up this file.

simensma-fresh (approved):

PR #3171, opened 2024-06-26
[MDS-5944] application submission success fail page

asinn134:
## Objective 

[MDS-5944](https://bcmines.atlassian.net/browse/MDS-5944)

- Added submission status page in common package to display if authorization jobs have succeeded or failed. The page's layout is determined by the ams_status_code of the authorizations, if any authorization has a "400" ams_status_code then we will display the fail page, else the success page will be displayed

Screenshot (363)

Screenshot (366)

- Made some styling changes based on suggestions by Denise

     1.  The input text stating a file upload field is required appears at the bottom of the file uploader and there is a bit of a gap between the text and the actual file uploader it is referring to. The change was to move the text closer to the uploader it refers to. 
Screenshot (364)

     2. Fixed a css issue in Major project's Purpose and Authorization section where the width would stretch depending on the length of the file's name being uploaded.

Screenshot (367)

PR #3183, opened 2024-07-10
[MDS-5944 ] application submission success fail page

asinn134:
## Objective 

[MDS-5944 ](https://bcmines.atlassian.net/browse/MDS-5944 )

-This contains the same changes found in the previous pr (https://github.com/bcgov/mds/pull/3171)
-There are a few other changes, but the biggest ones are around the `facility_coords_source` property. It was found that when we make a submission to the AMS API, we would keep getting a 400 status code error with the message `Business Errors Exist: Queued errors exist`. 

The cause turned out to be because in the Major project's Location Access and Land Use section, the Source of Data drop down, when say for example Survey is selected, we are actually sending the value "SUR" instead of "Survey" and AMS API only accepts the following values for the `facility_coords_source` property:

Screenshot (374)


services/core-web/src/styles/components/Callout.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.bcgov-callout--success>p,
+.bcgov-callout--warning>p,
+.bcgov-callout--danger>p {
+    margin-bottom: 0;
+}
+
+.bcgov-callout h1,
+h2,
```
taraepp: I see you're bringing this file over to CORE!

An issue that I noticed with this on Minespace: this selector doesn't work as intended. They meant for every `h1`, `h2`, etc, _within_ `.bcgov-callout` to be selected, but what actually happened was they selected the `h1` within bcgov-callout and also _every single other_ `h2`-`h4`! I changed the selector, and then made some other changes that sonarcloud or eslint wanted, in the PR that I put up today.

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -130,20 +137,42 @@ export const ProjectSummary: FC = () => {
...
+    );
+
+    await handleFetchData();
+    if (
+      tab === PROJECT_SUMMARY_WITH_AMS_SUBMISSION_SECTION &&
+      amsFeatureEnabled &&
+      projectSummaryResponse
+    ) {
+      const { data } = projectSummaryResponse;
+      const authorizations = data?.authorizations;
```
taraepp: should probably be `data?.authorizations ?? []` for the array operations below

taraepp (commented):
Looks good overall. 👍 
Couple minor things.

services/core-web/src/styles/components/Callout.scss
asinn134: Thanks for the heads up. I adjusted this to be aligned with the change you made in your pr.

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
asinn134: Thanks, this has been adjusted now

taraepp (approved):

simensma-fresh (approved):

PR #3184, opened 2024-07-12
[MDS-6074] - BE spatial bugfixes

matbusby-fw:
- Changed bundle_id column in mine_document_bundle to a serial from a guid
- A few other updates to the mine_document_bundle table- Added a mine_document_bundle and mine_document_bundle_id to the mine_document model
- Add response a response model for mine_document_bundle and included it in the mine_document response model

### DIFF
--- a/services/core-api/app/api/mines/response_models.py
+++ b/services/core-api/app/api/mines/response_models.py
@@ -125,6 +125,16 @@ INFORMATION_REQUIREMENTS_TABLE_DOCUMENT_XREF_MODEL = api.model(
         'information_requirements_table_document_type_code': fields.String
     })
 
+MINE_DOCUMENT_BUNDLE_MODEL = api.model(
+    'MineDocumentBundle', {
+        'bundle_id': fields.Integer,
+        'bundle_guid': fields.String,
+        'name': fields.String,
+        'geomark_id': fields.String,
+        'docman_bundle_guid': fields.String
+    }
+)
+
 MINE_DOCUMENT_MODEL = api.model(
     'MineDocument', {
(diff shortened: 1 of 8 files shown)


services/document-manager/backend/app/__init__.py
```diff
@@ -30,6 +30,9 @@ def create_app(test_config=None):
         dictConfig(Config.LOGGING_DICT_CONFIG)
     app = Flask(__name__)
 
+    if not os.path.exists('/tmp/spatial'):
+        os.makedirs('/tmp/spatial')
```
simensma-fresh, line 34: image
Looks like this doesn't match exactly with the path the files are downloaded to?
Think we might need to add a leading `/` there

simensma-fresh (changes requested):

simensma-fresh (dismissed):

henryoforeh-dev (dismissed):

services/core-api/app/api/mines/response_models.py
```diff
@@ -137,6 +147,7 @@ def format(self, value):
         'is_archived': fields.Boolean,
         'archived_date': fields.String,
         'archived_by': fields.String,
+        'mine_document_bundle': fields.Nested(MINE_DOCUMENT_BUNDLE_MODEL),
```
taraepp: I can see sending back the geomark_id, but I don't think any of the other fields are going to be used at all on the FE. I think I'd still rather do a fetch to get the geomark_id, as it's only used when we're displaying that particular bundle (one at a time).

simensma-fresh (approved):

taraepp (approved):

PR #3187, opened 2024-07-17
[FIX] Add in spatial bundle feature flag to FE

taraepp:
## Objective 
- put spatial bundle behind feature flag
- The way I did it _will_ result in a minor UI difference from pre-feature flag
  - before adding the spatial features, there was a single document table with a "category" column to indicate spatial/supporting
  - I figured it wouldn't be a problem to keep the tables separate like they will be when the flag is turned on


matbusby-fw (approved):
Yeah, I think that should work!

henryoforeh-dev (approved):

PR #3189, opened 2024-07-18
[MDS-6052] Added pipeline for extracting permit conditions

simensma-fresh:
## Objective 

[MDS-6052](https://bcmines.atlassian.net/browse/MDS-6052)

Added a new haystack ai pipeline in the permit service that accepts a permit PDF and extracts conditions from it.

image

The extraction happens in a couple of steps, 5 pages at a time

1. `pdf_converter`
Extracts raw text from pdf documents using `PyPDF`

2. `prompt_builder`
Creates prompts for GPT4 based on the jinja templates defined in [permit_condition_prompts.yaml](https://github.com/bcgov/mds/compare/permit_condition_extraction?expand=1#diff-684dcd364d59d43d7fd79d5d6ac094888d4bd6f5418c66a23d31304c54742a71) and the raw text extracted

3. `llm`
Queries OpenAI with the prompts created in the `prompt_builder`. Handles partial responses from GPT4 (stop_reason=length) by asking GPT4 to continue the response

4. `json_fixer`
GPT4 sometimes outputs Json that is not valid (e.g. an extra " or missing ]) this step uses the `json-fixer` library to attempt to correct this.

5. `validator`
Validates that the output from the previous step is valid json. Moves the pipeline back into the `prompt_builder` step if there are more pages to process.

*Haystack upgrade*
Upgraded haystack to the latest version which included a lot of breaking changes, so re-wrote the permit search functionality to support it.


services/permits/app/permit_conditions/pipelines/CachedAzureOpenAIChatGenerator.py
```diff
@@ -0,0 +1,139 @@
...
+logger = logging.getLogger(__name__)
+
+
+from haystack.components.generators.chat import AzureOpenAIChatGenerator
+from haystack.dataclasses import ChatMessage
+
+from haystack import component
+import pickle
+import os
+import os
```
taraepp: Imports look a bit strange! import os x 2, and then hashlib & struct are imported again below too.

henryoforeh-dev (approved):

taraepp (approved):

PR #3195, opened 2024-07-25
[MDS-6024] All Documents Tab

taraepp:
## Objective 
- add new content, layout, to all documents tab
- fix IRT nav on CORE
- is it loading? Make the loading bar more visible, and the spinner more consistent
- scroll side page wrapper can be used in different ways now

TO DO:
- ~~Minespace styling! Horizontal alignment in layout is _bad_ (it overlaps, and also the content extends beyond the right page boundary)~~
- ~~double check heading sizes, styling, spacing~~ (it's not perfect but added Row/Col/gutter, Alert, and fixed heading sizes)
- ~~would be nice if the random looking "top" values in stylesheet were at least composed of variables and didn't look completely arbitrary (& difficult to maintain if padding/margin changes)~~
- another ticket: make sure that archived AMS authorization documents come back from the document fetch and then turn on archiving
- the buttons in the bulk "Actions" dropdown look really weirdly styled
- implement the "Download All" on CORE for all the documents on the page (I think should be fairly simple?)
- and maybe the Actions (for _all_ of the documents) if it's doable


services/core-web/src/components/common/Loading.tsx
```diff
@@ -0,0 +1,11 @@
+/**
+ * @constant Loading is a full page loading spinner
+ */
+import React from "react";
+import { Loading as CommonLoading } from "@mds/common/components/common/Loading";
+
+const Loading = () => {
+  return <CommonLoading />;
```
asinn134, line 8: nice to see load spinner be more consistent across minespace and core

services/common/src/components/projects/ArchivedDocumentsSection.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return <></>;
+  }
+
+  const additionalColumns = [renderCategoryColumn("category_code", "Category", CATEGORY_CODE)];
+
+  return (
+    <div id={href}>
+      <Typography.Title level={titleLevel}>
+        <DeleteOutlined
+        // className="violet"
```
simensma-fresh: Can probably remove this

simensma-fresh (dismissed):

henryoforeh-dev (approved):

asinn134 (approved):

PR #3202, opened 2024-07-31
[MDS-6024] Fix issues on all documents

henryoforeh-dev:
## Objective 

1. Fixed issue with displaying archived files.
2. Fixed issue on ReplaceDocumentModal component.
3. Updated unit tests and snapshot test.


services/core-web/src/tests/components/modalContent/__snapshots__/ReplaceDocumentModal.spec.tsx.snap
```diff
@@ -1,138 +1,3 @@
...
-      type="button"
-    >
-      <span>
-        Replace
-      </span>
-    </button>
-  </div>
-</form>
-`;
+exports[`ReplaceDocumentModal renders properly 1`] = `undefined`;
```
simensma-fresh: Would be good to have this snapshot render properly instead of being `undefined` if possible

simensma-fresh (changes requested):

services/core-web/src/tests/components/modalContent/__snapshots__/ReplaceDocumentModal.spec.tsx.snap
henryoforeh-dev: Updated

services/common/src/components/documents/ReplaceDocumentModal.tsx
```diff
@@ -109,7 +109,7 @@ const ReplaceDocumentModal: FC<ReplaceDocumentModalProps> = (props) => {
         </Col>
       </Row>
 
-      <RenderFileUpload
+      <Field
```
taraepp, line 112: thanks, this was driving me _nuts._

services/common/src/components/projects/ProjectDocumentsTab.tsx
```diff
@@ -34,7 +34,6 @@ const ProjectDocumentsTab: FC<ProjectDocumentsTabProps> = ({ project }) => {
       dispatch(
         fetchMineDocuments(project.mine_guid, {
           is_archived: true,
-          project_guid: project.project_guid,
         })
```
taraepp: So, I tried doing this, and as I recall, it would give me _all_ of the mine's archived documents, and not filtered to the project. Are you 100% sure that we're not getting unrelated documents here? I think we just need to add a join to the ams docs table on the BE.

services/core-web/src/tests/components/modalContent/ReplaceDocumentModal.spec.tsx
```diff
@@ -1,33 +1,33 @@
...
-});
-
 beforeAll(() => {
   window.matchMedia = matchMedia;
 });
 
 describe("ReplaceDocumentModal", () => {
-  it("renders properly", () => {
-    const { container } = render(
+  it("renders correctly and matches the snapshot", () => {
```
taraepp, line 19: I am wondering why go back to the "shallow" render when the react-testing-library render gave us a better snapshot. The shallow one doesn't have anything useful in it, but the "render" gave us the full DOM.

simensma-fresh (approved):

taraepp (approved):

PR #3207, opened 2024-08-07
[CHORE] Clean up CORE mine permits table 

taraepp:
## Objective 
- make it function exactly the way it did before
- but look nicer/consistent
- with much less code, and easier to add/modify actions
   - applied existing functions for generating columns & actions menus
   - took out most of the params in the transform data functions (ex: we were assigning a function to a record, and calling it with `record.functionName()` in the column, and now it is just calling it with `functionName()`) 
- also filled in missing properties for PERMITS in dataMocks, which is making the changes in other snapshot results.

core-permit-actions


services/core-web/src/components/mine/Permit/MinePermitTable.tsx
```diff
@@ -96,94 +89,6 @@ const finalApplicationPackage = (amendment) => {
   return finalAppPackageCore.concat(finalAppPackageImported);
 };
 
-const renderDeleteButtonForPermitAmendments = (record) => {
```
matbusby-fw, line 99: Very happy with these changes.  I really dislike the previous pattern used here.

services/core-web/src/components/mine/Permit/MinePermitTable.tsx
```diff
@@ -18,16 +21,21 @@ import {
...
+  renderActionsColumn,
+  renderCategoryColumn,
+  renderTextColumn,
+} from "@mds/common/components/common/CoreTableCommonColumns";
+import { getDocumentDownloadToken } from "@mds/common/redux/utils/actionlessNetworkCalls";
+import { downloadDocument, waitFor } from "@/components/common/downloads/helpers";
+import { userHasRole } from "@mds/common/redux/selectors/authenticationSelectors";
 
 /**
  * @class  MinePermitTable - displays a table of permits and permit amendments
```
matbusby-fw, line 41: Tiny thing, but this isn't a class anymore I suppose

matbusby-fw (approved):
Love it!  Looks good to me!

services/core-web/src/components/mine/Permit/MinePermitTable.tsx
```diff
@@ -484,71 +255,177 @@ export const MinePermitTable: React.FC<MinePermitTableProps> = ({
...
+    userCanEditPermits && actionsMenu,
+  ].filter(Boolean);
+
+  const handleNavigateToPermitConditions = (record) => {
+    return history.push(
+      route.EDIT_PERMIT_CONDITIONS.dynamicRoute(record.mineGuid, record.permit_amendment_guid)
+    );
+  };
+
+  // from DownloadAllDocumentsButton/MineReportTable.js - I think there is a new way to accomplish this
```
simensma-fresh, line 319: Want to put migrating this over to use the new zip file download functionality ( I can't find one in the backlog)?
Think this must be one of the last bulk download that doesn't

simensma-fresh (approved):
Yay for cleaning up this!

services/core-web/src/components/mine/Permit/MinePermitTable.tsx
```diff
@@ -249,233 +157,96 @@ export const MinePermitTable: React.FC<MinePermitTableProps> = ({
...
+      key: "edit-site-properties",
+      label: "Edit Site Properties",
+      icon: <EditOutlined />,
+      clickFunction: (event, record) => openEditSitePropertiesModal(event, record.permit),
+    },
+    userIsAdmin && {
+      key: "delete-permit",
+      label: "Delete",
+      icon: <DeleteOutlined />,
+      clickFunction: (_, record) => {
```
henryoforeh-dev, line 241: Any particular need for _ ? Will it break if it is omitted ?

taraepp: It expects to receive an `event` param. The `_` is just to indicate that it's not being used. 🤷‍♀️ 
In hindsight, back when I first implemented that function, I should have switched the order, but it feels too late now.

henryoforeh-dev (approved):

PR #3209, opened 2024-08-09
[FIX] Match "is_historic" flag to current feature flag status

taraepp:
## Objective 
- make sure that records created after the migration is pushed to prod, but before the flag is turned on, are validated consistently by setting is_historic based on the flag.


asinn134 (approved):

henryoforeh-dev (approved):
Nice nice!

PR #3213, opened 2024-08-12
[MDS-6024] Fix redux form error

henryoforeh-dev:
## Objective 

- Fixed reduxform related error.
- Add null check on mine application documents.

image


services/common/src/components/documents/ReplaceDocumentModal.tsx
```diff
@@ -153,4 +153,14 @@ const mapDispatchToProps = (dispatch) =>
     dispatch
   );
 
-export default connect(null, mapDispatchToProps)(ReplaceDocumentModal);
+export default compose(
```
taraepp: I was thinking we could just use the FormWrapper here?
I don't think we can specify the form as FORM.ADD_EDIT_PROJECT_SUMMARY though, as it won't always be that form. Does it work to specify a new one like FORM.REPLACE_DOCUMENT?

services/core-web/src/tests/components/modalContent/ReplaceDocumentModal.spec.tsx
```diff
@@ -7,7 +7,7 @@ import { MINEDOCUMENTS } from "@mds/common/tests/mocks/dataMocks";
 import FormWrapper from "@mds/common/components/forms/FormWrapper";
 import { FORM } from "@mds/common/constants/forms";
 
-const props = {
+const props: any = {
```
taraepp: When I expand the test here- I'm thinking that the FormWrapper should _not_ be included as part of the test, since the component should handle the reduxForm on its own.

services/common/src/components/documents/ReplaceDocumentModal.tsx
matbusby-fw: I guess this is more a preference thing for me as what you have looks like it should work, but we do have a `FormWrapper` component that handles this quite nicely without having to use the compose feature.

You can simply wrap the elements in it like this:

```
<FormWrapper
              name={name}
              onSubmit={() => {}}
              initialValues={initialValues}
              isEditMode={isEditMode}
              reduxFormConfig={
                reduxFormConfig ?? {
                  touchOnBlur: true,
                  touchOnChange: false,
                  enableReinitialize: true,
                }
              }
            >
            <div>...</div>
            </FormWrapper>
```

matbusby-fw: Not a deal breaker, but I find it much easier to work with personally.  Especially by making the export simpler and cleaner for typescript types.

matbusby-fw (dismissed):
Looks good!  Had one suggested change, but not strictly necessary.

services/common/src/components/documents/ReplaceDocumentModal.tsx
henryoforeh-dev: Updated

henryoforeh-dev: Updated

services/core-web/src/tests/components/modalContent/ReplaceDocumentModal.spec.tsx
henryoforeh-dev: Updated.

services/common/src/components/documents/ReplaceDocumentModal.tsx
```diff
@@ -89,7 +91,7 @@ const ReplaceDocumentModal: FC<ReplaceDocumentModalProps> = (props) => {
   };
 
   return (
-    <Form layout="vertical" onFinish={() => props.handleSubmit(document).then(props.closeModal)}>
+    <FormWrapper name={FORM.REPLACE_DOCUMENT} onSubmit={handleReplaceSubmit}>
```
taraepp, line 94: Just want to check that taking out the `.then(closeModal)` bit still gets the modal closed on submit

taraepp: Although now that I look at what was taken out, it's just props.closeModal and not props.closeModal() so I wonder if it did anything to begin with. 😂

henryoforeh-dev: > Just want to check that taking out the `.then(closeModal)` bit still gets the modal closed on submit

Yes the modal gets closed after clicking the "Replace" button. I tested it before updating the PR.

henryoforeh-dev: This is the function (handleReplaceSubmit) that gets called when "Replace" button is clicked. 

image

matbusby-fw (approved):

services/common/src/components/documents/ReplaceDocumentModal.tsx
taraepp: Is that for _every_ instance of replacing documents? And so every time we implement replacing documents we'll have to remember to pass in the closeModal as part of the submit handler?
Personally, I would take it out of the submit handler and put it in the form, because the form knows it's a modal.

Nevermind! I see this is in the component, I thought it was in the parent. 😅

taraepp (approved):

PR #3214, opened 2024-08-12
[MDS-6011] - View geospatial bundles

matbusby-fw:
## Objective 

[MDS-6011](https://bcmines.atlassian.net/browse/MDS-6011)

The look of the view was already completed by @taraepp.  Previously the mapping information was stubbed in, so the work here was largely to addressing the fetching/saving of the geomark_ids and their respective mine_document_bundles.

One snag here was that we need to display the map before we have actually created the mine_docment_bundle to which we save the geomark_id, so had to do some work to accomodate different sources of that id depending on where in the process the user is.

TESTING UPDATE
I decided to add `msw` to the jest framework to simplify some kinds of tests.  I also had to upgrade a bunch of related settings packages to get it working, but it has a few nice benefits.

- Now we can mock any network response that would occur in the use of a component and specify it's return
- msw will warn you if if network request is happening that isn't mocked (and tell you exactly what it is)
- you can now interact with the dom through user click events to produce more robust tests
- if you're struggling with the data you need not being loaded immediately you can add checks for certain elements to be in the document before proceeding with further testing (ie snapshots).

I'll demonstrate the functionality when we have our testing meeting.

image


services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
```diff
@@ -101,12 +101,29 @@ const AddSpatialDocumentsModal: FC<AddSpatialDocumentsModalProps> = ({
...
+    const newFiles = currentValues[fieldName].map((f) => {
+      return {
+        ...f,
+        docman_bundle_guid,
+        geomark_id,
+        bundle_id: docman_bundle_guid,
+      };
+    });
+    dispatch(change(modalFormName, fieldName, [...initialDocuments, ...newFiles]));
+    dispatch(change(formName, fieldName, [...initialDocuments, ...newFiles]));
```
taraepp: ooooo one little maybe hiccup here-
what happens if the user uploads files, goes to review, and then cancels? That was the reason for only updating the parent form when `isFinalStep`.

services/common/src/redux/slices/spatialDataSlice.ts
```diff
@@ -6,52 +6,90 @@ import { formatDate } from "../utils/helpers";
 import { IGeoJsonFeature } from "@mds/common/interfaces/document/geojsonFeature.interface";
 import { ISpatialBundle } from "@mds/common/interfaces/document/spatialBundle.interface";
 
+interface ItempMineDocumentWithGeomarkId extends IMineDocument {
```
taraepp: I am trying to work out what an Itemp is? Typo?

services/common/src/redux/slices/spatialDataSlice.ts
```diff
@@ -6,52 +6,90 @@ import { formatDate } from "../utils/helpers";
...
-      geomark_id,
-      isSingleFile: false,
-    };
-  });
-  const individualFiles = temp_indiv
-    .filter((f) => !f.bundle_id)
-    .map((f) => {
+  const headers = createRequestHeader();
+
+  const spatial_bundles = await Promise.all(
```
taraepp: This was a little helper function before, but I think now it wants to be split up into a `fetch` function and a selector to separate concerns and follow the redux pattern.

matbusby-fw: This was the main issue with what took me so long.  I was trying to avoid refactoring all of your previous work since this function is providing the base for two seperate components.  One of those components just needs the file grouping and the other needs the geomark_id which sometimes needs to be fetched.

Since the state of the files (whether or not the mine_document_bundle has been created) is what dictates where it looks for the geomark_id it starts to get really janky to split up the concerns. 

I'll take another look at it.

services/common/src/redux/slices/spatialDataSlice.ts
matbusby-fw: it's a temporary MineDocument with a geomarkId

services/common/src/redux/slices/spatialDataSlice.ts
matbusby-fw: I do appreciate the eagerness, but this PR is not yet "Ready For Review"

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
matbusby-fw: Hmmm tricky.  

I see what you're saying.  Since it's these documents that get sent to the viewSpatial component they needed to contain these new values.  Maybe can pass the other set of documents.

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
```diff
@@ -95,18 +99,35 @@ const AddSpatialDocumentsModal: FC<AddSpatialDocumentsModalProps> = ({
...
+    geomark_id: string;
+  }) => {
+    const newFiles = currentValues[fieldName].map((f) => {
+      return {
+        ...f,
+        docman_bundle_guid,
+        geomark_id,
+      };
+    });
+    dispatch(change(modalFormName, fieldName, [...initialDocuments, ...newFiles]));
```
taraepp: I remember that I had to take the `initialDocuments` _out_ of the dispatch-change-modalFormName. Otherwise I'd upload a spacial bundle, save, go to upload another, and my first bundle would already be there. Have you tested it when there's already a bundle there and you're adding another?

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
```diff
@@ -95,18 +99,35 @@ const AddSpatialDocumentsModal: FC<AddSpatialDocumentsModalProps> = ({
...
+  };
+
   const handleSubmit = async (values) => {
     const isFinalStep = currentStep === stepContent.length - 1;
     const newFiles = values[fieldName];
-    const allFiles = [...newFiles, ...initialDocuments];
     if (isFinalStep) {
-      dispatch(change(formName, fieldName, allFiles));
+      dispatch(change(formName, fieldName, [...initialDocuments, ...newFiles]));
+      dispatch(change(formName, "spatial_documents", [...initialDocuments, ...newFiles]));
```
taraepp: Was this second dispatch added by mistake?

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
taraepp: I'm thinking that the geomark_id just wants to be a state variable or something. Then we're not having to add it to project summary documents and can keep it with the bundle.

services/common/src/components/documents/spatial/AddSpatialDocumentsModal.tsx
matbusby-fw: Indeed!

services/common/package.json
```diff
@@ -32,6 +32,7 @@
     "@syncfusion/ej2-react-pdfviewer": "26.1.35",
     "@syncfusion/ej2-splitbuttons": "26.1.35",
     "flagsmith": "3.19.1",
+    "msw": "2.3.5",
```
simensma-fresh: Does this need to be in `dependencies` or can it be moved to `devDependencies`?

services/common/package.json
```diff
@@ -60,6 +63,7 @@
     "terser-webpack-plugin": "5.3.9",
     "ts-jest": "29.1.2",
     "typescript": "4.7.4",
+    "undici": "^6.19.7",
```
simensma-fresh: Could we lock down this one?

services/common/src/redux/slices/spatialDataSlice.ts
```diff
@@ -99,19 +102,39 @@ const spatialSlice = createAppSlice({
...
           rejectHandler(action);
         },
       }
     ),
-    createSpatialBundle: create.asyncThunk(
+    fetchSpatialBundle: create.asyncThunk(
+      async (mine_document_bundle_id: string, thunkAPI) => {
+        thunkAPI.dispatch(showLoading());
+        const headers = createRequestHeader();
+        const url = `${ENVIRONMENT.apiUrl}/mines/document-bundle/${mine_document_bundle_id}`;
```
simensma-fresh: Could we put this into `@mds/common/constants/API` together with teh rest of our endpoints?
And on that note: Any way to update the mock server to also use those constants?

services/core-api/app/api/mines/documents/resources/mine_document_bundle.py
```diff
@@ -0,0 +1,18 @@
...
+from app.api.mines.documents.models.mine_document_bundle import MineDocumentBundle
+from app.api.utils.access_decorators import VIEW_ALL, MINESPACE_PROPONENT, requires_any_of
+from app.api.utils.resources_mixins import UserMixin
+from app.extensions import api
+
+
+class MineDocumentBundleResource(Resource, UserMixin):
+    @api.doc(description='Returns list of documents associated with mines')
+    @requires_any_of([VIEW_ALL, MINESPACE_PROPONENT])
+    def get(self, mine_document_bundle_id):
```
simensma-fresh, line 13: Could we add a test for those roles to `test_expected_auth.py`?

simensma-fresh: Would also be awesome to add some integration tests for this endpoint

simensma-fresh (changes requested):

henryoforeh-dev (dismissed):
Nice work! I intend to use the test package here in my upcoming PR too.

henryoforeh-dev (dismissed):

henryoforeh-dev (approved):

taraepp (approved):

simensma-fresh (approved):

PR #3218, opened 2024-08-15
[MDS-6026] View Permit Conditions

taraepp:
## Objective 
- flesh out the View Permit Conditions Page
  - it's scrolly now and has tab navigation
  - there's buttons added that were in the mockup, since we're behind a feature flag. Mostly they either do nothing or console log "not implemented" (please feel free to change my stubbed-in event handlers as necessary in future work!)
- Some things that really increased the files changed count with single changes: 
  - there is now a reusable header component. It wanted MS & CORE to have the same route name for MINE_DASHBOARD for GLOBAL_ROUTES. Updated all references. 
   - I made a CORE button component with defined custom types. And so I made it a scss file and moved rules in there on CORE. And then I got rid of an extraneous classname. And then I got rid of the CORE/MS DocumentTable component and replaced with references to the common one.

[MDS-6026](https://bcmines.atlassian.net/browse/MDS-6026)

image


simensma-fresh:
So close!
image

henryoforeh-dev (approved):

simensma-fresh (approved):

PR #3219, opened 2024-08-16
[MDS-6104] issue-credential v2 protocol

Jsyro:
## Objective 

[MDS-6104](https://bcmines.atlassian.net/browse/MDS-6104)

Issue-credential v1 protocols are being deprecated. 

Updated to use v2, I've reached out to Northern Block to see if they are prepared to accept this protocol. Might need to add feature flag if they aren't ready. 

It is feature-flagged, and Northern Block is actively working on AIP 2.0 compliance (which includes issue-credential-v2), so I will coordinate testing with them (and enable this flag in develop)


simensma-fresh (approved):

taraepp (commented):
LGTM!
Probably a follow-up ticket with some test coverage though?

Jsyro:
> LGTM! Probably a follow-up ticket with some test coverage though?

I currently don't have any tests for the webhook handling, because we do not control that acapy actually sends. To test it properly I would want to build tests that verify that ACA-py is sending the right thing (a true intergration test), and another test mocking the webhook and seeing how the handler processes them, which can be quite a complex sequence. 

Not saying it's not worth doing, and now that ACA-py is 1.0 released, maybe it is worth doing. 

Excluding the webhook handling the only thing to test would be that the feature flag properly switches the method called, which isn't a good test imo.

henryoforeh-dev (approved):

PR #3220, opened 2024-08-16
[MDS-6011] Fix buttons in modal footers, only show allowed type in replace modal

taraepp:
## Objective 
- this, along with prev PR should deal with _most_ button styling issues on CORE
   - specifically targeting buttons here in "ant-modal-footer" to add margins
- only show the matching file type in replace modal
- did a little cleaning of doc modals
   - switched to hooks
   - removed props like "openModal" from DocumentTableProps
   - added snapshot tests where missing
   - used FormWrapper so submit button can disable itself when submitting
   - cleaned up some async stuff: adjusted the prop to -> Promise<void>, put in awaits & Promise.all
   - noticed an issue with bulk actions vs actions on individual documents where we were allowing bulk archive after disallowing individual archive. Brought them in line. 

[MDS-6011](https://bcmines.atlassian.net/browse/MDS-6011)

image


simensma-fresh (approved):

simensma-fresh (changes requested):
Looks like cypress tests + common package tests are failing. Otherwise looks good

services/core-web/src/styles/settings/themeOverride.scss
```diff
@@ -551,6 +551,10 @@
   padding: 0;
 }
 
+.ant-modal-footer .ant-btn {
+  margin: 10px;
```
taraepp, line 555: A note for the future:

Writing it nested like:
`.ant-modal-footer { .ant-btn { margin: 10px }}`
did produce a build error in the javascript bundle. In order to reproduce the bundle error, just change it back to nested.

However, leaving it in, and adding `.ant-modal-footer .ant-btn {...` did not produce a build error. So it is not the presence or absence of that particular rule/format that breaks the build.

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #3223, opened 2024-08-21
[MDS-6141] Bug fix: use latest amendment on permit overview page

taraepp:
## Objective 
- use the latest amendment on the page, was using the first 
- made a selector for it because this is something we do constantly
- the selector is modelled after the places in the code where we do it already (search `latestAmendment =` in codebase for comparison)
- made an h4 an h3 on the project documents tab as well

[MDS-6141](https://bcmines.atlassian.net/browse/MDS-6141)


services/common/src/redux/selectors/permitSelectors.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  it("`getEditingPreambleFlag` calls `permitReducer.getEditingPreambleFlag`", () => {
+    const storeAction = storeEditingPreambleFlag(mockFlagsResponse);
+    const storeState = permitReducer({} as any, storeAction);
+    const localMockState = {
+      [PERMITS]: storeState,
+    };
+    expect(getEditingPreambleFlag(localMockState as RootState)).toEqual(editingPreambleFlag);
+  });
+  it("getPermitByGuid returns the correct permit", () => {
```
taraepp, line 45: Added these two tests! The others are pre-existing. Whole file was moved to common + TS-ed.

services/document-manager/backend/requirements.txt
```diff
@@ -20,7 +20,7 @@ requests==2.32.3
 SQLAlchemy==2.0.19
 uwsgi==2.0.23
 uwsgitop==0.11
-Werkzeug==2.3.6
+Werkzeug==3.0.3
```
taraepp, line 23: Build was failing in MS Cypress run due to Werkzeug version!

Upgraded to version on core-api and nris-api/backend! Checked out the PR where it was upgraded on core-api and based on that I'm hoping there's no other changes I'm missing: https://github.com/bcgov/mds/commit/2643944d725abd8e8c179a64ea2d856b63d93ad2 😅

henryoforeh-dev:
More coverage required for the **permitSelectors.ts**  file.

taraepp:
> More coverage required for the **permitSelectors.ts** file.

I did cover the one function that I added plus two functions that I added in the past. I'm not too keen on writing tests for functions like `getDraftPermitAmendmentForNOW` because I really don't know how they're used and what I should expect. It's a lot of research into unrelated areas for a bug fix.

asinn134 (approved):

services/document-manager/backend/requirements.txt
simensma-fresh: Ohh I seem to remember there was a reason for why this wasn't upgraded in Docman. Want to take it out of this PR and I'll see if I can dig it up 😅

simensma-fresh (changes requested):

services/document-manager/backend/requirements.txt
simensma-fresh: Actually it seems to be running fine, so should be OK to leave in (hopefully)

simensma-fresh (approved):

services/document-manager/backend/requirements.txt
taraepp: Alrighty! If it blows up test as soon as it's deployed (or doc upload/downloads down't work) we can do up a quick PR to revert.

PR #3227, opened 2024-08-22
[MDS-6073] Fix bugs on spatial docs upload.

henryoforeh-dev:
## Objective 

[MDS-6073](https://bcmines.atlassian.net/browse/MDS-6073)

Fixed a bug due to an oversight in my last PR. List of bugs fixed:

- Data was not saved in the mine_document_bundle table.
- Uploaded spatial bundle files are not displaying as expected.
- The hyperlink to the spatial standards URL is not functioning.


services/core-api/app/api/projects/major_mine_application/models/major_mine_application_document_xref.py
```diff
@@ -35,6 +35,8 @@ class MajorMineApplicationDocumentXref(Base):
     create_user = association_proxy('mine_document', 'create_user')
     versions = association_proxy('mine_document', 'versions')
     update_timestamp = association_proxy('mine_document', 'update_timestamp')
+    mine_document_bundle = association_proxy('mine_document', 'mine_document_bundle')
```
taraepp: Would want to avoid including the whole bundle with the document like in `project_summary_document_xref`.

henryoforeh-dev: Copied from project_summary_document_xref. 
image

taraepp: Oh no! I thought we took that out! Are we using it on the FE anywhere? If not, it's doing a much larger query than necessary, especially if it queries it every time.

henryoforeh-dev: I see mine_document_bundle_id as well. I know that is used in the FE. I will remove the other one from my code to see if it breaks anything.

asinn134 (approved):

taraepp (approved):

PR #3228, opened 2024-08-26
[MDS-6120] Add permit conditions version history

matbusby-fw:
Utilize the versioning implemention set up in Core-Api to enable versions of permit conditions.

The retrieval of versioned information is scoped to a future ticket.

## Objective 

[MDS-6120](https://bcmines.atlassian.net/browse/MDS-6120)

permit-condition-versions


henryoforeh-dev (dismissed):

simensma-fresh (changes requested):
Thoughts on adding a quick test or two here? E.g. editing a permit condition and making sure the history table also gets populated.

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #3234, opened 2024-09-04
[MDS-6095] Add BE Pagination/Filters to Mine Reports Table

matbusby-fw:
- Moved global reports pagination on the backend to a shared helper function (no changes made to the code within)
- Utilized this pagination on the mine-reports.py resource
- Updated front end components to fetch new reports on page and filter changes rather than simply handling this on the front end
- Refactored several FE components from JS to TS

## Objective 

[MDS-6095](https://bcmines.atlassian.net/browse/MDS-6095)


services/core-web/src/components/Forms/reports/ReportFilterForm.tsx
```diff
@@ -67,85 +72,78 @@ export class ReportFilterForm extends Component {
...
-    return (
-      <Form layout="vertical" onSubmit={this.props.handleSubmit} onReset={this.handleReset}>
+  return (
+    <FormWrapper
+      name={FORM.FILTER_REPORTS}
+      reduxFormConfig={{ touchOnBlur: false, enableReinitialize: true }}
+      onSubmit={onSubmit}
+      initialValues={initialValues}
+    >
+      <Form layout="vertical" onReset={handleReset}>
```
taraepp: I know you're still working on this but it caught my attention!
Tricky- I'm sure you're getting some console error like "invalid DOM Nesting: form cannot appear as a descendant of form", but there simply isn't an equivalent of `onReset` in redux-form. 🤔 

I wonder about using the `pristine` property of the form as a useEffect param? Or I guess there's always making the button just call `reset(); handleReset();` directly. Or could make FormWrapper take FormProps type param. 🤷‍♀️ 

Looks fabulous so far. ✨

services/common/src/interfaces/reports/mineReportPageData.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { IMineReport } from "@mds/common/interfaces";
+
+export interface IMineReportPageData {
+  records: IMineReport[];
```
taraepp: oh! We do have a generic `IPageData<T>` interface that can be used with `IPageData<IMineReport>` instead of creating a new interface if you like.

matbusby-fw: Oooh, nice!

services/core-web/src/components/Forms/reports/ReportFilterForm.tsx
matbusby-fw: Ah, good catch.  Went with just adding the onReset to the onClick function of the button.

simensma-fresh (approved):

henryoforeh-dev (approved):

PR #3244, opened 2024-09-11
[MDS-6132] UI Updates around Project Description

asinn134:
## Objective 

[MDS-6132](https://bcmines.atlassian.net/browse/MDS-6132)

- Adjusted Environmental Management Act text in Project Description Overview page

- I noticed the Project Description Overview tab only shows if ams_agent flag is on for Core, but wasn't set like this in Minespace. I made it also dependent on the ams agent flag in Minespace. Also made it so tables/sections on this page will not be displayed if there is no data related to it

- Made Project Description Overview tab only show after user has made a submission in this project

- Adjusted heading font sizes to match up with what was found here https://www.figma.com/design/t3bgkwadKPD4IwU5abtANk/MDS-Design-System-v1?node-id=1090-37761&node-type=text&m=dev. The change was really only in Minespace, it seems like Core heading already uses the same font sizes as the figma design page.


services/minespace-web/src/tests/components/project/projectPage/ProjectPage.spec.tsx
```diff
@@ -37,7 +38,13 @@ describe("ProjectPage", () => {
   it("renders properly", () => {
     const { container } = render(
       <ReduxWrapper initialState={initialState}>
-        <ProjectPage />
+        <FeatureFlagContext.Provider
```
taraepp: We do have this mocked in all 3 setupTests files- just curious if it's not working that way or added out of habit.

taraepp (dismissed):
LGTM
Just checking with the useFeatureFlag in earlier comment.
Checked coverage details and it looks like it's just ProjectStagesTable missing coverage but I think your added data brought ProjectPage up to 91.7% 🙌 (I'm okay with ProjectStagesTable being lower, iirc it's still buggy anyway)

services/minespace-web/src/tests/components/project/projectPage/ProjectPage.spec.tsx
asinn134: Hey, that was my mistake I added it out of habit, this is removed now.

matbusby-fw (dismissed):

matbusby-fw (approved):
🚀

henryoforeh-dev (approved):

PR #3245, opened 2024-09-11
[MDS-6129] streamline status actions

henryoforeh-dev:
## Objective 

Implemented logic to enable/disable fields with respect to status.

All conditions are in Miroboard link. https://miro.com/app/board/uXjVKnK62M0=/


services/common/src/constants/utils.ts
```diff
@@ -63,3 +65,43 @@ export const getProjectStatusDescription = (
...
+  // Return false if status = "" => "Not Started"
+  if (!projectSummaryStatusCode) return false;
+
+  const disabledStatuses = new Set(["WDN", "COM"]);
+  const enabledStatusMapping = {
+    [SystemFlagEnum.ms]: new Set(["DFT", "CHR"]),
+    [SystemFlagEnum.core]: new Set(["DFT", "ASG", "UNR", "CHR", "OHD"]),
+  };
+
+  if (disabledStatuses.has(projectSummaryStatusCode)) return true;
```
matbusby-fw, line 83: Nice.  Don't see `.has` used too much.  Slick.

matbusby-fw (changes requested):
Looks really nice.  My only thought is that we should probably add some snapshots for these project components (I realize for the most part you just added the disabled functionality on most of them, but maybe we could bump up the coverage a bit by adding snapshots for at least a couple of them?)

henryoforeh-dev:
> Looks really nice. My only thought is that we should probably add some snapshots for these project components (I realize for the most part you just added the disabled functionality on most of them, but maybe we could bump up the coverage a bit by adding snapshots for at least a couple of them?)

I added snapshot tests for all the components. While still short on recommended coverage, it will take more work to get each one to 80%.

matbusby-fw (approved):
Awesome!  Yeah, I think we're okay not always hitting the 80%.  Especially in cases like this where it requires new testing on previously created work.

asinn134 (approved):

PR #3247, opened 2024-09-16
[MDS-6149] - Update Ministry Contacts to Project Management

matbusby-fw:
## Objective 

- Changed `Ministry Contacts` tab of project summary to `Project Management`
- Added internal ministry comments to the Project Management page
- created the backend functionality for project summary internal ministry comments

[MDS-6149](https://bcmines.atlassian.net/browse/MDS-6149)

image

project-management


services/common/src/components/projectSummary/ProjectManagement.tsx
```diff
@@ -0,0 +1,195 @@
...
+            Project status is shared with the proponent via MineSpace. Changing the status will
+            notify the project’s Primary Contact.
+          </Paragraph>
+          {status_code && (
+            <Field
+              id="status_code"
+              name="status_code"
+              required
+              component={RenderSelect}
+              data={projectSummaryStatusCodes}
```
taraepp, line 95: Something to watch out for- the `required` attribute will put on the indicator, but it won't do the validating. Probably need a required validator in here too.

services/common/src/components/comments/CommentEditor.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { FC } from "react";
+import { useSelector } from "react-redux";
+import { Form } from "@ant-design/compatible";
+import "@ant-design/compatible/assets/index.css";
```
taraepp: it would be _amazing_ to update this bit (the compatible Form and the compatible stylesheet) but not sure of how much of a rabbit-hole it would be.
Like, I think if we put in a `<FormWrapper>` you'd end up with form within a form and that's sure to introduce jankiness. 
What if just the imports are updated? Taking out the stylesheet and importing Form together with Button/Input instead of from compatible?

services/common/src/components/comments/MinistryComment.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC } from "react";
+import { timeAgo } from "@mds/core-web/common/utils/helpers";
```
taraepp: This should probably be importing from @mds/common. Interesting that it works though!

services/common/src/components/comments/MinistryCommentPanel.tsx
```diff
@@ -1,46 +1,43 @@
...
+import React, { FC } from "react";
+import { LoadingOutlined, UserOutlined } from "@ant-design/icons";
+import { Avatar, Col, Divider, List, Row, Spin } from "antd";
 
 import CommentEditor from "@mds/common/components/comments/CommentEditor";
 import MinistryComment from "@mds/common/components/comments/MinistryComment";
-import * as Style from "@mds/common/constants/styles";
-import AuthorizationWrapper from "@mds/common/wrappers/AuthorizationWrapper";
+import * as Style from "@mds/core-web/src/constants/styles";
+import AuthorizationWrapper from "@mds/core-web/src/components/common/wrappers/AuthorizationWrapper";
```
taraepp: Strange imports!
I can see importing the style from core-web, but it makes me wonder if this component belongs in common since it's only used on core-web. But the AuthorizationWrapper I don't think we want to maintain a separate file for core-web

matbusby-fw: Yeah, it's causing all kinds of problems.  Didn't look at them too carefully at first, but I'm working on cleaning that up now.

services/common/src/components/projectSummary/ProjectManagement.tsx
```diff
@@ -0,0 +1,195 @@
...
+                  loading={false}
+                  maxLength={500}
+                  comments={ministryComments?.map((comment: IProjectSummaryMinistryComment) => ({
+                    key: comment.project_summary_ministry_comment_guid,
+                    author: comment.update_user,
+                    content: comment.content,
+                    actions: null,
+                    datetime: comment.update_timestamp,
+                  }))}
+                  createPermission={Permission.EDIT_PROJECT_SUMMARIES}
```
taraepp: So, I think the weird import from core-web instead of common for Permission is because I've resisted bringing the Permission file into common as it's essentially duplicating USER_ROLES. If you give it USER_ROLES.role_edit_project_summaries instead, you can import from common. `userHasRole` in authenticationReducer has been setup to take either USER_ROLES.x or Permission.x in order to be accommodating.

services/core-web/package.json
```diff
@@ -11,6 +11,7 @@
     "@fortawesome/fontawesome-pro": "6.4.2",
     "@fortawesome/fontawesome-svg-core": "6.5.1",
     "@fortawesome/pro-light-svg-icons": "6.5.1",
+    "@fortawesome/pro-regular-svg-icons": "6.5.1",
```
taraepp, line 14: Denise should be happy about this!

services/core-web/src/components/common/wrappers/AuthorizationWrapper.tsx
```diff
@@ -1,14 +1,13 @@
...
 import React from "react";
 import ReactDOMServer from "react-dom/server";
-import { PropTypes } from "prop-types";
 import { connect } from "react-redux";
 import { startCase, camelCase } from "lodash";
 import { getUserAccessData } from "@mds/common/redux/selectors/authenticationSelectors";
-import { USER_ROLES } from "@mds/common";
-import { detectDevelopmentEnvironment, detectProdEnvironment } from "@mds/common";
 import { Tooltip } from "antd";
 import * as Permission from "@/constants/permissions";
```
taraepp, line 8: I would be inclined to just update to the services/common AuthorizationWrapper instead of using this one.

services/common/src/components/comments/CommentEditor.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { FC } from "react";
+import { useSelector } from "react-redux";
+import { Button, Form, Input } from "antd";
```
taraepp, line 3: yay!

services/common/src/components/comments/CommentEditor.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    await onSubmit({ comment });
+    handleReset();
+  };
+
+  const handleChange = (e) => {
+    setComment(e.target.value);
+  };
+
+  const canAddComment = addCommentPermission
```
taraepp: I still think it would be more flexible to use userHasRole! Then you're not stuck using the Permission file at any point. just canAddComment = useSelector(userHasRole(addCommentPermission)) and then you can take out a bunch of imports too.

services/common/src/components/projectSummary/ProjectSummaryForm.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+const asyncSave = () => new Promise<void>(() => {});
+
+describe("Project Management", () => {
+  it("renders properly", async () => {
+    const { container } = render(
+      <ReduxWrapper initialState={initialState}>
+        <ProjectSummaryForm
+          initialValues={initialState.form[FORM.ADD_EDIT_PROJECT_SUMMARY].values}
+          isEditMode={false}
+          handleSaveData={asyncSave}
```
taraepp, line 38: just FYI I've got a little trick for this that might be simpler! `() => Promise.resolve()`
Not important though, it just looks like it was annoying to write

services/core-web/src/tests/components/common/comments/Comment.spec.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
-import { shallow } from "enzyme";
 import { Comment } from "@/components/common/comments/Comment";
+import { render } from "@testing-library/react";
 
-const dispatchProps = {};
-const props = {};
+const dispatchProps: any = {};
+const props: any = {};
```
taraepp, line 6: I think the props and dispatchProps here can just be omitted completely.

matbusby-fw: The props are used, but I'll remove the dispatch

services/common/src/components/projectSummary/ProjectSummaryForm.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+          handleTabChange={() => {}}
+          activeTab={"basic-information"}
```
asinn134, line 40: I don't think this matters too much, but shouldn't the activeTab be project management?

matbusby-fw: This is just for the general form snapshot, so it doesn't matter too much.  project management and basic information already have their own tests in place.

services/common/src/components/comments/CommentEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+    ? useSelector((state) => userHasRole(state, addCommentPermission))
```
taraepp: I do believe that putting a hook in a conditional statement is a hooks rules violation (gotta separate it out into its own statement)

taraepp (approved):

asinn134 (approved):

PR #3248, opened 2024-09-16
[MDS-6140] Project Primary Contact on Project Final Application

taraepp:
## Objective 
- on the "review & submit" page it wasn't showing the contact properly, or, I believe, the mine name
- previously I had thought that review & submit would always be within the form on MajorMineApplicationPage, but it's also on ProjectPage with a different set of props!
  - while I did upgrade the files to .tsx, I mostly kept things the same in how they worked
- except that formatting the documents in the final application was tedious so I extracted that into a selector, and moved the function to get the contact's name into a common file.

[MDS-6140](https://bcmines.atlassian.net/browse/MDS-6140)


services/common/src/redux/selectors/projectSelectors.ts
```diff
@@ -90,6 +111,32 @@ export const getAmsAuthorizationTypes = createSelector(
   }
 );
 
+export const getFormattedProjectApplication = createSelector(
```
matbusby-fw, line 114: Nice new selector!

matbusby-fw (approved):
Nice!  Yay more refactoring!

asinn134 (approved):

PR #3254, opened 2024-09-24
[MDS-6118] FE extract permits

taraepp:
## Objective 
- fulfill MVP for process
- I _did_ end up adding a disabled style for the action, because it was incredibly annoying for _me_ not to know which was disabled! It ain't pretty though.
- fix some hosts stuff on BE that was throwing errors
- tweak the FE to be more streamlined, less page refreshing when data changes

[MDS-6118](https://bcmines.atlassian.net/browse/MDS-6118)


services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -76,17 +86,16 @@ const PermitConditions: FC<PermitConditionProps> = ({ latestAmendment }) => {
     return Promise.resolve();
   };
 
-  if (fetchingPermits) {
+  if (isLoading) {
```
matbusby-fw, line 89: 🤨 Didn't like that name?

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -76,17 +86,16 @@ const PermitConditions: FC<PermitConditionProps> = ({ latestAmendment }) => {
...
 
-  const isExtractionInProgress =
-    permitExtraction?.task_status === PermitExtractionStatus.in_progress;
-  const canStartExtraction = !permitExtraction && !permitConditions?.length;
-
   if (isExtractionInProgress) {
     return <RenderExtractionProgress />;
   }
+  if (!isExtractionComplete && permitExtraction?.task_status) {
+    return <div>Permit extraction status: {permitExtraction?.task_status}</div>;
```
matbusby-fw, line 97: Nice addition!

services/core-web/src/components/mine/Permit/PermitConditions.tsx
taraepp: Basically it's because of something I added and then took out! And then I decided that I liked "isLoading" anyway and left it after I took it out.

If you're curious, the call to fetchPermits on the parent when the extraction was complete wasn't getting called properly at first (and so I didn't notice it was there) so I made a call _here_ to fetchPermitConditions and isLoading also was based on that. But that created a whole data issue because it wasn't stored in latestAmendment.permit_conditions where everything was looking for it.

services/permits/app/celery/elasticsearch_backend.py
```diff
@@ -27,7 +27,7 @@ def _get_server(self):
             http_auth = (self.username, self.password)
 
         return elasticsearch.Elasticsearch(
-            f'{self.scheme}://{self.host}:{self.port}',
+            hosts="https://elasticsearch:9200",
```
matbusby-fw: Is there a way to use dynamic variables here while keeping things working properly?  I know this was part of the change to get it working, but it would be better if we didn't rely on hardcoded values here.

matbusby-fw: Even just using ELASTICSEARCH_HOST

taraepp: oh oops I thought I fixed that!

matbusby-fw (approved):
Woohoo!

services/core-web/src/components/mine/Permit/PermitConditions.tsx
matbusby-fw: Ah, cool!

asinn134 (approved):

PR #3258, opened 2024-10-07
[MDS-6114] update AMS fields

henryoforeh-dev:
## Objective 

[MDS-6114](https://bcmines.atlassian.net/browse/MDS-6114)

1. Generated migration script to alter columns according to AMS specs.
2. Added new attributes to the AMS API service.
3. Updated fields in FE application to match AMS specs.

### DIFF
--- a/services/common/src/components/projectSummary/LegalLandOwnerInformation.tsx
+++ b/services/common/src/components/projectSummary/LegalLandOwnerInformation.tsx
@@ -104,7 +104,7 @@ export const LegalLandOwnerInformation: FC = () => {
                 label="Legal Land Owner Name"
                 component={RenderField}
                 required={!is_legal_land_owner}
-                validate={!is_legal_land_owner ? [required] : []}
+                validate={!is_legal_land_owner ? [required, maxLength(100)] : []}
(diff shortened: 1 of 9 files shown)


asinn134 (dismissed):

services/common/src/components/projectSummary/LegalLandOwnerInformation.tsx
```diff
@@ -104,7 +104,7 @@ export const LegalLandOwnerInformation: FC = () => {
                 label="Legal Land Owner Name"
                 component={RenderField}
                 required={!is_legal_land_owner}
-                validate={!is_legal_land_owner ? [required] : []}
+                validate={!is_legal_land_owner ? [required, maxLength(100)] : []}
```
taraepp: For this one and the one below- we only want the required validation when they're not the legal land owner, but shouldn't the other methods be in the second array? Like if it's not required, it should still have a max length, and below even if it's not required should still follow email validation. Existing issue I guess but may as well fix it?

asinn134 (approved):

taraepp (approved):

PR #3266, opened 2024-10-16
[MDS-6118] Added failure state to permit extraction flow + fix error when conditions are not matching a category

simensma-fresh:
## Objective 

[MDS-6118](https://bcmines.atlassian.net/browse/MDS-6118)

1. Update permit conditions with the general category if no sections could be matched to one (the case for old permits). Without it, an error is produced
2. Update the state of the extraction task stored in the core DB when an extraction task actually fails (for any reason)
3. Added an error state if extraction failed

### DIFF
--- a/services/core-web/src/components/mine/Permit/PermitConditionExtraction.tsx
+++ b/services/core-web/src/components/mine/Permit/PermitConditionExtraction.tsx
@@ -2,6 +2,8 @@ import React, { FC } from "react";
 import { Col, Row, Typography } from "antd";
 import { PERMIT } from "@/constants/assets";
 import LoadingOutlined from "@ant-design/icons/LoadingOutlined";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faCircleX } from "@fortawesome/pro-light-svg-icons";
 
 const { Title } = Typography;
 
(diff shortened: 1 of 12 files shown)


services/core-web/src/components/mine/Permit/PermitConditionExtraction.tsx
```diff
@@ -54,3 +56,29 @@ export const RenderExtractionProgress: FC = () => (
...
+    <Col span={24}>
+      <Title className="margin-none" level={2}>
+        Permit Conditions
+      </Title>
+    </Col>
+    <Col span={24}>
+      <Row gutter={10} style={{ background: "#fff" }} justify={"center"}>
+        <div className="null-screen fade-in" style={{ maxWidth: "728px" }}>
+          <div>
+            <p className="margin-medium--bottom">
```
matbusby-fw: Not a huge deal, but it might good to use the Typography elements here to make sure any styling specific to that is in place.

matbusby-fw (changes requested):
Looking good to me.  Just left one comment about using antD typography over `p` and `h3`.

asinn134 (approved):

matbusby-fw (approved):
🦛

PR #3268, opened 2024-10-17
[MDS-6114] Fix AMS bug

henryoforeh-dev:
## Objective 

[MDS-6114](https://bcmines.atlassian.net/browse/MDS-6114)

- Fixed: Regions were displaying the ID instead of the region name.
- Fixed: AMS submission bug where the incorporation number and address unit number attributes were omitted from the application object.
- 
Note: The phone number extension was also identified as missing from the application object. However, this attribute is not included in the latest AMS documentation.


asinn134 (approved):

matbusby-fw (approved):
Nice!

PR #3269, opened 2024-10-17
[MDS-6194] Major project duplication

matbusby-fw:
## Objective 

On the `ProjectSummaryPage` in Minespace the `projectFormTabs` variable was null.  As a result, when the system attempted to replace the current url with one containing the new project and project summary guids it would pass in that null value as part of the route and fail.

Loaded the projectFormTabs properly and it seems to work fine.

[MDS-6194](https://bcmines.atlassian.net/browse/MDS-6194)

major-project-duplicate

### DIFF
--- a/services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
+++ b/services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
@@ -36,9 +36,11 @@ import {
   AMS_STATUS_CODES_SUCCESS,
   AMS_STATUS_CODE_FAIL,
   AMS_ENVIRONMENTAL_MANAGEMENT_ACT_TYPES,
+  SystemFlagEnum,
 } from "@mds/common";
 import { useFeatureFlag } from "@mds/common/providers/featureFlags/useFeatureFlag";
 import { fetchRegions } from "@mds/common/redux/slices/regionsSlice";
+import { getSystemFlag } from "@mds/common/redux/selectors/authenticationSelectors";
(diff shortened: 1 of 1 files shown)


simensma-fresh (approved):

services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
```diff
@@ -71,7 +75,11 @@ export const ProjectSummaryPage = () => {
     : mine?.mine_guid === mineGuid;
   const [isLoaded, setIsLoaded] = useState(isDefaultLoaded);
   const [isEditMode, setIsEditMode] = useState(isDefaultEditMode);
-  const projectFormTabs = null;
+  const projectFormTabs = getProjectFormTabs(
+    amsFeatureEnabled,
+    isCore,
```
taraepp, line 80: Y'know, I even noticed that this function was throwing errors! 
Technically this should _always_ be false as this is a MineSpace file, but I'm not bothered by a bit of extra robustness.

taraepp (approved):

asinn134 (approved):

services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
matbusby-fw: Yeah, I thought might as well add the actual flag for that imaginary case where this component gets used in Core at some point for some weird unknown reason.

PR #3271, opened 2024-10-18
[MDS-4971] In-app help

taraepp:
## Objective 
- create in-app help feature
- I left in the stubbed image handler for rich text editor (it was a bit tricky to get it to call the function at all!)

[MDS-4971](https://bcmines.atlassian.net/browse/MDS-4971)


services/common/src/components/help/HelpGuide.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const initialValues =
+    hasHelpGuide && !defaultGuide ? helpGuide : { help_key: helpKey, page_tab: pageTab };
+  const mainContent = isEditMode ? (
+    <HelpGuideForm
+      initialValues={initialValues}
+      handleSaveGuide={handleSaveGuide}
+      pageTab={pageTab}
+    />
+  ) : (
+    parse(content)
```
simensma-fresh: I'm wondering if this might be vulnerable for an XSS injection? Seeing as it just renders the html stored in the DB directly.
If so, we might want to sanitize it using something like `sanitize-html` before we render it

simensma-fresh (changes requested):

services/common/src/components/help/HelpGuide.tsx
simensma-fresh: `DomPurify` also looks good!

taraepp: Just added!
Added a test on MS where the rendered (dataMocks) content contains both a script tag and a click handler to verify that they're stripped out.

services/core-api/app/api/help/resources/help_resource.py
```diff
@@ -0,0 +1,138 @@
...
+        system = request.args['system']
+        if is_minespace_user() and system != str(SystemFlag.ms):
+            raise BadRequest("Only MineSpace help guides may be requested.")
+        help_guides = Help.find_by_help_key(help_key, system)
+        return help_guides
+    
+    @requires_any_of([EDIT_HELPDESK])
+    @api.expect(parser)
+    @api.marshal_with(HELP_MODEL, code=201)
+    def post(self, help_key):
```
simensma-fresh, line 93: Would be great if we could add some tests to these endpoints, and add the resource + expected auth to `tests/auth/test_expected_auth.py`

simensma-fresh (changes requested):

services/core-web/webpack.config.ts
```diff
@@ -35,7 +35,8 @@ const PATHS = {
 };
 
 const BUILD_FILE_NAMES = {
-  css: "style/[name].[contenthash:4].css",
+  css: "style/[name].[contenthash:8].css",
+  cssChunkName: "style/[name].[chunkhash:8].css",
```
simensma-fresh, line 39: @taraepp  Reading through our webpack config I remembered we updated this in Minespace a while ago, but looks like we didn't in Core!

Do not know why it fixes the bundle error, but it does (at least when built in production mode locally). Thanks anonymous stackoverflow post

https://stackoverflow.com/questions/71142869/invalid-js-in-webpack-5-bundle-some-numbers-from-source-code-are-replaced-by-nu

simensma-fresh (approved):

services/minespace-web/src/constants/routes.ts
```diff
@@ -41,42 +41,49 @@ const ProjectSubmissionStatusPage = React.lazy(() =>
 export const HOME = {
   route: "/",
   component: LandingPage,
+  helpKey: "Home",
```
taraepp, line 44: Just thought I'd mention:
- currently, this is what the title of the guide is based on
- some of these pages are very difficult to access on prod without actually creating data
- I anticipate that "title" is going to be an editable field in the future to deal with the first problem
- as for accessing the pages, I anticipate that CORE users will be able to view a list of keys and pick one without actually going through the motions to get there
- I duplicated keys where I felt it was appropriate

I didn't worry about it too much about making them perfect right now, I figure that'll come out in testing.

services/core-web/webpack.config.ts
asinn134: nice catch :)

services/common/src/components/help/HelpGuide.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const initialValues =
+    hasHelpGuide && !defaultGuide ? helpGuide : { help_key: helpKey, page_tab: pageTab };
+  const mainContent = isEditMode ? (
+    <HelpGuideForm
+      initialValues={initialValues}
+      handleSaveGuide={handleSaveGuide}
+      pageTab={pageTab}
+    />
+  ) : (
+    parse(DOMPurify.sanitize(content))
```
matbusby-fw, line 136: nice!

services/common/src/components/help/HelpGuide.tsx
```diff
@@ -0,0 +1,203 @@
...
+            {title} Help Guide
+          </Typography.Title>
+          {isLoaded ? mainContent : <Loading />}
+        </div>
+      </Drawer>
+    </>
+  );
+};
+
+const HelpGuide: FC = () => {
```
matbusby-fw, line 172: I guess this is just personal preference, but I generally prefer to have functional components in their own files where possible.  I'll leave it up to you though.

services/common/src/redux/slices/helpSlice.ts
```diff
@@ -0,0 +1,168 @@
...
+
+interface HelpState {
+  helpGuides: ItemMap<HelpGuide[]>;
+}
+
+const initialState: HelpState = {
+  helpGuides: {},
+};
+
+const helpSlice = createAppSlice({
```
matbusby-fw, line 23: Big fan of the slices!

services/core-web/src/styles/components/Modal.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .ant-modal-wrap {
-  z-index: 1001;
+  z-index: 1005;
```
matbusby-fw, line 19: This one goes to 11 🎸

services/core-web/webpack.config.ts
matbusby-fw: Oh!  Nice!

matbusby-fw (approved):
Approved!  Really nice work!  Added one preference for separating functional components to their own files, but I'm not stuck on it, so it's up to you.

services/common/src/components/help/HelpGuide.tsx
taraepp: I think I'll leave it this time... The sub-component I only turned into an export so that I'd be able to test it easier!

services/common/src/redux/slices/helpSlice.ts
taraepp: sooo much simpler to write! 🔪

PR #3272, opened 2024-10-18
[CHORE] - update VerificationMethod and .env-example

Jsyro:
Env changes have been made in openshift, should be made locally for DEV to match. 

The verificationmethod had the wrong fragment in it, it was hard coded for how indy ledgers create diddocuments, but the tdw server does it differently, now take verification method and derive did from that.


services/core-api/app/config.py
```diff
@@ -291,9 +291,13 @@ def JWT_ROLE_CALLBACK(jwt_dict):
     CRED_DEF_ID_MINES_ACT_PERMIT = os.environ.get("CRED_DEF_ID_MINES_ACT_PERMIT",
                                                   "CRED_DEF_ID_MINES_ACT_PERMIT")
 
-    #The key pair in this did web MUST match the keypair of the did:indy:candy in Traction.
-    CHIEF_PERMITTING_OFFICER_DID_WEB = os.environ.get("CHIEF_PERMITTING_OFFICER_DID_WEB",
-                                                      "CHIEF_PERMITTING_OFFICER_DID_WEB")
+    CHIEF_PERMITTING_OFFICER_DID_WEB_VERIFICATION_METHOD = os.environ.get(
+        "CHIEF_PERMITTING_OFFICER_DID_WEB_VERIFICATION_METHOD",
+        "CHIEF_PERMITTING_OFFICER_DID_WEB_VERIFICATION_METHOD")
```
PatStLouis, line 296: why two lines here?

Jsyro: 2nd params is the default if not provided. 

App won't start if the second param is not provided. 

slightly more developer friendly this way, .env-example has the appropriate values.

simensma-fresh (approved):

matbusby-fw (approved):
🫎

PR #3278, opened 2024-10-23
[MDS-6103] Upgraded build related npm packages + upgraded codebase to support newest sass version

simensma-fresh:
## Objective 

[MDS-6103](https://bcmines.atlassian.net/browse/MDS-6103)

**SASS Upgrade**
- The newest SASS version provides a warning if `@import` statements are present (these conflict with the plain `css` version). Instead, we should use the SASS module system (good writeup here: https://css-tricks.com/introducing-sass-modules/ )
- Updated all use of global SASS functions (e.g. `darken`) to use the namespaced ones instead (e.g. color.adjust) to remove a bunch of warnings. This was done automatically using the `sass-migration` command
- I added a `base.scss` file that contains most of the old global utility `scss` (variables / theme etc.). I added this to make this migration a bit more achievable, but it might be a good idea to instead use direct imports where we can instead.
- De-duplicated a bunch of variables in the global scope that had the same names (especially colors / theme variables). The build now throws an error when duplicates are found

*Package upgrades*
- Updated a bunch of build related npm packages, and removed some that were not in use


taraepp (approved):
🙌

henryoforeh-dev (approved):
Holy smokes!

services/minespace-web/src/styles/tools/syncfusion_imports.scss
```diff
@@ -1,17 +1,17 @@
...
-@import "~@syncfusion/ej2-inputs/styles/material.scss";
-@import "~@syncfusion/ej2-navigations/styles/material.scss";
-@import "~@syncfusion/ej2-popups/styles/material.scss";
-@import "~@syncfusion/ej2-splitbuttons/styles/material.scss";
-@import "~@syncfusion/ej2-lists/styles/material.scss";
-@import "~@syncfusion/ej2-layouts/styles/material.scss";
-@import "~@syncfusion/ej2-grids/styles/material.scss";
-@import "~@syncfusion/ej2-react-pdfviewer/styles/material.scss";
-@import "~@syncfusion/ej2-react-filemanager/styles/material.scss";
+@import "~@syncfusion/ej2-base/styles/material.css";
```
matbusby-fw, line 1: Interesting.  Was this a change on the Syncfusion side of things?  Or were there never scss versions of these?

matbusby-fw (approved):
Amazing.  So many sweet sweet upgrades.  Love it!

services/common/src/redux/utils/helpers.ts
```diff
@@ -109,16 +109,15 @@ export const formatDateTimeUserTz = (dateTime: Date | string): string => {
 
 export const timeAgo = (dateTime, unit = "day") => {
   const startDate = dateTime;
-  const endDate = new Date().toUTCString();
```
asinn134, line 112: Was the issue with the test spec going between 9 to 10 because endDate was created as utc string?

asinn134 (approved):

services/common/src/redux/utils/helpers.ts
simensma-fresh: Exactly! End date being converted to UTC, whereas startDate was not

PR #3279, opened 2024-10-23
[MDS-6157] major projects uploads file name convention

asinn134:
## Objective 

[MDS-6157](https://bcmines.atlassian.net/browse/MDS-6157)

- Adds help text for Major projects document uploads
- Adds file name character limit and error notification


taraepp:
- I think it would be better for the BE to have the file name length as a constant to validate against instead of taking it as a request parameter. Sorry! I know it was a pain! FE shouldn't tell BE how to validate. I see that it's in document_manager_service though and not the project summary or whatever so maybe that's also quite tricky to implement. 🤔 
- That 9 days in the snapshot is going to have to be set to 10 days again. Simen's got a fix in the works (I think on his open PR)
- I like all the added error catching/logging. Nice.

services/core-web/src/styles/components/MajorMineApplication.scss
```diff
@@ -0,0 +1,5 @@
+.decision-package-uploads {
```
taraepp: This form item label is misaligned because it's using the old antd Form.Item from "@ant-design/compatible". Now, I'm not _sure_ because I can't see the DOM structure right now, but there is a very good chance that you can straight up _remove_ that Form.Item wrapper because it'll have another Form.Item nested within it from the RenderFileUpload component. I would try that first, and if it doesn't work, just update the import to be from "antd" instead.

(newer form input components have Form.Item within the component and don't need another around it. But if redux is grabbing the name from that Form.Item it might take more effort to fix, but probably just passing a name attribute to the Field component)

asinn134: Thanks I will try this out!

services/core-web/src/components/Forms/UploadProjectDecisionPackageDocumentForm.js
```diff
@@ -3,9 +3,7 @@ import PropTypes from "prop-types";
 import { compose } from "redux";
 import { Field, reduxForm, getFormValues } from "redux-form";
 import { connect } from "react-redux";
-import { Form } from "@ant-design/compatible";
-import "@ant-design/compatible/assets/index.css";
```
taraepp, line 7: Yessss to getting rid of this stylesheet too! 🙌

taraepp (approved):
👍

matbusby-fw (approved):
Nice!

henryoforeh-dev (approved):

PR #3285, opened 2024-10-28
[MDS-6178] CORE project summary edit mode bugs, contd

taraepp:
## Objective 
Project Dates:
- make validation messages more useful
- throw in some spacing
- allow user to clear the date
- sorry, lots of whitespace differences on this now!

Project Links:
- hide the input part of it altogether when in view mode
- doing so also fixed the state issues where (I believe) the currentSelection was staying the same between edit/view toggle, making it so that clicking the add button after toggling back to edit was adding what was previously there (even though it wasn't in the select box anymore). The re-render fixed it.
- and make sure that after adding the related project that the selection was cleared.

General Project summary form:
- reset the form on cancel edit

[MDS-6178](https://bcmines.atlassian.net/browse/MDS-6178)

image

### DIFF
--- a/services/common/src/redux/utils/Validate.ts
+++ b/services/common/src/redux/utils/Validate.ts
@@ -277,11 +277,16 @@ export const dateInFuture = (value) =>
   value && !moment(value).isAfter() ? "Date must be in the future" : undefined;
 
 // NOTE: modified from version in CORE- change from <= to <
-export const dateNotBeforeOther = memoize((other: string) => (value: string) => {
-  return value && other && value < other
-    ? `Date cannot be before ${moment(other).format("ddd MMM D YYYY")}`
-    : undefined;
-});
+// removed memoization because it was interfering with passing 2nd arg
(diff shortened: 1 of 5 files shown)


henryoforeh-dev (approved):

services/common/src/redux/utils/Validate.ts
```diff
@@ -277,11 +277,15 @@ export const dateInFuture = (value) =>
...
-  return value && other && value < other
-    ? `Date cannot be before ${moment(other).format("ddd MMM D YYYY")}`
-    : undefined;
-});
+export const dateNotBeforeOther = memoize((other: string, otherLabel?: string) => (value: string) => {
+  const invalid = value && other && value < other;
+  if (!invalid) {
+    return undefined;
+  }
+  const otherFormattedDate = moment(other).format("ddd MMM D YYYY");
```
matbusby-fw, line 285: I'd love to see these converted to using DateFns instead of moment at some point.  But I think we have a ticket to throw moment in the garbage already, so maybe it can wait for that.

matbusby-fw (approved):

PR #3289, opened 2024-10-30
[MDS-5697] Add method to forward UNTP DCC's through Orgbook Publisher

Jsyro:
## Objective 

[MDS-5697](https://bcmines.atlassian.net/browse/MDS-5697)

Implemented both options for now, can be trialed until decision is made there 

Option A) 
- MDS creates the UNTP credential, signs it with their own tenant 
- with the CPO keypair
- and is pushed to Orgbook publisher

Option B) 
- MDS sends abbreviated date to Orgbook publisher
- Publisher constructs UNTP DCC from fixed templating
- Publisher signs with a different keypair.
- Orgbook publisher injects and manages the BitStringStatusList(s).

It would save us from having to manage status lists and align our credentials with others on the service, but this work was already done, so i'm sharing this for now.


services/core-web/common/utils/helpers.ts
```diff
@@ -593,7 +590,7 @@ export const getHighestConsequence = (tsf) => {
 
   const highestRankedDam = tsf.dams.reduce((prev, current) =>
     CONSEQUENCE_CLASSIFICATION_RANK_HASH[prev.consequence_classification] >
-      CONSEQUENCE_CLASSIFICATION_RANK_HASH[current.consequence_classification]
+    CONSEQUENCE_CLASSIFICATION_RANK_HASH[current.consequence_classification]
```
taraepp, line 593: Just a note on this helpers.ts file (I know it's not ready for review and haven't really looked at the rest!)-

There is a newer version of the file in `services/common/src/redux/utils`. Lately, we tend to update _that_ version, and just update the reference in whatever's importing from it. Then you don't need to make changes in both core/ms versions. We are planning to do a big ticket at some point to update _all_ the references, and a change like this might get missed!

taraepp: Oh wait I see now that this is just a whitespace difference. Don't worry about it!

Jsyro: Yea, my vs studio formatter and the git pre-hook formatter are competing i think

taraepp (approved):
👍

matbusby-fw (approved):

PR #3292, opened 2024-11-01
[MDS-6227] Add administrative amendments and filter to now_gis_export API

taraepp:
## Objective 
- remove the filtering out of administrative amendments from now gis view
- add it as a column in the view, so it can be filtered _for_
- add it as a parameter on the now gis endpoint

[MDS-6227](https://bcmines.atlassian.net/browse/MDS-6227)

_Testing considerations_
- the automatic tests that I added can only look at the status code of the response, once you try to look at the _data_ everything kind of falls apart because the view depends on tables that don't exist in the test context
- so it was all tested manually/locally
1. changed 2/800 now_application_identity records to have an application_type_code of `ADA` instead of `NOW`
2. recreated the view locally (basically ran the migration in Postico, taking out `materialized`, `with data`, and `TABLESPACE pg_default`, and simplified the first statement to just a `DROP VIEW`)
3. added a console log in `now_application_gis_export_resource.py-> get-> generate` (logged `r`)
4. made a call in postman to `http://localhost:5000/exports/now-application-gis-export?application_type_code=ADA` and watched the logs in core-api container (postman can't easily handle _streaming_ responses and so only the header is visible!)
5. Saw that the two results (and only the two results) that I had changed to `ADA` were logged

### DIFF
--- a/services/core-api/tests/exports/test_export_success.py
+++ b/services/core-api/tests/exports/test_export_success.py
@@ -5,13 +5,13 @@ from app.api.exports.mines.resources.mine_summary_resource import PAGE_DEFAULT, 
 
 def test_mine_summary_export_success(test_client, db_session, auth_headers):
     get_resp = test_client.get(
-        f'/exports/mine-summary-csv', headers=auth_headers['full_auth_header'])
+        '/exports/mine-summary-csv', headers=auth_headers['full_auth_header'])
 
     assert get_resp.status_code == 200
 
 def test_mine_summary_export_paginated(test_client, db_session, auth_headers):
     get_resp = test_client.get(
-        f'/exports/mine-summary', headers=auth_headers['full_auth_header'])
+        '/exports/mine-summary', headers=auth_headers['full_auth_header'])
 
     assert get_resp.status_code == 200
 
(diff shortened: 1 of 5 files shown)


services/core-api/tests/exports/test_export_success.py
```diff
@@ -24,6 +24,21 @@ def test_mine_summary_export_paginated(test_client, db_session, auth_headers):
...
 def test_core_static_content_success(test_client, db_session, auth_headers):
     get_resp = test_client.get(
-        f'/exports/core-static-content', headers=auth_headers['full_auth_header'])
+        '/exports/core-static-content', headers=auth_headers['full_auth_header'])
 
-    assert get_resp.status_code == 200
\ No newline at end of file
+    assert get_resp.status_code == 200
+
+def test_now_application_gis_export_success_no_filter(test_client, db_session, auth_headers):
```
matbusby-fw, line 31: Yay more tests!

matbusby-fw (approved):
Everything looks solid to me.  Always a little bit annoying working with the stuff that only exists in Prod.

simensma-fresh (approved):

PR #3293, opened 2024-11-05
[MDS-6129] Clean-up in project summary and add react-redux typing

taraepp:
## Objective 
- making a PR before making the real PR with some work that might be out of scope...
- I like library imports at the top, before system imports. 😁 
- some linter issues, like React Fragments being used where they don't do anything
- there's also a loop that doesn't seem to be using its loopiness but I didn't want to try breaking it just yet. I let the linter turn it from a `let i` loop to a `for const` loop though.
- that `<a>` I took out wasn't linking anywhere
- I didn't like that the `isFieldDisabled` function is specific to projects but was in a generic utils file with a generic-sounding name (probably put in because `removeNullValuesRecursive` is there). But I also don't like that it's not specific to the field it's potentially disabling and therefore doesn't need to be called 88 times. Still deciding whether to just call it once (or max once per file) or turn it into a createSelector or memoize it.
- added react-redux typing
   - you can kind of tell by where the `@ts-ignore`s are added is where our current typing isn't working out
   - technically, it wasn't working out before, but now it's telling us about it!
   - my compiler shows 142 type errors now (from all over), and I'm pretty sure they weren't there before. 🤦‍♀️  
   - it doesn't like the `dispatch(args).then(func)`
   - it also doesn't like the state being passed to userHasRole and similar stuff
   - I briefly experimented with making FormWrapper a generic by copying the approach that antd uses (kind of neat! `<Values = any>`- so it's generic but if you don't give it a type it defaults to any) but it wasn't working and it was way too complicated with redux-form, which initializes FormData as {}, which really interferes with typing (hence the `as IProjectSummaryForm` instead of `formValues: IProjectSummaryForm`)
- I'm not dead-set on the way I set up those interfaces for IProjectSummaryForm. I just quickly added attributes that seemed new or different into a new interface.
- Also not dead-set on the name IProjectSummaryForm, but especially due to the changes in authorizations, it seemed better to separate out as 2 different interfaces.


simensma-fresh (approved):
Yay for more typings! For the future we should deal with those `ts-ignore`s, think we're just missing an `unwrap()` call. But that's probably best dealt with on a case by case basis

https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions

asinn134 (approved):

PR #3296, opened 2024-11-06
[MDS-5697] Orgbook Publisher, data typing issues and clean up

Jsyro:
one list issue and a date formatting issue resolved. 

reinstate the celery job of task rather than the api doing it. 

rename env var `ORGBOOK_PUBLISHER_BASE_URL` and new env var `ORGBOOK_PUBLISHER_API_KEY` as if the credential isn't signed, then the orgbook publisher needs a different way to authenticate. I have the value needed for `dev` and will provide to @simensma-fresh


matbusby-fw (dismissed):

matbusby-fw (approved):
🍏

simensma-fresh (approved):

PR #3300, opened 2024-11-07
[MDS-6244] Switched to use gunicorn for Core API to improve performance and ease debugging of `504`s

simensma-fresh:
## Objective 

[MDS-6244](https://bcmines.atlassian.net/browse/MDS-6244)

Running the Core API app in production with the `flask` cli is discouraged as it's not meant to be a production-ready tool. Switched to using `gunicorn` to enable
- better handling of parallel requests
- better visibility into what happens when a request times out, as it supports printing a stack trace when that happens

### DIFF
--- a/services/core-api/Dockerfile.ci
+++ b/services/core-api/Dockerfile.ci
@@ -33,4 +33,5 @@ USER $USERNAME
 # Run the server
 EXPOSE 5000
 # Works only with no reload. Flas should not be reloading in prod anyway - Tech debt. 
-CMD [ "opentelemetry-instrument", "flask", "run", "--no-reload"]
+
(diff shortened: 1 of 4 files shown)


Jsyro:
This is long overdue. 

Does running gunicorn locally still support hotreloading? or should gunicorn only be for Openshift deployments?

taraepp (approved):

matbusby-fw (approved):
🟢 🦄

simensma-fresh:
@Jsyro Gunicorn does support hot-reloading for local development, but haven't had a chance to really test it out yet. So sticking with the `flask` cli still for that

PR #1634, opened 2021-01-26
Issue permits as vc with mdt

Jsyro:
Still needs:
- [x] dev openshift secret
- [x] test openshift secret
- [x] backend tests
- [x] add to permit approval
- [x] error handling

Add button (admins only) to call [MDT's issuer controller](https://github.com/bcgov/mines-digital-trust) to issue Verifiable Credentials to businesses in orgbook.

PR #3301, opened 2024-11-08
[FIX] - catch missing orgbook entity and improve logging

Jsyro:


matbusby (approved):
👍

matbusby-fw (dismissed):

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -81,8 +83,9 @@ class W3CCred(BaseModel):
     credentialSchema: List[dict]
 
 
-def convert_date_to_iso_datetime(date: date) -> str:
-    return datetime(date.year, date.month, date.day, 0, 0, 0, tzinfo=ZoneInfo("UTC")).isoformat()
+def convert_date_to_iso_datetime(datetime: datetime) -> str:
+    return datetime(
+        datetime.year, datetime.month, datetime.day, 0, 0, 0, tzinfo=ZoneInfo("UTC")).isoformat()
```
taraepp: Have to check, just because we have a million issues with UTC times- this _converts_ the datetime to UTC, not tells it that the timezone it has is UTC, right?

Jsyro: datetime objects are natively timezone-unaware. 

the python language is deprecating some of these methods in favour or producting timezone-aware datetime objects

well formed VCDM credentials required timezone codes, i could just append `T00:00:00` to the string, but this is a more pythonic way to do that and is better understood by future devs

Jsyro: What that means is existing datetime objects based on UTC don't know they are based on UTC, they are date+time with no timezone information. it's up to the code to know that it's based on UTC, or maybe postgres is automatically adding UTC information? 

it's probably something that should become it's own task. to update the python to start using the new timezone-aware methods and classes, update the validators to require timezone-aware objects (or add it), then update remaining datetime values to have timezone information.

taraepp: 👍 
Sounds good. The db does automatically save it as UTC. I appreciate the documentation!

services/core-api/tests/reports/resource/test_mine_reports_resource.py
```diff
@@ -61,28 +60,28 @@ def test_get_code_required_reports_for_mine(test_client, db_session, auth_header
...
     get_data = json.loads(get_resp.data.decode())
     assert get_resp.status_code == 200
 
     for report in get_data['records']:
         received_date = datetime.strptime(report['received_date'], '%Y-%m-%d')
 
-        assert (start_date <= received_date.date())
-        assert (received_date.date() <= end_date)
+        assert (start_date <= received_date)
+        assert (received_date <= end_date)
```
taraepp: wait, did you fix our extremely flaky python test??

taraepp (dismissed):
🎉

matbusby-fw (dismissed):

services/core-api/tests/reports/resource/test_mine_reports_resource.py
Jsyro: I'm trying too, but i think so?

the TODAY generator in the factory file is being very odd....

Jsyro:
This line scares me.... 

https://github.com/bcgov/mds/blob/fffc9aa8ef8f64c41b28e6a278894aa74f5407df/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py#L123

taraepp:
> This line scares me....
> 
> https://github.com/bcgov/mds/blob/fffc9aa8ef8f64c41b28e6a278894aa74f5407df/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py#L123

*clicks git blame*
https://github.com/bcgov/mds/pull/1634 😂 
It was you!

On the bright side: you've learned something in the last 3 years.

Jsyro:
> > This line scares me....
> > https://github.com/bcgov/mds/blob/fffc9aa8ef8f64c41b28e6a278894aa74f5407df/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py#L123
> 
> _clicks git blame_ #1634 😂 It was you!
> 
> On the bright side: you've learned something in the last 3 years.

that's not funny at all. 

~~it implies that dates and datetimes are both being used in the code, but the original intention was to only ever store datetime objects. ~~
^This is wholly incorrect apparently, need to do more digging.  

this one hurts, and on a friday no less.

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -81,8 +83,8 @@ class W3CCred(BaseModel):
     credentialSchema: List[dict]
 
 
-def convert_date_to_iso_datetime(date: date) -> str:
-    return datetime(date.year, date.month, date.day, 0, 0, 0, tzinfo=ZoneInfo("UTC")).isoformat()
+def convert_date_to_iso_datetime(dt: datetime) -> str:
+    return datetime(dt.year, dt.month, dt.day, 0, 0, 0, tzinfo=ZoneInfo("UTC")).isoformat()
```
taraepp, line 87: this function I think I might like to see in a generic helper/utils/whatever file. I think we might have more use for it!

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -461,17 +463,44 @@ def produce_untp_cc_map_payload(cls, did: str,
...
+            pmt_appt_start_date = None
+            if not pmt_appt.start_date:
+                pmt_appt_start_date = datetime(1900)
+            elif isinstance(pmt_appt.start_date, date):
+                pmt_appt_start_date = pmt_appt.start_date
+            elif isinstance(pmt_appt.start_date, datetime):
+                pmt_appt_start_date = pmt_appt.start_date.date()
+            else:
+                print(type(pmt_appt.start_date))
+                print("WHAT?")
```
taraepp: 😂 
Miiiiiight want something else for an error log message. A little more context for the rest of us.

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: It's currently converting it to a string that is needed specifically for vcdm. so maybe it's name should be clearer.

taraepp (approved):

simensma-fresh (approved):
Great! Thanks for cleaning up those date <-> datetime inconsistencies 🙌

PR #3302, opened 2024-11-08
[MDS-6205] Permit condition specific reports

matbusby-fw:
## Objective 

- Created a mine_report_permit_requirement table to set up permit required reports with frequency that can be attached to a permit condition.
- Updated front and backend to interface with this new table in the permit conditions view
- Set up create of a mine_report_permit_requirement flow

[MDS-6205](https://bcmines.atlassian.net/browse/MDS-6205)

permit-reports


services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
```diff
@@ -0,0 +1,195 @@
...
+  REPORT_TYPE_CODES,
+} from "@mds/common";
+import { required } from "@mds/common/redux/utils/Validate";
+import FormWrapper from "@mds/common/components/forms/FormWrapper";
+import RenderSelect from "@mds/common/components/forms/RenderSelect";
+import RenderDate from "@mds/common/components/forms/RenderDate";
+import RenderCancelButton from "@mds/common/components/forms/RenderCancelButton";
+import RenderField from "@mds/common/components/forms/RenderField";
+import RenderRadioButtons from "@/components/common/RenderRadioButtons";
+import { requiredRadioButton } from "@common/utils/Validate";
```
taraepp: would want to update the import for requiredRadioButton and RenderRadioButtons to be from @mds/common

services/core-web/src/components/common/RenderRadioButtons.tsx
```diff
@@ -46,8 +47,17 @@ const RenderRadioButtons: FC<RenderRadioButtonsProps> = (props) => {
         value={input.value}
         onChange={handleRadioChange}
         id={id as string}
-        options={options}
-      />
+      >
```
taraepp, line 50: the common version has isVertical instead of vertical for a parameter. Just something to keep in mind

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
```diff
@@ -1,32 +1,185 @@
...
       <div className={expandClass}>
-        <p className={className}>
+        <p>
           {condition.step} {condition.condition}
         </p>
-        {condition?.sub_conditions?.map((condition) => {
+        {sectionEdit && (
+          <Row justify="space-between">
+            <Row gutter={8} className="condition-edit-buttons" align="middle">
+              <Button
```
taraepp: I think that all the children of `<Row>` need to be nested under a `<Col>` element(s). It may work a little better/as expected and make some of the added CSS unnecessary.

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -65,14 +74,43 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
+        return {
+          ...condition,
+          stepPath,
+          mineReportPermitRequirement,
+          sub_conditions,
+        };
+      };
+
+      // Initialize the step paths for all top-level conditions
+      const updatedConditions = conditions.map((condition) => getStepPath(condition));
```
taraepp: perhaps a bit of a nitpick, but are these _updated_ conditions, as in they have updates that the user is making that need to be saved? Or are they _formatted_?

taraepp:
I am still a bit unsure if what the condition requirement should be mapped to is the permit or the permit amendment (granted, the permit condition is _already_ mapped to the permit amendment which is mapped to the permit but I get trying to avoid excessively complicated queries). But I also haven't tried running or rendering it. The decider to me would be "What happens when a permit has 2 amendments, and each of them have conditions with different report requirements?" <-- if it can handle that as is, then I'm happy. May just need to have dataMocks be a little more fleshed out with such a scenario.

services/core-web/src/components/mine/Permit/PermitConditions.tsx
matbusby-fw: Fair enough.  I'll update it!

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
matbusby-fw: Ah, yep.  Just missed this one.

matbusby-fw:
> I am still a bit unsure if what the condition requirement should be mapped to is the permit or the permit amendment (granted, the permit condition is _already_ mapped to the permit amendment which is mapped to the permit but I get trying to avoid excessively complicated queries). But I also haven't tried running or rendering it. The decider to me would be "What happens when a permit has 2 amendments, and each of them have conditions with different report requirements?" <-- if it can handle that as is, then I'm happy. May just need to have dataMocks be a little more fleshed out with such a scenario.

It worked since it was just using it to grab the requirements for that permit, and since the important connection was between permit_condition and the requirement it would maintain the correct usage.  That said, I think you're right that amendment is actually more appropriate, so I updated it to adhere to that.

services/common/src/redux/selectors/permitSelectors.ts
```diff
@@ -84,3 +85,18 @@ export const getPermits = createSelector([getUnformattedPermits], (permits) => {
   const formattedPermits = permits.map((permit) => formatPermit(permit));
   return formattedPermits;
 });
+
+export const getMineReportPermitRequirements = (permitGuid) =>
+  createSelector(
+    [getPermitByGuid(permitGuid)],
```
taraepp: minor- but I would recommend replacing getPermitByGuid with getLatestAmendmentByPermitGuid here.

services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
```diff
@@ -0,0 +1,87 @@
...
+
+    mine_report_permit_requirement_id: int = db.Column(db.Integer, primary_key=True, server_default=FetchedValue())
+    due_date_period_months: int = db.Column(db.Integer, nullable=False)
+    initial_due_date: Optional[date] = db.Column(db.Date, nullable=True)
+    active_ind: bool = db.Column(db.Boolean, nullable=False, server_default=FetchedValue())
+    cim_or_cpo: Optional[CimOrCpo] = db.Column(db.Enum(CimOrCpo, name='cim_or_cpo_type'), nullable=True)
+    ministry_recipient: Optional[list[OfficeDestination]] = db.Column(
+        ARRAY(db.Enum(OfficeDestination, name='ministry_recipient_type')), nullable=True)
+    permit_condition_id: int = db.Column(db.Integer, db.ForeignKey('permit_conditions.permit_condition_id'))
+    permit_amendment_id: int = db.Column(db.Integer, db.ForeignKey('permit_amendment.permit_amendment_id'))
```
taraepp, line 42: Wow, so much typing! 🎨

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
```diff
@@ -1,32 +1,206 @@
...
-        {condition?.sub_conditions?.map((condition) => {
+        {sectionEdit && (
+          <Row justify="space-between" align="middle">
+            <Col>
+              <Row gutter={8} className="condition-edit-buttons" align="middle">
+                <Col>
+                  <Button
+                    onClick={(event) => {
+                      event.stopPropagation();
+                      console.log("Not yet implemented");
```
taraepp, line 117: Just curious what the feature flag status of this is. I'm seeing a Feature.MODIFY_PERMIT_CONDITIONS (separate from Feature.PERMIT_CONDITIONS_PAGE) that should possibly be blocking these buttons.

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -184,7 +270,7 @@ const PermitConditions: FC<PermitConditionProps> = ({
           </Col>
         </Row>
       }
-    />
```
taraepp: I think this one wants to stay a self-closing tag? It doesn't take any children

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
```diff
@@ -0,0 +1,74 @@
...
+from app.extensions import api
+
+from flask_restx import Resource
+
+from app.api.mines.reports.models.mine_report_permit_requirement import CimOrCpo, MineReportPermitRequirement
+from app.api.utils.custom_reqparser import CustomReqparser
+from app.api.utils.resources_mixins import UserMixin
+
+
+class MineReportPermitRequirementResource(Resource, UserMixin):
```
simensma-fresh, line 21: Would be great if we could add an entry for this endpoint in  `test_expected_auth.py`

services/core-api/app/api/utils/helpers.py
```diff
@@ -2,6 +2,9 @@
 from io import BytesIO
 import re
 from datetime import datetime
+
+from flask_restx import ValidationError
+from flask_restx.fields import Raw
```
simensma-fresh: Intentional to add these?

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
matbusby-fw: Isn't this whole page behind a feature flag at this point in time?  I should check that.

simensma-fresh (changes requested):

services/core-web/src/components/mine/Permit/PermitConditions.tsx
matbusby-fw: That's weird.  I made no concious decision to change that.

services/core-api/app/api/utils/helpers.py
matbusby-fw: Ah, I removed the thing that used them.  I'll remove the imports

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
matbusby-fw: 👍🏻

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
taraepp: Technically the page is behind PERMIT_CONDITIONS_PAGE (which is currently off) but we have modifications controlled by the MODIFY flag separately. So we could make it possible to _view_ data and not _change_ anything. I think it would be nice to be able to hide the unimplemented features, specifically.

simensma-fresh (approved):

taraepp (approved):

asinn134 (approved):

PR #3305, opened 2024-11-13
[MDS-6129] Project Summary emails and notification fixes

taraepp:
## Objective 
- in status update emails, only send when status has changed, and make the email specific to the new status
- send emails when there's new documents
- check properly if there's _new_ documents
- send email/notification updates only when the status indicates, and only to those recipients that should be getting them
- fix a spelling error and non-pythonic variable name
- these changes would have possibly/likely broken the notifications for mine applications and IRT (by requiring the status code in the notify function), but that ticket is coming up, soo.... 
- BONUS: include the "original recipients" in email subject line (except for prod) so that logs don't have to be checked to confirm
- BONUS: make BE logs readable on local (ie: don't include trace id and keycloak id)

### DIFF
--- a/services/core-api/app/api/mines/documents/resources/mine_document_resource.py
+++ b/services/core-api/app/api/mines/documents/resources/mine_document_resource.py
@@ -136,24 +136,26 @@ class MineDocumentArchiveResource(Resource, UserMixin):
             project = None
             doc = documents[0]
             mine_document_guid = doc.mine_document_guid
-            isNotifiableDoc = False
+            is_notifiable_doc = False
+            status_code = None
 
             if doc.major_mine_application_document_xref:
                 project = MajorMineApplication.find_by_mine_document_guid(mine_document_guid).project
-                isNotifiableDoc = True
+                is_notifiable_doc = True
             elif doc.project_summary_document_xref:
                 project = ProjectSummary.find_by_mine_document_guid(mine_document_guid).project
-                isNotifiableDoc = True
+                is_notifiable_doc = True
+                status_code = project.project_summary.status_code
             elif doc.project_decision_package_document_xref:
                 project = ProjectDecisionPackage.find_by_mine_document_guid(mine_document_guid).project
-                isNotifiableDoc = True
+                is_notifiable_doc = True
             elif doc.information_requirements_table_document_xref:
                 project = InformationRequirementsTable.find_by_mine_document_guid(mine_document_guid).project
-                isNotifiableDoc = True
+                is_notifiable_doc = True
 
             # If one of the *xref value is not None that means the notification should be sent.
-            if isNotifiableDoc:
-                ProjectUtil.notifiy_file_updates(project, mine)
+            if is_notifiable_doc:
+                ProjectUtil.notify_file_updates(project, mine, status_code)
 
         return None, 204
 
--- a/services/core-api/app/api/mines/documents/resources/mine_document_resource.py
+++ b/services/core-api/app/api/mines/documents/resources/mine_document_resource.py
@@ -7,7 +7,7 @@ from app.api.projects.project_summary.models.project_summary import ProjectSumma
 from app.api.projects.project_decision_package.models.project_decision_package import ProjectDecisionPackage
 from app.api.projects.information_requirements_table.models.information_requirements_table import InformationRequirementsTable
 
-from flask_restx import Resource, reqparse, fields
+from flask_restx import Resource, reqparse
 from datetime import datetime
 from werkzeug.exceptions import BadRequest, NotFound
 
@@ -22,7 +22,7 @@ from app.api.mines.documents.mine_document_search_util import MineDocumentSearch
 from app.api.mines.response_models import ARCHIVE_MINE_DOCUMENT, MINE_DOCUMENT_MODEL, DOCUMENT_MANAGER_ZIP
 
 from app.api.services.document_manager_service import DocumentManagerService
-from app.api.projects.project.project_util import ProjectUtil
+from app.api.projects.project.project_util import notify_file_updates
 
 class MineDocumentListResource(Resource, UserMixin):
     @api.doc(description='Returns list of documents associated with mines')
@@ -155,7 +155,7 @@ class MineDocumentArchiveResource(Resource, UserMixin):
 
             # If one of the *xref value is not None that means the notification should be sent.
             if is_notifiable_doc:
-                ProjectUtil.notify_file_updates(project, mine, status_code)
+                notify_file_updates(project, mine, status_code)
 
         return None, 204
 
--- a/services/core-api/app/api/parties/party/models/party.py
+++ b/services/core-api/app/api/parties/party/models/party.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from flask import current_app
 import re
+from typing import Self
 
 from sqlalchemy import func, case, and_
 from sqlalchemy.schema import FetchedValue
@@ -196,7 +197,7 @@ class Party(SoftDeleteMixin, AuditMixin, Base):
                     else_=cls.party_name)
 
     @classmethod
-    def find_by_party_guid(cls, party_guid):
+    def find_by_party_guid(cls, party_guid) -> Self:
         return cls.query.filter_by(party_guid=party_guid, deleted_ind=False).one_or_none()
 
     @classmethod
--- a/services/core-api/app/api/projects/project/project_util.py
+++ b/services/core-api/app/api/projects/project/project_util.py
@@ -2,15 +2,12 @@ from app.api.activity.models.activity_notification import ActivityRecipients
 from app.api.utils.feature_flag import is_feature_enabled, Feature
 from app.api.activity.utils import trigger_notification
 from app.api.activity.models.activity_notification import ActivityType
-from flask import current_app
 
-class ProjectUtil:
 
-    @staticmethod
-    def notify_file_updates(project, mine, status_code = None):
-        status_codes = ["SUB", "ASG", "CHR"]
+def notify_file_updates(project, mine, status_code = None):
+    status_codes = ["SUB", "ASG", "CHR"]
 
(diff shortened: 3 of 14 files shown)


services/core-api/app/api/projects/project/project_util.py
```diff
@@ -6,8 +6,11 @@
 
 class ProjectUtil:
 
-    def notifiy_file_updates(project, mine):
-        if is_feature_enabled(Feature.MINE_APPLICATION_FILE_UDPATE_ALERTS):
+    @staticmethod
+    def notify_file_updates(project, mine, status_code = None):
+        status_codes = ["SUB", "ASG", "CHR"]
```
taraepp: for IRT/application, if the status codes are different, may want to pass in some indication of IRT/app/summary. 🤷‍♀️ 

Would be an improvement to use an Enum. I think an enum wouldn't be out of place in this file.

services/core-api/app/api/projects/project_summary/resources/project_summary.py
```diff
@@ -308,9 +270,39 @@ def put(self, project_guid, project_summary_guid):
...
-            project = Project.find_by_project_guid(project_guid)
-            ProjectUtil.notifiy_file_updates(project, mine)
+        # notify of status changes
+        if prev_status != project_summary.status_code:
+            message = ''
+            extra_data = {'project': {'project_guid': str(project.project_guid)}}
+            if prev_status == 'DFT' and project_summary.status_code == 'SUB':
+                message = f'A new major project description for ({project.project_title}) has been submitted for ({project.mine_name})'
+
+            if project_summary.status_code == 'ASG':
```
taraepp, line 280: here I mostly just put all the repetitive stuff at the top/bottom (data at top, calls at bottom).
But the ASG status was missing.

services/core-api/app/templates/email/projects/emli_project_summary_email.html
```diff
@@ -1,114 +1,144 @@
...
+            <tr>
+              <td class="wrapper" style="vertical-align: top; box-sizing: border-box; padding: 20px;" valign="top">
+                <table role="presentation" cellpadding="0" cellspacing="0"
+                  style="border-collapse: separate; mso-table-lspace: 0pt; mso-table-rspace: 0pt; width: 100%;"
+                  width="100%">
+                  <tr>
+                    <td
+                      style="vertical-align: top; text-align: center; font-family: sans-serif; font-size: 24px; font-weight: 700; line-height: 30px; color: #5E46A1; padding: 16px 0px 24px;"
+                      valign="top">
+                      {{message}}
```
taraepp, line 78: This is the only change (making message dynamic). Rest is formatting.

services/core-api/app/api/mines/documents/resources/mine_document_resource.py
```diff
@@ -136,24 +136,26 @@ def patch(self, mine_guid):
...
             elif doc.information_requirements_table_document_xref:
                 project = InformationRequirementsTable.find_by_mine_document_guid(mine_document_guid).project
-                isNotifiableDoc = True
+                is_notifiable_doc = True
 
             # If one of the *xref value is not None that means the notification should be sent.
-            if isNotifiableDoc:
-                ProjectUtil.notifiy_file_updates(project, mine)
+            if is_notifiable_doc:
+                ProjectUtil.notify_file_updates(project, mine, status_code)
```
taraepp: This whole thing should probably go in project utils instead of mine document resource.

services/core-api/app/api/projects/project/project_util.py
Jsyro: the staticmethod here seems odd, staticmethod in python is a classmethod (instead of an instance method). 

the first parameter is usually the class itself and is provided automatically.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -166,6 +166,11 @@ def project_summary_lead_party_guid(self):
         if self.project.project_lead_party_guid:
             return self.project.project_lead_party_guid
         return None
+    
+    @hybrid_property
+    def project_lead_email(self):
+        project_lead: Party = Party.find_by_party_guid(self.project_summary_lead_party_guid)
```
Jsyro:
```suggestion
        project_lead = Party.find_by_party_guid(self.project_summary_lead_party_guid)
```
don't need to type the local variable as it's being set right away, it should be infered by the `Party.find_by_party_guid` method's return type, so make the change there and let it infer. 

typing that would be helpful if you needed to initialize it to null then reassign it later.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -166,6 +166,11 @@ def project_summary_lead_party_guid(self):
         if self.project.project_lead_party_guid:
             return self.project.project_lead_party_guid
         return None
+    
+    @hybrid_property
+    def project_lead_email(self):
```
Jsyro:
```suggestion
    def project_lead_email(self) -> str | None:
```
I think is how python 3.12 does optional params now, using the logical OR operator `|`.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -226,6 +231,29 @@ def find_by_mine_document_guid(cls, mine_document_guid):
 
         except ValueError:
             return None
+        
+    @staticmethod
+    def has_new_documents(documents, ams_authorizations) -> bool:
```
Jsyro, line 231: Nice typing!

Jsyro (changes requested):
Couple small python comments, can't speak the functional change, but looks good overall.

taraepp:
> Couple small python comments, can't speak the functional change, but looks good overall.

Totally what I requested your review for! Thanks, I'll make those changes when I've got a minute.

services/core-api/app/api/projects/project/project_util.py
taraepp: Well, here's my reasoning, and feel free to let me know where I'm going wrong!

- Might not be obvious that this class is ProjectUtils and not Project
- before it was neither static nor class method, and it was complaining that the first parameter (project) should be renamed to self, which was a hint to me that something was _terribly wrong_.
- when choosing between class and static methods, my shorthand is "do I need to access the instance of the class (ie: the self parameter)" and yes = class and no = static method. This class has no attributes and is really just a (very small) collection of functions, so maybe there's a better way to set that up.

services/core-api/app/api/parties/party/models/party.py
```diff
@@ -196,7 +197,7 @@ def name(cls):
                     else_=cls.party_name)
 
     @classmethod
-    def find_by_party_guid(cls, party_guid):
+    def find_by_party_guid(cls, party_guid) -> Self:
```
Jsyro, line 200: good find. https://typing.readthedocs.io/en/latest/spec/generics.html#self

services/core-api/app/api/projects/project/project_util.py
```diff
@@ -6,8 +6,11 @@
 
 class ProjectUtil:
 
-    def notifiy_file_updates(project, mine):
-        if is_feature_enabled(Feature.MINE_APPLICATION_FILE_UDPATE_ALERTS):
+    @staticmethod
+    def notify_file_updates(project, mine, status_code = None):
```
Jsyro:
```suggestion
    def notify_file_updates(cls, mine, status_code = None):
```

Then i would recommend this.

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -1253,8 +1281,53 @@ def delete(self, commit=True):
             doc.mine_document.delete(False)
         return super(ProjectSummary, self).delete(commit)
 
-    def send_project_summary_email(self, mine):
-        emli_recipients = PROJECT_SUMMARY_EMAILS
+    def send_project_summary_document_email(self, mine):
```
Jsyro:
```suggestion
    def send_project_summary_document_email(self, mine) -> None:
```
If no return value.

services/core-api/app/api/projects/project/project_util.py
taraepp: the class isn't actually a project though. It's a completely different class.
Like I'd actually have to change it to `def notify_file_updates(cls, project, mine, status_code):` and not use the cls parameter. 

I think I actually want to refactor the file to be more like `services/core-api/app/activity/utils.py`

Jsyro: I'm totally wrong. you've got this right.

services/core-api/app/templates/email/projects/emli_project_summary_email.html
simensma-fresh: I really really want to create some reusable email templates out of these...

One day... 🙏

simensma-fresh (dismissed):

services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -1253,8 +1281,53 @@ def delete(self, commit=True):
...
+                        "mine_no": mine.mine_no,
+                    },
+                    "message": message,
+                    "core_project_summary_link": f'{Config.CORE_WEB_URL}/pre-applications/{self.project.project_guid}/overview'
+                }
+                EmailService.send_template_email(subject, email_recipients, emli_body, emli_context, cc=cc)
+
+
+    def send_project_summary_email(self, mine, message):
+        current_app.logger.info(f'HI TARA Sending email with message {message}')
```
asinn134: Is this log meant to be removed?

asinn134 (dismissed):

services/core-api/app/api/projects/project_summary/models/project_summary.py
taraepp: Oops, I thought I caught all of these! Will remove.

Jsyro (approved):
python looks good

asinn134 (approved):

PR #3314, opened 2024-11-22
[FIX] - Include end dates that are null or Dec-31-9999

Jsyro:
this adds many results to the query that should be there!


matbusby-fw (approved):
Take that, people living in the year 10,000!

taraepp (approved):

PR #3320, opened 2024-11-26
[CHORE] - add cleanup job for failed publish records. add validUntil for superceded records

Jsyro:
If there are issues, we don't want to retry records for ever, but we do want to report on records with problems and be able to clean it up easily.


simensma-fresh (dismissed):

simensma-fresh (approved):

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -279,11 +279,23 @@ def push_untp_map_data_to_publisher():
     #token is valid for an hour currently.
     publisher_service = OrgbookPublisherService()
 
-    for row in permit_amendment_query_results:
+    for index, row in enumerate(permit_amendment_query_results):
```
matbusby-fw, line 282: Oh this is cool.  I like it.

matbusby-fw (approved):

PR #3327, opened 2024-11-28
[MDS-6264] Updating activity feed when status changes

asinn134:
## Objective 

[MDS-6264](https://bcmines.atlassian.net/browse/MDS-6264)

- Updated activity feed to change without page refresh when project summary status is changed.


services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -239,6 +242,8 @@ export const ProjectSummary: FC = () => {
         handleTabChange(newActiveTab);
         setIsLoaded(true);
       }
+
+      dispatch(fetchActivities(userInfo?.preferred_username));
```
asinn134, line 246: The only real change is having this dispatch here so when we use the `getActivities` selector in the NotificationDrawer component it will have the latest activity feed information.

services/core-web/src/components/navigation/NotificationDrawer.tsx
```diff
@@ -246,19 +243,4 @@ const NotificationDrawer: FC<INotificationDrawerProps> = (props) => {
...
-    {
-      fetchActivities,
-      markActivitiesAsRead,
-      storeActivities,
-    },
-    dispatch
-  );
-
-export default connect(mapStateToProps, mapDispatchToProps)(NotificationDrawer);
+export default NotificationDrawer;
```
matbusby-fw, line 246: Yeah!  Get outta here mapStateToProps and mapDispatchToProps!

matbusby-fw (approved):
Nice!  Looks great!

services/core-web/src/components/navigation/NotificationDrawer.tsx
asinn134: Thanks again for talking it through with me when I was too focused on the infinite loop I was getting!

simensma-fresh (approved):

PR #3334, opened 2024-12-05
[MDS-6263] Adjustments around Assigned status not behaving as aspected

asinn134:
## Objective 

[MDS-6263](https://bcmines.atlassian.net/browse/MDS-6263)

- Changed unassigning project lead to not send blank activity for Core and Minespace

- Changed assigning project lead to not send notifications for Minespace

- Added new text provided in mockup

- Adjusted Minespace fields to be editable when a project summary that hasn't been submitted yet has an assigned project lead

- Adjusted Core's project management to be able to save a new status and also make an adjustment to the project lead input without the status being forced to be "Submit" or "Assigned"

- Adjusted declaration section's submission in Core to change status to "Submit" when a draft project summary is completed


matbusby-fw (dismissed):
Nice work!  Looks solid!

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
```diff
@@ -215,19 +215,15 @@ export const ProjectSummary: FC = () => {
...
     if (!status_code || isNewProject) {
       status_code = "DFT";
-    } else if (!newActiveTab) {
-      if (isCore) {
-        status_code = formValues.status_code;
-      } else {
-        status_code = "SUB";
-      }
+    } else if (!newActiveTab && status_code === "DFT") {
+      status_code = "SUB";
```
taraepp, line 213: Noticed a sonarcloud issue- just some unused stuff that wants to be removed (isCore wants to be removed, and then probably related imports like the SystemFlagEnum and the selector)

services/common/src/components/projects/projectUtils.spec.ts
```diff
@@ -9,8 +9,8 @@ const TEST_PARAMETERS = [
         testFunction: areFieldsDisabled,
         coreDisabledStatuses: ["WDN", "COM"],
         coreEnabledStatuses: ["DFT", "SUB", "ASG", "UNR", "CHR", "OHD"],
-        msDisabledStatuses: ["SUB", "ASG", "UNR", "WDN", "OHD", "COM"],
-        msEnabledStatuses: ["DFT", "CHR"],
+        msDisabledStatuses: ["SUB", "UNR", "WDN", "OHD", "COM"],
+        msEnabledStatuses: ["DFT", "CHR", "ASG"],
```
taraepp: I do find it confusing that Assigned was added to enabled- like, generally they're still not supposed to be able to edit when it's in Assigned status, right? Is the status even assigned when there's no project lead assigned?

taraepp: Hmm, so is the flow that's causing issues basically:
1. an MS/CORE user makes a draft project summary
2. they assign a project lead
3. the project gets put in Assigned status even though it's not submitted

I guess it's pretty wacky! I'm thinking that for the tests though, maybe pass a parameter to indicate that it _has_ been submitted so that we know what to expect generally, and then add another test for the special case where it's assigned but not submitted?

services/core-web/cypress/e2e/majorprojects.cy.ts
```diff
@@ -18,6 +18,12 @@ describe("Major Projects", () => {
   it("should upload and download a document successfully", () => {
     const fileName = "dummy.pdf";
 
+    cy.get('body').then((element) => {
```
asinn134, line 21: Depending on the status of the project summary, clicking the button for project description can bring us to the project description tab or the actual project description form page. Just have to include this to make sure we click the `View Project Description Details` button if we do get sent to the Project Description tab.

services/core-web/src/components/mine/Projects/ProjectSummary.tsx
asinn134: This should be removed now.

services/common/src/components/projects/projectUtils.spec.ts
asinn134: I reverted the change to the arrays, and a new test has been added now.

matbusby-fw (approved):

taraepp (approved):

PR #3335, opened 2024-12-05
[MDS-6206] Tweak permit condition category input field sizes

simensma-fresh:
## Objective 

[MDS-6206](https://bcmines.atlassian.net/browse/MDS-6206)

Updated size of step + description fields for permit categories to
- reflect that step should likely only be one character
- added a minWidth to the autocomplete description field
image

### DIFF
--- a/services/core-web/src/components/mine/Permit/PermitConditionCategory.tsx
+++ b/services/core-web/src/components/mine/Permit/PermitConditionCategory.tsx
@@ -67,7 +67,7 @@ export const EditPermitConditionCategoryInline = (props: IPermitConditionCategor
     <FormWrapper scrollOnToggleEdit={false} name={formName} onSubmit={handleSubmit} initialValues={props.category} isEditMode={isEditMode}>
       <Row className="flex" style={{ gap: '0.5em' }}>
         <Form.Item style={{ marginRight: 0, maxWidth: '40px' }}>
-          <Field name="step" component={RenderField} required={true} validate={[required]} style={{ maxWidth: '100px' }} />
+          <Field name="step" component={RenderField} required={true} validate={[required]} />
         </Form.Item>
         <PermitConditionCategorySelector showLabel={false} />
         <Button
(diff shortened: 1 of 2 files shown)


asinn134 (dismissed):

services/core-web/src/components/mine/Permit/PermitConditionCategory.tsx
```diff
@@ -66,7 +66,9 @@ export const EditPermitConditionCategoryInline = (props: IPermitConditionCategor
   return (
     <FormWrapper scrollOnToggleEdit={false} name={formName} onSubmit={handleSubmit} initialValues={props.category} isEditMode={isEditMode}>
       <Row className="flex" style={{ gap: '0.5em' }}>
-        <Field name="step" component={RenderField} required={true} validate={[required]} />
+        <Form.Item style={{ marginRight: 0, maxWidth: '40px' }}>
+          <Field name="step" component={RenderField} required={true} validate={[required]} />
```
taraepp: The Field component does already have a Form.Item within it, so I really don't think there should be another one wrapping it. Does it want to be a div? A Col? Probably a Col because it's a child of Row and Row likes all its direct children to be Cols. I did something similar on my branch, and to save you looking up the antd props, the Row probably wants `wrap={false}`

simensma-fresh: Updated. Kept the wrap so it wraps nicely when squished. Kept the buttons in the same column so they wrap together

asinn134 (dismissed):

matbusby-fw (dismissed):

services/core-web/src/components/mine/Permit/PermitConditionCategory.tsx
```diff
@@ -3,7 +3,7 @@ import RenderField from "@mds/common/components/forms/RenderField";
 import RenderSubmitButton from "@mds/common/components/forms/RenderSubmitButton";
 import { FORM } from "@mds/common/constants/forms";
 import { IPermitConditionCategory } from "@mds/common/interfaces";
-import { Button, Popconfirm, Row, Tooltip, Typography } from "antd";
+import { Button, Col, Form, Popconfirm, Row, Tooltip, Typography } from "antd";
```
taraepp: sonarcloud complaining about unused imports!

asinn134 (approved):

taraepp (approved):

PR #3337, opened 2024-12-06
[CHORE] Refactor reducers

taraepp:
## Objective 
Make sure to "hide whitespace changes"! There's a lot of whitespace changes in files that I just did name updates.
- make it so you can use typed network reducers/selectors
   - Actual changes:
      - copied reducerTypes into networkReducerTypes, separated the contents by type, made NRT an enum
      - networkReducer.ts <-- made it TS-y, threw a selector in there. If I felt like making it more work, could have converted to a slice, but didn't want to do that today. But it should be _easier_ to do that now if we wanted to. Actually, if we did that, we wouldn't need a reducer for each request type, could just have networkReducer with currentRequests or something, and maybe that's the plan down the road, but then I need to find all the instances of state[GET_X].isFetching and then it gets bigger.... one thing at a time
      - rootReducerShared.ts <-- made it loop through all the network reducers to create them so they all end up being defined (some were missing)
- added a selector for getIsFetching to the reducer file and maybe it should be in its own file (I'll move it if anyone wants it moved, but personally I don't really care)
- removed some unused references
- basically just a WHOLE LOT of find + replace
- I did verify that the reference updates are all good. 👍 
   - I was missing a couple, but they came up as searchable console warnings
- snuck in a bug fix for MajorMineApplicationTab (the route name was changed but not changed on the file, I thought it was on a previous PR but must have got missed)


simensma-fresh (approved):

matbusby-fw (approved):
Groovy!

PR #3338, opened 2024-12-09
[FIX] Other references to "final-app"

taraepp:
## Objective 
Think these got missed. All my editor could find.


simensma-fresh (approved):

matbusby-fw (approved):
I swear we updated these already (with the exception of the notification drawer maybe).  Did those changes get lost somewhere in a rebaes or something?

PR #3341, opened 2024-12-12
[MDS-6294] - User management

matbusby-fw:
- Created a user table that is versioned and auto updated on login (currently only set up in Core).
- Removed legacy unused user components
- Created a new userSlice for fetching/updating the new user
- Added this request to the initial loading state requests that need to resolve before load completes
- Refactored Home.js to Home.tsx in Core.

## Objective 

[MDS-6294](https://bcmines.atlassian.net/browse/MDS-6294)


services/core-api/app/api/users/resources/user_resource.py
```diff
@@ -0,0 +1,43 @@
...
+from app.api.users.response_models import USER_MODEL
+from app.api.utils.include.user_info import User as UserUtils
+from app.api.utils.resources_mixins import UserMixin
+from app.extensions import api
+
+
+class UserResource(Resource, UserMixin):
+    @api.doc(description='Retrive a list of alerts for a mine')
+    @api.marshal_with(USER_MODEL, code=200)
+    def get(self):
```
simensma-fresh, line 19: Do you think we maybe should restrict this to a base role?
Thinking we might not want to create a user record if they actually don't have the role to view core

simensma-fresh (changes requested):

services/core-api/app/api/users/resources/user_resource.py
matbusby-fw: Done!

simensma-fresh (approved):

migrations/sql/V2024.12.10.15.15__create_users_table.sql
```diff
@@ -0,0 +1,26 @@
...
+    email             VARCHAR                                NOT NULL,
+    given_name        VARCHAR                                NOT NULL,
+    family_name       VARCHAR                                NOT NULL,
+    display_name      VARCHAR                                NOT NULL,
+    idir_username     VARCHAR                                NOT NULL,
+    identity_provider VARCHAR                                NOT NULL,
+    idir_user_guid    VARCHAR                                NOT NULL,
+    last_logged_in    TIMESTAMPTZ,
+    create_user       VARCHAR(255)                           NOT NULL,
+    create_timestamp  timestamp with time zone DEFAULT now() NOT NULL,
```
taraepp, line 13: Just curious what the difference is between "TIMESTAMPTZ" and "timestamp with time zone" (or are they the same but just copied and pasted from different places? Because that is 100% perfectly fine)

migrations/sql/V2024.12.11.22.33__add_user_version_history_table_backfill.sql
```diff
@@ -0,0 +1,6 @@
+-- This file was generated by the generate_history_table_ddl command
+-- The file contains the data migration to backfill history records for the {table} table
+with transaction AS (insert into transaction(id) values(DEFAULT) RETURNING id)
+insert into user_version (transaction_id, operation_type, end_transaction_id, "create_user", "create_timestamp", "update_user", "update_timestamp", "deleted_ind", "sub", "email", "given_name", "family_name", "display_name", "idir_username", "identity_provider", "idir_user_guid")
+select t.id, '0', null, "create_user", "create_timestamp", "update_user", "update_timestamp", "deleted_ind", "sub", "email", "given_name", "family_name", "display_name", "idir_username", "identity_provider", "idir_user_guid"
+from "user",transaction t;
```
taraepp, line 6: For future me/us- does omitting `last_logged_in` from here prevent new records from being created when there's a new `last_logged_in`?

services/common/src/redux/actionCreators/userActionCreator.js
```diff
@@ -1,20 +0,0 @@
...
-import CustomAxios from "../customAxios";
-
-// This file is anticipated to have multiple exports
-// eslint-disable-next-line import/prefer-default-export
-export const fetchCoreUsers = () => (dispatch) => {
-  dispatch(request(NetworkReducerTypes.GET_CORE_USERS));
-  return CustomAxios()
-    .get(ENVIRONMENT.apiUrl + API.CORE_USER, createRequestHeader())
-    .then((response) => {
-      dispatch(success(NetworkReducerTypes.GET_CORE_USERS));
```
taraepp, line 16: I am delighted to delete these files, buuuut can we also then delete, say, NetworkReducerTypes.GET_CORE_USERS and any other unused references?

services/common/src/redux/actions/userActions.js
```diff
@@ -1,8 +0,0 @@
-import * as actionTypes from "@mds/common/constants/actionTypes";
-
-// This file is anticipated to have multiple exports
-// eslint-disable-next-line import/prefer-default-export
-export const storeCoreUserList = (payload) => ({
-  type: actionTypes.STORE_CORE_USERS,
```
taraepp, line 6: like this one

services/common/src/redux/reducers/userReducer.js
```diff
@@ -1,26 +0,0 @@
-import * as actionTypes from "@mds/common/constants/actionTypes";
-import { USERS } from "@mds/common/constants/reducerTypes";
```
taraepp, line 2: probably this one too. 🤩

services/core-api/app/api/users/models/user.py
```diff
@@ -0,0 +1,76 @@
+from datetime import datetime
+from pytz import utc
+from app.api.utils.models_mixins import SoftDeleteMixin, Base, AuditMixin
+from app.extensions import db
+
+
+class User(SoftDeleteMixin, AuditMixin, Base):
+    __tablename__ = "user"
+    __versioned__ = {
+        'exclude': ['last_logged_in']
```
taraepp, line 10: oh wait I see it now. Nice, clear and simple. 👍

migrations/sql/V2024.12.10.15.15__create_users_table.sql
matbusby-fw: Just two ways of referring to the same data type.  It resolves to the same thing.

migrations/sql/V2024.12.11.22.33__add_user_version_history_table_backfill.sql
matbusby-fw: It exludes last_logged_in from the versioning altogether, so it doesn't trigger a version when it's updated and the versions don't even have that field.  When it's fetched, it grabs the value from the non-versioned user entity.

services/common/src/redux/actionCreators/userActionCreator.js
matbusby-fw: Yeah...I started going down a delete path and got nervous about being overzealous.  Probably though.

simensma-fresh (changes requested):
Looks like there's a couple of backend tests that needs fixing, but otherwise looks good

asinn134 (dismissed):

asinn134 (approved):

simensma-fresh (approved):

PR #3343, opened 2024-12-12
[MDS-6273] Project file deletion

sggerard:
## Objective 

Allow users when the application is still in draft status to delete documents on both the 'Purpose and Authorization' and 'Document Upload' tabs of the project application process in both CORE & Minespace.

[MDS-6273](https://bcmines.atlassian.net/browse/MDS-6273)

image

- Added onDeleteDocument methods to both the AuthorizationsInvolved and DocumentUpload components.
- When project is not in a draft status or edit mode then the delete option is not shown.
- When a document is deleted a friendly message is shown, and the document is removed from the document table.
- Updated the project summary delete document endpoint and tests to support amended documents
- Minor change to project uploader to update existing files on file removal to fix the replace functionality.
- Minor change to DocumentTable to remove mutating code.
- Minor scss fix

### DIFF
--- a/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
+++ b/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
@@ -85,10 +85,10 @@ const RenderEMAPermitCommonSections = ({ code, isAmendment, index, isDisabled })
     setShowExemptionSection(value);
   };
 
-  const onDeleteDocument = async (event, key: string)  =>{
+  const onDeleteDocument = (event, key: string) => {
     const document = tableDocuments.find( (doc) => key === doc.key);
     if(document){
-      await dispatch(
+      dispatch(
         removeDocumentFromProjectSummary(
           project_guid,
           project_summary_guid,
--- a/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
+++ b/services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
@@ -58,7 +58,7 @@ import { SystemFlagEnum } from "@mds/common/constants/enums";
 import { getSystemFlag } from "@mds/common/redux/selectors/authenticationSelectors";
 import { FormContext } from "../forms/FormWrapper";
 import { ProjectSummaryFormComponentProps } from "./ProjectSummaryForm";
-import { areAuthEnvFieldsDisabled, areDocumentFieldsDisabled } from "../projects/projectUtils";
+import { areAuthEnvFieldsDisabled, areDocumentFieldsDisabled, isDocumentDeletionEnabled } from "../projects/projectUtils";
 import { removeDocumentFromProjectSummary } from "@mds/common/redux/actionCreators/projectActionCreator";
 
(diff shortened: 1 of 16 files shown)


services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
```diff
@@ -231,6 +247,7 @@ const RenderEMAPermitCommonSections = ({ code, isAmendment, index, isDisabled })
         documents={tableDocuments}
         documentParent="project summary authorization"
         documentColumns={documentColumns}
+        removeDocument={!isDisabled && isEditMode ? onDeleteDocument : undefined}
```
taraepp: not sure if here we want to use isDisabled or if we want to use docFieldsDisabled (as above in `AuthorizationSupportDocumentUpload`). See ticket 6129

services/common/src/components/projectSummary/DocumentUpload.tsx
```diff
@@ -67,9 +68,10 @@ const RenderOldDocuments = ({
 
 interface DocumentUploadProps {
   docFieldsDisabled: boolean;
+  fieldsDisabled: boolean;
```
taraepp: Yeah.... I'm questioning this one. 
- the purpose of it seems to be controlling _deletion_, so if it's separate from docFieldsDisabled I think it should be named explicitly like "canDelete" or "deleteDisabled" or whatever (keyword: delete)
- tying it to the authorization fields in the parent component and then a different value in the test is kind of confusing. 
- but it's getting late in the day and maybe I'm overcomplicating things

services/minespace-web/src/styles/components/SteppedForm.scss
```diff
@@ -30,6 +30,10 @@
   line-height: 25px;
 }
 
+.ant-dropdown-menu-item:hover, .ant-dropdown-menu-item button:hover{
```
taraepp: nitpick- might be more at home with Tables.scss. There's some other styling of the dropdown menu there.

services/common/src/components/projectSummary/AuthorizationsInvolved.tsx
sggerard: I guess I need to confirm what state's document deletion should be allowed in. docFieldsDisabled is still enabled even if a project is no longer in draft which is why I have been using the fieldsDisabled flag instead. (isDisabled = fieldsDisabled on line 660).

services/common/src/components/projectSummary/DocumentUpload.tsx
sggerard: Happy to change the field name but I was just keeping it the same as all the other tabs

image

sggerard: Although it looks like I did use authFieldsDisabled instead of fieldsDisabled.... oops

asinn134 (dismissed):

simensma-fresh (dismissed):

simensma-fresh (approved):

taraepp (approved):

PR #3345, opened 2024-12-13
[CHORE] Update imports to not import from common package root

simensma-fresh:
## Objective 

[MDS-CHORE](https://bcmines.atlassian.net/browse/MDS-CHORE)

The "@mds/common" namespace is polluted with a lot of exports. Some of these have naming collisions that in some cases might cause unforeseen bugs.

This PR updates a fair few imports to be more specific with the goal of eventually get rid of those warnings, which in turn will unblock further improvements such as the use of rspack.

Note: There's still a lot of imports left to fix, but gotta start somewhere


matbusby-fw (dismissed):
Nice!  Yeah, I've been trying to clean these up when I come accross them, but it's nice to get more of them at once!

taraepp:
Nice. How do you feel about either removing or disabling the `Compare Common Redux folder / common-folder-compare` PR check? I think it's outlived its usefulness.

taraepp (dismissed):

simensma-fresh:
@matbusby-fw @taraepp Updated this to deal with the rest of the imports. Also nuked the exports in `@mds/common/constants`, which fixed all the conflicting export warnings.

Agreed that it would be nice to get rid of the common package check. I updated it to ignore whitespace changes to make it a bit less intrusive, but think we should hold off removing it until all the TSF components are out of there

services/common/src/index.ts
taraepp, line 1: Should this file simply be deleted?

taraepp (approved):

matbusby-fw (approved):

PR #3347, opened 2024-12-13
[FIX] Fix reports resource test flakiness

simensma-fresh:
## Objective 

[MDS-FIX](https://bcmines.atlassian.net/browse/MDS-FIX)

Fixes flakiness in the test_get_reports test. Cause: we do the comparison that filtering worked using received_date, but performed the actual filtering on `due_date`. Also made sure we're always comparing datetimes

Before:
image

After:

image


matbusby-fw (approved):
Woot!

asinn134 (approved):

sggerard (approved):

PR #3348, opened 2024-12-14
[MDS-6285] ESUP amendments detonator magazine missing

asinn134:
## Objective 

[MDS-6285](https://bcmines.atlassian.net/browse/MDS-6285)

- Adjustments made in the esup's generate preview certificate work flow to use the explosive permit amendment magazines if user is trying to preview an amendment. Previously it would only use explosive permit magazine data even when we are dealing with a explosive permit amendment.


services/core-api/app/api/mines/explosives_permit/resources/explosives_permit_document_type.py
```diff
@@ -70,12 +71,13 @@ def post(self, document_type_code):
 
             explosives_permit_guid = data['explosives_permit_guid']
             explosives_permit = ExplosivesPermit.find_by_explosives_permit_guid(explosives_permit_guid)
+            explosives_permit_amendment_id = data.get('explosives_permit_amendment_id', None)
```
simensma-fresh, line 76: I think we might want to perform a check here to make sure the explosives_permit_amendment_id is actually for this explosives_permit. Just a precaution.

simensma-fresh (changes requested):

matbusby-fw (dismissed):
I like Simen's suggestion, but otherwise looks great!

services/core-api/app/api/mines/explosives_permit/resources/explosives_permit_document_type.py
asinn134: A check for this has been added now.

simensma-fresh (approved):

matbusby-fw (approved):
👍🏻

PR #3351, opened 2024-12-17
[MDS-6207] Show PDF of permit amendment on Permit Conditions page + auto-focus on click

simensma-fresh:
## Objective 

[MDS-6207](https://bcmines.atlassian.net/browse/MDS-6207)

With all the data about a permit condition now coming over from the permit service, we now have access to the bounding box of where it came from in the original PDF. This PR adds a side-by-side view of the conditions when you click "Open Permit in Document Viewer", where we also highlight the condition in the original PDF on click. Note: This is behind a feature flag, and just added so we can see how well it works/ if it's worth polishing.

image

### DIFF
--- a/services/common/src/components/syncfusion/DocumentViewer.tsx
+++ b/services/common/src/components/syncfusion/DocumentViewer.tsx
@@ -51,6 +51,13 @@ const getAjaxRequestSettingsHeaders = (obj) => {
   return ajaxRequestSettingsHeaders;
 };
 
+function getAjaxRequestSettings() {
+  return {
+    ajaxHeaders: getAjaxRequestSettingsHeaders(createRequestHeader().headers),
+    withCredentials: false,
+  };
+}
+
 export const OPENABLE_DOCUMENT_TYPES = ["PDF"];
 
 export const isDocumentOpenable = (documentName) =>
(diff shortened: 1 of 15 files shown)


services/core-web/src/tests/components/mine/Permit/PreviewPermitAmendmentDocument.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+jest.mock("@mds/common/redux/utils/actionlessNetworkCalls", () => ({
+    getDocument: jest.fn()
+}));
+
+const mockPermitAmendment = MOCK.PERMITS[0].permit_amendments[0];
+
+const mockSelectedCondition = {
+    ...MOCK.PERMITS[0].permit_amendments[0].conditions[0],
+    permit_condition_guid: "test-condition",
+    meta: {
```
taraepp, line 18: I get a TS error about the type on meta- looks like it's missing 'role'?

services/common/src/redux/slices/permitServiceSlice.ts
```diff
@@ -34,6 +34,7 @@ const permitExtractionStatusMap = {
 export interface PermitExtraction {
   task_status: PermitExtractionStatus;
   task_id: string;
+  permit_amendment_document_guid: string;
```
taraepp, line 37: this bit makes a TS error on line 148 (where it fetches task status). Should the permit_amendment_document_guid be optional? Either that or it wants to be set to a value on that line (which probably doesn't make sense)

taraepp:
Other than some TS issues, looks very cool and works well!

services/common/src/components/syncfusion/DocumentViewer.tsx
```diff
@@ -85,22 +92,46 @@ interface ViewPDFProps {
...
   documentPath,
   ajaxRequestSettings,
+  id = "pdfviewer-container",
+  onInit = null
 }) => {
+  let pdfViewer;
   return (
     <PdfViewerComponent
-      id="pdfviewer-container"
+      id={id || "pdfviewer-container"}
```
matbusby-fw: Could probably remove the "pdfviewer-container" part here since it's set as a default in the props.

services/common/src/interfaces/permits/permitCondition.interface.ts
```diff
@@ -1,5 +1,23 @@
 import { IMineReportPermitRequirement } from "@mds/common/interfaces";
 
+export interface IPermitConditionCategory {
```
matbusby-fw: Oh, I think this already exists lower in this file.

matbusby-fw (changes requested):
Looks awesome!  

Just noticed a duplicated interface and a suggestion on removing an unnecessary `||` block.

simensma-fresh:
@matbusby-fw Fixed 👍

taraepp (approved):

matbusby-fw (approved):

asinn134 (approved):

PR #3352, opened 2024-12-18
[MDS-6304] Tighten up Archive/Replace docs for Major Mine Application

taraepp:
## Objective 
- don't allow replace on spatial table
- lock down actions on the form according to status instead of assuming they can't navigate to it
- removed some unused imports
- made a couple little tweaks to the document table...
   - I thought it would show Download instead of Actions so I added another condition to the logic to make it so
   - in order to put the header in a row with the Actions/Download button, passed it as a param
   - and then added spacing between the button and the table (on MS only. I seem to recall it already has spacing on CORE)

[MDS-6304](https://bcmines.atlassian.net/browse/MDS-6304)

image


sggerard (dismissed):

asinn134 (dismissed):

simensma-fresh (changes requested):
Looks like there's a merge conflict, might have to re-generate that snap. Otherwise looks good!

simensma-fresh (approved):

asinn134 (approved):

sggerard (approved):

PR #3353, opened 2024-12-19
[MDS-6265] Project Summary: Change from Change Requested -> Under Review on final submit only

simensma-fresh:
## Objective 

[MDS-6265](https://bcmines.atlassian.net/browse/MDS-6265)

Updated the Project Summary submission logic to change the status of the application from Change Requested to Under Review only when the declaration is re-submitted, not on updates on the other tabs.

### DIFF
--- a/services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
+++ b/services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
@@ -183,14 +183,20 @@ export const ProjectSummaryPage = () => {
     history.push(url);
   };
 
-  const handleSaveData = async (formValues, newActiveTab?: string) => {
+  const handleSaveData = async (formValues, newActiveTab?: string, currentTab?: string) => {
     let message = newActiveTab
       ? "Successfully updated the project description."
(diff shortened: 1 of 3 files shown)


matbusby-fw (approved):
So weird.  This must have been lost in a rebase somewhere, but I added almost this exact logic already. 🤷🏻

services/minespace-web/src/components/pages/Project/ProjectSummaryPage.tsx
```diff
@@ -183,14 +183,20 @@ export const ProjectSummaryPage = () => {
...
 
-    if (status_code === "CHR" && formValues.confirmation_of_submission) {
+    if (
+      status_code === "CHR" &&
+      formValues.confirmation_of_submission &&
+      currentTab === "declaration"
+    ) {
+      // If a proponent re-submits the declaration, when changes are made to the project description,
+      // change it back to under review.
       status_code = "UNR";
```
simensma-fresh, line 200: Would love for this business logic to live in the API instead so you can't get around it, but that's something for another time 🙃

sggerard (approved):

PR #3354, opened 2024-12-19
[CHORE] Provide types for react-redux hooks

simensma-fresh:
## Objective 

[CHORE](https://bcmines.atlassian.net/browse/MDS-CHORE)

This PR adds typed versions of common react-redux hooks as per https://react-redux.js.org/using-react-redux/usage-with-typescript.

Using those hooks will have the types defined in your action creators propagate or if those are untyped, will be `any`
redux-form actions is supported in addition to our actions

Before:

image
image
image

After:
image
image
image

### DIFF
--- a/services/common/src/redux/rootState.ts
+++ b/services/common/src/redux/rootState.ts
@@ -1,9 +1,10 @@
 import { sharedReducer } from "@mds/common/redux/reducers/rootReducerShared";
-import { configureStore } from "@reduxjs/toolkit";
+import { configureStore, Dispatch } from "@reduxjs/toolkit";
 import { loadingBarReducer } from "react-redux-loading-bar";
 
 import type { TypedUseSelectorHook } from 'react-redux'
 import { useDispatch, useSelector, useStore } from 'react-redux'
+import type { FormAction } from 'redux-form';
 
 export const getStore = (preloadedState = {}) =>
   configureStore({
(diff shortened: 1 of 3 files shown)


asinn134 (dismissed):

simensma-fresh:
If anyone feels like a good cleanup here's a regex (thanks copilot) that (maybe) will replace all our useDispatch hooks to use the new one...

Search:
`import\s*\{\s*([^,}]*)\s*,?\s*\buseDispatch\b\s*,?\s*([^,}]*)\s*\}\s*from\s*['"]react-redux['"];`

Replace:
`import { useAppDispatch as useDispatch } from "@mds/common/redux/rootState";\nimport { $1$2 } from 'react-redux';`

services/common/src/redux/rootState.ts
```diff
@@ -14,6 +18,22 @@ export const getStore = (preloadedState = {}) =>
...
+/**
+ * Provide typed versions for global redux functions
+**/
+
 type RootState = ReturnType<typeof store.getState>;
 
+type AppAction = UnknownAction | FormAction;
+
+// Infer the `RootState` and `AppDispatch` types from the store itself + make sure redux-form acti are captured as well.
+export type AppDispatch = typeof store.dispatch & Dispatch<AppAction>;
```
simensma-fresh, line 30: Sonarcloud doesn't like this `&` operator. Unfortunately needed as `AppAction` is coming from `redux-form` and couldn't find a way of avoiding it without providing our own `redux-form` typing

asinn134 (approved):

taraepp (approved):
🙌

PR #3355, opened 2024-12-19
[MDS-6204] - condition review assignment

matbusby-fw:
## Objective 

[MDS-6204](https://bcmines.atlassian.net/browse/MDS-6204)

- Added component to search for users
- Added functionality to assign a user to a condition category
- Can also remove an assignment (if you are the assignee or you have the correct role)
- Locked editing of pemit conditions unless you are assigned as a reviewer
- Included an update to the partySelect component to show a loading indicator when searching for a party (this was part of the early work before adding the user table, so I kept it)

image


services/core-api/app/api/mines/permits/permit_conditions/models/permit_condition_category.py
```diff
@@ -21,6 +21,13 @@ class PermitConditionCategory(SoftDeleteMixin, AuditMixin, Base):
     active_ind = db.Column(db.Boolean, nullable=False, server_default=FetchedValue())
     display_order = db.Column(db.Integer, nullable=False)
 
+    user_sub = db.Column(db.String, db.ForeignKey('user.sub', ondelete='SET NULL', onupdate="CASCADE"))
+
+    assigned_review_user = db.relationship(
+        'User',
+        back_populates='permit_condition_categories'
```
simensma-fresh: I'm wondering if we should maybe make this a joined, or selectin query to avoid having this as a separate query for each permit condition category (lazy='joined' or lazy='selectin')

simensma-fresh (changes requested):

asinn134 (dismissed):
Nice work!

services/core-api/app/api/mines/permits/permit_conditions/models/permit_condition_category.py
matbusby-fw: Sure, that makes sense!

simensma-fresh (approved):

asinn134 (approved):

PR #3358, opened 2024-12-27
[MDS-6306] Fix minespace user mine management

sggerard:
## Objective 

[MDS-6306](https://bcmines.atlassian.net/browse/MDS-6306)

As a core Administrator, it is hard to edit a minespace user's list of mines due to the mine guid showing instead of the mine name. This is due to only the first 50 mines being returned by the api instead of the complete list which would cause a large slow response.

As an alternative solution, I have added all the mines that user's currently have assigned to the list so that they show correctly. Additionally I have converted files to typescript, fixed a bug where an update would trigger a call multiple times and sorted the list of mines.

Possible breaking changes:
- Removed boolean from IOption as a boolean cannot be used as a key on a <select <option. This could have potential ramifications elsewhere. **update**: Have added IRadioOption as a replacement interface for the multiple implementations of RenderRatioButtons
- Added IMineSearch and replaced IMine in MineReducer for mineNameList as search results appear to return a slightly different response to IMine. This is used in multiple places and could have ramifications elsewhere.


services/core-web/src/components/Forms/EditMinespaceUser.tsx
```diff
@@ -9,18 +8,22 @@ import { nullableStringSorter, resetForm } from "@common/utils/helpers";
 import RenderField from "@/components/common/RenderField";
 import * as FORM from "@/constants/forms";
 import { renderConfig } from "@/components/common/config";
-import CustomPropTypes from "@/customPropTypes";
+import { IMine } from "@mds/common";
```
simensma-fresh: Could we import this from `@mds/common/interfaces` instead? Hoping to get rid of that root global export

services/core-web/src/components/common/RenderMultiSelect.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { FC, useRef } from "react";
+import { Form } from "@ant-design/compatible";
+import "@ant-design/compatible/assets/index.css";
+import { Select } from "antd";
+import { caseInsensitiveLabelFilter } from "@common/utils/helpers";
+import { debounce} from "lodash";
+import { WrappedFieldMetaProps } from "redux-form";
+import { IOption } from "@mds/common";
```
simensma-fresh: Could we import this from `@mds/common/interfaces` instead?

simensma-fresh (changes requested):
Yay for more js->ts conversion!

simensma-fresh (dismissed):

simensma-fresh (approved):

services/core-web/src/components/common/RenderMultiSelect.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { FC, useRef } from "react";
```
taraepp, line 1: Would say just skip updating the core-web MultiSelect. (I deleted it on my branch already). Use common version instead.

asinn134 (approved):

PR #3362, opened 2025-01-07
[MDS-6299] Add "Manage My Mines" dropdown to MinesPage UI

matbusby-fw:
- Added a dropdown menu on the MinesPage allowing users to request access to or removal from mines directly.
- Adjusted styling to be inline with newer designs

## Objective 

[MDS-](https://bcmines.atlassian.net/browse/MDS-6299)

image

### DIFF
--- a/services/minespace-web/src/components/pages/MinesPage.tsx
+++ b/services/minespace-web/src/components/pages/MinesPage.tsx
@@ -109,10 +109,11 @@ export const MinesPage = () => {
                     key: "add",
                     label: (
                       <Button
+                        className="actions-dropdown-button menu-item-button fa-icon-container"
                         type="link"
                         href="https://submit.digital.gov.bc.ca/app/form/submit?f=0cdcf6c4-bbad-429e-b17b-4031d8960ae3"
                         target="_blank"
-                        icon={<FontAwesomeIcon icon={faPlus} className="margin-medium--right" />}
+                        icon={<FontAwesomeIcon icon={faPlus} />}
                       >
                         Request Access To Mine
                       </Button>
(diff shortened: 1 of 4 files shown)


services/minespace-web/src/components/pages/MinesPage.tsx
```diff
@@ -88,17 +92,58 @@ export const MinesPage = () => {
...
+                        }
+                      >
+                        Request Removal From Mine
+                      </Button>
+                    ),
+                  },
+                ],
+              }}
+            >
+              <Button className="mine-page-dropdown">Manage My Mines</Button>
```
taraepp: For consistency across similar elements, would recommend trying:
- make the dropdown button type primary
- mine-page-dropdown --> actions-dropdown-button
- for the item buttons, className="actions-dropdown-button menu-item-button fa-icon-container"
- I _think_ you can use className="color-error" for the same effect as "mines-page-remove-mine-link"

The `faIcon`s might make it so the spacing isn't quite right that way, but should help keep styles in line and then you don't have to fuss with some of the styling details that have already been handled. Not 100% sure that all those styles have been carried over from Core, but _pretty sure_.

matbusby-fw: Yeah, nice suggestions.  I had looked for something that had the error color but somehow missed that one.  Good call on the `actions-dropdown-button menu-item-button fa-icon-container`.  It's a good fit.  Kind of what I hate about css though, that it depends on someone remembering where a certain class being reused would be appropriate. 🤷🏻

asinn134 (approved):

services/minespace-web/src/components/pages/MinesPage.tsx
taraepp: I figured it would want to be the same as the actions menu button but yeah it's a total pain! Like, I do just remember at this point that we have color-x and bg-color-x rules for the main theme colours but it's not intuitive at all and then the scss setup makes it even harder to find because it's like `.color { &-error}`. Ahhh CSS...

taraepp (approved):

PR #3372, opened 2025-01-14
[MDS-6307] DRAFT - Added support for non-redux forms using Antd forms to existing Form components

simensma-fresh:
## Objective 

### DIFF
--- a/docs/forms.md
+++ b/docs/forms.md
@@ -1,0 +1,168 @@
+# Forms Migration Guide
+
+## Overview
+
+This application supports two form implementations.
+
+- Legacy: `FormWrapper` uses redux-form for state management. You'll see this in a lot of forms using things like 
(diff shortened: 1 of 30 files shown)


docs/forms.md
```diff
@@ -0,0 +1,168 @@
...
+     onSubmit={handleSubmit}
+     reduxFormConfig={{touchOnChange: true}}
+   >
+
+   // New
+   const [form] = Form.useForm(); // This form instance is what can be used to replace things like getting access to form values, setting form values programmatically, etc.
+   <AntdFormWrapper
+     form={form} 
+     name="my_form"
+     onSubmit={handleSubmit}
```
matbusby-fw, line 29: Might be nice to add a reference to `antdOptions` here as the alternative to the previous `reduxFormConfig`

PR #3373, opened 2025-01-14
[MDS-6321] Update Ministry Name

taraepp:
## Objective 
It is confusing to have old ministry names in our codebase, but also not ideal to be updating them whenever there is a change in names. So, my plan here was to:
- change EMLI -> MCM (and their expanded versions) in places where users see it, operating with the assumption that in those places, we are displaying the specific ministry name deliberately and should continue doing so
- change EMLI -> Ministry in places where only developers see it- doing it once so it never has to be done again (hopefully)

[MDS-6321](https://bcmines.atlassian.net/browse/MDS-6321)


migrations/sql/V2025.01.14.10.47__update_emli_to_ministry.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE mine_report_notification 
+    DROP CONSTRAINT mine_report_notification_contact_guid_fkey;
+
+ALTER TABLE emli_contact
```
simensma-fresh: I think we should hold off on renaming these tables / columns. They may be in use by the analytics teams (in Metabase or PowerBI), so might cause some issues

simensma-fresh (changes requested):

migrations/sql/V2025.01.14.10.47__update_emli_to_ministry.sql
taraepp: that's a good point. I'll take a look after lunch

simensma-fresh (approved):

matbusby-fw (approved):

asinn134 (approved):

PR #3374, opened 2025-01-14
[MDS-6308] Add permit condition status

sggerard:
## Objective 

[MDS-6308](https://bcmines.atlassian.net/browse/MDS-6308)

Add statuses to permit conditions. 

- Update sidebar to display summary of each categories statuses (Complete, In Progress or Not started)
- Allow assigned user to Complete Review 
- Reset status when any condition (including children) is updated

Please note - when testing you will need to run the database migration and generate new permit conditions as there are additional fields that need to be set.

image


migrations/sql/V2025.01.07.22.23__add_permit_condition_status.sql
```diff
@@ -0,0 +1,48 @@
...
+        ('COM', 'Complete', 30, 'system-mds', 'system-mds')
+    ON CONFLICT DO NOTHING;
+
+ALTER TABLE permit_conditions
+    ADD COLUMN IF NOT EXISTS permit_condition_status_code character varying(3),
+    ADD COLUMN IF NOT EXISTS top_level_parent_permit_condition_id integer,
+ADD CONSTRAINT permit_condition_status_code_fkey FOREIGN KEY (permit_condition_status_code)
+    REFERENCES permit_condition_status_code(permit_condition_status_code),
+ADD CONSTRAINT top_level_parent_permit_condition_id_fkey FOREIGN KEY (top_level_parent_permit_condition_id)
+    REFERENCES permit_conditions(permit_condition_id);
```
taraepp, line 40: would it make sense to set permit condition status code to NOT NULL and default it to Not Started?

migrations/sql/afterMigrate__1__seed_data.sql
```diff
@@ -381,6 +381,20 @@ VALUES
...
+    display_order,
+    create_user,
+    update_user
+    )
+VALUES
+    ('NST', 'Not Started', 10, 'system-mds', 'system-mds'),
+    ('INP', 'In Progress', 20, 'system-mds', 'system-mds'),
+    ('COM', 'Complete', 30, 'system-mds', 'system-mds')
+ON CONFLICT DO NOTHING;
+
```
taraepp: I'm wondering if it's necessary to add to this file!
We've run into issues in the past where it's difficult to maintain things that are put here. If it's not necessary, would avoid it

services/common/src/interfaces/permits/permitCondition.interface.ts
```diff
@@ -26,6 +26,8 @@ export interface IPermitCondition {
   display_order: number;
   mineReportPermitRequirement?: IMineReportPermitRequirement;
   meta?: IPermitConditionMeta
+  permit_condition_status_code: string;
+  top_level_parent_permit_condition_id: number;
```
taraepp: curious since top_level_parent_permit_condition_id isn't optional in the interface- does a top level condition refer to itself here?

services/core-api/app/api/mines/permits/permit_conditions/models/permit_condition_status_code.py
```diff
@@ -0,0 +1,30 @@
...
+    @classmethod
+    def find_by_permit_condition_status_code(cls, _id):
+        return cls.query.filter_by(permit_condition_status_code=_id).first()
+
+    @validates('permit_condition_status_code')
+    def validate_permit_condition_status_code(self, key, permit_condition_status_code):
+        if not permit_condition_status_code:
+            raise AssertionError('Permit amendment status code is not provided.')
+        if len(permit_condition_status_code) > 3:
+            raise AssertionError('Permit amendment status code must not exceed 3 characters.')
```
taraepp: error messages might be confusing with "permit **amendment** status code"

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -127,6 +131,12 @@ const PermitConditions: FC<PermitConditionProps> = ({
     "condition_category_code"
   );
 
+  const PERMIT_CONDITION_STATUS = {
+    complete: {icon: faCheckCircle, color: "#45a766", text: "Complete"},
+    in_progress: {icon: faClock, color: COLOR.violet, text: "In Progress"},
+    not_started: {icon: faBan, color: "#bbb", text:"Not Started"}
```
taraepp: I'd take a look at helpers.scss for colours. The core-web version is missing `&-success` which could be added (this shade of green has a variable of $ok-green it looks like). But I think these could be class names of color-success, color-primary, and color-gov-grey.

docker-compose.M1.yaml
```diff
@@ -122,12 +122,28 @@ services:
...
       - otelcollector
       - keycloak
+      - core_api_celery
     healthcheck:
       test: [ "CMD", "curl", "localhost:5000/health" ]
       interval: 5s
       timeout: 5s
       retries: 5
 
+  core_api_celery:
```
matbusby-fw, line 132: 🙌🏻

migrations/sql/afterMigrate__1__seed_data.sql
matbusby-fw: Counterpoint to that, this is the kind of thing that would get annoying to remind people to enter in themselves.  I'd prefer to have it seeded personally.

matbusby-fw (dismissed):
Nice work!

A bit of a dive into lots of the more complicated parts of the code.  I agree with some of what Tara was saying in her comments.  

- Would probably be good to make that interface property optional
- The one permit amendment message likely won't come up much, but "Permit amendment status code is not provided" specifically could probably be a bit more user friendly.  Something like "Permit condition status code missing from request"?

migrations/sql/V2025.01.07.22.23__add_permit_condition_status.sql
```diff
@@ -0,0 +1,48 @@
...
+        display_order smallint,
+        active_ind boolean DEFAULT true NOT null,
+        create_user character varying(60) NOT NULL,
+        create_timestamp timestamp with time zone DEFAULT now() NOT NULL,
+        update_user character varying(60) NOT NULL,
+        update_timestamp timestamp with time zone DEFAULT now() NOT NULL
+    );
+
+    ALTER TABLE permit_condition_status_code OWNER TO mds;
+    ALTER TABLE ONLY permit_condition_status_code
```
asinn134: Just wondering but how come we need to specify `ONLY` here?

migrations/sql/afterMigrate__1__seed_data.sql
simensma-fresh: Absolutely.
We're just adding these data migrations together with schema migrations instead of using the afterMigrate functionality. So would suggest maybe adding in another.

(example: V2024.06.11.10.03__add_new_document_types_to_project_summary_document_type.sql)

services/common/src/interfaces/permits/permitCondition.interface.ts
sggerard: Have made it optional

services/core-api/app/api/mines/permits/permit_conditions/models/permit_condition_status_code.py
sggerard: Fixed

migrations/sql/V2025.01.07.22.23__add_permit_condition_status.sql
sggerard: Done

migrations/sql/afterMigrate__1__seed_data.sql
sggerard: I had it in both the migration and seed_data. Have removed it from the seed data now.

simensma-fresh (approved):

matbusby-fw (approved):
Woo!

asinn134 (approved):

PR #3383, opened 2025-01-22
[MDS-6192] Add notifications for project status and IRT updates

matbusby-fw:
Added notifications for Project IRTs and Applications tied to status changes and document submission.
Specific designs for these emails do not exist, so modeled them off of the existing project email designs.

## Objective 

[MDS-6192](https://bcmines.atlassian.net/browse/MDS-6192)

image
image
image
image
image
image


services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -80,6 +84,7 @@ def find_by_mine_document_guid(cls, mine_document_guid):
         except ValueError:
             return None
 
+    @classmethod
     def send_irt_submit_email(self):
         project_lead_email = self.project.project_lead.email if self.project.project_lead else None
         recipients = [MAJOR_MINES_OFFICE_EMAIL, project_lead_email
```
taraepp: wait, if this is a classmethod, then shouldn't the first parameter of the function be the class rather than the instance?

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -92,6 +97,51 @@ def send_irt_submit_email(self):
...
+
+        message = f'The status of the IRT for the project {project.project_title} for {project.mine_name} has been updated to {parse_status_code_to_text(status_code)}.'
+        extra_data = {'project': {'project_guid': str(project.project_guid)}}
+        trigger_notification(message, ActivityType.project_irt_status_updated, mine,
+                             'InformationRequirementsTable', self.irt_guid, extra_data)
+
+
+    @classmethod
     def send_irt_approval_email(self):
         recipients = [contact.email for contact in self.project.contacts]
```
taraepp, line 144: yeah I find the addition of the @classmethod decorators here confusing since it does seem to be referencing the self rather than the class.

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -92,6 +97,51 @@ def send_irt_submit_email(self):
         body += f'<p>View IRT in Core: <a href="{link}" target="_blank">{link}</a></p>'
         EmailService.send_email(subject, recipients, body)
 
+    def send_irt_status_notifications(self, project_guid: str, status_code: str):
+        project: Project = Project.find_by_project_guid(project_guid)
```
taraepp: should be able to do one less query with self.project to get the project

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
matbusby-fw: Ah, good catch.  Meant to remove that.

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
matbusby-fw: Tried that, and it wasn't returning with the full project including the project lead.

taraepp (approved):

asinn134 (approved):

PR #3384, opened 2025-01-23
[MDS-6291] Update development postgres instances

sggerard:
## Objective 

[MDS-6291](https://bcmines.atlassian.net/browse/MDS-6291)

Aim: Update development setups to run Postgresql version 16. (17 is too new for crunchyDB)
- Updated postgres dockerfile to postgis 16-3.5 and FDW client
- Added new postgres_update container to both the M1 and standard(untested!) docker-compose files. (Not the ci one at this time)
- Added make updatedb command to automate the process of backing up, updating and re-creating postgres db.
- Had to change how the postgresql.conf and ph_hba.conf were being applied to the container as they are on a volume and initdb only runs if there is no existing postgresql install (Which there is after an upgrade!)
- Fixed a migration, which means the checksum will need to be updated if you want to run a flyway verify. `update public.flyway_schema_history set checksum = '-702230991' where version = '2021.06.03'`

To run locally:
- Pull this branch
- Run `make updatedb`. This will stop the running container, make a volume backup, delete the old image, perform the update and build the new postgres image.

@simensma-fresh Can you please review my changes and be my test dummy before anyone else tries to run this locally

### DIFF
--- a/docker-compose.M1.yaml
+++ b/docker-compose.M1.yaml
@@ -70,6 +70,7 @@ services:
       interval: 5s
       timeout: 5s
       retries: 5
+    stop_grace_period: 30s
 
   postgres_update:
     platform: linux/amd64
--- a/docker-compose.M1.yaml
+++ b/docker-compose.M1.yaml
@@ -64,6 +64,8 @@ services:
     ports:
       - "5432:5432"
     volumes:
+      - ./services/postgres/postgresql.conf:/var/lib/postgresql/data/postgresql.conf
+      - ./services/postgres/pg_hba.conf:/var/lib/postgresql/data/pg_hba.conf
       - postgres_data:/var/lib/postgresql/data
     healthcheck:
       test: [ "CMD", "pg_isready" ]
--- a/docker-compose.M1.yaml
+++ b/docker-compose.M1.yaml
@@ -56,7 +56,7 @@ services:
     user: postgres
     build:
       context: services/postgres
-      dockerfile: Dockerfile_new
+      dockerfile: Dockerfile
     environment:
       - POSTGRES_USER=mds
       - POSTGRES_PASSWORD=test
@@ -346,4 +346,3 @@ volumes:
   core_api_logs: {}
   core_web_logs: {}
   document_manager_logs: {}
-  
--- a/docker-compose.M1.yaml
+++ b/docker-compose.M1.yaml
@@ -83,7 +83,7 @@ services:
     - POSTGRES_PASSWORD=test
     - POSTGRES_INITDB_ARGS=-U mds
     volumes:
-    - mds_postgres_data_bkp:/var/lib/postgresql/13/data
+    - postgres_data_bkp:/var/lib/postgresql/13/data
     - postgres_data:/var/lib/postgresql/16/data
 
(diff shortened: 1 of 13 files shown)


sggerard:
Found an issue, initdb won't run if the .conf files have already been copied across. ugh

services/postgres/pg_hba.conf
```diff
@@ -0,0 +1,15 @@
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     trust
+# IPv4 local connections:
+host    all             all             127.0.0.1/32            trust
+host    all             mds             172.20.0.0/24           password
+host    all             mds             172.17.0.0/24           password
```
simensma-fresh: Could we add in a `host all all all trust`? My docker ip seems to change whenever the instance is restarted. Would be nice not having to deal with an update every time that happens

services/postgres/updateConfig.sh
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 
-cat /tmp/postgresql.conf >/var/lib/postgresql/data/pgdata/postgresql.conf
+cat /tmp/postgresql.conf >/var/lib/postgresql/data/postgresql.conf
+cat /tmp/pg_hba.conf >/var/lib/postgresql/data/pg_hba.conf
```
simensma-fresh: Do we need to copy over the file? Seeing as it's being mounted into the container. If we need to make changes to it in dev/test/prod we can always mount it as a config map

simensma-fresh (changes requested):

docker-compose.M1.yaml
```diff
@@ -64,12 +64,30 @@ services:
     ports:
       - "5432:5432"
     volumes:
+      - ./services/postgres/postgresql.conf:/var/lib/postgresql/data/postgresql.conf
+      - ./services/postgres/pg_hba.conf:/var/lib/postgresql/data/pg_hba.conf
```
simensma-fresh: Think we'll need to mount these in the `docker-compose.yaml` as well

simensma-fresh (changes requested):

services/postgres/pg_hba.conf
sggerard: Done. Newer pg_hba.conf files will include a host all scram-sha-256 line but the passwords from our version were using md5 so didn't match.

services/postgres/updateConfig.sh
sggerard: Have now removed updateConfig.sh

docker-compose.M1.yaml
sggerard: Oops, good catch!

simensma-fresh (approved):
Can confirm that the backend + DB runs well locally using `docker-compose.yml` after running `make updatedb`

simensma-fresh (changes requested):
Oh looks like the integration tests are failing (some flyway error), so will have to get those resolved before we can merge this

simensma-fresh (approved):

matbusby-fw (approved):
This looks great!  And complex!  Really nice work!

PR #3386, opened 2025-01-23
[MDS-6174] Show difference between permit conditions generated in Core and extracted via AI

taraepp:
## Objective 
- I may have done some of the work that's probably meant for a follow-up ticket. 😅 
- Would like to note that the first BadRequest raised in `PermitCOnditionsResource->put` prevents "generated in core" conditions from being review confirmed, _that_ at least I decided was out of scope

[MDS-6174](https://bcmines.atlassian.net/browse/MDS-6174)


matbusby-fw (approved):
🥇

sggerard (approved):

PR #3388, opened 2025-01-23
[CHORE] Re-export redux-form references

taraepp:
## Objective 
- put the redux form exports we use in a file and re-export them so that we can make changes to them in the future
- but make this its own PR so that any further work will actually be readable and not spammed with this stuff
- the main change is adding `form.tsx` and the rest is text change on import.
   - this might be a bad file name in a bad location. IDK. Opinions/suggestions welcome. Will just find and replace again, no biggie. Contemplated calling it formSlice or something but didn't want to have the name signal "we're going to re-write redux-form" 
- I didn't export any redux-form components that we don't use anywhere.

I reverted this before pushing, but this is roughly where this is going (also why it's a .tsx file and not .ts).
image


matbusby-fw (dismissed):
Nice!  First step!

simensma-fresh (dismissed):

matbusby-fw (approved):

simensma-fresh (approved):

asinn134 (approved):

PR #3389, opened 2025-01-27
[MDS-6328] - Select permit document to extract conditions from 🍰 

matbusby-fw:
## Objective 

- Added logic to `ViewPermit` to check if there are multiple documents associated with the permit, and if so, open a modal to select which one to extract from.  
- If there is only a single document, it will imediately start the extraction on that document as before.

[MDS-6328](https://bcmines.atlassian.net/browse/MDS-6328)

image


services/core-web/src/components/mine/Permit/PermitConditionsSelectDocumentModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+        rowKey="document_manager_guid"
+        dataSource={documents}
+        columns={[
+          documentNameColumn("document_name", "File Name"),
+          renderTextColumn("create_user", "Created By"),
+          renderDateColumn("create_timestamp", "Uploaded"),
+          {
+            key: "action",
+            render: (text, record: IPermitAmendmentDocument) => {
+              console.log(record);
```
taraepp: psst. Stray console log.
Also, fun fact: when there's no dataIndex on a column, the first argument of the render function is also the record. You should be able to take out "text" without anything changing

services/core-web/src/components/mine/Permit/ViewPermit.tsx
```diff
@@ -60,13 +62,14 @@ const ViewPermit: FC = () => {
...
+  const isReviewComplete = latestAmendment?.conditions?.every(
+    (con) => con.permit_condition_status_code === PERMIT_CONDITION_STATUS_CODE.COM
+  );
 
   const canStartExtraction =
-    ((documents.length > 0 && !permitExtraction?.status) ||
+    ((documents.length > 0 && !permitExtraction?.task_status) ||
       [PermitExtractionStatus.error, PermitExtractionStatus.not_started].includes(
-        permitExtraction?.status
+        permitExtraction?.task_status
```
taraepp, line 72: Y'know, I _had_ noticed that something was funky with how it displayed when there was no extraction- was that it? We were using status when the actual property name was task_status? Ahh, to get TS working properly with redux...

services/core-web/src/components/mine/Permit/PermitConditionsSelectDocumentModal.tsx
matbusby-fw: Damn you console log!!

Oh right, I forgot about that dataIndex behaviour!

services/core-web/src/components/mine/Permit/ViewPermit.tsx
matbusby-fw: It was!  I only noticed it though because Typescript showed the error and I followed the thread to find out it was right!  Yay Typescript!

simensma-fresh (dismissed):

sggerard:
Looks great!

asinn134 (approved):

sggerard (approved):

PR #3391, opened 2025-01-28
[MDS-6259] Add ability to delete and update permit report requirements

asinn134:
## Objective 

[MDS-6259](https://bcmines.atlassian.net/browse/MDS-6259)

- Moved report requirements to be displayed under their condition
- Added ability to delete and update report requirements by user assigned to review condition

Screenshot (573)

Screenshot (579)


services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
```diff
@@ -168,30 +202,44 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
...
-                cancelFunction={!modalView ? () => setIsEditMode(false) : undefined}
-              />
-              <Button type="primary" htmlType="submit">
-                {mineReportPermitRequirement ? "Update" : "Add"} Report
+        <Row>
+          <Col md={12} sm={24}>
+            {(isEditMode && mineReportPermitRequirement) && (
+              <LinkButton
+                class="report-delete-button"
+                onClick={() => handleDeleteReportRequirement(mineReportPermitRequirement)}
```
taraepp: would recommend deleteConfirmWrapper or similar

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
```diff
@@ -168,30 +202,44 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
...
+              >
+                Delete Report
+              </LinkButton>
+            )
+            }
+          </Col>
+          <Col md={12} sm={24}>
+            {isEditMode ? (
+              <div className="float-right">
```
taraepp: is the float-right to replace the Row justify="end"?

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
```diff
@@ -83,3 +86,108 @@ def post(self, mine_guid):
...
+        if permit_amendment is None:
+            raise NotFound("Permit not found")
+        if permit_amendment:
+            permit_amendment._context_mine = mine
+            if permit_amendment.mine_guid != mine.mine_guid:
+                raise BadRequest(
+                    "The report requirement to be deleted is not part of a permit associated with the given mine"
+                )
+        
+        mine_report_permit_requirement.deleted_ind = True
```
matbusby-fw: Since `mine_report_permit_requirement` model uses the `SoftDeleteMixin` you should be able to actually just call:

```
mine_report_permit_requirement.delete()
```

It won't handle the active_ind, but do we need to update that as well if the record is deleted?

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
```diff
@@ -87,6 +95,22 @@ const PermitConditionLayer: FC<PermitConditionLayerProps> = ({
     await handleMoveCondition(condition, false);
   };
 
+  const getConditionsWithRequirements = (conditions: IPermitCondition[]) => {
```
taraepp: I thought this function looked familiar- it's in the codebase twice already!
Should extract it to some helper file. Not sure if we have one already with permit stuff but I feel like we do.

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
matbusby-fw:
```suggestion
        mine_report_permit_requirement.delete()
```

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
```diff
@@ -83,3 +86,108 @@ def post(self, mine_guid):
...
+        mine_report_permit_requirement.active_ind = False
```
matbusby-fw:
```suggestion
```

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
```diff
@@ -83,3 +86,108 @@ def post(self, mine_guid):
...
+        current_app.logger.info(f'Deleting {mine_report_permit_requirement}')
+
+        mine_report_permit_requirement.save()
```
matbusby-fw:
```suggestion
```

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
taraepp: Okay, I see what's up!
To do the spacing here using grid structure, you want something like:
`<Row justify="space-between {or end if no delete button}">`
`<Col>{deleteButton}</Col>`
`<Col>{other buttons}</Col>`
`</Row>`

services/common/src/constants/strings.tsx
```diff
@@ -474,7 +474,7 @@ export const WASTE_DISCHARGE_AUTHORIZATION_PROCESS =
 export const REPORT_REGULATORY_AUTHORITY_CODES_HASH = {
   CPO: "Chief Permitting Officer",
   CIM: "Chief Inspector of Mines",
-  Both: "Both",
+  BOTH: "Both",
```
taraepp: In the database, records are saved as "Both", so if you're going to change the hash, you'll have to update the records in the database (I probably wouldn't bother)

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
```diff
@@ -168,30 +202,44 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
...
-            <div>
-              <RenderCancelButton
-                cancelFunction={!modalView ? () => setIsEditMode(false) : undefined}
-              />
-              <Button type="primary" htmlType="submit">
-                {mineReportPermitRequirement ? "Update" : "Add"} Report
+        <Row>
+          <Col md={12} sm={24}>
+            {(isEditMode && mineReportPermitRequirement) && (
+              <LinkButton
```
matbusby-fw: Was there any discussion about putting a PopConfirm component around this?  Seems like a good place to add that, but really it's more of a design question for Denise I guess.

services/common/src/constants/strings.tsx
taraepp: Ahhh... I see that compliance_article uses "Both", and permit_condition_report_requirement uses "BOTH". I'd lean towards changing permit_condition_report_requirement as it has no prod records.
I guess this is why we use status tables and so on for things like this.

matbusby-fw (changes requested):
Looks awesome.  I'd just add that popconfirm on the delete button and take a look at using the soft delete mixin on the delete (unless there's a strong need to also update that active_ind that I'm not seeing)

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
asinn134: Thanks for the recommendation this is being used now.

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
asinn134: Nice, the way I did it was overly complicated. This is changed now

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
asinn134: Thanks, that makes sense and is cleaner, this is changed now.

services/core-web/src/components/mine/Permit/PermitConditionLayer.tsx
asinn134: This has been moved to the helper file now

services/core-api/app/api/mines/reports/resources/mine_report_permit_requirement.py
asinn134: Thanks! this area should be adjusted now to use the soft delete mixin.

services/common/src/constants/strings.tsx
asinn134: This has been adjusted, in the database I changed the data type that was being used in the mine_report_permit_requirement table's cim_or_cpo column. Also made some small changes around the `_parse_cim_cpo` function and the `CimOrCpo` class. The value should show as "Both" in the cim_or_cpo column instead of "BOTH" now.

services/core-web/src/components/Forms/reports/ReportPermitRequirementForm.tsx
asinn134: It wasn't mentioned in the mockup but it is a good idea, thanks! I ended up using the `deleteConfirmWrapper` that Tara recommended

matbusby-fw (approved):

taraepp (approved):

PR #3397, opened 2025-01-29
[MDS-6334] Fix condition category bad request

sggerard:
## Objective 

[MDS-6334](https://bcmines.atlassian.net/browse/MDS-6334)

Small PR to fix a bug where newly extracted permit conditions were unable to be updated. This was caused by the validation checking the static data instead of live data for the conditon_category_code, likely a remnant from the original implementation which only included 5 standard categories and not custom/unique categories per extraction.

I have removed the validation from the marshalling and instead added checks on the create and update PermitConditions endpoints. There may be other endpoints that also need this validation but I'm not sure which.

image


asinn134 (approved):

matbusby-fw (approved):
Nice catch!

PR #3398, opened 2025-01-30
[MDS-6255] view mine report definitions

taraepp:
## Objective 
- put mine report definitions on their own table 
- make it its own request
- take it out of core static content

TO ~~DO~~ Done:
- ~~some work with sorting and filtering~~
- ~~odd test failures locally, seeing if happens in pipeline~~ (it was an import thing)
- ~~changed instances where we used the static content reducer to use the new slice, but those instances _aren't fetching data_~~
- ~~write more tests!~~

[MDS-6255](https://bcmines.atlassian.net/browse/MDS-6255)


services/common/src/interfaces/common/searchParams.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface ISearchParams {
```
matbusby-fw, line 1: Feels strange that we don't already have this.  Nice!

services/common/src/redux/reducers/staticContentReducer.js
```diff
@@ -113,8 +112,6 @@ export const getMajorMinesApplicationStatusCodes = (state) =>
   state[STATIC_CONTENT].majorMineApplicationStatusCodes;
 export const getMajorMinesApplicationDocumentTypes = (state) =>
   state[STATIC_CONTENT].majorMineApplicationDocumentTypes;
-export const getMineReportDefinitionOptions = (state) =>
```
matbusby-fw, line 116: Yeah!

matbusby-fw (dismissed):
🤨  I tried to find something to pick apart...but it looks pretty solid. 😁

matbusby-fw:
Looks like there's a possibly legit error in a backend test:

```
>           assert (start_date <= received_date.date())
E           TypeError: can't compare datetime.datetime to datetime.date
```

matbusby-fw (approved):

asinn134 (approved):

asinn134:
Nice work!😃

PR #3400, opened 2025-02-01
[MDS-6347] Fix to add report bug and other css issues

asinn134:
## Objective 

[MDS-6347](https://bcmines.atlassian.net/browse/MDS-6347)

- Disable add report button for a condition when condition already has a report
- Adjusted spacing around conditions and reports section. Also adjusted font weight for some report fields


matbusby-fw (approved):
👍🏻

simensma-fresh (approved):

PR #3410, opened 2025-02-07
[MDS-6346] Fix unable to add custom category

sggerard:
## Objective 

[MDS-6346](https://bcmines.atlassian.net/browse/MDS-6346)

Fixes:
- On small screens/permits withs lots of permit conditions the '+ Add Condition Category' button disappears off the screen. Fixed by moving the extraItems into the Anchor element which has the scroll overflow.
- When opening the Add Condition Category modal the page would scroll to the top losing your position on the page. Fixed by setting scrollOnToggleEdit to false.
- Adding a new Condition Category was not causing a state update/ page re-render. I'm assuming this worked at some point and has broken since moving to custom categories. Added an additonal dispatch on create & delete.

image

Other bugs discovered:
- On a small screen when scrolled all the way down the sidebar will bottom out and go over the footer. This is due to the sidebar being fixed and how Antd handles the overflow. The fix for this is to replace the sidebar--fixed class with a new sidebar--sticky class. However this breaks countless rules including Z-indexes and would require the page layout to be fixed so that the content and sidebar sit nicely next to each other. (I tried and put this in the too hard basket)

### DIFF
--- a/services/common/src/components/common/ScrollSideMenu.tsx
+++ b/services/common/src/components/common/ScrollSideMenu.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect } from "react";
+import React, { FC, ReactNode, useEffect } from "react";
 import { useParams, useHistory, useLocation } from "react-router-dom";
 import { Anchor } from "antd";
 import { ISideMenuOption } from "../../interfaces/common/sideMenuOption.interface";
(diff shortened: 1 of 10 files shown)


asinn134 (approved):
Nice!

taraepp (approved):

matbusby-fw (approved):
🦡

PR #3413, opened 2025-02-10
[MDS-5995] - create mine report definitions

matbusby-fw:
## Objective 

- Added modal to create mine report definitions that can then be linked to a compliance code
- Added api endpoint and update to model to intake new definitions
- Updated mine_report_definition.report_name to be unique
- Added a get endpoint to get mine_report_definition_due_date_types (previously only accessed in the backend)

[MDS-5995](https://bcmines.atlassian.net/browse/MDS-5995)

image


services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -1,13 +1,19 @@
 import { Alert, Button, Col, Row, Typography } from "antd";
 import React, { FC, ReactNode, useEffect, useState } from "react";
-import { useAppDispatch as useDispatch, useAppSelector as useSelector } from "@mds/common/redux/rootState";
-import { arrayPush, change, Field, FieldArray, getFormValues } from "@mds/common/components/forms/form";
+import {
+  useAppDispatch as useDispatch,
+  useAppSelector as useSelector,
+} from "@mds/common/redux/rootState";
+import { arrayPush, change, Field, FieldArray, getFormValues } from "redux-form";
```
taraepp: ooo can we change this import back

services/common/src/components/reports/ReportDetailsForm.tsx
```diff
@@ -236,21 +254,27 @@ const ReportDetailsForm: FC<ReportDetailsFormProps> = ({
...
-      .map((report) => {
-        return {
-          label: formatComplianceCodeReportName(report),
-          value: report.mine_report_definition_guid,
-        };
-      })
-      .sort((a, b) => a.label.localeCompare(b.label));
-    setFormattedMineReportDefinitionOptions(uniqBy(newFormattedMineReportDefinitionOptions, "value"));
+    if (mineReportDefinitionOptions.length) {
+      console.log("mine_report_definition_options", mineReportDefinitionOptions);
```
taraepp: whoops stray console log

services/common/src/redux/slices/complianceReportsSlice.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  },
+}));
+
+describe("complianceReportsSlice", () => {
+  let store;
+
+  // Mock responses
+  const mockComplianceReports = {
+    data: {
+      records: [
```
taraepp, line 38: probably a type on here- just so it gets flagged as an error if some of these properties change in the future.

services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -3,129 +3,197 @@ import { ENVIRONMENT } from "@mds/common/constants/environment";
...
-};
+  is_prr_only?: boolean[];
+  regulatory_authority?: "CPO" | "CIM" | "Both" | "NONE"[];
+  active_ind?: boolean[];
+  section?: string;
+}
 
 export const reportParamsGetAll: ComplianceReportParams = {
-    active_ind: [true, false]
+  per_page: 0,
```
taraepp: I thought that I had made this `per_page` parameter unnecessary.

services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -3,129 +3,197 @@ import { ENVIRONMENT } from "@mds/common/constants/environment";
...
 export const getReportDefinitionsLoaded = (params: ComplianceReportParams) =>
-    createSelector([getReportSearchParams], (currentParams) => {
-        return isEqual(params, currentParams);
-    })
-
-export const {
-    fetchComplianceReports
-} = complianceReportSlice.actions;
+  createSelector([getReportSearchParams], (currentParams) => {
+    return params === currentParams;
```
taraepp: this should be the lodash isEqual. I think some weird stuff happened with a merge resolve with this file, not sure why semicolons and stuff are being removed too.

services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -3,129 +3,197 @@ import { ENVIRONMENT } from "@mds/common/constants/environment";
...
+          return resp.data;
+        } catch (error) {
+          throw error;
+        } finally {
+          thunkApi.dispatch(hideLoading());
+        }
+      },
+      {
+        fulfilled: (state: ComplianceReportState, action) => {
+          notification.success({
```
taraepp: is it just me that prefers to use the successToastMessage of CustomAxios? It might be just me...

services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -3,129 +3,197 @@ import { ENVIRONMENT } from "@mds/common/constants/environment";
...
+        thunkApi.dispatch(showLoading());
+        try {
+          const resp = await CustomAxios().post(
+            `${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITION}`,
+            reportDefinition,
+            headers
+          );
+          return resp.data;
+        } catch (error) {
+          throw error;
```
taraepp: This block is a bit odd. I think you can take out the try-catch (`rejected` automatically catches, basically) and dispatch hideLoading before returning resp.data (does it work to return the function in the try block and then do something else in the finally?)

services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -3,129 +3,197 @@ import { ENVIRONMENT } from "@mds/common/constants/environment";
...
+          rejectHandler(action);
+          // don't show loading forever if there's an error
+          state.params = action.meta.arg;
         },
-        getComplianceReportPageData: (state): IPageData<IMineReportDefinition> => {
-            return state.reportPageData;
+      }
+    ),
+    createMineReportDefinition: create.asyncThunk(
+      async (reportDefinition, thunkApi) => {
```
taraepp: Type would be nice here

services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
```diff
@@ -0,0 +1,117 @@
+import React from "react";
+import FormWrapper from "@mds/common/components/forms/FormWrapper";
+import { ADD_REPORT_DEFINITION } from "@/constants/forms";
+import { Col, Row, Typography } from "antd";
+import { Field } from "redux-form";
```
taraepp: this import too

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
```diff
@@ -1,204 +1,233 @@
...
-    const isLoaded = useSelector(getReportDefinitionsLoaded(queryParams));
-
-    const fetchData = () => {
-        dispatch(fetchComplianceReports(queryParams));
-    };
+  const dispatch = useDispatch();
+  const reportPageData = useSelector(getComplianceReportPageData);
+  const reportDefinitions = reportPageData?.records ?? [];
+  const [sectionSearchText, setSectionSearchText] = useState(null);
+  const [queryParams, setQueryParams] = useState<ComplianceReportParams>(defaultParams);
```
taraepp: Looks like there were a lot of issues with merge conflict resolution in this file. All the stuff with parsing the query from the url (useLocation, useHistory, initialParams)

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
```diff
@@ -1,204 +1,233 @@
...
-            sort_field,
-            ...sortParams,
-            ...activeFilters,
-        }
-        setQueryParams(newParams)
-    };
+  const onTableChange = (pagination, filters, sorter) => {
+    const { current: page, pageSize: per_page } = pagination;
+    const { order, field: sort_field } = sorter;
+    const activeFilters = removeNullValues(filters);
```
taraepp: this too needs to be reverted to include sectionSearchText

services/common/src/components/reports/ReportDetailsForm.tsx
matbusby-fw: Gar!  Missed that on the merge conflicts.  Will do!

services/common/src/redux/slices/complianceReportsSlice.ts
matbusby-fw: Weird.  For the most part with the rebase I just accepted the incoming changes and reapplied my work where necessary.  Something must have gone wrong with it though as I didn't really touch any of this stuff.  I'll go over that again.

services/common/src/redux/slices/complianceReportsSlice.ts
matbusby-fw: I guess it's just more of aesthetic thing for me.  Since we're using a `fulfilled` function here, I always like to see all of the side efffects like that in one place.  It helps to prevent the "Why is this happening?" kind of issue.

That said, it's certainly not a hill I care to die on.  Both work.

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
```diff
@@ -14,191 +20,205 @@ import { removeNullValues } from "@mds/common/constants/utils";
...
+  const initialParams = queryString.parse(search);
+  const reportPageData = useSelector(getComplianceReportPageData);
+  const reportDefinitions = reportPageData?.records ?? [];
+  const [sectionSearchText, setSectionSearchText] = useState(null);
+  const [queryParams, setQueryParams] = useState<ComplianceReportParams>({ ...defaultParams, ...initialParams });
+  const isLoaded = useSelector(getReportDefinitionsLoaded(queryParams));
+
+  const fetchData = () => {
+    dispatch(fetchComplianceReports(queryParams));
+    dispatch(fetchMineReportDueDateTypes({}));
```
taraepp: might actually be nice to separate this call out from fetchData. fetchData is called every time that the table sort/filter/pagination is changed. But I don't think the due date types change with any kind of frequency like that. Could make sense to put it in its own useEffect to only call it once on initial render

taraepp: that or have a loading indicator so that it tests if it's loaded in the state before fetching it

taraepp (approved):

asinn134 (approved):
nice!

PR #3419, opened 2025-02-13
[MDS-6341] Issues with assigning reviewers to conditions

taraepp:
## Objective 
- make the review assignment its own table with some version history
- change FE/BE to reference this

[MDS-6341](https://bcmines.atlassian.net/browse/MDS-6341)


services/common/src/redux/slices/permitConditionCategorySlice.ts
```diff
@@ -118,21 +172,61 @@ const searchConditionCategoriesSlice = createAppSlice({
...
+);
+
+export const getUserReviewAssignmentsByAmendment = (permitAmendmentId: number) => createSelector(
+  [getReviewAssignmentsByAmendment(permitAmendmentId), getUser], (assignments, currentUser) => {
+    if (!currentUser?.sub) { return []; }
+    return assignments.filter((assignment) => assignment.assigned_review_user.sub === currentUser.sub);
+  }
+);
+
+export const isUserAssignedToReviewCategory = (permitAmendmentId: number, condition_category_code: string) => createSelector(
```
taraepp, line 224: I maybe went overboard making selectors. I just think they're neat!

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_amendment_condition_category_user_resource.py
```diff
@@ -1,19 +1,45 @@
...
 from app.api.utils.access_decorators import requires_role_edit_standard_permit_conditions, \
     requires_role_view_all, EDIT_STANDARD_PERMIT_CONDITIONS
-from app.extensions import db, api, jwt
+from app.extensions import api, jwt
 from app.api.utils.include.user_info import User as UserUtils
 
 
 class AssignUserToPermitConditionCategory(Resource):
-    @api.doc(description='Assign a user to a permit condition category')
+    @api.doc(
```
simensma-fresh, line 16: Alternative to adding `api.doc` here so we don't have to define it twice while still having swagger pick it up:

- Move `parser` out of the method and onto the class
- And do `@api.expects(parser)`

simensma-fresh (approved):

services/common/src/redux/slices/permitConditionCategorySlice.ts
simensma-fresh: Can never have enough selectors

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_amendment_condition_category_user_resource.py
taraepp: I don't think I can in this case as the different endpoints take different parameters. 🤔

sggerard (approved):

PR #3422, opened 2025-02-18
[MDS-6369] Report definition bugs (+[MDS-6388])

taraepp:
## Objective 
A minor bug bash.
- sorting by "section" excluded all results without a corresponding compliance article. Inspecting the query showed that it was because the emitted SQL was doing `JOIN compliance_article` (an inner join) so adding `isouter` flag fixed the issue. Didn't test this, but would have also affected any filter or sort that resulted in that join
- cancel button on create report definition form wasn't working. Told it was a modal, now it works.
  - I passed this in the modal wrapper with the idea that a form may or may not be in the modal, but the modal wrapper can always tell it so. 
- mine report definition form wanted a character limit- there are limits on the BE for report name & description, so just added FE validation to match (100 for report name, 300 for description)
- fixed the alignment on the compliance code view/edit that has been bothering me

[MDS-6369](https://bcmines.atlassian.net/browse/MDS-6369)
[MDS-6388](https://bcmines.atlassian.net/browse/MDS-6388)

Here's the resulting alignment on compliance code view/edit. 
View:
image
Edit:
image

### DIFF
--- a/services/core-api/tests/reports/resource/test_mine_reports_resource.py
+++ b/services/core-api/tests/reports/resource/test_mine_reports_resource.py
@@ -66,7 +66,7 @@ def test_get_code_required_reports_for_mine(test_client, db_session, auth_header
     for report in get_data['records']:
         received_date = datetime.strptime(report['received_date'], '%Y-%m-%d')
 
-        assert (start_date <= received_date.date())
+        assert (start_date.date() <= received_date.date())
         assert (received_date.date() <= end_date)
 
 
--- a/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
+++ b/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
@@ -10,7 +10,7 @@ import { useAppSelector } from "@mds/common/redux/rootState";
 import { getMineReportDueDateTypes } from "@mds/common/redux/slices/complianceReportsSlice";
 import RenderCancelButton from "@mds/common/components/forms/RenderCancelButton";
 import RenderSubmitButton from "@mds/common/components/forms/RenderSubmitButton";
-import { required, requiredRadioButton } from "@mds/common/redux/utils/Validate";
+import { maxLength, required, requiredRadioButton } from "@mds/common/redux/utils/Validate";
 import { Field } from "@mds/common/components/forms/form";
 
 const { Title, Paragraph } = Typography;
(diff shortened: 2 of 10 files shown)


services/core-api/tests/reports/resource/test_mine_reports_resource.py
```diff
@@ -66,7 +66,7 @@ def test_get_code_required_reports_for_mine(test_client, db_session, auth_header
     for report in get_data['records']:
         received_date = datetime.strptime(report['received_date'], '%Y-%m-%d')
 
-        assert (start_date <= received_date.date())
+        assert (start_date.date() <= received_date.date())
```
taraepp, line 69: this has been failing intermittently lately... Pretty strange that it's only sometimes just trying this out to see if it fixes it

services/core-web/src/components/admin/complianceCodes/ComplianceCodeViewEditForm.tsx
```diff
@@ -91,42 +91,39 @@ const ComplianceCodeViewEditForm: FC<{
...
-            {steps.map((step) => (
-              <Steps.Step key={step.title} title={step.title} />
-            ))}
-          </Steps>
+        <Steps current={currentStep}>
+          {steps.map((step) => (
+            <Steps.Step key={step.title} title={step.title} />
+          ))}
+        </Steps>
+        {steps[currentStep].content}
```
taraepp, line 99: this looks like a big change but it's really just indentation from taking out a `<Row>` or two

sggerard (approved):

services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
```diff
@@ -25,26 +25,23 @@ const AddReportDefinitionForm = ({ handleSubmit }) => {
...
-      </Paragraph>
+    <FormWrapper name={ADD_REPORT_DEFINITION} onSubmit={handleSubmit} isModal={isModal}>
+      <Title level={3}>Report Details</Title>
       <Field
         required
         component={RenderField}
         name="report_name"
         label="Report Name"
-        validate={[required]}
+        validate={[required, maxLength(100)]}
```
matbusby-fw, line 35: Ah, good catch!

matbusby-fw (approved):

PR #3423, opened 2025-02-19
[MDS-6374] Change addressing duplicate report on extraction

asinn134:
## Objective 

[MDS-6374](https://bcmines.atlassian.net/browse/MDS-6374)

- Change to check report name to address duplicate report on permit extraction
- Change to address blank page showing when clicking a report
- Change to address report name sometimes not showing up
- I noticed when doing a permit amalgamation using the same document as the original permit extraction sometimes the previous report doesn't show in the comparison modal. Included a change for this.

image

### DIFF
--- a/services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
@@ -51,6 +51,13 @@ class MineReportPermitRequirement(SoftDeleteMixin, Base, AuditMixin):
             return cls.query.filter_by(mine_report_permit_requirement_id=_id, deleted_ind=False).first()
         except ValueError:
             return None
+        
+    @classmethod
+    def find_by_permit_condition_id(cls, _id) -> "MineReportPermitRequirement":
+        try:
+            return cls.query.filter_by(permit_condition_id=_id, deleted_ind=False).first()
+        except ValueError:
+            return None
 
     @classmethod
     def find_by_report_name(cls, _report_name) -> "MineReportPermitRequirement":
--- a/services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
@@ -46,23 +46,23 @@ class MineReportPermitRequirement(SoftDeleteMixin, Base, AuditMixin):
         return '<MineReportPermitRequirement %r>' % self.mine_report_permit_requirement_id
 
     @classmethod
-    def find_by_mine_report_permit_requirement_id(cls, _id) -> "MineReportPermitRequirement":
+    def find_by_mine_report_permit_requirement_id(cls, id) -> "MineReportPermitRequirement":
         try:
-            return cls.query.filter_by(mine_report_permit_requirement_id=_id, deleted_ind=False).first()
+            return cls.query.filter_by(mine_report_permit_requirement_id=id, deleted_ind=False).first()
         except ValueError:
             return None
         
     @classmethod
-    def find_by_permit_condition_id(cls, _id) -> "MineReportPermitRequirement":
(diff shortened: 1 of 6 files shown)


services/core-web/src/components/mine/Permit/ComparePermitConditionHistoryModal.tsx
```diff
@@ -31,7 +31,7 @@ const ComparePermitConditionHistoryModal: FC<ComparePermitConditionHistoryModalP
         getMineReportPermitRequirementsByAmendment(props.permitGuid, props.previousAmendment?.permit_amendment_guid)
     );
 
-    const oldReports = getConditionsWithRequirements([props.currentAmendmentCondition], previousMineReportPermitRequirements);
+    const oldReports = getConditionsWithRequirements([props.previousAmendmentCondition], previousMineReportPermitRequirements);
```
asinn134, line 34: This is mainly the change for the previous report not showing in the comparison modal when amalgamating with same document

services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
```diff
@@ -51,6 +51,13 @@ def find_by_mine_report_permit_requirement_id(cls, _id) -> "MineReportPermitRequ
             return cls.query.filter_by(mine_report_permit_requirement_id=_id, deleted_ind=False).first()
         except ValueError:
             return None
+        
+    @classmethod
+    def find_by_permit_condition_id(cls, _id) -> "MineReportPermitRequirement":
+        try:
+            return cls.query.filter_by(permit_condition_id=_id, deleted_ind=False).first()
```
taraepp: minor convention thing- what's with the underscore in `_id`?

services/core-web/src/components/mine/Permit/PermitConditionReportRequirements.tsx
```diff
@@ -23,31 +23,41 @@ const PermitConditionReportRequirements: FC<PermitConditionReportRequirementsPro
...
-                        permitGuid={permitGuid}
-                        mineReportPermitRequirement={cond.mineReportPermitRequirement}
-                        canEditPermitConditions={canEditPermitConditions}
-                        refreshData={refreshData}
-                        currentAmendment={currentAmendment}
-                        mineGuid={mineGuid}
-                    />
-                </Collapse.Panel>
-            ))}
+            {conditionsWithRequirements.map((cond, index) => {
```
taraepp: is the condition no longer `IPermitCondition`? Wondering why typing was removed. Does it make sense to add optional property `mineReportPermitRequirement` to the interface?

services/core-web/src/components/mine/Permit/ComparePermitConditionHistoryModal.tsx
taraepp: Good catch! Looking at it, this seems obvious. 😂

services/core-api/app/api/mines/reports/models/mine_report_permit_requirement.py
asinn134: I think I just saw other functions doing that in this file and followed along. The underscore has been removed now.

services/core-web/src/components/mine/Permit/PermitConditionReportRequirements.tsx
asinn134: I ran into an issue here with the `getConditionsWithRequirements` function, in it's if condition where `requirements && condition?.permit_condition_id` the code there would return an object of type `IMineReportPermitRequirement`. So sometimes the function would return a type of `IMineReportPermitRequirement[]` and sometimes a type of `IPermitCondition[]`. 

I think the way I tried to handled it initially in my PR isn't the best, I adjusted it now so `getConditionsWithRequirements` function will always return `IPermitCondition[]`

simensma-fresh (approved):

taraepp (approved):

PR #3428, opened 2025-02-20
[FIX] Cancel edit conditions

taraepp:
## Objective 
- sometimes it doesn't cancel the edit but you're not in edit mode so then you can't cancel it and you have to refresh to go back to being able to edit
- solution: just cancel the edit whenever we refresh the data to get rid of weird edge cases


taraepp:
oh bah I accidentally have the changes from the other branch I made in here too

taraepp:
okay I fixed it to just be the intended change

asinn134 (approved):

sggerard (approved):

PR #3430, opened 2025-02-20
[FIX] Don't scroll on open View Changes modal

taraepp:
## Objective 
the scroll is annoying here


simensma-fresh (approved):

matbusby-fw (approved):
Approved, but  I think we should look at switching the default behaviour of the scrollOnToggleEdit to false.  Especially since we've now attached it to so many pre-existing forms.  It should be treated as an optional enhancement rather than a forced change.

PR #3434, opened 2025-02-25
[MDS-6395] Remove Skeleton from Permit Conditions

taraepp:
## Objective 
- for a smoother user experience
- there's a lot of subforms and stuff, so added loading/setLoading to the context provider instead of props passing
- snuck in some useDispatch --> useAppDispatch
- Just a note that on the icon buttons with no text- using "loading" doesn't fit in the button, so it renders a spinner _below_ the button (ew) so I used disabled for those buttons, although I generally prefer loading > disabled

[MDS-6395](https://bcmines.atlassian.net/browse/MDS-6395)
conditions-loading


sggerard:
Changes work great!

sggerard (approved):

simensma-fresh (approved):

PR #3439, opened 2025-02-27
[MDS-6394] - Improve performance of forms in the permit condition editor

matbusby-fw:
- Memoized several components used in the permit condition editor
- Updated permit condition form to touchOnBlur but not on Change

I still wasn't able to fully prevent the rerendering of each form on input, but it should be significantly reduced.

## Objective 

[MDS-6394](https://bcmines.atlassian.net/browse/MDS-6394)

### DIFF
--- a/services/common/src/components/forms/FormWrapper.tsx
+++ b/services/common/src/components/forms/FormWrapper.tsx
@@ -2,7 +2,13 @@ import React, { FC, useEffect } from "react";
 import { Form } from "antd";
 import { compose } from "@reduxjs/toolkit";
 import { connect, useDispatch, useSelector } from "react-redux";
-import { reduxForm, submit, getFormSubmitErrors, InjectedFormProps, ConfigProps } from "@mds/common/components/forms/form";
+import {
+  reduxForm,
+  submit,
+  getFormSubmitErrors,
+  InjectedFormProps,
+  ConfigProps,
+} from "@mds/common/components/forms/form";
 
 export interface IFormContext {
   isEditMode: boolean;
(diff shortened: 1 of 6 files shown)


services/common/src/components/forms/FormWrapper.tsx
```diff
@@ -85,11 +91,15 @@ const FormWrapper: FC<FormWrapperProps & InjectedFormProps<any>> = ({
...
-    onReset
+    onReset,
   };
-  console.log('open form', props.name)
+  console.log("open form", props.name);
   const dispatch = useDispatch();
-  const formErrors = useSelector(getFormSubmitErrors(props.name));
+
+  const formErrors = useSelector(
+    (state) => getFormSubmitErrors(props.name)(state),
```
taraepp: does this still work if it's just getFormSubmitErrors(props.name) with the state param being implicitly passed in? Just thinking in case we want to update the selector at some point.

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -103,8 +103,15 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
 
-  const reviewAssignments = useAppSelector(getReviewAssignmentsByAmendment(currentAmendment.permit_amendment_id));
-  const userReviewAssignments = useAppSelector(getUserReviewAssignmentsByAmendment(currentAmendment.permit_amendment_id));
+  const reviewAssignments = useAppSelector(
+    getReviewAssignmentsByAmendment(currentAmendment.permit_amendment_id)
+  );
+  const userReviewAssignments = useAppSelector(
+    useMemo(
+      () => getUserReviewAssignmentsByAmendment(currentAmendment.permit_amendment_id),
+      [currentAmendment.permit_amendment_id]
```
taraepp, line 112: What happens when you assign yourself to review the conditions? Do you have to refresh the page?

services/common/src/components/forms/FormWrapper.tsx
matbusby-fw: Ah, this actually didn't yield any real benefit.  I'm going to revert it.

services/core-web/src/components/mine/Permit/PermitConditions.tsx
matbusby-fw: Nope.  It still updates as expected, and you can immediately edit if you assign yourself.

sggerard (approved):

taraepp (approved):

PR #3440, opened 2025-02-27
[MDS-6397] Disable "Edit permit conditions" except for NoW permits

taraepp:
## Objective 

[MDS-6397](https://bcmines.atlassian.net/browse/MDS-6397)


taraepp:
So, this testing is kinda lousy (counting the number of dispatches instead of the resulting state, for example)
And the way I set up the state is also kind of lousy (I'm not sure how exactly I'd like to see it set up)

But the intention here is basically:
- get ready to convert permitSelectors/reducer/actionCreator into a slice
  - then revamp testing
  - maybe re-jigger the state setup too (consumers of selectors should not have to be aware of the state structure!)
- and also get things a little closer to making more granular network requests, like just for the conditions, or just for the amendment, in other situations

matbusby-fw (approved):

asinn134 (approved):

PR #3444, opened 2025-03-04
[MDS-6386] Update postgres github actions

sggerard:
## Objective 

[MDS-](https://bcmines.atlassian.net/browse/MDS-6386)

Before continuing the work on update postgres we need to ensure that dev, test and productions are all using different image tags so that we can do a gradual roll out. This is the first step, tagging the images separately. Next I need to update the deployment configs in the tenant gitops repo to point to these tags.

Currently there is no deployment step in these actions, the images are created and pushed to openshift but the stateful set is not redeployed. Thoughts?

### DIFF
--- a/.github/workflows/postgres.build.dev.yaml
+++ b/.github/workflows/postgres.build.dev.yaml
@@ -1,4 +1,4 @@
-name: Postgres Image Build
+name: Postgres Image Build DEV
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


simensma-fresh (dismissed):

simensma-fresh:
Yeah I think this should be fine for now seeing how infrequently we apply the postgres upgrades. We will have to revisit this regardless if we end up moving to CrunchyDB

taraepp (approved):

taraepp (changes requested):
I know I just approved it but now I gotta ask that ubuntu-20.04 --> ubuntu-24.04 because first one is deprecated

taraepp (approved):

matbusby-fw (approved):
Nice!

PR #3448, opened 2025-03-05
[MDS-6412] Fix add party relationship form

taraepp:
## Objective 
- making sure that AddPartyRelationship works from mine contacts, and with various types of contacts
- and also that it works on TSFs
- and not break minespace

[MDS-6412](https://bcmines.atlassian.net/browse/MDS-6412)


services/common/src/components/forms/RenderLargeSelect.tsx
```diff
@@ -22,7 +22,7 @@ const RenderLargeSelect: FC<LargeSelectProps> = ({
   dataSource = [],
   ...props
 }) => {
-  const selectedInData = dataSource.includes((opt) => props.selectedOption?.value === opt.value);
+  const selectedInData = dataSource.find((opt) => props.selectedOption?.value === opt.value);
```
taraepp, line 25: I kept seeing that I had like 5 copies of the `props.selectedOption` in the list, turns out that `dataSource.includes` _always returns false_- it doesn't want a predicate.
I had a hard time figuring out what exactly was the issue so random stuff like this got fixed just because it looked sus.

services/common/src/constants/enums.ts
```diff
@@ -91,9 +91,29 @@ export enum OperatingStatusEnum {
...
+  MOR = "MOR",
+  MOW = "MOW",
+  MRC = "MRC",
+  PAY = "PAY",
+  PMT = "PMT",
+  SHB = "SHB",
+  SVR = "SVR",
+  THD = "THD",
+  TQP = "TQP",
+  URP = "URP",
```
taraepp, line 116: turns out we were missing like, all of these. Who knew!

simensma-fresh (approved):

services/common/src/components/forms/RenderLargeSelect.tsx
matbusby-fw: Oh yeah.  That makes sense.  I've never seen `includes` used in that way.  I think it only accepts single instance of the data source passed in to work properly.

services/core-web/src/components/common/PartySelectField.tsx
```diff
@@ -189,13 +197,10 @@ export const PartySelectField: FC<PartySelectFieldProps> = ({
     }
   };
 
-  const handleSelect = (value, option) => {
+  const handleSelect = (_value: string, option: IPartySelectOption) => {
```
asinn134, line 200: Was just wondering what does the underscore in front of `value ` do, is it just naming convention?

matbusby-fw (approved):

asinn134 (approved):
Looks great! Just had a question a small question in the comments, not a bug or anything.

services/core-web/src/components/common/PartySelectField.tsx
taraepp: The underscore indicates that it's an unused variable. :)
In python I think it often indicates that it's private? I forget exactly but that's a different rabbit-hole.
Either you can prefix with underscore, or or replace with underscore. It makes the things-that-yell-at-me-about-my-code shut up about unused variables. I didn't want to just take it out entirely because it's passed as a parameter elsewhere.

asinn134: Ohh I didn't know that, thanks!

PR #3449, opened 2025-03-05
[MDS-6366]fee exemption error

asinn134:
## Objective 

[MDS-6366](https://bcmines.atlassian.net/browse/MDS-6366)

- Fix for error “Exemption fee should be 'Yes' for this permit” potentially blocking creation and update of Permit
- Converted EditPermitForm file to typescript

### DIFF
--- a/services/core-web/src/components/Forms/AddPermitForm.tsx
+++ b/services/core-web/src/components/Forms/AddPermitForm.tsx
@@ -131,7 +131,7 @@ export const AddPermitForm: FC<AddPermitFormProps> = (props) => {
       dispatch(change(formName, "exemption_fee_status_code", statusCode));
     }
 
-  }, [permit_status_code, permit_type, mine_tenure_type_code]);
+  }, [permit_status_code, permit_type, mine_tenure_type_code, mine_disturbance_code]);
 
   const isCoalOrMineral =
     mine_tenure_type_code === "COL" ||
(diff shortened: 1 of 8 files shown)


services/core-web/src/components/Forms/AddPermitForm.tsx
```diff
@@ -131,7 +131,7 @@ export const AddPermitForm: FC<AddPermitFormProps> = (props) => {
       dispatch(change(formName, "exemption_fee_status_code", statusCode));
     }
 
-  }, [permit_status_code, permit_type, mine_tenure_type_code]);
+  }, [permit_status_code, permit_type, mine_tenure_type_code, mine_disturbance_code]);
```
asinn134, line 134: Found that mine disturbance can affect the exemption status so without mine_distubance_code as a dependent here,  changes to the field won't update us with the correct exemption status.

services/core-web/src/components/Forms/EditPermitForm.tsx
```diff
@@ -15,40 +13,43 @@ import RenderSelect from "@mds/common/components/forms/RenderSelect";
...
-      props.initialValues.permit_prefix,
-      props.initialValues.site_properties?.mine_tenure_type_code,
+      initialValues.permit_prefix,
+      initialValues.site_properties?.mine_tenure_type_code,
       isExploration,
-      props.initialValues.site_properties?.mine_disturbance_code
+      initialValues.site_properties?.mine_disturbance_code
     );
-    dispatch(change("exemption_fee_status_code", feeStatus));
+    dispatch(change(FORM.EDIT_PERMIT, "exemption_fee_status_code", feeStatus));
```
asinn134, line 46: The reason we were getting an error when updating the permit is because the dispatch was missing a form name so our exemption status didn't update correctly. Every other change in this file is just typescript changes.

simensma-fresh (approved):

taraepp (approved):
🎨

PR #3451, opened 2025-03-06
[MDS-6397] Fix permit conditions not saving on NoW editor

taraepp:
## Objective 
- whoops, no props.location. Gave an error in the submit
- make location be defined

[MDS-6397](https://bcmines.atlassian.net/browse/MDS-6397)


simensma-fresh (changes requested):
See that the `AddCondition` test is failing, otherwise 👍

taraepp:
The test stubbornly resisted being turned into a `render` test. That is, I would have had to do a refactor to get it to return anything but `<div />` due to some things being set as defaults with `useState` blocking proper renders. Too much of a pain for a little bug fix.

asinn134 (approved):

matbusby-fw (approved):

simensma-fresh (approved):

PR #3453, opened 2025-03-07
[FIX] Core typescript errors

sggerard:
## Objective 

Fix almost all the typescript errors in CORE.

There are only 2 remaining, MineAlert and MinePermitInfo which both need to be rewritten as Functional Components.
Using the connector props won't fix these as they need to be using the typed Dispatch.
image

**Key changes:**
- Added interfaces (Including to Search!)
- Prettier may have had a go at some indentation argh
- Swapped useDispatch and useSelector for useAppDispatch and useAppSelector where needed
- Cast formValues to Interfaces
- Did some ugly connector props joining 
- Updated the redux-form-input-masks library to fix bad typing.


services/common/src/interfaces/common/option.interface.ts
```diff
@@ -13,3 +13,9 @@ export interface IGroupedDropdownList {
   groupName: string | number;
   opt: IOption[];
 }
+export interface DropdownOption {
+  value: string;
+  label: string;
+  isActive?: boolean;
```
taraepp: I wonder if we just want to add these properties to IOption? Because value/label is required in both and the rest is optional

services/core-web/src/components/Forms/incidents/IncidentForm.tsx
```diff
@@ -56,20 +56,20 @@ export const IncidentForm: FC<IncidentFormProps> = (props) => {
...
+  const inspectorOptions = useAppSelector(getDropdownInspectors) || [];
   const selector = formValueSelector(FORM.ADD_EDIT_INCIDENT);
-  const documents = useSelector((state) => selector(state, "documents")) || [];
-  const formValues = useSelector((state) => getFormValues(FORM.ADD_EDIT_INCIDENT)(state)) || {};
-  const dropdownIncidentStatusCodeOptions = useSelector(getDropdownIncidentStatusCodeOptions);
-  const isPristine = useSelector((state) => state.form[FORM.ADD_EDIT_INCIDENT]?.pristine);
+  const documents = useAppSelector((state) => selector(state, "documents")) || [];
+  const formValues = useAppSelector((state) => getFormValues(FORM.ADD_EDIT_INCIDENT)(state)) || {};
+  const dropdownIncidentStatusCodeOptions = useAppSelector(getDropdownIncidentStatusCodeOptions);
+  const isPristine = useAppSelector((state) => state.form[FORM.ADD_EDIT_INCIDENT]?.pristine);
```
taraepp: perhaps a little out of scope, but there is a redux-form selector `isPristine(formName)` that doesn't depend on knowing the state structure. You would have to rename the `const isPristine` though, to use it. Actually, seeing how we have `formValueSelector` on ln 68, used in 69, and then followed by `getFormValues`, this file probably needs a little more love in general.

services/common/src/interfaces/common/option.interface.ts
taraepp: Yeeeeah it looks like where it's used is being passed into RenderSelect/RenderMultiSelect, which expects IOption for it. May as well make it the same

sggerard: Cool, done

services/core-web/src/components/Forms/incidents/IncidentForm.tsx
sggerard: I have refactored it a little :)

services/core-web/src/components/Forms/incidents/IncidentForm.tsx
```diff
@@ -164,7 +163,7 @@ export const IncidentForm: FC<IncidentFormProps> = (props) => {
           />
           <br />
           <IncidentFormInternalDocumentComments
-            documents={documents}
+            documents={formValues?.documents}
```
taraepp: I think we might want to null coalesce these to []- looks like both `IncidentFormDocuments` and `IncidentFormInternalDocumentComments` does a `documents.filter` which will probably blow up with undefined

taraepp (approved):

matbusby-fw (approved):

PR #3454, opened 2025-03-07
[MDS-6411] TSF navigation weirdness

taraepp:
## Objective 
- noticed some strange stuff navigating around TSFs 

[MDS-6411](https://bcmines.atlassian.net/browse/MDS-6411)

### DIFF
--- a/services/common/src/components/forms/SteppedForm.tsx
+++ b/services/common/src/components/forms/SteppedForm.tsx
@@ -110,8 +110,9 @@ const SteppedForm: FC<SteppedFormProps> = ({
       if (handleSaveData && (await saveCheck())) {
         await handleSaveData(getValues(formValues), tab, `${tabs[tabIndex]}`);
       }
-      if (errors.length > 0) return;
-      setTabIndex(indexOf(tabs, tab));
+      if (errors.length === 0 && tab) {
+        setTabIndex(indexOf(tabs, tab));
+      }
     } finally {
       setIsSubmitting(false);
     }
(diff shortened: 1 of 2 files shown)


services/common/src/components/forms/SteppedForm.tsx
```diff
@@ -110,8 +110,9 @@ const SteppedForm: FC<SteppedFormProps> = ({
       if (handleSaveData && (await saveCheck())) {
         await handleSaveData(getValues(formValues), tab, `${tabs[tabIndex]}`);
       }
-      if (errors.length > 0) return;
-      setTabIndex(indexOf(tabs, tab));
+      if (errors.length === 0 && tab) {
+        setTabIndex(indexOf(tabs, tab));
+      }
```
taraepp, line 115: making sure that the "setIsSubmitting" in the finally gets called instead of returning from the function. 
There was something going on where I think the form was reset, I would navigate to the next page, and then it would _never_ let me go to the next page.

services/common/src/components/tailings/TailingsSummaryPage.tsx
```diff
@@ -176,17 +183,23 @@ export const TailingsSummaryPage: FC = () => {
...
+                GLOBAL_ROUTES?.EDIT_TAILINGS_STORAGE_FACILITY.dynamicRoute(
+                    newTsf?.data.mine_tailings_storage_facility_guid || tsfGuid,
+                    mineGuid,
+                    newActiveTab,
+                    isUserActionEdit
+                )
+            );
+        } else {
+            submissionComplete();
+        }
```
taraepp, line 197: form navigated to EoR page at the end...
I figure we'll want to do something more complicated at some point when we're done with it, so I made it its own tiny function for now.

matbusby-fw (approved):
Looks good.  I can fiddle with the submit function a bit more in my PR

simensma-fresh (approved):

PR #3458, opened 2025-03-12
[MDS-6360] Create EOR/QP report requests on TSF submit

matbusby-fw:
- Added logic to create report requests for QP and EOR when a new one is assigned and a TSF is submitted
- Added find logic to search for a report by it's sections (this is needed to find the right reports to request here)
- Added a post-submit page after submitting a TSF that directs users to reports or back to TSFs
- Removed document logic from the EOR page as this is now handled in the reports.
- Updated UI on EOR and QP pages to be inline with new UI

## Objective 

[MDS-6360](https://bcmines.atlassian.net/browse/MDS-6360)

image
image

### DIFF
--- a/services/common/src/components/forms/RenderDate.tsx
+++ b/services/common/src/components/forms/RenderDate.tsx
@@ -14,6 +14,7 @@ interface DateInputProps extends BaseInputProps {
   yearMode?: boolean;
   disabledDate?: (currentDate) => boolean;
   formatViewDate?: boolean;
+  fieldEditMode?: boolean;
 }
 
 const RenderDate: FC<DateInputProps> = ({
@@ -29,11 +30,12 @@ const RenderDate: FC<DateInputProps> = ({
   yearMode = false,
   disabledDate,
   formatViewDate = false,
+  fieldEditMode,
 }) => {
   return (
     <FormConsumer>
       {(value) => {
-        if (!value.isEditMode) {
+        if (!value.isEditMode || (typeof fieldEditMode !== "undefined" && !fieldEditMode)) {
           return (
             <BaseViewInput
               label={label}
--- a/services/common/src/components/forms/RenderDate.tsx
+++ b/services/common/src/components/forms/RenderDate.tsx
@@ -14,7 +14,6 @@ interface DateInputProps extends BaseInputProps {
   yearMode?: boolean;
   disabledDate?: (currentDate) => boolean;
   formatViewDate?: boolean;
-  fieldEditMode?: boolean;
 }
 
 const RenderDate: FC<DateInputProps> = ({
--- a/services/common/src/components/tailings/QualifiedPerson.tsx
+++ b/services/common/src/components/tailings/QualifiedPerson.tsx
@@ -1,242 +1,212 @@
 import { Alert, Button, Col, Empty, Popconfirm, Row, Typography } from "antd";
 import { change, Field, getFormValues } from "@mds/common/components/forms/form";
-import React, { FC, useContext, useEffect, useState } from "react";
+import React, { FC, useContext } from "react";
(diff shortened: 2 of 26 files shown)


services/common/src/components/forms/RenderDate.tsx
```diff
@@ -14,6 +14,7 @@ interface DateInputProps extends BaseInputProps {
   yearMode?: boolean;
   disabledDate?: (currentDate) => boolean;
   formatViewDate?: boolean;
+  fieldEditMode?: boolean;
```
matbusby-fw: This is to allow specific fields within a form to individually set their editMode (ie. Nested in a step form)

services/core-api/app/api/mines/tailings/resources/tailings.py
```diff
@@ -113,26 +128,52 @@ def put(self, mine_guid, mine_tailings_storage_facility_guid):
...
 
             if is_minespace_user():
                 # EORs created through minespace should have a status of "pending"
                 new_status = MinePartyAppointmentStatus.pending
                 mine_party_acknowledgement_status = MinePartyAcknowledgedStatus.not_acknowledged
+
+                if is_feature_enabled(Feature.TSF_V2):
+                    if is_submitting:
+                        calculated_due_date = datetime.now() + timedelta(hours=72)
+                        mine_report_definition = MineReportDefinition.find_one_by_section('10', '4', '1', '2')
```
matbusby-fw, line 153: Was trying to find the best way to find these specific reports that wouldn't be dependent on the environment.  I don't love hard coding the section values here, but it seemed like the most stable element of a report definition (compliance_article) to search by.

Open to ideas for a better way to handle it though.

services/common/src/components/forms/RenderDate.tsx
taraepp: I think this might make sense to move into `BaseInputProps`

services/common/src/components/tailings/QualifiedPerson.tsx
```diff
@@ -1,242 +1,212 @@
...
+              component={RenderDate}
+            />
+          </Col>
         </Row>
-    );
+        <Row>
+          <Col span={24}>
+            <PartyAppointmentTable
+              canEditTSF={canEditTSFAndEditMode}
+              eor_or_qp="qualified_persons"
```
taraepp, line 203: I'd be tempted to use the MinePartyAppointmentTypeCodeEnum here! Just because it's a constant we're using elsewhere. Seems harmless.

matbusby-fw: Unfortunately `qualified_persons` and `engineers_of_record` don't fit with that (note the plural), as they're the field name of the grouping of those items for a tsf.

services/core-api/app/api/mines/tailings/resources/tailings.py
taraepp: Oof. Yeah ngl this is gross. But then I looked at prod mine_report_definitions and there is literally "Engineer of Record (confirmation)" and "Engineer of Record (departure)" so the other option is to hardcode the report name and that is just as gross. I have no suggestions.

sggerard (dismissed):
Looks good! 🦫

services/common/src/components/forms/RenderDate.tsx
matbusby-fw: Good call!

services/core-api/app/api/mines/tailings/resources/tailings.py
matbusby-fw: Making a ticket to improve this and make it more reliable/maintainable.

sggerard (approved):

taraepp (approved):

asinn134 (approved):

PR #3459, opened 2025-03-12
[MDS-6385] NoD decision document upload conditions

asinn134:
## Objective 

[MDS-6385](https://bcmines.atlassian.net/browse/MDS-6385)

- In NoD adjusted the requirement to upload a decision document to be only for statuses of "determined not substantial" and "determined substantial"

- Uploading decision document not required when there is an existing one.


services/core-web/src/components/modalContent/NoticeOfDepartureModal.tsx
```diff
@@ -209,8 +220,16 @@ const NoticeOfDepartureModal: React.FC<NoticeOfDepartureModalProps> = (props) =>
         };
       })
     );
+
+    setCurrentStatus(props?.initialValues?.nod_status);
```
taraepp: I'm wondering if you can simplify things here with something like `currentStatus = formValues?.nod_status` since it always seems to be equal to that. Setting the initialValues on the form should also set the formValues.

asinn134: You're right it does work with that change, its much cleaner.

taraepp (approved):

sggerard (approved):

PR #3461, opened 2025-03-13
[MDS-6349] Permit Condition Search POC

simensma-fresh:
## Objective 

[MDS-6349](https://bcmines.atlassian.net/browse/MDS-6349)

## Search Improvements
Made some improvements to the search interface itself. Now you're able to view the PDF a condition was extracted from, view conditions of the permit in a modal, or navigate to the new condition editor. 

#### Highlighting
Enabled highlighting of search results (e.g. bold on keyword matches), this was done using functionality provided by Azure AI Search, and rendering the search results as markdown.

#### Search relevance
Plugged in Azure AI Search score as a "relevance" indicator on each result so you can more easily see what actually was a good match to your search.

#### Search More Fields
Enabled proper hybrid search (keyword + semantic). This opens up for queries like "give me a couple of conditions for Elkview mine" where the information of the mine isn't necessarily present in the condition itself.

- Hybrid search is be performed on the conditions content
- Keyword search is performed on any of its metadata (such as mine name / permit no)
- Semantic reranker is enabled to combine those matches

#### Recency boost
Enabled a little boost of score based on the issue date of the amendment the condition is part of, hoping to make more recent amendments bubble up in search results before previous ones.

#### Expand content
Modified prompt + search pipeline to include context about the conditions around the condition you searched for. The LLM is provided a limited number of parent/child/sibling conditions to have as much info as it can to provide a relevant answer

#### Filter sidebar
Moved the filters into a sidebar. They became a bit unwieldily when on the page even with 10 permits, hoping a sidebar will make that experience a bit better. 

#### AI-Generated Response
A couple of changes were made to this. Basic styling for common markdown elements were added to improve the look of LLM responses. Switched the search endpoint / pipeline to use Server-Sent Events (SSE) to stream back results to the frontend. This allows us to display the hybrid search results as soon as they're available and allow the LLM response to continue generating. 

#### Haystack upgrade
Upgraded haystack to the latest version which broke a bunch of things I had to fix, the library doesn't seem to be very concerned about backward compatibility 😓 

## Indexing improvements
Switched out the more manual process of indexing permit conditions / generating embeddings to using Azure Search Indexers. Basically the indexing process for indexing conditions for search is

1. Export permit conditions as a csv (added a utility script for this)
2. Upload csv to a Azure Storage Blob (added an endpoint to do this)
3. Trigger indexing (part of the endpoint above)

The indexing will use an integrated embedder skillset to generate vector embeddings and map the csv rows to an Azure search index. This also opened up the ability to use integrated embeddings when actually performing a vector search, so we don't have to manually generate embeddings.

### Bulk Export / import scripts
Added some utility scripts for exporting PDFs / mine/permit/amendment metadata from Prod and importing it to local/test. Otherwise actually testing things out proved to be a pain.

Also added some scripts to trigger permit condition extraction in bulk using a celery task

### Permit Condition Validation / Extraction step
I added in another step to the permit condition indexing pipeline that tries to validate the structure of conditions we identified, fix any issues it can using an llm, or retry extracting the structure if significant issues were found. This is behind a future flag and the goal is to better support documents that don't follow the nice structure amalgamated permits have.

image

image

image
image
image


services/common/src/components/common/ActionMenu.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+    return {
+        ...originalModule,
+        Modal: {
+            ...originalModule.Modal,
+            confirm: jest.fn(),
+        },
+    };
+});
+
+describe('ActionMenu', () => {
```
taraepp, line 18: oh nice test addition! Yeah... this probably should have already been in there... I'll have to double-check the output but I bet this takes out a lot of the uncovered lines in coverage reports.

services/common/src/redux/slices/permitSearchSlice.ts
```diff
@@ -30,60 +37,149 @@ const permitSearchSlice = createAppSlice({
...
+
+                let eventSource = null;
+
                 try {
-                    const response = await CustomAxios().post(
-                        `${ENVIRONMENT.apiUrl}/search/permit-conditions`,
-                        payload,
-                        headers
-                    );
+                    eventSource = createEventSource({
```
taraepp, line 118: I'm curious about this createEventSource. For streaming results?

services/common/src/redux/slices/permitSearchSlice.ts
```diff
@@ -30,60 +37,149 @@ const permitSearchSlice = createAppSlice({
...
+                            'Accept': 'text/event-stream',
+                            'Content-Type': 'application/json'
+                        },
 
+                        onMessage: ({ event, data }) => {
+
+                            const payload = JSON.parse(data);
+                            let updatedState: RootState = thunkApi.getState();
+
+                            console.error('got event', event);
```
taraepp: prooooobably don't want the console.error to stay?

services/common/src/redux/slices/permitSearchSlice.ts
```diff
@@ -30,60 +37,149 @@ const permitSearchSlice = createAppSlice({
...
+            state.documentLoading = action.payload;
+        }),
+        /**
+         * Search for permit condtions and parse the results from a Server-Sent Events stream (SSE).
+         * 
+         * Handles the following events sent by the server:
+         *  - documents: Matching Permit Conditions
+         *  - ai_start: Sent when the AI is processing the query
+         *  - prompt: AI Generated result
+         *  - ai_complete: Sent when the AI has finished processing the query
```
taraepp, line 80: these event values look like they might want to be an enum

services/common/src/redux/slices/permitSearchSlice.ts
simensma-fresh: Exactly. It'll listen for `Server-Sent Events` that the backend streams back to the client. Can see them being generated in the `permit_condition_search_resource.py`. Basically just in use so we can see search results as they're ready before the full AI generated response finishes (which can take 5+ seconds)

services/core-web/src/components/mine/Permit/PermitConditions.tsx
```diff
@@ -442,8 +443,168 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
+            </CoreButton>
+          </Col>
+        </Row>
+      </Col>
+      <Col span={24}>
+        <div className="core-page-content">
+          <Row gutter={[16, 16]}>
+            {formattedPermitConditionCategories.map((category, idx) => {
+              return (
+                <React.Fragment key={category.href}>
```
matbusby-fw, line 519: Since the fragment is the child of the row, is the row even recognizing the Col?  It looks like it wouldn't.

matbusby-fw: I guess that's not new code, but 🤷🏼

services/core-web/src/index.js
```diff
@@ -88,26 +88,28 @@ export const Index = (props) => {
...
+      }}
+      onTokenExpired={() => {
+        if (!isIdle()) {
+          keycloak.updateToken();
+        }
+      }}
+      LoadingComponent={<Loading />}
+      isLoadingCheck={(kc) => !kc || !environment}
+    >
+      <FeatureFlagProvider>
```
matbusby-fw, line 105: Interesting, was this the source of some of our errors?

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment_document.py
```diff
@@ -39,6 +38,10 @@ class PermitAmendmentDocument(SoftDeleteMixin, AuditMixin, Base):
     @classmethod
     def find_by_permit_amendment_document_guid(cls, _guid):
         return cls.query.filter_by(permit_amendment_document_guid=_guid, deleted_ind=False).first()
+    
+    @classmethod
+    def find_by_permit_amendment_document_name(cls, document_name):
```
taraepp: I'm a little wary of searching by the document name. Just did a quick check on metabase for duplicates:
image

matbusby-fw (dismissed):
This is looking really good.  There's quite a lot here that's difficult to cognitively connect, but the comments you've provided help mitigate that, and I think it's more a result of the large scale of this and the amount of wholly new content.  Be good to have a dev demo at some point to walk through the code.

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment_document.py
taraepp: If it's gotta be found by the document name:
- a lot of these (duplicates) are missing a permit amendment id
- and I think a lot of them are replacements

So could be made more stable by adding a check for permit_amendment_id is not null (since it looks for the document.permit_amendment later where function is used- or pass in a reference to the amendment as a function parameter) & order to find the most recent match.

services/core-api/app/cli_commands/prepare_data.py
```diff
@@ -0,0 +1,285 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+class ImportMinesAndPermits:
+    # Utility script to import mines, permits, amendments, and associated documents from a CSV file.
+    # It will take the documents specified and upload them to S3 using the document manager.
+    # The purpose for this script is to test out permit condition search in a real-ish environment with actual data.
+    #
+    # WARNING: DO NOT RUN THIS IN PRODUCTION
```
taraepp, line 28: Juuust to be extra-safe, could we do a `if Config.ENVIRONMENT_NAME === 'production' : throw new Error(NO! Not allowed!)`

asinn134:
> This is looking really good. There's quite a lot here that's difficult to cognitively connect, but the comments you've provided help mitigate that, and I think it's more a result of the large scale of this and the amount of wholly new content. Be good to have a dev demo at some point to walk through the code.

Would love to have a dev demo sometime to walk through the code!

services/core-web/src/components/mine/Permit/PermitConditions.tsx
simensma-fresh: It does indeed. It's just a way of grouping multiple children without adding extra dom elements (there's multiple `Col`s in that statement)

image

matbusby-fw (approved):

sggerard (approved):

asinn134 (approved):
Nice work Simen!

PR #3463, opened 2025-03-14
[MDS-5758] Dam History

taraepp:
## Objective 
- record and show history of the dam
- the damReducer was pretty lightweight so I turned it into a slice
- made DiffTable re-usable
- I wanted to be able to use functions and hashes for displaying data, so added that in. 
- made HistoryMixin on the BE

[MDS-5758](https://bcmines.atlassian.net/browse/MDS-5758)

The modal:
image
All of the fields:
image


services/core-api/app/api/dams/models/dam.py
```diff
@@ -15,14 +15,14 @@ class ConsequenceClassification(Enum):
     NRT = 'NRT'
 
     def __str__(self):
-        return self.value
+        return self.value[0]
```
taraepp, line 18: this was interesting! I kept getting errors in the bit that converted these values to strings. As it turned out, `self.value` actually returned a Tuple. So it looked like `[ValueIWant,]`.

services/core-api/app/api/dams/dto.py
```diff
@@ -1,23 +1,22 @@
...
     'mine_tailings_storage_facility_guid': fields.String,
-    'dam_type': fields.String(enum=DamType, attribute='dam_type.name'),
+    'dam_type': fields.String,
     'dam_name': fields.String,
     'latitude': fields.Float,
     'longitude': fields.Float,
-    'operating_status': fields.String(enum=OperatingStatus, attribute='operating_status.name'),
-    'consequence_classification': fields.String(enum=ConsequenceClassification, attribute='consequence_classification.name'),
+    'operating_status': fields.String,
+    'consequence_classification': fields.String,
```
taraepp, line 12: I don't know why, but these enums worked perfectly fine when using `DAM_MODEL` but exploded the swagger when it was cloned into `DAM_HISTORY_MODEL`. Turned them into `fields.String` on the base model on the off-chance that another enum attribute is added at some point or someone goes to use it as a reference.

services/common/src/components/history/DiffColumn.interface.tsx
```diff
@@ -6,12 +6,13 @@ export interface IDiffColumn {
...
 export interface IDiffEntry {
   updated_by: string;
-  updated_at: Date;
+  updated_at: string;
   changeset: IDiffColumn[];
 }
 
 /**
  * Used to map the diff titles and values to a more user-friendly format
+ * Supply *one of* hash, tranform, data
```
matbusby-fw, line 15: Autobots, TRANFORM! 😆

services/common/src/redux/slices/damSlice.ts
```diff
@@ -0,0 +1,123 @@
...
+    dams: ItemMap<IDam>;
+}
+
+const initialState: DamState = {
+    dams: {}
+};
+
+const createRequestHeader = REQUEST_HEADER.createRequestHeader;
+
+const damSlice = createAppSlice({
```
matbusby-fw, line 25: Nice!  Loving the slice format!

services/core-api/app/api/dams/dto.py
matbusby-fw: Interesting.  Should work fine though in this model context.

services/core-api/app/api/utils/models_mixins.py
```diff
@@ -478,6 +479,30 @@ class AuditMixin(object):
     update_timestamp = db.Column(
         db.DateTime, nullable=False, default=datetime.utcnow, onupdate=datetime.utcnow)
 
+class HistoryMixin(object):
```
matbusby-fw, line 482: Oh nice, good idea!

matbusby-fw (approved):
Looks good!

sggerard (approved):
Nice Dam work! 🦫

PR #3465, opened 2025-03-17
[MDS-6406] Filter out mine report definitions with expired compliance codes

matbusby-fw:
- Added a filter in the mine_report_definitions list to remove definitions with an expired compliance code from the list
- Added a param to `show_expired` codes and set it to true on the report management page in Core

## Objective 

[MDS-6406](https://bcmines.atlassian.net/browse/MDS-6406)

### DIFF
--- a/services/core-api/app/api/mines/reports/models/mine_report_definition.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report_definition.py
@@ -157,7 +157,7 @@ class MineReportDefinition(Base, AuditMixin):
         return query
 
     @classmethod
-    def _apply_filters(cls, query, regulatory_authority, is_prr_only, active_ind, section):
+    def _apply_filters(cls, query, regulatory_authority, is_prr_only, active_ind, section, show_expired):
         filters = []
         if regulatory_authority:
             reg_auth_filter = []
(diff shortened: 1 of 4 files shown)


taraepp (approved):

services/core-api/app/api/mines/reports/models/mine_report_definition.py
```diff
@@ -196,6 +196,19 @@ def _apply_filters(cls, query, regulatory_authority, is_prr_only, active_ind, se
                 field_name = section_order[index]
                 filters.append(field_name.ilike(part))
 
+        if not show_expired:
+            now = datetime.now(timezone('US/Pacific'))
```
simensma-fresh, line 200: Will this tz cause some issues? Seeing as the dates would be stored utc and the server would be running in utc

sggerard (approved):

services/core-api/app/api/mines/reports/models/mine_report_definition.py
matbusby-fw: Ah, good point.  I'll check quickly buy I suspect you're right.

simensma-fresh (approved):

PR #3472, opened 2025-03-21
[MDS-5041] Don't allow EOR or QTF to be an organization

sggerard:
## Objective 

[MDS-5041](https://bcmines.atlassian.net/browse/MDS-5041)

Engineers of record and Qualified persons are not allowed to be organizations.

In turns out this change was a lot simpler than I initially thought, I was attempting to add party filtering on the backend but filtering already existed on the frontend.

In minespace the AddContactForm appears to have no support for creating organizations and has a really bad time if you select an organization from the existing contacts so just filtered out organizations entirely.


asinn134 (approved):

matbusby-fw (approved):
Nice!

PR #3474, opened 2025-03-24
[MDS-6329] Moved permit condition components to the common package

matbusby-fw:
## Objective 

- Moved all new view permit/permit conditions components into the Common service to prepare them to be used in Minespace
- Components are largely unchanged other than the imports
- Had to make some small updates to tests to mock global routes and allow them to work.
- All functionality/views remain unchanged at this point
- Deleted `ViewPermitReports.tsx` as it wasn't actually being used anymore
- Added the Permit.svg to the common service


services/common/src/tests/permits/PermitConditions.spec.tsx
```diff
@@ -190,72 +221,74 @@ describe("PermitConditions", () => {
...
-      expect(addReport).toBeInTheDocument();
-      const addListItem = screen.queryByText("List Item");
-      expect(addListItem).toBeInTheDocument();
-      const conditionInput = container.querySelector('[name="condition"]');
-      expect(conditionInput).toBeInTheDocument();
-    });
-
   });
+
+  // it("enables adding a condition category in a modal", async () => {
```
sggerard: Is this the test causing you issues or not currently needed?

services/common/src/components/forms/RenderField.tsx
```diff
@@ -44,6 +45,7 @@ const RenderField: FC<BaseInputProps> = ({
           >
             <>
               <Input
+                data-testid={input.name}
```
matbusby-fw, line 48: I added this while trying to fix that odd test.  Decided to leave it, as it will allow any RenderField instances to be found in tests by using `getByTestId('{input.name}')`.

simensma-fresh (approved):

sggerard (approved):

PR #3479, opened 2025-03-26
[MDS-6462] Add bcgovpubcode.yml

sggerard:
## Objective 

[MDS-6462](https://bcmines.atlassian.net/browse/MDS-6462)

bcgovpubcode.yml is a metadata standard for repositories containing software developed or acquired by the Public Administration, aimed at making them easily discoverable and thus reusable by other entities.

https://pubcode.apps.silver.devops.gov.bc.ca/

### DIFF
--- a/bcgovpubcode.yml
+++ b/bcgovpubcode.yml
@@ -1,0 +1,47 @@
+version: 1
+data_management_roles:
+  data_custodian: Colin Squirrell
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
Not familiar with the required structure for these yamls, but all of the values look right to me!

simensma-fresh (approved):

PR #3487, opened 2025-04-03
[MDS-6329] - Minespace View Permits/Permit Conditions

matbusby-fw:
## Objective 

- Copied various styles from Core to Minespace
- Updated styles where necessary
- Updated banner text for Minespace users
- Updated logic to restrict Minespace users to view only
- Renamed CorePageHeader to CommonPageHeader

image


services/common/src/components/permits/PermitReviewBanner.tsx
```diff
@@ -1,55 +1,76 @@
...
+        justify="center"
+        align="middle"
+        gutter={16}
+        className={`permit-status-banner permit-status-banner--${paramsToUse.className}`}
+        style={{
+          height,
+        }}
+      >
+        <Col>{<FontAwesomeIcon icon={paramsToUse.icon} />}</Col>
+        <Col span={23}>{paramsToUse.text}</Col>
```
taraepp, line 67: Just checking- visually looks the same/better in Core as before?

services/minespace-web/src/components/dashboard/mine/permits/ViewPermitRedirect.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  if (permit) {
+    console.log(
+      "should redirect",
+      mineGuid,
+      permitGuid,
+      permit.permit_amendments[0].permit_amendment_guid,
+      tab
+    );
+    console.log(
```
taraepp: Great console log testing, I am 100% that the redirect is working. 😂 
But probably should be taken out.

services/minespace-web/src/constants/routes.ts
```diff
@@ -301,3 +305,26 @@ export const REPORT_VIEW_EDIT = {
   component: ReportPage,
   helpKey: "Report",
 };
+
+export const VIEW_MINE_PERMIT_AMENDMENT = {
+  route: "/mines/:id/permits/:permitGuid/permit-amendment/:permitAmendmentGuid/:tab",
```
taraepp, line 310: I know it's kind of annoying- but can we use mineGuid instead of id in these new routes? `id` is just so vague that in the context of the page you have to think an extra second about which object it refers to.

services/common/src/components/permits/PermitReviewBanner.tsx
matbusby-fw: Looks the same for the most part.  Could end up off by a few pixels, but without it the long text was taking up the whole width and leaving the icon on it's own row.

services/minespace-web/src/constants/routes.ts
matbusby: Yeah, I hear you.  I generally prefer mineGuid as well, but left it as id here to match the usage of the identical route in Core.  But, I guess there's no reason they have to match.

services/minespace-web/src/components/dashboard/mine/permits/ViewPermitRedirect.tsx
matbusby-fw: Ha, yes.  It is working!  I had a conflict in the route paths that were failing to hit the redirect path at one point, so had to troubleshoot. :grin:

sggerard (dismissed):

taraepp (dismissed):

sggerard (approved):

taraepp (approved):

PR #3488, opened 2025-04-03
[MDS-6430] TSF reports page blank

sggerard:
[MDS-6430](https://bcmines.atlassian.net/browse/MDS-6430)

Mine reports were filtered by mine report definitions with a category of "TSF", and the issue was that mine report definitions were not being loaded and therefore all the reports were getting filtered out.

Initially I fixed this with:
`  useEffect(() => {
    if (!reportDefinitionsLoaded) {
      dispatch(fetchComplianceReports(reportParamsGetAll));
    }
  }, []);
`
However, since we are now only fetching mine reports with a type of Strings.MINE_REPORTS_TYPE.tailingsReports I believe all this filtering logic is unnecessary,.

I don't fully understand the existing logic so please correct me this assumption is incorrect and I will revert back to my initial fix.

### DIFF
--- a/services/core-web/src/components/mine/Tailings/MineTailingsInfoTabs.tsx
+++ b/services/core-web/src/components/mine/Tailings/MineTailingsInfoTabs.tsx
@@ -17,7 +17,7 @@ import {
 import { getMineReports, getReportsPageData } from "@mds/common/redux/selectors/reportSelectors";
 import { getMineById } from "@mds/common/redux/selectors/mineSelectors";
 import { closeModal, openModal } from "@mds/common/redux/actions/modalActions";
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
Yeah, I can't remember exactly why it was set up this way, but I think you're right that we don't need that extra filter.

taraepp:
Yeah looks good to me.
Looks like we were only using the `mineReportDefinitionOptions` to filter the mine reports down to the TSF type, which was totally unnecessary because the request for mine reports already did that. I bet it was some redundant code that was left in at some point. 🤷‍♀️

taraepp (approved):

PR #3490, opened 2025-04-04
[MDS-6340] Show environment authorization statuses in minespace and core

asinn134:
## Objective 

[MDS-6340](https://bcmines.atlassian.net/browse/MDS-6340)

- Adjusted create new auth and amend auth endpoints to be aligned with AMS service's new error response structures
- Adjusted it so when user enters Project Description Tab environment authorization status data will be retrieved from AMS
- Adjusted Project Description Tab's environment authorization table to display latest status data and added modal to let user download documents associated with the specific authorization

**View of table with adjusted columns**
image

**No documents:**
image

**Has document:**
image


services/core-api/app/api/projects/project_summary/models/project_summary.py
```diff
@@ -1285,11 +1291,17 @@ def update(self,
                                                                          authorization.get(
                                                                              'project_summary_authorization_guid'))
                     if ams_tracking_details:
-                        # if result does not have a statusCode attribute, it means the outcome is successful.
-                        authorization['ams_status_code'] = ams_tracking_details.get('statusCode', '200')
+                        ams_error_details = ams_tracking_details.get('detail', None)
```
taraepp: wondering if some of this duplication be extracted to a function

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -211,64 +262,76 @@ const ProjectDescriptionTab = () => {
...
-      AMS_WATER_SUSTAINABILITY_ACT_TYPES,
-      setWaterSustainabilityActData
-    );
-    loadOtherActPermitData(
-      project.project_summary.authorizations,
-      AMS_FORESTRY_ACT_TYPE,
-      setForestryActData
-    );
-    loadEnvironmentActPermitData(project.project_summary.authorizations);
+    if (tab === 'project-description') {
```
taraepp: is the tab === 'project-description' bit used to determine if the ProjectDescriptionTab is the one being displayed and only load data when we're on that tab? I wonder if something like the `destroyInactiveTabPane` on `<Tabs />` might accomplish the same thing a bit more simply.

asinn134: Great recommendation! `destroyInactiveTabPane` works as intended. The tab 'project-description' prop and check have been removed now.

services/core-api/app/api/projects/project_summary/models/project_summary.py
asinn134: That makes sense, this portion has been extracted now to another function.

services/common/src/components/project/ProjectDescriptionTab.tsx
```diff
@@ -264,7 +310,21 @@ const ProjectDescriptionTab = () => {
...
     );
-    loadEnvironmentActPermitData(project.project_summary.authorizations);
+    const amsAuthorizations = project?.project_summary.authorizations?.filter(
+      auth => auth.ams_tracking_number && auth.ams_tracking_number !== "0"
+    );
+
+    const amsTrackingNumbers = amsAuthorizations?.map(auth => auth.ams_tracking_number);
+    if (amsTrackingNumbers.length > 0) {
+      setIsLoaded(false);
+      // @ts-ignore
```
taraepp: wondering if the ts-ignore would be unnecessary if the dispatch was from useAppDispatch

services/common/src/components/project/ProjectDescriptionTab.spec.tsx
```diff
@@ -44,7 +44,7 @@ describe("ProjectDescriptionTab", () => {
     const { container } = render(
       <ReduxWrapper initialState={initialState}>
         <BrowserRouter>
-          <ProjectDescriptionTab />
+          <ProjectDescriptionTab tab={'project-description'} />
```
taraepp: just a little artefact left in here I think

services/common/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -105,17 +106,18 @@ export const renderActionsColumn = ({
...
     key: "actions",
     fixed: fixed ? ("right" as const) : null,
     className: "actions-column",
     render: (record) => {
       const filteredActions = recordActionsFilter ? recordActionsFilter(record, actions) : actions;
       const items = generateActionMenuItems(filteredActions, record);
-
       return (
-        <div>
+        <div title={title}>
```
matbusby-fw: I tend to prefer using the antD `<Tooltip>` for this kind of thing as it maintains more consistent styling, but we do use this method in various places as well, so it's not a huge deal.

taraepp: The div title populates the label in the table when it's viewed on a smaller screen! I like the addition. 🤷‍♀️  
image

services/common/src/components/project/ProjectDescriptionTab.tsx
asinn134: It does make it unnecessary! This has been changed to use useAppDispatch instead now.

services/common/src/components/project/ProjectDescriptionTab.spec.tsx
asinn134: Thanks, this has been removed now.

services/common/src/components/common/CoreTableCommonColumns.tsx
asinn134: I was testing this out just now and I ended up sticking with `<Tooltip>` usage. I noticed if I use div for this some tables will not display the actions menu in smaller screen so just going to stick with consistent styling for the time being.

taraepp: That makes sense!
I suppose that for "Actions" we don't really need "Actions: [Actions button]" in small screens

services/common/src/components/common/CoreTableCommonColumns.tsx
```diff
@@ -105,17 +106,18 @@ export const renderActionsColumn = ({
...
     key: "actions",
     fixed: fixed ? ("right" as const) : null,
     className: "actions-column",
     render: (record) => {
       const filteredActions = recordActionsFilter ? recordActionsFilter(record, actions) : actions;
       const items = generateActionMenuItems(filteredActions, record);
-
       return (
-        <div>
+        <Tooltip title={title}>
```
taraepp: wait- what if there's no title passed in? Probably want to default it to Actions?

taraepp: Although- the snaps look like it isn't adding a tooltip with something like `title=undefined` - can you confirm?

asinn134: I think the past design around actions column doesn't have a title so I didn't include a default value for Actions because of this. When there isn't a title passed it ends up looking like this:

image

edit: It doesn't include the title text or a tooltip when `title=undefined`

asinn134: I'm actually now thinking I don't need to pass any title text to the Tooltip component. It seems a bit redundant to have a column titled "Documents" and then when the user hovers over the action menu a tooltip displaying "Documents" shows up:

image

I do still think keeping Tooltip over div works better for now just because with div the action menu doesn't disappear in smaller screens.

taraepp (dismissed):

taraepp (approved):

matbusby-fw (approved):

PR #3491, opened 2025-04-05
[MDS-6429] Fix mine compliance filters

sggerard:
## Objective 

[MDS-6429](https://bcmines.atlassian.net/browse/MDS-6429)

All of the filters on the inspection orders didn't handle nulls, I'm not entirely sure how they ever worked. Easy fix (slap a lodash isEmpty() on the front). Also decided to update the class to Typescript. I've tested locally by intercepting the browser api calls and everything appears to work now :)

image


services/core-web/src/components/mine/Compliance/MineComplianceInfo.tsx
```diff
@@ -0,0 +1,187 @@
+import React, { FC, useEffect, useState } from "react";
+import queryString from "query-string";
+import { useSelector } from "react-redux";
+import { Divider } from "antd";
+import { isEmpty } from "lodash";
+import { formatParamStringToArray, formatDate, getFiscalYear } from "@common/utils/helpers";
```
taraepp: minor thing, but we have a version of this file in @mds/common- can we use that import? (it's a file that would be amazing to migrate to common usage)

services/core-web/src/components/mine/Compliance/MineComplianceInfo.tsx
```diff
@@ -0,0 +1,187 @@
...
+  };
+
+  const handleComplianceFilter = (values) => {
+    if (isEmpty(values)) {
+      history.push(routes.MINE_INSPECTIONS.dynamicRoute(mineGuid));
+    } else {
+      const { violation, ...rest } = values;
+      history.push(
+        routes.MINE_INSPECTIONS.dynamicRoute(mineGuid, {
+          violation: violation && violation.join(","),
```
taraepp: SonarCloud wants this `violation` bit to be an optional chain. Do we need it to be `false` if not violation? If we don't need it to be false then I'd agree.

services/core-web/src/components/mine/Compliance/MineComplianceInfo.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const fiscalYear = getFiscalYear();
+    return (
+      <div>
+        <div>
+          <h4>COMPLIANCE OVERVIEW</h4>
+          <Divider />
+          <div className="compliance--container">
+            <LoadingWrapper condition={isLoaded}>
+              <div>
+                {mineComplianceInfo && mineComplianceInfo.last_inspection ? (
```
taraepp: it also wants an optional chain here. I think that's it...

taraepp (approved):

matbusby-fw (approved):
Nice!

PR #3492, opened 2025-04-07
[MDS-6329] Refactor Minespace permit header styles

matbusby-fw:
Updated Minespace permit styles to not interfere with other pages using parts of the css that was already in place

## Objective 

### DIFF
--- a/services/minespace-web/src/styles/components/CommonPageHeader.scss
+++ b/services/minespace-web/src/styles/components/CommonPageHeader.scss
@@ -1,55 +1,54 @@
 @use "../base.scss" as *;
 
 .common-page {
-    background-color: #F9F9F9;
+  background-color: #f9f9f9;
 
-    .view--header {
-        h1.ant-typography {
-            line-height: 54px;
-            letter-spacing: -0.5px;
-        }
+  .view--header {
+    z-index: 5;
+    h1.ant-typography {
+      line-height: 54px;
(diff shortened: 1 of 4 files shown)


services/minespace-web/src/styles/components/CommonPageHeader.scss
```diff
@@ -48,4 +48,8 @@
         background-color: $pure-white;
 
     }
+
+    .common-page-header {
+        height: calc(65px + $header-height);
```
taraepp: I am wondering if it was necessary to set the height on common-page-header and the `top` property on `permit-status-banner`- it looks like those properties are already correct on test (on the permit conditions page and on the project overview page). Just thinking about how fiddly those can be to maintain in coordination with the ant-tabs ones.

matbusby-fw: When I reverted the styles in the fixed-tab-container there were a lot of trickle down styles that messed up the positioning of the top banner in permits.
You're correct that these are annoyingly fiddly to maintain, but I think these are all necessary unless we separate out some of the other styling from the fixed-tab section which would also have to be updated on Core.

sggerard (dismissed):

taraepp (approved):

sggerard (approved):

asinn134 (approved):

PR #3493, opened 2025-04-07
[MDS-6472] Local Environment Improvements

taraepp:
## Objective 
- populate all necessary local environment variables with data from openshift for a faster dev set-up
   - put it in a loop to do all the variables listed, some changes there to handle dynamic values, also handle more characters, newlines
- there was no SYNCFUSION_LICENSE_KEY in the MS env.example file- added that in so there won't be a license error on local MS FE, now that we're going to be using that function
- label & auto-forward all the ports so that it's super obvious what is and isn't running, and ports don't have to be forwarded manually
- added some files to .gitignore that are used in environment setup
- updated documentation and the messages output in `make env` to be more accurate (running make env multiple times isn't actually destructive- trust me, I re-ran it _a lot._ But it will wipe out anything custom you've got there)
   - `make init` is pretty smooth now. auto-answering `yes` to all the questions wasn't a good thing though so that's gone now 

[MDS-6472](https://bcmines.atlassian.net/browse/MDS-6472)
ports tab looks like this now. I thought it would be nice to easily see what they all are without having to cross-reference the Dockerfile
image


bin/setenv.sh
```diff
@@ -78,21 +96,49 @@ function loadExternalSecrets() {
...
+            ESCAPED_SECRET=$(printf '%s\n' "$SECRET" | sed ':a;N;$!ba;s/[&/\]/\\&/g;s/\n/\\n/g')
+            if [ "$KEY" = "ELASTICSEARCH_CA_CERT" ]; then
+                ESCAPED_SECRET="\"$ESCAPED_SECRET\""
+            fi
+            sed -i "s|$KEY=.*|$KEY=$ESCAPED_SECRET|g" $SERVICES_PATH/$S/.env || echo -e "Failed to set $KEY for $S service"
+        done
+    done
 
+    # Read secrets where those env keys are not consistently named to match the local-dev-secrets key
+    OBJECT_STORE_ACCESS_KEY=$(kubectl get secret local-dev-secrets --namespace 4c2ba9-dev -o go-template='{{.data.OBJECT_STORE_ACCESS_KEY | base64decode}}')
```
matbusby-fw, line 121: I've gotten so used to having to set these manually.  This will be lovely.

matbusby-fw (approved):
Nice.  Very happy with the devcontainer stuff especially.  Since I don't generally use vsCode, it will be nice to have that setup more automated.

asinn134 (approved):

PR #3496, opened 2025-04-10
[MDS-6431] Draft functionality for TSF

taraepp:
## Objective 
- make a DraftMixin
- use it on both mine party appt & tsf
- make the mpa draft dependent on the tsf (have them be drafts when tsf is draft and not draft when it's submitted)
- save in draft state for old form and on the last page of the new form, otherwise save in draft

[MDS-6431](https://bcmines.atlassian.net/browse/MDS-6431)


services/core-api/app/api/mines/tailings/models/tailings.py
```diff
@@ -136,7 +158,8 @@ def create(cls,
             storage_location=storage_location,
             facility_type=facility_type,
             tailings_storage_facility_type=tailings_storage_facility_type,
-            mines_act_permit_no=mines_act_permit_no
+            mines_act_permit_no=mines_act_permit_no,
+            is_draft=is_feature_enabled(Feature.TSF_V2)
```
taraepp, line 162: I'm pretty sure this line won't be necessary when we remove the feature flag- the db column default should take over correctly

services/core-api/app/api/mines/tailings/resources/tailings.py
```diff
@@ -57,48 +55,40 @@ class MineTailingsStorageFacilityResource(Resource, UserMixin):
...
     parser.add_argument(
         'mines_act_permit_no',
         type=str,
         help='Mines Act Permit Number',
-        location='json',
         store_missing=False)
     parser.add_argument(
         'is_submitting',
         type=bool,
-        location='json',
```
taraepp, line 101: I was noticing really weird stuff with tests!
I was trying to send through is_submitting=True on the test and it just was _not picking up._ Additionally it was missing a bunch of other properties, and the lat/lon would always be None (lat/lon has a lambda). It was the location=json. I don't know how it worked on the application with that still in there, but it did not with tests. Works in the application without it.

services/core-api/app/api/parties/party_appt/models/mine_party_appt.py
```diff
@@ -149,6 +149,7 @@ def json(self, relationships=[]):
             str(self.start_date) if self.start_date else None,
             'end_date':
             str(self.end_date) if self.end_date else None,
+            'is_draft': str(self.is_draft),
```
taraepp, line 152: out of scope, but the mine party appointment endpoint doesn't marshal with the response model, it sends back the json. Also there's 2 response models. 🤷‍♀️

services/core-api/tests/mines/tailings/resources/test_mine_tailings_resource.py
```diff
@@ -134,10 +134,6 @@ def test_tsf_history_creates_record_on_update(test_client, db_session, auth_head
     entry = put_data['history'][1]
     changeset = entry['changeset']
 
-    print('hii')
```
matbusby-fw, line 137: Can't delete that.  This line is the glue that holds the entire app together.

matbusby-fw (approved):
Looks great!  Love the mixin!

asinn134 (approved):

PR #3501, opened 2025-04-15
[MDS-6413] 1 report to many conditions

taraepp:
## Objective 
🚧 👷 ⚠️ 
As this is a ticket with 3 subtasks that can be worked on by multiple devs, I'm creating this as a base PR!
Draft status until the sub-tasks are done, as I think it would be _unwise_ to merge things in parts.

- added some missing env variables, what I needed to get extraction going locally

[MDS-6413](https://bcmines.atlassian.net/browse/MDS-6413)


matbusby-fw (approved):
👍🏼

sggerard (approved):
Good work team!

PR #3502, opened 2025-04-15
[MDS-6451] Report Condition Migration

taraepp:
## Objective 
- create the new mapping table
- migrate existing data
- clean up the table data & columns afterwards
- add unique constraint, but don't apply it when report_name is null

[MDS-6451](https://bcmines.atlassian.net/browse/MDS-6451)

Little demo of the unique index:
- when trying to insert duplicate report name for a permit amendment 
image

- but I excluded null report names because there were just too many and I couldn't assume they were actual duplicates
image


services/common/src/redux/slices/mineReportPermitRequirementSlice.ts
```diff
@@ -27,11 +27,16 @@ const mineReportPermitRequirementSlice = createAppSlice({
         const headers = createRequestHeader();
         thunkApi.dispatch(showLoading());
 
+        const formatPayload = {
+          ...payload.values,
+          permit_condition_ids: [payload.values.permit_condition_id]
```
taraepp, line 32: just the minimum to run the post request- will most likely change when starting on FE

services/core-api/app/api/mines/permits/permit_extraction/create_permit_condition_report_requirement.py
```diff
@@ -71,13 +74,17 @@ def create_permit_condition_report_requirement(
...
         cim_or_cpo=cim_or_cpo,
         due_date_period_months=due_date_period_months or 0,
         initial_due_date=initial_due_date,
         ministry_recipient=None,  # Not specified in permits themselves.
     )
+    xref = MineReportReqPermitConditionXref(
+        mine_report_permit_requirement=mine_report_permit_requirement,
+        permit_condition_id=condition.permit_condition_id
+    )
+    mine_report_permit_requirement.mine_report_req_permit_conditions.append(xref)
```
taraepp, line 87: haven't really tested this out (beyond running existing tests), but it is quite possible that the extraction part of it is done. I just wanted the tests to pass...

services/core-api/app/api/mines/permits/permit_extraction/create_permit_conditions.py
```diff
@@ -87,7 +87,8 @@ def create_permit_conditions_from_task(task: PermitExtractionTask):
         return True
 
     except Exception as e:
-        current_app.logger.error("Failed to create permit conditions from task", e)
+        current_app.logger.error("Failed to create permit conditions from task")
+        current_app.logger.error(e)
```
taraepp, line 91: every time that the test hit this error (I think about 4 tests) it would output as "TypeError: not all arguments converted during string formatting" rather than the actual error, which was not very helpful, so I separated it out into its own line.

matbusby-fw (approved):
That's some nice SQL!  Also love the python typing!

PR #3507, opened 2025-04-23
[MDS-6487] TSF bugs: data loading, parties selector, pop confirm, new slice

taraepp:
## Objective 
- fix some of the issues that were cropping up with TSFs
- neither Core/MS would show updated data when navigating back to table
   - this is what the new slice is for- putting all the TSF data in one place
   - unfortunately, clearing the TSFs created big problems, so state may get a bit bogged down there
   - on the plus side, closer to remove TSFs from the mine response/FE state
- Core wouldn't let you add a party unless at least one party already existed
- MS would only show EoR/QP that already existed for that TSF specifically, not just the mine
- there was no confirmation to leave the page when the form was dirty

[MDS-6487](https://bcmines.atlassian.net/browse/MDS-6487)


sggerard (dismissed):
Looks good!

services/common/src/redux/actionCreators/mineActionCreator.ts
```diff
@@ -92,78 +88,6 @@ export const removeMineType = (mineGuid, mineTypeGuid, tenure) => (dispatch) =>
     .finally(() => dispatch(hideLoading("modal")));
 };
 
-export const createTailingsStorageFacility = (
```
matbusby-fw, line 95: Yeah, I remember this bugging me in the past.  Slice is better, and you don't belong here TSF!

services/common/src/redux/slices/tailingsSlice.ts
```diff
@@ -0,0 +1,180 @@
...
+                        successToastMessage: "Successfully created new Tailings Storage Facility",
+                    }).post(`${ENVIRONMENT.apiUrl}${API.MINE_TSFS(payload.mine_guid)}`, payload, headers);
+                } finally {
+                    thunkApi.dispatch(hideLoading());
+                }
+                return resp.data;
+            },
+            {
+                fulfilled: (state: TailingsState, action) => {
+                    const newTsf = action.payload;
```
matbusby-fw, line 63: Might be good to use your helper function here since the logic is almost identical.

services/core-web/src/components/common/PartySelectField.tsx
```diff
@@ -150,7 +150,7 @@ export const PartySelectField: FC<PartySelectFieldProps> = ({
   };
 
   useEffect(() => {
-    if (searchResults?.party?.length > 0) {
```
matbusby-fw, line 153: I know sonar cloud generally prefers the optional chaining here, is this no longer an array?

matbusby-fw (dismissed):
Not blocking, but left a suggestion and a commnet.

services/core-web/src/components/common/PartySelectField.tsx
taraepp: this was a bug that was driving me crazy!
Within this if-block is the bit that puts in the "Add a new party" button. So if searchResults.party == [], you would not have the option to add a new party. It's undefined when the search hasn't completed yet, so the falsey check works.

services/common/src/redux/slices/tailingsSlice.ts
taraepp: oh shoot, I missed one! I'm going to change this. 👍

sggerard (approved):

matbusby-fw (approved):

PR #3517, opened 2025-05-02
[MDS-6193] Correct NOW Datetime issues

sggerard:
## Objective 

[MDS-6193](https://bcmines.atlassian.net/browse/MDS-6193)

There is a bunch of automated formatting that is slightly obscuring this PR, there are really only two small changes that I have made to fix two issues I was able to find and confirm with the end user Amy.

last_updated_date: This dateTime is set by the backend when a record is updated, and stores a UTC timestamp. However the UTC information was being stripped when being returned by the API and the frontend also wasn't handling UTC. This meant that the last updated date in both the Preamble and on line 1.1 would be wrong in the afternoon.
Screenshot 2025-05-02 at 9 37 58 AM

preamble_date: This date stored on attached documents is not UTC (only stored as a Date in the db) would actually show on the permit correctly, however when displayed in the table it was being treated as a UTC Datetime so was showing incorrectly. This has been corrected by treating preamble_date as only a Date in both the backend and frontend.
image

### DIFF
--- a/services/common/src/redux/utils/helpers.ts
+++ b/services/common/src/redux/utils/helpers.ts
@@ -94,6 +94,9 @@ export const formatTractionDate = (dateString: string) => {
 export const formatDate = (dateString) =>
   dateString && dateString !== "None" && moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");
 
+export const formatDateUTC = (dateString) =>
+  dateString && dateString !== "None" && moment(dateString).format("MMM DD YYYY");
+
 export const formatTime = (timeStamp) => timeStamp && moment(timeStamp).format("h:mm a");
 
 export const formatDateTime = (dateTime) => dateTime && moment(dateTime).format("lll");
--- a/services/common/src/redux/utils/helpers.ts
+++ b/services/common/src/redux/utils/helpers.ts
@@ -16,6 +16,7 @@ import {
   ItemMap,
 } from "@mds/common/interfaces";
 import { MINE_REPORT_SUBMISSION_CODES } from "@mds/common/constants/enums";
+import { format, parseISO } from 'date-fns';
 
 /**
  * Helper function to clear redux form after submission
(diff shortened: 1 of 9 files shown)


asinn134 (dismissed):

asinn134 (commented):
Looks good, just have some errors with tests

services/common/src/redux/utils/helpers.ts
```diff
@@ -94,6 +94,9 @@ export const formatTractionDate = (dateString: string) => {
 export const formatDate = (dateString) =>
   dateString && dateString !== "None" && moment(dateString, "YYYY-MM-DD").format("MMM DD YYYY");
 
+export const formatDateUTC = (dateString) =>
+  dateString && dateString !== "None" && moment(dateString).format("MMM DD YYYY");
```
matbusby-fw: Another gradual migration we've introduced is using date-fns rather than moment for dateTime logic since moment has long since been in maintenance mode only.  Not the biggest deal since it's still being used in lots of places, but date-fns is much leaner.

Something like:
```
import { format, parseISO } from 'date-fns';

export const formatDateUTC = (dateString) => 
  dateString && dateString !== "None" ? format(parseISO(dateString), 'MMM dd yyyy') : null;
```

Didn't test that, so it might need some adjustment.

taraepp: I'm down with switching helper date handling to date-fns. I think there's a bunch of tests (mostly snapshots of tables) that should verify it's working as expected- would be nice so long as it doesn't break/change any of those. 👍

asinn134 (dismissed):

services/common/src/redux/utils/helpers.ts
sggerard: I had a look at moving all the date formatting, and timezone stuff in helpers.ts to date-fns and I think it's a little out of the scope of this ticket. I'm going to update just the method that I have added to date-fns and create a ticket to replace moment with date-fns and date-fns/tz.

matbusby-fw (approved):
🙌🏼

taraepp (approved):

PR #3534, opened 2025-05-13
[MDS-6499] Make Minespace autoselect role when creating new EoR and TSF

asinn134:
## Objective 

[MDS-6499](https://bcmines.atlassian.net/browse/MDS-6499)

- When user is creating a new EoR or a new TSF QP, job title is now changed to auto select to be EoR or TSF QP.
- When user selects an existing contact the job title will auto select to that contact's existing job title.


simensma-fresh (approved):

simensma-fresh (changes requested):
Looks like there's a snapshot that needs updating. Otherwise 👍

asinn134:
> Looks like there's a snapshot that needs updating. Otherwise 👍

I found the snapshot would require an update to a date value for every new day. I ended up changing the mock NoW mock data so we don't potentially have to deal with this every time someone makes a PR.

matbusby-fw (approved):

taraepp (approved):

simensma-fresh (approved):

PR #3538, opened 2025-05-15
[CHORE] Make command for user access data

taraepp:
## Objective 
Does exactly what the code comments says it does
So that devs don't need to hunt for mines or manually make subscriptions or MS user access
Maybe needs a reference in setup documentation

Results in Core:
image
Results in MS:
image

### DIFF
--- a/services/core-api/app/commands.py
+++ b/services/core-api/app/commands.py
@@ -243,9 +243,8 @@ def register_commands(app):
         - core user isn't a foreign key so no user created/queried
 
         Example usage:
-            flask seed_user_data IDIR=myidir
-            flask seed_user_data BCEID=mybceid
-            flask seed_user_data IDIR=myidir BCEID=mybceid        
+            flask seed_user_data myidir true
+            flask seed_user_data mybceid false    
         """
         _seed_user_data(user_name, is_idir)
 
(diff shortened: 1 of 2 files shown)


services/core-api/app/commands.py
```diff
@@ -230,6 +231,62 @@ def do_generate_table_command(table):
...
+    def do_seed_user_data(user_name, is_idir):
+        """
+        Creates subscriptions for MS/Core users
+        - first 5 (alphabetical) major mines
+        - last 5 regional mines
+        - uses MS user if they already exist, or else creates a new one
+        - core user isn't a foreign key so no user created/queried
+
+        Example usage:
+            flask seed_user_data IDIR=myidir
```
sggerard: I think the example usage here might be incorrect, this format is for the make command. For flask it would be something like  'flask seed_user_data sggerard false' I think?

taraepp: you're totally right. This is what I get for letting copilot write comments for me...

matbusby-fw (dismissed):
Nice!

asinn134 (approved):
Nice 👍

sggerard (approved):

PR #3539, opened 2025-05-16
[MDS-6497] Fix for mine showing alerts from another mine

asinn134:
## Objective 

[MDS-6497](https://bcmines.atlassian.net/browse/MDS-6497)

- Changed the default value of `loaded` react state so we can trigger the fetch to retrieve alert data when viewing a different mine.
- Also small change to get rid of unnecessary `title` in `NoticeOfWorkTable`.


sggerard (approved):
Nice fix!

simensma-fresh (approved):

PR #3543, opened 2025-05-21
[MDS-6062] Fix NOW transfer

sggerard:
## Objective 

Fix transferring a Notice of Work between mines where a Draft permit has already been created

[MDS-6062](https://bcmines.atlassian.net/browse/MDS-6062)

- Updates the Draft permit on transfer, including creating a new xref if needed.
- Updates the mine_guid of documents
- Migration to correct any exisiting draft permits with mismatching mine_guids

### DIFF
--- a/migrations/sql/V2025.05.15.15.40__update_mine_permit_xref.sql
+++ b/migrations/sql/V2025.05.15.15.40__update_mine_permit_xref.sql
@@ -1,4 +1,3 @@
-CREATE UNIQUE INDEX mine_permit_xref_unique ON mine_permit_xref (mine_guid, permit_id) WHERE deleted_ind = false;
 
 -- MDS-6062, Cleanup draft permits of transferred now applications
 CREATE TEMPORARY TABLE original_now_permits AS (
--- a/services/core-api/app/api/now_applications/models/now_application_identity.py
+++ b/services/core-api/app/api/now_applications/models/now_application_identity.py
@@ -64,19 +64,37 @@ class NOWApplicationIdentity(Base, AuditMixin):
         self.now_number = NOWApplicationIdentity.create_now_number(mine, self.now_number_year)
         self.mine = mine
 
+        if self.now_application.documents:
+            for document in self.now_application.documents:
+                document.mine_document.mine_guid = mine.mine_guid
+                document.mine_document.save()
+
         if self.now_application.draft_permit:
-            #Create new xref, replace, then delete old one
-            MinePermitXref.create(
-                permit_id = self.now_application.draft_permit.permit_id,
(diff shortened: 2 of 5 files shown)


taraepp (approved):

migrations/sql/V2025.05.15.15.40__update_mine_permit_xref.sql
```diff
@@ -0,0 +1,42 @@
+
+-- MDS-6062, Cleanup draft permits of transferred now applications
+CREATE TEMPORARY TABLE original_now_permits AS (
```
matbusby-fw, line 3: Probably not a big deal, but if this only needs to be run in Prod, then you could move this file to the `migrations/sql-prod` folder.

services/core-api/app/api/now_applications/models/now_application_identity.py
```diff
@@ -56,9 +57,45 @@ def __repr__(self):
         return f'{self.__class__.__name__} {self.now_application_guid}'
 
     def transfer(self, mine):
+
+        if self.now_application.now_application_status_code in ['AIA','REJ', 'WDN', 'NPR']:
+            raise BadRequest(f'Unable to transfer NOW application with status {self.now_application.now_application_status_code}')
```
matbusby-fw, line 62: Might be good to parse the code here into the human readable version since the abbreviations aren't always obvious.

matbusby-fw (changes requested):
Looks awesome!  Just a couple of minor things.  Moving the sql likely doesn't matter too much, but I think it would be good to show the status name in the user error rather than the code.

migrations/sql/V2025.05.15.15.40__update_mine_permit_xref.sql
sggerard: Ooo didn't know this was a thing! I think I do want it to run in dev/test just to make sure no errors are thrown for some reason even if there is no records altered.

asinn134 (approved):

services/core-api/app/api/now_applications/models/now_application_identity.py
sggerard: I totally agree with you, but we don't really have any examples of using enums on the backend as far as I can see currently. So maybe a wider discussion for the team moving forwards?

migrations/sql/V2025.05.15.15.40__update_mine_permit_xref.sql
matbusby-fw: Sure, sounds good!

matbusby-fw (approved):

PR #3549, opened 2025-05-27
[MDS-6491] Condition search - Fix now documents, add tenure to index

sggerard:
## Objective 

[MDS-6491](https://bcmines.atlassian.net/browse/MDS-6491)

Address acceptance criteria that was missed from the original PR.

1. Missing NOW documents. I've added some logic to grab the first mine_document from the now documents. Is this the best logic to use?
image

2. Add Tenure as a filter. I have presumed the logic is the same for this regardless if its a NOW or Major Mine permit.
image

### DIFF
--- a/services/common/src/interfaces/search/facet-search.interface.ts
+++ b/services/common/src/interfaces/search/facet-search.interface.ts
@@ -20,6 +20,7 @@ export interface HaystackDocumentMeta {
         sibling_contexts?: SiblingContexts;
         child_contexts?: ContextItem[];
     }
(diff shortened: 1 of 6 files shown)


matbusby-fw (approved):
Looks good from what I can see.
Regarding your comments in your description, I honestly don't know which is the best file to use.  The first seems as good a choice as any though if Rebecca doesn't have a strong preference.
And yes, the tenure should behave the same for both types of permits.

asinn134 (approved):

PR #3551, opened 2025-05-28
[MDS-6516] Add mandatory report name in extraction

asinn134:
## Objective 

[MDS-6516](https://bcmines.atlassian.net/browse/MDS-6516)

- adds setup of report name for conditions that have a required report but the report has no name
- format is Report#1 - <condition_category> - <condition_section>

image

image


taraepp (approved):
Looks good to me!

PR #3553, opened 2025-05-29
[MDS-6522] Add Review Started to NOW table

sggerard:
## Objective 

[MDS-6522](https://bcmines.atlassian.net/browse/MDS-6522)

Add a column to the minespace NOW table for when the Technical review was started.

image

### DIFF
--- a/services/common/src/constants/enums.ts
+++ b/services/common/src/constants/enums.ts
@@ -204,6 +204,15 @@ export enum MAJOR_MINE_APPLICATION_AND_IRT_STATUS_CODES {
   UNR = "Under review",
   CHR = "Change Requested",
 }
+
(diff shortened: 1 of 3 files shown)


asinn134 (approved):

services/common/src/constants/enums.ts
```diff
@@ -204,6 +204,15 @@ export enum MAJOR_MINE_APPLICATION_AND_IRT_STATUS_CODES {
   UNR = "Under review",
   CHR = "Change Requested",
 }
+
+export enum NOW_APPLICATION_PROGRESS_STATUS_CODES {
```
taraepp, line 208: nice

taraepp (approved):

PR #3562, opened 2025-06-05
[MDS-6541] Handle deleted conditions

sggerard:
## Objective 

[MDS-6541](https://bcmines.atlassian.net/browse/MDS-6541)

- Delete xrefs when a condition is deleted
- Delete xrefs when a permit extraction is deleted
- Filter out deleted conditions/xrefs so permit extraction doesn't regenerate them
- Prevent reports with no conditions from being selectable on the FE
- Prevent reports with no conditions from being submitted on the BE
- Prevent conditions from being added to multiple reports by disabling them in the tree


taraepp (approved):

matbusby-fw (approved):
Nice work!

PR #3565, opened 2025-06-09
[MDS-6442] Update document templates

matbusby-fw:
Updated the templates for the NoW Letter of Acknowledgement and Permit Enclosed Letter

- updated the 'form_spec_json' column for both of these templates in the database
- added mapping of agent contact in the api for the letter of acknowledgement
- updated the docx templates to match the new requests

## Objective 

[MDS-6442](https://bcmines.atlassian.net/browse/MDS-6442)


taraepp (approved):

sggerard (approved):
Good job getting through this!

PR #3574, opened 2025-06-17
[MDS-6546] env final app db

taraepp:
## Objective 
- add data tables for ams final application
- add API resource for get/post/put the final application
- test it

[MDS-6546](https://bcmines.atlassian.net/browse/MDS-6546)


migrations/sql/V2025.06.09.13.28__ams_final_app.sql
```diff
@@ -0,0 +1,64 @@
+-- 1. -- a table for the final application itself
+CREATE TABLE IF NOT EXISTS ams_final_application (
+    ams_final_application_guid    uuid DEFAULT gen_random_uuid()          PRIMARY KEY,
+    project_summary_authorization_guid          uuid                                    NOT NULL,
+    submitter_name                              VARCHAR(255)                            NOT NULL,
+    is_agent                                    boolean DEFAULT false                   NOT NULL,
+    is_draft                                    boolean DEFAULT true                    NOT NULL,
+    pre_submitted_files                         text[] DEFAULT ARRAY[]::text[]          NOT NULL,
```
matbusby-fw, line 8: I'm curious why you went with an array here rather than creating an xref.  Aren't all presubmitted documents referencing documents that were added elsewhere within Minespace/Core so we'd have access to them somewhere?

taraepp: We're just asking them to check off a list of which document types they already submitted, rather than getting them to point out which document it is exactly. So I don't have the reference to the document. I _could_ do an xref to ams_final_application_document_type but it felt like over-engineering.

matbusby-fw: Okay, yeah that makes sense then.  I couldn't remember what we landed on in our conversation.

matbusby-fw (approved):

asinn134 (approved):

PR #3575, opened 2025-06-17
[MDS-6215]fix to have order number be taken from nris xml data

asinn134:
## Objective 

[MDS-6215](https://bcmines.atlassian.net/browse/MDS-6215)

- Adjusted inspection order_number to be retrieved from external NRIS XML data instead of a variable in our backend
- Adjusted sorting of order_number in Minespace and Core inspection order table


services/common/src/utils/helpers.ts
```diff
@@ -140,4 +140,22 @@ export const transformPermitReportRequirement = (report: IMineReportPermitRequir
     }
   }
   return null;
-}
\ No newline at end of file
+}
+
+export const InspectionOrderNumberSorter = (a, b, dataIndex) => {
```
taraepp: nitpick- the case makes me expect that `InspectionOrderNumberSorter` is a class. I'd want the i to be lowercase.

taraepp:
Other than the little nitpick I had about PascalCase vs camelCase, looks really good to me!

matbusby-fw (approved):
Looks good.  It is awkward using alembic for migrations on some of our services and flyway on others, but I guess we don't make changes to the alembic ones very often 🤷🏼‍♂️

taraepp (approved):

PR #3577, opened 2025-06-19
[MDS-6493] Auto index Major Mine conditions

sggerard:
## Objective 

[MDS-6493](https://bcmines.atlassian.net/browse/MDS-6493)

- Auto indexes permit conditions for a Major mine permit amendment on initial extraction as well as whenever all conditions are verified.
- The file name uploaded to Azure is now friendly and human readable instead of the random name of the tempfile.
- Added verification_status filter to show whether a condition has been verified by a human or not. When a condition is not verified a little warning symbol shows:

image


matbusby-fw (dismissed):
Nice!  Looks great!

matbusby-fw (approved):

taraepp (approved):

PR #3583, opened 2025-06-25
[MDS-6435] AMS final application UI

taraepp:
## Objective 
- FE for the AMS final application
- I wanted to use the CommonPageHeader but hadn't used it without tabs before, so I modified it a bit so you can pass in custom content.

[MDS-6435](https://bcmines.atlassian.net/browse/MDS-6435)

Basic information:
image
Documents:
image
Declaration:
image
MineSpace:
image
For MineSpace, I did change the mine tag in the header to be blue instead of transparent, to match Core styling (and make it look more like an interactive element)


services/common/src/components/project/envApplication/EnvDocumentsTab.tsx
```diff
@@ -0,0 +1,144 @@
...
+                                    placeholder="Select a Document Category"
+                                    component={RenderSelect}
+                                    data={documentTypeOptions}
+                                    required
+                                    validate={[required]}
+                                    allowClear={false}
+                                />
+                            </Col>
+                        </Row>
+                        {isSaved && <Button
```
taraepp, line 68: this bit I put in because when you remove a document that hasn't been saved as a mine document yet, it would still be sitting in the uploader, and that was weird. There wasn't a simple way to remove them from the uploader when the "Delete File" button was clicked, so instead I removed the Delete File button and made the user remove them from the uploader instead.
There's no confirmation when you delete from the uploader, but that's nothing new. 🤷‍♀️

services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
```diff
@@ -1,13 +1,32 @@
...
 from app.api.projects.response_models import AMS_FINAL_APPLICATION_MODEL
+from app.api.projects.project_summary.models.project_summary import ProjectSummary
 from app.api.projects.project_summary.models.project_summary_authorization import ProjectSummaryAuthorization
 from app.api.projects.ams_final_application.models.ams_final_application import AmsFinalApplication
+from app.api.mines.documents.models.mine_document import MineDocument
+from app.api.mines.mine.models.mine import Mine
 from app.api.utils.custom_reqparser import CustomReqparser
+from app.api.services.document_manager_service import DocumentManagerService
+
+class AmsFinalApplicationDocumentResource(Resource, UserMixin):
```
taraepp, line 17: there was a bit of document handling missing on the BE.

services/minespace-web/src/styles/overrides/antd-overrides.scss
```diff
@@ -23,10 +23,14 @@
...
   z-index: 8;
   width: 100%;
-  min-height: 80px;
+  min-height: $layout-footer-height;
   bottom: 0;
   padding: 0;
   height: $layout-footer-height;
+
+  @include screen-lg {
+    height: unset;
```
taraepp, line 32: footer bug!
Any screen size smaller than lg (somewhere ~1000px) on MS, you lose footer items and it appears to "float". It does this at the breakpoint where the items change from a horizontal list to vertical and so the height _should_ change. This is actually a pretty reasonable screen width to use with MS so I dealt with it.

services/common/src/components/project/envApplication/EnvDocumentsTab.tsx
```diff
@@ -0,0 +1,144 @@
...
+        dispatch(arrayPush(
+            formName,
+            "documents",
+            {
+                document_name,
+                document_manager_guid
+            }
+        ));
+    };
+    console.log('HI TARA formValues', formValues)
```
asinn134: This console log might be something we can remove, unless it was intended!

asinn134 (dismissed):
Great job! 👍

services/common/src/components/project/envApplication/EnvDocumentsTab.tsx
taraepp: haha oops! I thought I got all of those! Yeah I'll take it out

services/common/src/components/common/CommonPageHeader.tsx
```diff
@@ -21,7 +21,10 @@ interface CommonPageHeaderProps {
   mineGuid: string;
   current_permittee: string; // would be ideal to get this from the mine
   breadCrumbs?: BreadCrumb[];
+  // pass in either tabProps for tabbed view or pageContent for any styled content
```
matbusby-fw, line 24: Might be nice to make this an actual check in the component?  If both of these exist, explode?

sggerard (approved):
Good work!

matbusby-fw (approved):
Approvinated!

PR #3590, opened 2025-06-30
[MDS-6547] Flagsmith local fixes

taraepp:
## Objective 
- I made some typos. 😅 
- an empty value was being parsed as truthy. Made sure it came in as false if either the value was false or it was missing.

[MDS-6547](https://bcmines.atlassian.net/browse/MDS-6547)

### DIFF
--- a/services/common/src/constants/environment.ts
+++ b/services/common/src/constants/environment.ts
@@ -14,7 +14,7 @@ export const DEFAULT_ENVIRONMENT = {
   keycloak_url: "https://test.loginproxy.gov.bc.ca/auth",
   flagsmithKey: "4Eu9eEMDmWVEHKDaKoeWY7",
   flagsmithUrl: "https://mds-flags-dev.apps.silver.devops.gov.bc.ca/api/v1/",
-  flagsmithLocal: process.env.FLAGSMITH_ENABLE_LOCAL_EVALUATION || "false",
+  flagsmithLocal: process.env.FLAGSMITH_ENABLE_LOCAL_EVALUATION === "true" || false,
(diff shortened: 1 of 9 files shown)


services/common/src/constants/environment.ts
```diff
@@ -14,7 +14,7 @@ export const DEFAULT_ENVIRONMENT = {
   keycloak_url: "https://test.loginproxy.gov.bc.ca/auth",
   flagsmithKey: "4Eu9eEMDmWVEHKDaKoeWY7",
   flagsmithUrl: "https://mds-flags-dev.apps.silver.devops.gov.bc.ca/api/v1/",
-  flagsmithLocal: process.env.FLAGSMITH_ENABLE_LOCAL_EVALUATION || "false",
+  flagsmithLocal: process.env.FLAGSMITH_ENABLE_LOCAL_EVALUATION === "true" || false,
```
taraepp, line 17: this was coming in as the string "false" (whether from the env or from it being blank) which == true. Eek.

services/common/src/constants/environment.ts
```diff
@@ -110,6 +111,7 @@ export function setupEnvironment(
   ENVIRONMENT.environment = environment || "development";
   ENVIRONMENT.flagsmithKey = flagsmithKey;
   ENVIRONMENT.flagsmithUrl = flagsmithUrl;
+  ENVIRONMENT.flagsmithLocal = flagsmithLocal;
```
taraepp, line 114: aaaand when it was missing it was coming in as "<FLAGSMITH_ENABLE_LOCAL_EVALUATION>". Needed to add it to this function and then pass it to it from ms/core

matbusby-fw (approved):

sggerard (approved):

PR #3591, opened 2025-07-03
[MDS-6563] Add termination report trigger for EOR and TQP appointments

matbusby-fw:
## Objective 

- Added a trigger to generate report requests for the termination (change) of EOR/QP assignments when a new assignment is created, and an active one already exists.
- These will trigger when the new assignment is made from Core or Minespace.

[MDS-6563](https://bcmines.atlassian.net/browse/MDS-6563)

image

### DIFF
--- a/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
+++ b/services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
@@ -184,7 +184,7 @@ class MinePartyApptResource(Resource, UserMixin):
             mine_party_acknowledgement_status = MinePartyAcknowledgedStatus.not_acknowledged
         elif mine_party_appt_type_code == 'EOR':
             mine_party_acknowledgement_status = MinePartyAcknowledgedStatus.acknowledged
-        
+
         if end_current and (not is_minespace_user() or mine_party_appt_type_code == 'TQP'):
             new_status = MinePartyAppointmentStatus.active
             MinePartyAppointment.end_current(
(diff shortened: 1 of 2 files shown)


services/core-api/app/api/parties/party_appt/resources/mine_party_appt_resource.py
```diff
@@ -250,6 +250,16 @@ def post(self, mine_party_appt_guid=None):
...
 
+        if end_current and mine_party_appt_type_code in ['EOR', 'TQP']:
+            if mine_party_appt_type_code == 'EOR':
+                current_eor_list = MinePartyAppointment.find_current_appointments(mine_guid, 'EOR', None, tsf.mine_tailings_storage_facility_guid)
+                if len(current_eor_list) > 0:
+                    current_eor_list[0].request_termination_report_if_required(True)
+            if mine_party_appt_type_code == 'TQP':
+                current_tqp_list = MinePartyAppointment.find_current_appointments(mine_guid, 'TQP', None, tsf.mine_tailings_storage_facility_guid)
+                if len(current_tqp_list) > 0:
+                    current_tqp_list[0].request_termination_report_if_required(True)
```
sggerard: I feel like this is repeated logic doing the same thing for both EOR & TQP, could you instead not pass the mine_party_appt_type_code into the find_current_appointments() function call? You have already checked its valid

matbusby-fw: Sure.  That's a good suggestion!

matbusby-fw: Done

asinn134 (approved):
Nice!

sggerard (approved):

PR #3593, opened 2025-07-07
[MDS-6557] Permit Condition Editor UI

taraepp:
## Objective 
- make the form more usable
- more clear how things are structured, and what they'll be editing

[MDS-6557](https://bcmines.atlassian.net/browse/MDS-6557)

- new condition: moved the placeholder text to a label, made it required. Reference the parent condition, so that user doesn't lose context
- broke up the white between each category so they're visually separate
image

- put a border around each level so that it's clear when an item is a child of another
- changed the text in the "+" button to reflect the type of condition being added
image

- allow user to change which condition they're editing without clicking the "X" button to cancel changes
- if the first form that they were on is modified, bring up a "discard changes" confirmation
image

- delete condition modal
- basically the same as NoW delete modal
image

Not included in any screenshots:
- when you hover over a condition to edit it, it'll say something like "Edit Condition General 2.a" now, instead of "Edit Condition"

Couple little things that I neglected or didn't solve:
- you can click another condition to change what you're editing, but not "Add Condition" (stays disabled)
- clicking the cancel button on the "Add Condition" form always triggers the required validation on first click and has to be clicked again. 🤦‍♀️ 
  - e.stopPropagation did not stop it so I don't know what's going on there.
  - this only occurs when you click "Add Condition" (to a category, not a sub-condition), don't enter text, and then cancel it, so I decided it was trivial

### DIFF
--- a/services/common/src/components/forms/RenderCancelButton.tsx
+++ b/services/common/src/components/forms/RenderCancelButton.tsx
@@ -60,7 +60,11 @@ const RenderCancelButton: FC<RenderCancelButtonProps> = ({
   const { formName, isModal, isEditMode } = useContext(FormContext);
   const isFormDirty = useSelector(isDirty(formName));
 
-  const handleCancel = () => {
+  const handleCancel = (e?) => {
+    if (e) {
+      e?.stopPropagation();
+      e?.preventDefault();
+    }
     if (cancelFunction) {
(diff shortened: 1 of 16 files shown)


sggerard (approved):
Great work!

asinn134 (approved):
👍

PR #3595, opened 2025-07-07
[MDS-6562] Fix filtering of permit conditions

simensma-fresh:
## Objective 

[MDS-6562](https://bcmines.atlassian.net/browse/MDS-6562)

There's a bug in the Haystack Azure AI Search integration where it doesn't handle collection index field types correctly.
Issue: Filters on a field of the type Collection(...) fails with an error (it works for single fields).

Expected Odata filter syntax for a collection field is: 

`field/any(field: <filter>)` where `field` is the name of the field and `value` is the value to filter on. The haystack integration spits out just the <filter> part.

Example: `search.in(field, 'value1,value2', ',')` should be `field/any(field: search.in(field, 'value1,value2', ','))`, but is not.

The following code overrides the comparison operators to handle a collection index field type.
Why override and not fix the bug in Haystack? This aims to fix it specifically for our use case,
a more general fix would require more considerations (e.g. handle 'any' vs 'all', complex types etc.).

Reference: https://learn.microsoft.com/en-us/azure/search/search-query-odata-filter


services/permits/app/pipelines/permit_condition_search/stores/ai_search_document_store.py
```diff
@@ -79,29 +114,12 @@ def _convert_search_result_to_documents(
...
-
-        return index_fields
-
     def _hybrid_retrieval(
         self,
         query: str,
         query_embedding: List[float],
         top_k: int = 25,
-        filters: Optional[Dict[str, Any]] = None,
+        filters: Optional[str] = None,
```
simensma-fresh, line 122: The `filters` param here is now just the Odata expression, so str is correct

services/permits/app/pipelines/permit_condition_search/stores/ai_search_document_store.py
```diff
@@ -1,18 +1,50 @@
...
+# The following code overrides the comparison operators to handle a collection index field type.
+# Why override and not fix the bug in Haystack? This aims to fix it specifically for our use case,
+# a more general fix would require more considerations (e.g. handle 'any' vs 'all', complex types etc.).
+
+# Reference: https://learn.microsoft.com/en-us/azure/search/search-query-odata-filter
+
+fields_dict = {field.name: field for field in fields}
+og_fltr = fltrs.COMPARISON_OPERATORS
+
+def override_comparison_operators(op, func):
```
simensma-fresh, line 33: Don't like this one bit, but here we are 😅

sggerard (approved):
Yay! Good work Simen, even if it's not pretty I'm glad you found a solution

taraepp (approved):

matbusby-fw (approved):
😬.  Nice workaround.  Too bad we need it though.

PR #3597, opened 2025-07-09
[MDS-6492] index NoW issued permit for search

asinn134:
## Objective 

[MDS-6492](https://bcmines.atlassian.net/browse/MDS-6492)

- Makes use of indexing functionality Sam had previously added to index NoW permit when it is issued

image

image

image


sggerard (approved):

matbusby-fw (approved):
👍🏼

PR #3599, opened 2025-07-11
[MDS-6560] - Fix mine alerts persisting in state

matbusby-fw:
- Mine Alerts from a previously visited mine were persisting when visiting another mine with no alerts in the same session.
- (unreported, but also fixed) Mine Alerts were not disappearing on delete until the user refreshed.

## Objective 

[MDS-6560](https://bcmines.atlassian.net/browse/MDS-6560)

mine-alert-persistance


services/core-web/src/constants/actionTypes.js
```diff
@@ -6,3 +6,4 @@ export const STORE_DOCUMENT_COMPRESSION_PROGRESS = "STORE_DOCUMENT_COMPRESSION_P
 export const STORE_MINE_ALERTS = "STORE_MINE_ALERTS";
 export const STORE_GLOBAL_MINE_ALERTS = "STORE_GLOBAL_MINE_ALERTS";
 export const CLEAR_MINE_ALERT = "CLEAR_MINE_ALERT";
```
taraepp: I'm curious if this `CLEAR_MINE_ALERT` action is used anywhere- it kind of looks like it's orphaned? Might be nice to remove it entirely.

matbusby-fw: Good call.  Done.

taraepp (approved):

simensma-fresh (approved):

PR #3600, opened 2025-07-14
[MDS-6537] Permit Condition Tags

sggerard:
## Objective 

[MDS-6537](https://bcmines.atlassian.net/browse/MDS-6537)

Add the ability to manually tag permit conditions.

- Behind feature flag "permit_condition_tags"
- Migrations to create permit_condition_tag and permit_condition_tag_xref tables
- New admin page to Create, Edit and Delete tags
- Multi select added to permit condition form

image

image


services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { FC, useEffect, useState } from "react";
 import { useAppDispatch, useAppSelector } from "@mds/common/redux/rootState";
 import { change, Field, isDirty, reset } from "@mds/common/components/forms/form";
-import { Row, Col, Button, Typography, Modal } from "antd";
+import { Row, Col, Button, Typography, Modal, Select } from "antd";
```
taraepp: I'm guessing you experimented with `<Select />` but didn't end up using it?

services/core-web/src/components/admin/permitConditions/TagEditForm.tsx
```diff
@@ -0,0 +1,53 @@
+import { IPermitConditionTag } from "@mds/common/interfaces"
+import React from "react"
+import { FC } from "react"
```
taraepp: probably want to combine these 2 onto the same line...

services/core-web/src/styles/components/PermitConditions.scss
```diff
@@ -111,6 +111,28 @@ div.condition-layer {
   align-items: center;
 }
 
+.condition-tag-select {
```
taraepp, line 116: I'm wondering if this styling is something we can move into the multiselect component?
Something like where we pass in a prop like `tagMode=true` and it handles putting the classname on it. Just so we can easily use it in other places and keep it consistent

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -307,6 +316,18 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
                                                 : "Add Report Requirement"}
                                         </Button>
                                     </Col>
+                                    {enablePermitConditionTags && conditionTags &&
+                                        <Col className="condition-tag-select">
+                                            <Field
+                                                name="condition_tags"
```
taraepp, line 335: I'm wondering, since this is another item added to a row of a bunch of buttons, how wide does your screen have to be to keep it all in one line? Does it deserve its own line?

matbusby-fw: I think it looks decent on smaller screens as is.  It wraps when things get smaller so it never seems to look worse than anything else considering we're not generally optimized for mobile.

image

services/common/src/redux/actionCreators/permitActionCreator.ts
```diff
@@ -686,3 +687,81 @@ export const updateStandardPermitCondition = (
       .catch(() => dispatch(error(NetworkReducerTypes.UPDATE_PERMIT_CONDITION)))
       .finally(() => dispatch(hideLoading()));
   };
+
+export const fetchPermitConditionTags = (): AppThunk => (dispatch) => {
```
matbusby-fw: Would be a bit of a bigger overhaul, but I think we're tending in the direction lately of using the newer `slice` format for setting up actions like this.  It consolidates things quite a bit, so there's less need to update 4 or 5 different files when you need to make a change.  
This format is still valid, but it's worth considering.

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_conditions_resource.py
```diff
@@ -213,3 +219,30 @@ def set_audit_metadata(permit_amendment, commit=True):
...
+
+def updateConditionTags(old_tags, new_tags, condition):
+    if old_tags != new_tags:
+
+        # Remove tags
+        for tag in old_tags:
+            if tag not in new_tags:
+                xref = PermitConditionTagXref.find_by_guid_and_condition_id(tag,condition.permit_condition_id)
+                if xref:
+                    xref.deleted_ind = True
```
matbusby-fw: Since you have the SoftDeleteMixin here, you should be able to simply call `xref.delete(True)` I think, and it will handle the soft delete as well as the save.

sggerard: image

I'm not entirely sure why it doesn't work, maybe only in the model?

services/core-api/app/api/mines/permits/permit_conditions/resources/permit_conditions_resource.py
```diff
@@ -213,3 +219,29 @@ def set_audit_metadata(permit_amendment, commit=True):
     permit_amendment.permit_conditions_last_updated_by = User().get_user_username()
     permit_amendment.permit_conditions_last_updated_date = datetime.now(timezone.utc)
     permit_amendment.save(commit=commit)
+
+def updateConditionTags(old_tags, new_tags, condition):
```
taraepp: Sonarcloud wants this function to follow the more pythonic snake_case naming convention.

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -307,6 +314,18 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
...
+                                                component={RenderMultiSelect}
+                                                placeholder="Select tags"
```
taraepp, line 337: Larger concern- the tags don't show up in view mode, only in edit mode. Not sure if that was on purpose, but it does strike me as very odd. I'd still put it on its own row so that it isn't inside the bit to hide it in view mode, and then it won't wrap the buttons to a second line as a bonus.

taraepp:
image
I had to check what would happen if I tagged a condition, then deleted the tag.

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -307,30 +329,46 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
                                                 : "Add Report Requirement"}
                                         </Button>
                                     </Col>
+                                    {enablePermitConditionTags && conditionTags &&
+                                        <Col className="condition-tag-select">
```
matbusby-fw, line 333: Personally, I'm happy with how this is looking!

services/common/src/redux/slices/permitConditionTagSlice.ts
```diff
@@ -0,0 +1,124 @@
...
+
+interface permitConditionTagsState {
+    permitConditionTags: IPermitConditionTag[];
+}
+
+const initialState: permitConditionTagsState = {
+    permitConditionTags: [],
+};
+
+const permitConditionTagSlice = createAppSlice({
```
matbusby-fw, line 21: 🙌🏼

matbusby-fw (dismissed):
Looking great!

matbusby-fw (approved):

taraepp (approved):

PR #3603, opened 2025-07-16
[MDS-6601]fix for index runner being passed search endpoint that isn't a string

asinn134:
## Objective 

[MDS-6601](https://bcmines.atlassian.net/browse/MDS-6601)

- Adjusted `create_permit_condition_search_indexing_pipeline` function's `search_endpoint` property being passed to `IndexRunner` to be the string value instead of `EnvVarSecret` object

image


simensma-fresh (approved):

matbusby-fw (approved):
Nice fix!

PR #3611, opened 2025-07-25
[MDS-6590] Changes to support azure ai search in prod

asinn134:
## Objective 

[MDS-6590](https://bcmines.atlassian.net/browse/MDS-6590)

- Created new blob storage container 
- Added new secrets in dev, test, and prod env
- Changed hardcoded azure search config values to be dynamic
- Note: Will have to wait until code is in production to run the command flask cli command `update_search_index` to create prod specific index, indexer, data source, and skillset. Then will have to test in prod to make sure everything works.

image

image


asinn134:
Hi @simensma-fresh I was wondering if we need to set a string value in the `@permit_condition_search.command()` to run the cli command properly when in production?

image

simensma-fresh:
> Hi @simensma-fresh I was wondering if we need to set a string value in the `@permit_condition_search.command()` to run the cli command properly when in production?
> 
> image

I think we do not. You can confirm if you shell into the container and run `flask` - if it shows up in the available commands we're all good

asinn134:
> > Hi @simensma-fresh I was wondering if we need to set a string value in the `@permit_condition_search.command()` to run the cli command properly when in production?
> > image
> 
> I think we do not. You can confirm if you shell into the container and run `flask` - if it shows up in the available commands we're all good

I found the flask command is not available when I shell into the container. I think this command `python -m app.cli --help`
will show a list of commands, but I noticed there were some type errors trying to access some config.search Secret type properties when I run that command. I added `.resolve_value()` to hopefully fix this. I think the command to create the azure index, indexer, etc is this one here 
`python -m app.cli permit-condition-search update-search-index`.

sggerard (approved):

simensma-fresh (approved):

PR #3621, opened 2025-07-30
[MDS-6558] Notice of Work Permit Conditions Editor

taraepp:
## Objective 
- make the notice of work permit drafting use new editor
- upgrade a couple of key components in the area to TS
   - still a can of worms, but a bit better, mostly focussed on being able to tell if component props were incorrect. Objects are frequently mutated, making consistent typing difficult. The typing we do have represents an inaccurate view of the data- a property that is realistically a string representing a number is typed as a number, or a string date as a Date.  
   - missing properties are returned as empty objects from the BE but are typed as required in the interface, and components that use those properties expect them there without optional chaining. Party address is typed as an address but _sometimes_ actually returned as an array... Lots of effects on other components if those properties are "fixed"
- fix a couple of UI issues, mostly relating to buttons having no spacing between them
   - add css class names, but also quit with rendering nested spans and empty divs
   - move a couple of tooltip labels into the label property of the field
 - combined the AuthorizationWrapper components in core & common by importing the common one into core, moved its test into common, removed props that were never used anywhere, deprecate "inTesting" prop as we use flagsmith for features now

[MDS-v](https://bcmines.atlassian.net/browse/MDS-6558)


services/core-web/src/components/mine/ExplosivesPermit/MineExplosivesPermitTable.tsx
```diff
@@ -167,10 +159,6 @@ const MineExplosivesPermitTable: FC<RouteComponentProps & MineExplosivesPermitTa
 
         return (
           <div className="btn--middle flex">
-            {isApproved && !hasDocuments && isCoreSource && (
```
taraepp, line 170: This change probably looks out of place. 😅 
Basically, I typed AuthorizationWrapper so that `children` was a required prop. Which brought up this typescript error because here it is/was with no children! And then when I removed it there were other unused variables/imports.
Nothing has actually changed here.

sggerard (approved):
Love your work!

services/common/src/redux/selectors/permitSelectors.ts
```diff
@@ -229,6 +245,38 @@ export const getPermitConditionCategories = (permitGuid, permitAmendmentGuid) =>
...
+      if (!conditions || !categories) {
+        return {
+          conditionMap: {},
+          categoriesWithConditions: []
+        };
+      }
+      const conditionMap: { [permit_condition_id: string]: IPermitCondition } = {}
+
+      const categoriesWithConditions = categories.map((cat) => {
+        const catConditions = conditions.filter((c) =>
```
asinn134, line 262: 🐈

asinn134 (approved):

PR #3652, opened 2025-08-11
[MDS-6606] Standard Conditions Tags & Report Requirements

taraepp:
## Objective 
1. make it so admin users can add tags & report requirements to the standard conditions
2. and then when the standard conditions are copied over to a NoW when a draft permit is started, those tags & report requirements are also copied over

[MDS-6606](https://bcmines.atlassian.net/browse/MDS-6606)


sggerard (approved):
Great work!

asinn134 (approved):

PR #3658, opened 2025-08-27
[MDS-6461] ENV Process Tracking

taraepp:
## Objective 
- new project overview tab, shared between Core & MS

[MDS-6461](https://bcmines.atlassian.net/browse/MDS-6461)

MS screenshots:
Project with only non-AMS authorizations, IRT & MMA started but in draft:
image
- changes made here: Under permit & authorization added "Authorization Type" column, changed 1st column named "Permit" to "Permit Type" and decided to show all authorizations
- for non-AMS final application, made it match IRT for consistency (may want to clarify with Rebecca here if final application should only be for mines act permit auth?)

Project with only AMS authorizations. Project summary has been submitted
image

Project with both auth types. Project summary still in draft
- can't start IRT & final app yet
image

Core screenshot:
- AMS final applications show up on page but disabled when they don't exist in AMS tracking system yet (like IRT & MMA)
- didn't fuss with button text, it's always View. That can be changed when rules are in place to limit editing so it stays consistent.
image


services/common/src/components/projects/ProjectContacts.tsx
```diff
@@ -0,0 +1,66 @@
...
+                        <Typography.Text strong className="ministry-contact-title taratest4">
+                            {jobTitle}
+                        </Typography.Text>
+                    )}
+                    <br />
+                    {c.is_project_lead_contact && !c.first_name ? (
+                        <Typography.Text>Project Lead has not been assigned</Typography.Text>
+                    ) : (
+                        <>
+                            {name.length > 0 && <><Typography.Text className="taratest1">{name ?? Strings.EMPTY_FIELD}</Typography.Text>
```
simensma-fresh: className can probably be removed (or renamed) 😅
Same for the phone/email paragraphs as well

simensma-fresh (changes requested):

services/common/src/components/projects/ProjectContacts.tsx
taraepp: whoooops yep forgot that in there. I'll deal with it after meetings. 😅 
...I was having a hard time figuring out why I had a big old gap between some of the line items!

asinn134 (approved):

simensma-fresh (approved):

PR #3659, opened 2025-08-27
[MDS-6596] Add ema application tab and banner to major mine application tab

asinn134:
## Objective 

[MDS-6596](https://bcmines.atlassian.net/browse/MDS-6596)

- Added Ema Applications tab for Core and Minespace
- Renamed Major Applications tab to Mines Act Application tab
- Added banner in Mines Act application tab page to notify user of any Ema Auths associated with the application

image

image

image


services/common/src/components/project/EmaApplicationsTab.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const dispatch = useAppDispatch();
+    const history = useHistory();
+    const [isLoaded, setIsLoaded] = useState(true);
+    const systemFlag = useAppSelector(getSystemFlag);
+    const { isFeatureEnabled } = useFeatureFlag();
+    const amsFinalAppEnabled = isFeatureEnabled(Feature.AMS_FINAL_APPLICATION);
+    const isCore = systemFlag === SystemFlagEnum.core;
+    const project = useAppSelector(getProject);
+    const authorizations = project?.project_summary?.authorizations;
+    const hasMinesActApp = authorizations?.some(auth => !auth.ams_tracking_number);
```
taraepp: I will say this is not quite an accurate way of finding this. An AMS auth that hasn't been submitted yet won't have an ams tracking number either. The authorization type is more reliable.

asinn134: This is adjusted now to use the type instead.

services/common/src/components/project/EmaApplicationsTab.tsx
```diff
@@ -0,0 +1,194 @@
...
+                                                </span>
+                                            }
+                                            size="middle"
+                                            column={2}
+                                        >
+                                            <Descriptions.Item label="Type">{AMS_AUTHORIZATION_TYPES_TEXT[item.project_summary_permit_type[0]]}</Descriptions.Item>
+                                            <Descriptions.Item label="" style={{ float: "right", marginLeft: "auto" }}>
+                                                <Badge
+                                                    status={AMS_STATUS_CODES_SUCCESS as PresetStatusColorType}
+                                                    text={AMS_APPROVED_STATUSES[item.status]}
```
taraepp: the AMS_STATUS_CODES_SUCCESS is something that irritated me a bit! It will _only_ show success/green. And similarly, the AMS_APPROVED_STATUSES will only give the text for Approved/Completed/New.
- I did add a "AMS_STATUSES_TYPES" that'll give you the success/warning/error/default for all possible statuses that you can use instead of AMS_STATUS_CODES_SUCCESS or doing checks.
- As for the text, if you look at AMS_<APPROVED/WARNING/STOPPED>_STATUSES, every single key is exactly the same as the value it returns, with the sole exception of Rejected/AMS Rejected, so you can just show `item.status`. I am _pretty sure_ that when you look at ProjectDescriptionTab that "Rejected" is something that we're creating ourselves, not returned from AMS.
- But you might want to look at `ProjectDescriptionTab->processEnvironmentalActAuthorization`, or I tried to simplify it in `ProjectOverviewTab->amsAuthTableData`

services/core-web/src/components/mine/Projects/MajorMineApplicationTab.tsx
```diff
@@ -38,6 +38,8 @@ const MajorMineApplicationTab: FC = () => {
   const mineDocuments = useSelector(getMineDocuments);
   const dropdownMajorMineAppStatusCodes = useSelector(getDropdownMajorMinesApplicationStatusCodes);
   const archivedDocuments = mineDocuments?.map((doc) => new MajorMineApplicationDocument(doc))
+  const authorizations = project?.project_summary?.authorizations;
+  const hasEmaApp = authorizations?.some(auth => auth.ams_tracking_number);
```
taraepp, line 42: I think there's a few instances here of `hasEmaApp` in different files. Unless you are only looking at submitted to AMS applications, definitely use the authorization type instead.
Probably a good idea to make a draft project summary with some EMA authorizations, don't submit it, and make sure that everything functions as expected and doesn't error out.

asinn134: I think we are suppose to show only the successfully created ema auths? The ticket mentions this

"Can we only show the list of EMA authorizations that were successfully created in the project description stage"

and I interpreted successfully created EMA authorizations as successfully submitted to AMS.

simensma-fresh (approved):

sggerard (approved):

PR #3660, opened 2025-08-27
[MDS-6580] Add variable condition box

sggerard:
## Objective 

[MDS-6580](https://bcmines.atlassian.net/browse/MDS-6580)

Add variable condition box to PermitConditionViewEdit and only show it when editing NOW condition and StandardPermitConditions. 
Have renamed the old menu to VariableConditionMenuOld.js
Added back highlighting.
Inserts the variable at the mouse cursor position

Please pull the branch and have a play around let me know your thoughts!

 
image

image

image


services/common/src/components/permits/VariableConditionMenu.tsx
```diff
@@ -0,0 +1,167 @@
...
+    }
+    const isPreambleText = isEmpty(conditionFormValues);
+    const text = isPreambleText ? generatePermitFormValues.preamble_text ?? "" : conditionFormValues.condition ?? "";
+    const pos = inputRef.current?.resizableTextArea?.textArea?.selectionStart ?? text.length;
+    const newText = `${text.slice(0, pos)} ${event.key} ${text.slice(pos)}`.replace(/ {2,}/g, ' ');
+
+    if (!isPreambleText) {
+      await dispatch(change(conditionForm, "condition", newText));
+    } else {
+      await dispatch(change(FORM.GENERATE_PERMIT, "preamble_text", newText));
```
taraepp, line 45: I would be tempted to pass in formName/fieldName if there's no other differences between the two cases (and especially if the typing isn't all that important).

sggerard: isPreambleText is also used on line 38 for fetching the text

taraepp (approved):

matbusby-fw (approved):
Nice one!  Played with it a bit, and it looks like it's working great!

PR #3666, opened 2025-09-10
[MDS-6580] Bug fix, variable condition menu showing in incorrect places

sggerard:
## Objective 

[MDS-6580](https://bcmines.atlassian.net/browse/MDS-6580)

Bug fix, don't show Variable Condition menu unless it's on the Now editor or StandardConditionEditor.

Open to renaming the stepEditDisabled variable to something better!

### DIFF
--- a/services/common/src/components/permits/PermitConditionForm.tsx
+++ b/services/common/src/components/permits/PermitConditionForm.tsx
@@ -322,7 +322,7 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
                 </Col>
                 <Col className="condition-column" {...editableProps}>
                     <Col className="condition-editor">
-                        <VariableConditionMenu inputRef={conditionInputRef} isManagementView={Boolean(standardConditionType)} conditionForm={editingFormName}/>
+                        {stepEditDisabled && <VariableConditionMenu inputRef={conditionInputRef} isManagementView={Boolean(standardConditionType)} conditionForm={editingFormName}/>}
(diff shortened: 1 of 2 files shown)


taraepp (dismissed):
haha stepEditDisabled was created to determine that editing the step was disabled, which I guess follows the same logic but doesn't make as much sense here. I suppose you _could_ create another variable like `const hideVariableMenu = stepEditDisabled` or something like that, idk

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -322,7 +322,7 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
                 </Col>
                 <Col className="condition-column" {...editableProps}>
                     <Col className="condition-editor">
-                        <VariableConditionMenu inputRef={conditionInputRef} isManagementView={Boolean(standardConditionType)} conditionForm={editingFormName}/>
+                        {stepEditDisabled && <VariableConditionMenu inputRef={conditionInputRef} isManagementView={Boolean(standardConditionType)} conditionForm={editingFormName}/>}
```
matbusby-fw: Yeah, I kind of would prefer to rename this to something like 'showVariableConditionMenu'. Even if it means creating a second identical boolean by a different name.  This feels a bit confusing here.

matbusby-fw (approved):
Woo!

taraepp (approved):

PR #3674, opened 2025-09-16
[MDS-6657] Fix bug where variable inserted into the wrong form

sggerard:
## Objective 

[MDS-6657](https://bcmines.atlassian.net/browse/MDS-6657)

Fixed a bug spotted by Rebecca where if you are editing the preamble and a condition at the same time the menu may insert a variable meant for the preamble into a condition. 

Replaced using the currently editing form variable with FORM.GENERATE_PERMIT for the preamble to ensure it always goes to the correct place. Improved some of the logic in VariableConditionMenu too.


matbusby-fw (dismissed):
👍🏼

asinn134 (dismissed):

matbusby-fw (approved):

asinn134 (approved):

PR #3676, opened 2025-09-18
[MDS-6659] Fix keycloak error on hot reload

sggerard:
## Objective 

[MDS-6659](https://bcmines.atlassian.net/browse/MDS-6659)

Keycloak throws an error when a webpack rebuild/ hot reload happens in development. Removing the singleton appears to stop this error being thrown, however there was an original reason we added this which I have been unable to replicate so this requires some testing before merging! (Maybe apply this change locally and leave it for a bit)

The alternative fix will be to attach the singleton to the window.

### DIFF
--- a/services/common/src/keycloak.tsx
+++ b/services/common/src/keycloak.tsx
@@ -1,12 +1,7 @@
 import Keycloak from "keycloak-js";
 import { KEYCLOAK } from "@mds/common/constants/environment";
 
-let kc: Keycloak;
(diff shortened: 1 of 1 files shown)


matbusby-fw (approved):
Played with it for a bit.  I haven't been able to get it to throw the keycloak error, and I haven't found any other breaking changes.  Love it!

simensma-fresh (approved):

PR #3680, opened 2025-09-22
[MDS-6651] Ams Design

taraepp:
## Objective 
- make it super obvious how the stages of a project are divided
   - make sections collapsible
   - put numbers on them
   - some borders in there too 
- fix some issues with buttons/actions buttons
   - I maybe should have added the "button-col" to CoreTableCommonColumns (like with renderTextColumn etc)... 
- plus some sneaky table changes
   - some just to make Core/MS have same styles: Core had text wrapping onto a new line but MS didn't so I copied that to MS, MS had a really tall line-height and I took that out
   - and then reduced padding, made headers not bold, but indicated sorted column with a bold header 
- and a quick typo fix

[MDS-6651](https://bcmines.atlassian.net/browse/MDS-6651)
MineSpace:
image
(I tweaked the line-height after taking this screenshot, so the space in the "Permit Type" column now matches Core)
image
(General table changes)

Core:
image
(Straight up "violet" on the collapse bar looked _awful_ so picked a different variant. Basically same as MS)
image
(General table. Aside from spacing just changed the sorted column header BG colour because it was blending badly with its neighbours)

### DIFF
--- a/.github/workflows/common-package.unit.yaml
+++ b/.github/workflows/common-package.unit.yaml
@@ -55,6 +55,6 @@ jobs:
             services/common/coverage
             services/common/reports/test-reporter.xml
       - name: Sonar Reporting
-        uses: SonarSource/sonarcloud-github-action@v2.1.1
+        uses: SonarSource/sonarqube-scan-action@v6.0.0
         with:
           projectBaseDir: services/common
--- a/services/minespace-web/src/styles/settings/variables.scss
+++ b/services/minespace-web/src/styles/settings/variables.scss
@@ -28,6 +28,7 @@ $light-grey: #ded9d9;
 $lightest-grey: #f4f0f0;
 $lightest-grey-alt: #f8f7f7;
 $light-grey-cool: #F7F8FA;
+$light-mid-grey: #e6e8ec;
 $med-grey: #bcbec5;
 $white: #fff;
 $off-white: #fcf8f8;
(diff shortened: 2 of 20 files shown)


.github/workflows/common-package.unit.yaml
```diff
@@ -55,6 +55,6 @@ jobs:
             services/common/coverage
             services/common/reports/test-reporter.xml
       - name: Sonar Reporting
-        uses: SonarSource/sonarcloud-github-action@v2.1.1
+        uses: SonarSource/sonarqube-scan-action@v6.0.0
```
taraepp, line 58: For future reference/git-blaming: old action was running into Node.js error and going to its page, said it was deprecated and to use sonarqube-scan-action instead. 6.0.0 is the current newest version.

services/minespace-web/src/styles/settings/variables.scss
```diff
@@ -28,6 +28,7 @@ $light-grey: #ded9d9;
 $lightest-grey: #f4f0f0;
 $lightest-grey-alt: #f8f7f7;
 $light-grey-cool: #F7F8FA;
+$light-mid-grey: #e6e8ec;
```
taraepp, line 31: seriously running out of names to call new shades of grey. 😂

matbusby-fw (approved):
Looks Great!

simensma-fresh (approved):
🖌️

PR #3681, opened 2025-09-23
[FIX] untp endpoint - param fix, mine_Type null check

Jsyro:
This is feature flagged: Not sure if this matches the UNTP construction exactly, but it should work. 

This isn't: Add orgbook entity to party if it exists.


simensma-fresh (approved):

matbusby-fw (approved):

taraepp:
Just a note on the failing test: I ran into the same thing yesterday. The PR is still open but made a fix for it. You can check it out here if you want: https://github.com/bcgov/mds/pull/3680/commits/6dc2dac2ca9ad6bd0e2c71f21b6ec43648804569

PR #3682, opened 2025-09-23
[MDS-6633] Adding final app minespace edit toggle control in core

asinn134:
## Objective 

[MDS-6633](https://bcmines.atlassian.net/browse/MDS-6633)

- Added in Core a button to toggle the MineSpace user's ability to edit a final application.
- Toggles for individual EMA final application
- Final Application submission form changed to view mode when application not editable
- Final application documents in the **All Documents** tab's **Env Application** section changed not to be replaceable or archive-able when not editable.
- Added backend check to see if in MineSpace a final application is editable during the new final application submission form and during the action of archiving and replacing files

image

### DIFF
--- a/services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
+++ b/services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
@@ -4,7 +4,7 @@ from werkzeug.exceptions import BadRequest, NotFound
 from app.api.utils.resources_mixins import UserMixin
 
 from app.extensions import api
-from app.api.utils.access_decorators import MINESPACE_PROPONENT, requires_any_of, VIEW_ALL, MINE_ADMIN, EDIT_MAJOR_MINE_APPLICATIONS
+from app.api.utils.access_decorators import MINESPACE_PROPONENT, requires_any_of, VIEW_ALL, MINE_ADMIN, EDIT_MAJOR_MINE_APPLICATIONS, is_minespace_user
 from app.api.projects.response_models import AMS_FINAL_APPLICATION_MODEL
 from app.api.projects.project_summary.models.project_summary import ProjectSummary
 from app.api.projects.project_summary.models.project_summary_authorization import ProjectSummaryAuthorization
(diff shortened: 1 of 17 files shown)


services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
```diff
@@ -79,7 +82,27 @@ def put(self, project_summary_guid, project_summary_authorization_guid):
         final_app = final_app.update(submitter_name, documents, is_agent, pre_submitted_files, is_submitting)
         return final_app, 200
 
-    
+class AmsFinalApplicationMineSpaceEditResource(Resource, UserMixin):
```
taraepp, line 109: I don't think this functionality actually requires a separate endpoint/update method/etc, and could just be in the existing put method.

matbusby-fw (dismissed):

services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
```diff
@@ -1,18 +1,36 @@
...
-from app.api.projects.response_models import AMS_FINAL_APPLICATION_MODEL
-from app.api.projects.project_summary.models.project_summary import ProjectSummary
-from app.api.projects.project_summary.models.project_summary_authorization import ProjectSummaryAuthorization
-from app.api.projects.ams_final_application.models.ams_final_application import AmsFinalApplication
-from app.api.mines.documents.models.mine_document import MineDocument
-from app.api.mines.mine.models.mine import Mine
-from app.api.utils.custom_reqparser import CustomReqparser
-from app.api.services.document_manager_service import DocumentManagerService
+
+def check_mine_access(mine_guid):
```
simensma-fresh, line 28: Added an explicit Mine Check on all the AMS final app endpoints to force the minespace subscription check, seeing as the mine is not looked up anywhere else

services/core-api/app/api/projects/ams_final_application/resources/ams_final_application_resource.py
simensma-fresh: Decided It kind off makes sense as a separate endpoint (toggling of whether it can be edited being separate from actually editing the record), so OK with this

matbusby-fw (approved):

taraepp (approved):

PR #3693, opened 2025-10-06
[MDS-6635] Email Templates

taraepp:
## Objective 
- use templates for consistency, less code duplication
- converted template emails (but not non-template emails) to use the same format
- endpoint that FE can hit that gives a quick preview of these emails

- did not try to make the usage of those templates perfectly consistent
   - they aren't all the same with text alignment, titles, margin, etc
   - but without knowing exactly what the standard that we want to be is, didn't want to guess (also out of scope)
   - that can be a next step 
[MDS-6635](https://bcmines.atlassian.net/browse/MDS-6635)

Here's a recording of the preview endpoint (put on My Dashboard page on Core) and showing what the emails look like.
test-emails


services/core-api/app/api/email_preview/resources/email_preview_resource.py
```diff
@@ -0,0 +1,230 @@
...
+from app.api.utils.access_decorators import VIEW_ALL, MINESPACE_PROPONENT
+from app.api.constants import CORE_PURPLE_LOGO_BASE64_ENCODED, MINESPACE_LOGO_BASE64_ENCODED, BC_GOV_LOGO_BASE64_ENCODED
+
+
+class EmailPreviewResource(Resource, UserMixin):
+    """
+    Email Preview API endpoint for development
+    """
+    
+    @requires_any_of([VIEW_ALL, MINESPACE_PROPONENT])
```
simensma-fresh: Would suggest locking this down to just core users (maybe the `core_admin` role)

services/core-api/app/api/email_preview/resources/email_preview_resource.py
```diff
@@ -0,0 +1,230 @@
...
+    def get(self, template_name):
+        """
+        Preview an email template with test data
+        GET /api/email-preview/<template_name>
+        """
+        try:
+            # Get template with proper path
+            template_path = f"email/{template_name}"
+            template = current_app.jinja_env.get_template(template_path)
```
simensma-fresh, line 52: I'm wondering if this could be vulnerable for a path traversal attack (depending on how `get_template` implements it) - `f"email/.../.../.../etc/passwd`

Maybe we can define all the templates (and mapping to what file) as constants and use that as validation, think that would clean up the sending functionality a bit as well:

`EmailService.send_template_email(subject, recipients, EMAILS.INCIDENTS_RECEIVED_MS, email_context, cc=None)`

services/core-api/app/api/email_preview/resources/email_preview_resource.py
```diff
@@ -0,0 +1,230 @@
...
+    def get(self):
+        """
+        List all available email templates for preview
+        GET /api/email-preview/
+        """
+        try:
+            import os
+            from pathlib import Path
+            
+            templates_dir = Path(current_app.root_path) / 'templates' / 'email'
```
simensma-fresh: Would suggest using `os.path.join` instead of `/` here (in case anyone decides to try this on windows 🪟)

simensma-fresh (changes requested):

services/core-api/app/api/email_preview/resources/email_preview_resource.py
taraepp: took a slightly different strategy than named constants but should work effectively the same (compares against a query of existing templates)

matbusby-fw (approved):
Looks really good!  I think I'll need to do some merge conflict resolving in the email_service, but nothing crazy.  Haven't worked with Jinja before, but I think I'm getting the patterns now. Nice!

simensma-fresh (approved):

PR #3695, opened 2025-10-06
[FIX] Added missing resources to auth unit test

simensma-fresh:
## Objective 

Found that a lot of resources were missing from the `test_expected_auth` backend test. This pr adds the missing ones, and adds a test to verify that all Flask resources are covered.
Added in a test to verify that all endpoints either have a `requires` role decorator, or a `public_endpoint` decorator (for sanity sake)

image


matbusby-fw (dismissed):
Oooh, nice update!

matbusby-fw (approved):

alazar-aot:
## Updates Sept 2026:
This PR has been open for nearly a year, and in that time `develop` saw 224 commits missing from this branch, so it needed some changes to reconcile drift before it could be reviewed and merged

**Changes Made:**
- Merged in `develop` and resolved the 4 resulting conflicts
- Updated the auth coverage table for things that changed on develop since this PR opened: a couple of resources were renamed/removed (in MDS-6974), and additionally some endpoints were added without being included in the test table (`test_expected_auth.py`), and two entries had gone stale.
- I noticed that 4 new Minespace endpoints (related to signup) and one document-upload-status endpoint were missing role decorators entirely, and their uses cases didn't fit out existing patterns:
    - The 4 signup endpoints (introduced in MDS-6681) need a logged-in user but no specific role (expected for a brand-new user with only login auth) - so I created a new `@requires_authentication` decorator for exactly this case (valid token required, no role check), rather than exempting them from the test.
    - `DocumentUploadStatusResource.get` was made intentionally public in MDS-6681, so I've marked it a`@public_endpoint` so the test reflects reality.

PR #3696, opened 2025-10-07
[MDS-6569] Hide template conditions from extracted permits

simensma-fresh:
## Objective 

[MDS-6569](https://bcmines.atlassian.net/browse/MDS-5659)

Only show insert template condition button for NoW permits


matbusby-fw (approved):
Ah, good catch!

PR #3700, opened 2025-10-09
[MDS-6632] - Email Tracking

matbusby-fw:
## Objective 

Implement the ability to track the status of emails that have sent through CHES. 

- Added a new table `email_tracking`
  - Created as a generic so this could be used by any other existing entity and referenced
  - Added Tracking for the status of the send, various timestamps and tracking ids
  - Tracks the template name if this is a template email and a reference_email_type for non templates
- Added a celery task that will poll CHES for the status of email messages
- Integrated the email_tracking into the email_service if a track_email boolean is passed when initiating a send
- Updated documents flow for major applications so that it will only notify Minespace users of document updates once every 24 hours.

[MDS-6632](https://bcmines.atlassian.net/browse/MDS-6632)

### DIFF
--- a/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
+++ b/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
@@ -16,10 +16,10 @@ CREATE TYPE recipient_type_enum AS ENUM (
 
 -- Create generic email tracking table for all entities
 CREATE TABLE email_tracking (
-                                email_tracking_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
+                                email_tracking_guid uuid DEFAULT gen_random_uuid() PRIMARY KEY,
 
     -- Generic reference to any entity this email relates to
-                                reference_id uuid NOT NULL,
+                                reference_id varchar NOT NULL,
                                 reference_table varchar(100) NOT NULL,
                                 reference_email_type varchar(100),
 
@@ -57,7 +57,7 @@ CREATE TABLE email_tracking (
 
 -- Add comments for documentation
 COMMENT ON TABLE email_tracking IS 'Generic email tracking table for all entities using CHES service';
-COMMENT ON COLUMN email_tracking.reference_id IS 'UUID of the entity this email relates to (can reference any table)';
+COMMENT ON COLUMN email_tracking.reference_id IS 'Primary key of the entity this email relates to (can reference any table)';
 COMMENT ON COLUMN email_tracking.reference_table IS 'Name of the table that the reference_id points to';
 COMMENT ON COLUMN email_tracking.reference_email_type IS 'Type of email this is (e.g. documents uploaded, submission confirmation, etc.).  Needed for non template emails to differentiate between email types';
 COMMENT ON COLUMN email_tracking.email_template_name IS 'Name/identifier of the email template used';
--- a/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
+++ b/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
@@ -3,7 +3,7 @@ CREATE TYPE email_status_enum AS ENUM (
     'sent',
     'accepted',
     'cancelled',
-    'completed'
+    'completed',
     'failed',
     'pending'
 );
--- a/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
+++ b/migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
@@ -1,0 +1,81 @@
+-- Create enum types for email tracking
+CREATE TYPE email_status_enum AS ENUM (
+    'sent',
+    'accepted',
+    'cancelled',
+    'completed'
+    'failed',
+    'pending'
+);
+
+CREATE TYPE recipient_type_enum AS ENUM (
+    'primary',
+    'cc',
+    'bcc'
+);
+
+-- Create generic email tracking table for all entities
+CREATE TABLE email_tracking (
+                                email_tracking_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
+
+    -- Generic reference to any entity this email relates to
+                                reference_id uuid NOT NULL,
+                                reference_table varchar(100) NOT NULL,
+                                reference_email_type varchar(100),
+
+    -- Email content and template information
+                                email_template_name varchar(255),
+                                email_subject varchar(500),
+
+    -- Recipient information
+                                recipient_email varchar(320) NOT NULL, -- RFC 5321 max email length
+                                recipient_name varchar(255),
+                                recipient_type recipient_type_enum NOT NULL DEFAULT 'primary',
+
+    -- Email status tracking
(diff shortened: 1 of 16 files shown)


services/core-api/app/api/email_tracking/email_status_tasks.py
```diff
@@ -0,0 +1,113 @@
...
+            # Don't retry for 404s, but don't mark as failed either
+            return status_result
+
+        if status_result["status"] == "unknown_status":
+            # Don't update for unknown statuses
+            return status_result
+
+        if status_result["status"] == "error":
+            # Will be handled by the exception block for retries
+            raise Exception(status_result["message"])
```
taraepp, line 48: might be a nice place for a console log of the whole `status_result` so we can get more context from logs if things fail (probably won't happen very often... I don't think I hear of emails just failing)

migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
```diff
@@ -0,0 +1,84 @@
...
+
+CREATE TYPE recipient_type_enum AS ENUM (
+    'primary',
+    'cc',
+    'bcc'
+);
+
+-- Create generic email tracking table for all entities
+CREATE TABLE email_tracking (
+                                email_tracking_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
```
taraepp: Generally I assume that a column name ending in `_id` is an integer and expect `_guid` for uuid. 🤔 
If it's not tons of work, I'd consider making the names consistent with usage patterns.

services/core-api/tests/email_tracking/test_email_tracking.py
```diff
@@ -0,0 +1,239 @@
+import uuid
+from datetime import datetime, timedelta
+from pytz import utc
+
+from app.api.email_tracking.models.email_tracking import EmailTracking, EmailStatus, RecipientType
+from tests.factories import EmailTrackingFactory, MajorMineApplicationFactory, ProjectSummaryFactory
+
+
+def test_email_tracking_create(db_session):
+    """Test creating an EmailTracking record"""
```
taraepp: this is a personal preference thing- copilot looooooves to add doc strings to tests and I strongly dislike them here. 😆 
It makes it output the docstring instead of the test name in the test output, and it is harder to search doc strings than test names (can't double-click them to select the whole name and then search up the test)

services/core-api/app/api/projects/information_requirements_table/models/information_requirements_table.py
```diff
@@ -128,9 +136,27 @@ def send_irt_status_notifications(self, project_guid: str, status_code: str):
...
-            EmailService.send_template_email(subject, core_recipients, core_template, project_context)
+            EmailService.send_template_email(
+                subject,
+                core_recipients,
+                core_template,
+                project_context,
+                reference_id=self.irt_guid,
+                reference_table='information_requirements_table',
+                email_template_name='ministry_irt_status_notification',
+                create_tracking_record=True
```
taraepp: I wonder if the template and "email_template_name" can be combined as a parameter (split out the name between / and .), and just default create_tracking_record=True (or- I wonder if the presence of reference_id/reference_table is an indication that the email should be tracked? I guess I'm only seeing changes to calls where the email is being tracked so it looks like we want it to be true)

taraepp:
Overall it looks good to me! I have some nitpicks, which I suspect are more directed at copilot.

services/core-api/app/api/email_tracking/email_status_tasks.py
```diff
@@ -0,0 +1,113 @@
+from datetime import datetime, timezone
+from flask import current_app
+from celery import Task
+
+from app.api.email_tracking.models.email_tracking import EmailTracking, EmailStatus
+from app.api.services.email_service import EmailService
+from app.tasks.celery import celery
+
+
+class EmailTaskBase(Task):
```
simensma-fresh: I'm wondering if we could move this into a shared component, seeing as we've got the exact same logic for the `PermitConditionTaskBase`

services/core-api/app/api/email_tracking/email_status_tasks.py
```diff
@@ -0,0 +1,113 @@
...
+                    # CHES accepted the message
+                    timestamp_field = 'sent_timestamp'
+
+                tracking_record.update_status(new_status, timestamp_field=timestamp_field)
+
+        # Schedule next poll if email is not in final state
+        if new_status in [EmailStatus.pending, EmailStatus.accepted, EmailStatus.sent]:
+            # Schedule next poll in 5 minutes
+            current_app.logger.info(f"Scheduling next poll for email {ches_message_id} in 5 minutes")
+            poll_ches_email_status.apply_async(args=[ches_message_id], countdown=300)
```
simensma-fresh: I'm wondering if it would be better to do a `poll_ches_email_status.retry` instead of apply_async to make it count against` max_retries, so we don't get into an infinte loop here if the status in ches for some reason never updates.

services/core-api/app/api/services/email_service.py
```diff
@@ -127,7 +258,13 @@ def send_email(cls,
...
                    priority=EmailPriority.NORMAL.value,
                    tag=None,
-                   send_to_proponent=False):
+                   send_to_proponent=False,
+                   # Email tracking parameters
+                   reference_id=None,
+                   reference_table=None,
+                   email_template_name=None,
+                   reference_email_type=None,
+                   create_tracking_record=False):
```
simensma-fresh: I would suggest maybe lets turn on tracking as default, unless you see a reason to not track any emails

simensma-fresh (changes requested):

services/core-api/app/api/email_tracking/email_status_tasks.py
matbusby-fw: Sounds good!

migrations/sql/V2025.09.23.09.45__create_email_tracking_table.sql
matbusby-fw: Good point.  Mostly just a find and replace.  Shouldn't be a problem.  
For the reference table id, it could be either since it depends on what that table would have as a primary key, so I'm actually thinking I should change it to a varchar.
I want to keep the naming convention the same for the 'ids' coming from CHES since that's what they're called coming from them even though they're guids.

services/core-api/tests/email_tracking/test_email_tracking.py
matbusby-fw: Ah, yeah, that's totally a copilot thing.  Happy to get rid of that.

services/core-api/app/api/email_tracking/email_status_tasks.py
matbusby-fw: Good thought.  Will do.

services/core-api/app/api/email_tracking/email_status_tasks.py
matbusby-fw: Sure, that makes good sense

taraepp (approved):

simensma-fresh (approved):

PR #3706, opened 2025-10-16
[MDS-6629] Mines Act Application uploaders

taraepp:
## Objective 
- basically add a subcategory to mines act applications

## KNOWN ISSUE
- gosh I really don't like how we have those xref objects attached to mine documents that sometimes has category info, but also sometimes it's somewhere else completely different (without a consistent name!)
- but it felt seriously out of scope to fix it
- Suggestion for the future: rework mine document BE & FE
   - all those `object_xref` with `object_type_code` properties? That only show up for archived documents? _category_code_. Put it in a hybrid property. Return a simpler object. Have another one that indicates the object type if necessary.
   - and then same for returning major mine application documents- use those same property names so it's like `object_code: MMA, category_code: APX, subtype_code: BIF`
   - so we can treat all documents the same, but also know what kind of document it is when we need to filter
   - I wrote some really gnarly code and I was not happy about it. It works, but...

[MDS-6629](https://bcmines.atlassian.net/browse/MDS-6629)


migrations/sql/V2025.10.08.01.47__project_application_document_types.sql
```diff
@@ -0,0 +1,48 @@
+INSERT INTO major_mine_application_document_type 
+    (major_mine_application_document_type_code, description, display_order, create_user, update_user) 
+VALUES ('APX', 'Appendeix', 15, 'system-mds', 'system-mds');
```
simensma-fresh: Typo? `Appendeix`

migrations/sql/V2025.10.08.01.47__project_application_document_types.sql
```diff
@@ -0,0 +1,48 @@
...
+         ('MIP', 'APX', 'Mine Plan', 30, 'system-mds', 'system-mds'),
+         ('RCP', 'APX', 'Reclamation and Closure Plan', 40, 'system-mds', 'system-mds'),
+         ('WQM', 'APX', 'Water Quality Mitigation and Water Modeling', 50, 'system-mds', 'system-mds'),
+         ('EDE', 'APX', 'Effluent Discharges to the Environment', 60, 'system-mds', 'system-mds'),
+         ('EEA', 'APX', 'Environmental Effects Assessment', 70, 'system-mds', 'system-mds'),
+         ('EMO', 'APX', 'Environmental Monitoring', 80, 'system-mds', 'system-mds'),
+         ('MGP', 'APX', 'Management Plans', 90, 'system-mds', 'system-mds'),
+         ('RLE', 'APX', 'Reclamation Liability Cost Estimate', 100, 'system-mds', 'system-mds'),
+         
+         ('QPD', 'SPR', 'Qualified Professional Declaration Form', 10, 'system-mds', 'system-mds'),
```
simensma-fresh: Think the display order should probably continue on to 110 / 120 etc instead of resetting, or maybe even just have it as 1,2,3,4,5,6 etc

simensma-fresh (changes requested):

migrations/sql/V2025.10.08.01.47__project_application_document_types.sql
taraepp: I'll bump them up. They're not actually used anywhere (I was going with an existing pattern... which also doesn't actually get used) but I like being able to insert in between values.

services/common/src/components/projects/ProjectDocumentsTab.tsx
```diff
@@ -135,14 +135,16 @@ const ProjectDocumentsTab: FC<ProjectDocumentsTabProps> = ({ project }) => {
...
+        new MajorMineApplicationDocument({
           ...d,
           category: CATEGORY_CODE[d.major_mine_application_document_type_code],
         })
     ) ?? [];
 
   const primaryDocuments = majorMineAppDocs.filter((doc) => doc.category === CATEGORY_CODE.PRM);
+  const mmaAppendixDocuments = majorMineAppDocs.filter((doc) => doc.category === CATEGORY_CODE.APX);
   const mmaSpatialDocuments = majorMineAppDocs.filter((doc) => doc.category === CATEGORY_CODE.SPT);
+  console.log({ documents: project?.major_mine_application?.documents, mmaAppendixDocuments })
```
matbusby-fw: console.log left in here

matbusby-fw (approved):

simensma-fresh (approved):

PR #3717, opened 2025-10-28
[MDS-6679] Show MS users who have access to a mine

taraepp:
## Objective 
- on Core & MS, show a list of MS users who have access to the mine
- replaced the `minespaceReducer` with redux slice, updated imports, usages (manually tested each function)
- as a little bonus, when a user or ministry contact is created/updated/deleted, they are now getting added to the state, so it is unnecessary to fetch users again
- and as another bonus, a new user/ministry contact is added to the top of the list, so admins can find them immediately.
- made MineNavigation TS, and switched to newer Menu usage
- didn't TS the user management pages, but I did check for unused props being passed to modals/subcomponents and removed those

[MDS-6679](https://bcmines.atlassian.net/browse/MDS-6679)

MineSpace:
image
Core:
image


services/core-api/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -20,14 +21,26 @@ class MinespaceUserListResource(Resource, UserMixin):
...
+        is_admin = getJwtManager().contains_role([MINE_ADMIN])
+
+        if not is_admin and mine_guid is None:
+            raise BadRequest("mine_guid is a required argument")
+        
         if request.args.get('email_or_username'):
             ms_users = [MinespaceUser.find_by_email(request.args.get('email_or_username'))]
-        else:
+        elif mine_guid:
+            mine = Mine.find_by_mine_guid(mine_guid)
```
matbusby-fw, line 39: It might be good to add a check to make sure the user fetching the list is included in the returned list (if they aren't an admin)

matbusby-fw (changes requested):
Looks great.  Just add a suggestion to add a check in the BE.

services/core-api/app/api/users/minespace/resources/minespace_user.py
taraepp: that's the real reason for the `mine = Mine.find_by_mine_guid`! It'll make sure they have access.

matbusby-fw (approved):

PR #3723, opened 2025-11-03
[MDS-6706] Restore route for Mine Permits in MineDashboardRoutes

matbusby-fw:
## Objective 

The route to the permits page in core was accidentally removed when adding a new route.
This just restores it.

[MDS-6706](https://bcmines.atlassian.net/browse/MDS-6706)

image


taraepp (approved):
whoops, I think this was me. 😅 
Good job hunting it down so fast!

PR #3724, opened 2025-11-04
[MDS-6691] Save more data about MS users

taraepp:
## Objective 
- add more fields to DB table and FE, similar to core users
   - in addition, add versioning (like core users)
- every single user in prod has "username_or_email" ending in "@bceid", so migrated data accordingly (renamed the column to bceid_username)
- made some FE & BE updates to reflect that this was no longer a field that may contain an email address

[MDS-6691](https://bcmines.atlassian.net/browse/MDS-6691)

The User Access table (same on Core):
image

Admin page on Core (got bonus columns and quick TS update):
image


services/core-api/app/api/users/minespace/models/minespace_user.py
```diff
@@ -42,20 +51,51 @@ def find_by_mine_guid(cls, mine_guid):
...
+        
+        return user
+
+
+    @validates('bceid_username')
+    def validate_email(self, key, bceid_username):
+        if not bceid_username:
             raise AssertionError('Identifier is not provided.')
-        return email_or_username
+        if not bceid_username.endswith('@bceid'):
```
taraepp, line 98: added a BE error when there's a bad BCeID passed in from the FE. I didn't add validation on the FE, figuring that the form will be revamped.

services/core-api/app/api/users/models/user.py
```diff
@@ -1,10 +1,10 @@
 from datetime import datetime
 from pytz import utc
-from app.api.utils.models_mixins import SoftDeleteMixin, Base, AuditMixin
+from app.api.utils.models_mixins import SoftDeleteMixin, Base, AuditMixin, HistoryMixin
 from app.extensions import db
 
 
-class User(SoftDeleteMixin, AuditMixin, Base):
+class User(HistoryMixin, SoftDeleteMixin, AuditMixin, Base):
```
taraepp, line 7: was mostly experimenting when I was trying to figure out why versioning wasn't working! Anyway, this adds a history attribute that'll have nicely parsed data, so I left it in.

services/core-api/app/api/users/resources/user_resource.py
```diff
@@ -23,19 +24,49 @@ def get(self):
...
-                "display_name": user_info.get("display_name", ""),
-                "idir_username": user_info.get("idir_username", ""),
-                "identity_provider": user_info.get("identity_provider", ""),
-                "idir_user_guid": user_info.get("idir_user_guid", ""),
-                "last_logged_in": datetime.now(tz=utc),
-            }
-
-            user = User.create_or_update_user(**user_data)
+            if is_minespace_user():
+                # bceid given/family name may be combined
```
taraepp: my particular token data has the following:
given_name: "Tara Epp"
display_name: "Tara Epp"
family_name: ""

It's a pretty small sample size, but prod data can be checked later on to see how the parsing is going and possibly modify it from there. The display_name is what's used by the FE.

services/core-api/app/auth.py
```diff
@@ -36,18 +35,18 @@ def get_permission(self, mine_id: UUID):
 
 
 def get_mine_access():
+    from .api.users.minespace.models.minespace_user import MinespaceUser
```
taraepp, line 38: the imports from MinespaceUser before the app is initialized is what caused versioning to not work, so now they're loaded in the functions. There is a BE test added to catch the versioning breaking again.

services/core-api/app/api/users/resources/user_resource.py
```diff
@@ -23,19 +24,49 @@ def get(self):
...
+                family_name = user_info.get("family_name", "")
+                display_name = user_info.get("display_name", "")
+
+                bceid_username_data = user_info.get("bceid_username", None)
+                bceid_username = bceid_username_data + "@bceid" if bceid_username_data is not None else ""
+
+                if given_name == display_name and family_name == "":
+                    name = display_name.split()
+                    given_name = name[0] if len(name) > 0 else ""
+                    family_name = name[1] if len(name) > 1 else ""
```
simensma-fresh: I'm thinking we might just want to accept empty family name if not provided in the bceid mapping, as in some cases this wouldn't be correct (e.g. in my case!)

Given Name: Simen Fivelstad
Family Name: Smaaberg

services/core-api/app/api/users/resources/user_resource.py
```diff
@@ -23,19 +24,49 @@ def get(self):
...
+                    "family_name": family_name,
+                    "display_name": user_info.get("display_name", ""),
+                    "bceid_username": bceid_username,
+                    "identity_provider": user_info.get("identity_provider", ""),
+                    "bceid_user_guid": user_info.get("bceid_user_guid", ""),
+                    "last_logged_in": datetime.now(tz=utc),
+                }
+
+                user = MinespaceUser.update_from_token_data(**user_data)
+                print(user)
```
simensma-fresh: print

simensma-fresh (changes requested):

services/core-api/app/api/users/minespace/resources/minespace_user.py
```diff
@@ -33,8 +33,8 @@ def get(self):
         if not is_admin and mine_guid is None:
             raise BadRequest("mine_guid is a required argument")
         
-        if request.args.get('email_or_username'):
-            ms_users = [MinespaceUser.find_by_email(request.args.get('email_or_username'))]
+        if request.args.get('bceid_username'):
+            ms_users = [MinespaceUser.find_by_username(request.args.get('bceid_username'))]
```
simensma-fresh: Hmm something feels a bit odd with this logic here reading this whole function. I think we should probably also filter this by the mine_guid you pass in if you're a minespace proponent?

taraepp: this is some 5 year old code, which, based on the FE code, I think is actually just _dead_. There's no calls to the API matching this resource where we do a GET with anything other than mine_guid. I'll take it out.

asinn134 (dismissed):

services/core-api/app/api/users/resources/user_resource.py
taraepp: Hmm yeah. Yours would parse as given (the if would be false) but we may as well see how weird the data is before trying to clean it. Using display_name on the FE prevents the weirdness from showing up anywhere visible to any users.

services/core-api/app/api/users/minespace/models/minespace_user.py
```diff
@@ -42,20 +51,51 @@ def find_by_mine_guid(cls, mine_guid):
...
+        
+        for key, value in kwargs.items():
+            setattr(user, key, value)
+        user.save()
+        
+        return user
+
+
+    @validates('bceid_username')
+    def validate_email(self, key, bceid_username):
```
matbusby-fw: Just for consistency, can we rename this to validate bceid?

matbusby-fw (changes requested):
Looks great.  Just one small change request on a function name for consistency.

asinn134 (approved):

simensma-fresh (approved):

matbusby-fw (approved):

PR #3732, opened 2025-11-07
[FIX] Celery error. Also TS error

taraepp:
## Objective 
- include the task in the init.py so it gets registered
- also squash those TS errors
- looks like there were some snap updates with last updates but the checks didn't get triggered.  Snap updates here


simensma-fresh (dismissed):
👍

asinn134 (dismissed):

matbusby-fw (approved):
Nice!

PR #3740, opened 2025-11-17
[MDS-6696] - Add Party fix

matbusby-fw:
## Objective 

Another similar state issue when _**adding**_ a mine party appointment.  Added logic to explicitly pass in the variable being checked with fallbacks.

[MDS-6696](https://bcmines.atlassian.net/browse/MDS-6696)

### DIFF
--- a/services/core-web/src/components/mine/ContactInfo/ViewPartyRelationships.tsx
+++ b/services/core-web/src/components/mine/ContactInfo/ViewPartyRelationships.tsx
@@ -77,18 +77,10 @@ export const ViewPartyRelationships: FC<ViewPartyRelationshipsProps> = ({ mine }
     setUploadedFiles(uploadedFiles.filter((fileArr) => fileArr[0] !== fileItem.serverId));
   };
 
-  const onSubmitAddPartyRelationship = async (
-    values: Partial<IPartyAppt>,
-    explicitTypeCode?: MinePartyAppointmentTypeCodeEnum
-  ) => {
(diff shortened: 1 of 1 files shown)


services/core-web/src/components/mine/ContactInfo/ViewPartyRelationships.tsx
```diff
@@ -77,11 +77,18 @@ export const ViewPartyRelationships: FC<ViewPartyRelationshipsProps> = ({ mine }
...
   };
 
-  const onSubmitAddPartyRelationship = async (values: Partial<IPartyAppt>) => {
-    if (!selectedPartyRelationshipType) return;
+  const onSubmitAddPartyRelationship = async (
+    values: Partial<IPartyAppt>,
+    explicitTypeCode?: MinePartyAppointmentTypeCodeEnum
+  ) => {
+    const effectiveTypeCode =
+      explicitTypeCode ||
```
taraepp: Looks like the same value is being passed to the form and used in the initial values- is it not getting passed to the submit handler as values.mine_party_appt_type_code?

matbusby-fw: It's really wonky.  I was trying to avoid a complete redo of this file, but the way things are set up with refs and such I'm only 95% sure that all use cases would be covered by using that value.

taraepp (approved):

PR #3741, opened 2025-11-19
[MDS-6720] include non repeating reports in task

taraepp:
## Objective 
- _include_ reports with initial due date but no due_date_period_months
- _exclude_ reports where the permit amendment is not verified

[MDS-6720](https://bcmines.atlassian.net/browse/MDS-6720)


matbusby-fw (approved):
Approvinated!

PR #3747, opened 2025-11-27
[MDS-6731] Fixed logic to associate the correct NOW document to a condition when indexing for search

simensma-fresh:
## Objective 

[MDS-6731](https://bcmines.atlassian.net/browse/MDS-6731)

Currently, when indexing now permit conditions for search, we associate it with `amendment.now_application_documents[0].mine_document` which may not always be correct depending on the order of when the documents were generated/updated. This PR ensures we link the condition to the latest generated NoW application document ('PMT', 'PMA' codes), that are part of the issued permit (related_documents).

Also modified the bulk index command to add a `dry-run` option which lets you view the export as a csv so it can be verified before updating the search index

### DIFF
--- a/services/core-api/app/cli_commands/export_permit_conditions.py
+++ b/services/core-api/app/cli_commands/export_permit_conditions.py
@@ -7,7 +7,11 @@ from app.api.mines.permits.permit_amendment.models.permit_amendment import (
 from app.api.mines.permits.permit_conditions.models.permit_conditions import (
     PermitConditions,
 )
-from app.api.now_applications.now_template_transformer import calculate_liability, replace_condition_value_with_data, transform_variables_to_data
+from app.api.now_applications.now_template_transformer import (
+    calculate_liability,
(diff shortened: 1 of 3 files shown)


services/core-api/app/cli_commands/export_permit_conditions.py
```diff
@@ -18,6 +22,40 @@
...
+        return None
+
+    # If amendment has related documents, try to find a match
+    if amendment.related_documents:
+        related_guids = {doc.document_manager_guid for doc in amendment.related_documents}
+        
+        # Find NoW docs that are also in related_documents
+        issued_docs = [
+            doc for doc in now_permit_docs 
+            if doc.mine_document.document_manager_guid in related_guids
```
simensma-fresh, line 47: The `PermitAmendmentDocument` don't have a type unfortunately, so gotta do this janky thing to actually figure out if the document was an auto-generated permit document

taraepp (approved):

PR #3786, opened 2025-12-15
[MDS-6741] - View Mine Report Definition Button Does Nothing

matbusby-fw:
It seems we never implemented the view modal for viewing a mine report definition.  Converted the existing create modal to have an optional view only mode

## Objective 

[MDS-6741](https://bcmines.atlassian.net/browse/MDS-6741)

image

### DIFF
--- a/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
+++ b/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
@@ -118,7 +118,7 @@ const AddReportDefinitionForm: FC<{
 
       <div className="ant-modal-footer">
         <RenderCancelButton />
-        {isEditMode ? <RenderSubmitButton buttonText="Save Report" disableOnClean={false} /> : null}
+        <RenderSubmitButton buttonText="Save Report" disableOnClean={false} />
       </div>
     </FormWrapper>
   );
--- a/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
+++ b/services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { FC } from "react";
 import FormWrapper from "@mds/common/components/forms/FormWrapper";
 import { ADD_REPORT_DEFINITION } from "@/constants/forms";
 import { Col, Row, Typography } from "antd";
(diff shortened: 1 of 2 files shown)


services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
```diff
@@ -105,7 +118,7 @@ const AddReportDefinitionForm = ({ handleSubmit, isModal = false }) => {
 
       <div className="ant-modal-footer">
         <RenderCancelButton />
-        <RenderSubmitButton buttonText="Save Report" disableOnClean={false} />
+        {isEditMode ? <RenderSubmitButton buttonText="Save Report" disableOnClean={false} /> : null}
```
taraepp: I am 90% sure that the submit button will hide itself when not in edit mode! But I'm not too worried about it.

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
```diff
@@ -115,7 +115,32 @@ const ComplianceReportManagement: FC = () => {
   };
 
   const openViewModal = (record) => {
-    console.log("record", record);
+    const reportDefinition = reportDefinitions.find(
+      (r) => r.mine_report_definition_guid === record.mine_report_definition_guid
+    );
+    if (!reportDefinition) {
```
taraepp: I'm wondering if there is a scenario where reportDefinition isn't found. Like, the table data is the (transformed, but not filtered) reportDefinitions, and the record is part of that list.

services/core-web/src/components/admin/complianceCodes/AddReportDefinitionForm.tsx
matbusby-fw: I had actually added that at one point because certain props were missing in view only mode, but that's not a problem anymore so I can remove this.

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
matbusby-fw: Yes, I ran into some problems there testing locally which is why I added this, but I forgot to add an early return with the notification.  Added that now.

asinn134 (dismissed):

services/core-web/src/components/admin/complianceCodes/ComplianceReportManagement.tsx
taraepp: Hmm, I'm just looking at the transformData where the record is defined. I don't think it should be necessary to do reportDefinitions.find at all because of how transformData does a spread to copy the whole object into the table record. I wonder if there's something else strange going on here

taraepp (approved):

PR #3791, opened 2026-01-07
[MDS-6755] Allow submitted or assigned minespace projects to add files

asinn134:
## Objective 

[MDS-6755](https://bcmines.atlassian.net/browse/MDS-6755)

- Adjustment to allow submitted, and assigned minespace projects to add files.

### DIFF
--- a/services/common/src/components/projectSummary/ProjectSummaryForm.tsx
+++ b/services/common/src/components/projectSummary/ProjectSummaryForm.tsx
@@ -138,7 +138,7 @@ export const ProjectSummaryForm: FC<ProjectSummaryFormProps> = ({
     <SteppedForm
       name={FORM.ADD_EDIT_PROJECT_SUMMARY}
       initialValues={initialValues}
-      isEditMode={isEditMode}
+      isEditMode={isCore ? isEditMode : (isEditMode || !docFieldsDisabled)}
       handleSaveData={handleSaveData}
       handleTabChange={handleTabChange}
       transformPayload={handleTransformPayload}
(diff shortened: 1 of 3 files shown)


services/common/src/components/projectSummary/ProjectSummaryForm.tsx
```diff
@@ -138,7 +138,7 @@ export const ProjectSummaryForm: FC<ProjectSummaryFormProps> = ({
     <SteppedForm
       name={FORM.ADD_EDIT_PROJECT_SUMMARY}
       initialValues={initialValues}
-      isEditMode={isEditMode}
+      isEditMode={isCore ? isEditMode : (isEditMode || !docFieldsDisabled)}
```
taraepp, line 141: wait- would this enable editing _any_ fields in MS if `!docFieldsDisabled`?

asinn134: I found if `!docFieldsDisabled` is true it does enable edit mode but it doesn't necessarily enable all fields, because in `ProjectSummaryForm.tsx` the different types of fields being enabled are controlled separately (`areFieldsDisabled()`, `areDocumentFieldsDIsabled()`, etc) but dependent on the `status_code`. I found for `status_code` "submit", and "assigned" all the fields are disabled but the file upload fields are available. When the `status_code` is "draft" or "change requested" the input fields will be enabled including file upload ones, but I think this is what we want for "draft" and "change requested".

matbusby-fw (approved):

PR #3807, opened 2026-01-24
[MDS-6764] - NRIS duplicate entries

matbusby-fw:
Our NRIS api was creating multiple records from the raw data with new internal ids, but the same external one which shouldn't happen.  Updated it to ensure idempotency of the records.

## Objective 

[MDS-6764](https://bcmines.atlassian.net/browse/MDS-6764)

### DIFF
--- a/services/nris-api/backend/app/nris/etl/nris_etl.py
+++ b/services/nris-api/backend/app/nris/etl/nris_etl.py
@@ -90,6 +90,9 @@ def etl_nris_data():
 
     has_next_page = True
     i = 0
+    processed = 0
+    skipped_missing_status = 0
+    skipped_deleted = 0
 
     print('Parsing {} assessments'.format(nris_data.total))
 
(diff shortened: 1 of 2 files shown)


services/nris-api/backend/app/nris/etl/nris_etl.py
```diff
@@ -132,7 +145,12 @@ def _parse_nris_element(input):
         assessment_sub_type = data.find('assessment_sub_type')
         business_area = data.find('businessArea').find('business_area_name')
 
-        inspection = Inspection(external_id=_parse_element_text(assessment_id))
+        # Find or create the inspection
+        inspection = db.session.query(Inspection).filter_by(external_id=external_id).first()
+        if inspection is None:
```
matbusby-fw, line 150: This is the core of the change.  Checking for an inspection within the session that already contains the external_id and updating it if necessary.  The rest of the changes generally follow on this behaviour to ensure idempotency.

alva-fresh (approved):

simensma-fresh (approved):

PR #3813, opened 2026-01-27
[Core+MS] MDS-6613: make it mandatory to add in number of injuries for reportable incidents

alva-fresh:
## Objective 

[MDS-6613](https://bcmines.atlassian.net/browse/MDS-6613)

**Summary:**
- Added a new validation check for the "Number of injuries" field to make it required;
- Added placeholder to describe input;
- Removed "(optional)" label;


simensma-fresh (changes requested):
I see there's a few failing snapshots. Otherwise 👍

simensma-fresh (approved):

PR #3825, opened 2026-02-07
[MDS-6761] Fix for signature not being able to change

asinn134:
## Objective 

[MDS-6761](https://bcmines.atlassian.net/browse/MDS-6761)

- Updated filepond-plugin-file-encode to version 2.1.14 to fix this error 

`can't access property "data", base64Cache[item.id] is undefined` 

I found only PartySignatureUpload file uses this library. I tried testing some other file upload areas too just in case and didn't notice any breaking bugs from version upgrade.

-  Adjusted how the FilePond file property is feed data in the PartySignatureUpload file.

### DIFF
--- a/services/core-web/src/components/Forms/parties/PartySignatureUpload.js
+++ b/services/core-web/src/components/Forms/parties/PartySignatureUpload.js
@@ -32,11 +32,13 @@ registerPlugin(
 );
 
 export const PartySignatureUpload = (props) => {
-  const [files, setFiles] = useState([]);
+  const [files, setFiles] = useState(() => {
+    return props.signature ? [props.signature] : [];
+  });
 
   return (
     <FilePond
-      files={props.signature ? [props.signature] : files}
+      files={files}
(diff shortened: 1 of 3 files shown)


services/core-web/src/components/Forms/parties/PartySignatureUpload.js
```diff
@@ -32,19 +32,31 @@ registerPlugin(
...
 
 export const PartySignatureUpload = (props) => {
-  const [files, setFiles] = useState([]);
+  const [files, setFiles] = useState(() => {
+    return props.signature ? [props.signature] : [];
+  });
 
   return (
     <FilePond
-      files={props.signature ? [props.signature] : files}
```
asinn134, line 39: The way this was previously setup meant props.signature would override the files state with the old signature even after the user clicks the X button to have it removed.

simensma-fresh (approved):

alva-fresh (approved):

PR #3833, opened 2026-02-20
[MDS-6803] Debounce global search

simensma-fresh:
## Objective 

[MDS-6803](https://bcmines.atlassian.net/browse/MDS-6803)

This PR fixes a couple of issues noticed during testing of the new global search feature

- Added a debounce to global search to avoid searches on each keystroke
- Don't navigate to the first search result if you still have focus in the search bar
- Fix facet count not always being correct
- Fix text on "View All" button when search is empty
- Tweak styling of search modal
- Added some logging to help debug connection issue for longer searches


matbusby-fw (dismissed):

matbusby-fw (approved):
Nice!

asinn134 (approved):

PR #3834, opened 2026-02-20
[CHORE] CLI Command to replicate now_applications in local, dev or test

matbusby-fw:
## Objective 

This introduces a command that can be run locally to copy the pertinent data from a production NoW onto either the local, dev or test environment.  

It uses the Openshift CLI to connect to the production environment and run a read query to get the necessary data and then allows the user to select a target environment against which to apply that data patch.
- The user will be prompted to provide an Openshift login token (unless they are already logged in)
- Then they are prompted to enter the NoW number that they wish to copy from prod
- Then they are prompted to choose a target environment to place the data (defaults to local)
- Then they are prompted to enter a mine guid from the target environment to which the now_application will be added
- It will then insert the data into the chosen environment and mine.  
- This copies most relevant data from the original now_application, but leaves out personal details and protected data.  This includes copying the now_application_guid which can be used to test data fix migrations on now_application records across environments
- This can be run multiple times, and the script will remove the previous record and re-insert

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,10 @@
     "@syncfusion/ej2-pdfviewer": "31.2.5"
   },
   "scripts": {
-    "postinstall": "husky install"
+    "postinstall": "husky install",
+    "core-web": "yarn workspace @mds/core-web serve",
+    "minespace-web": "yarn workspace @mds/minespace-web serve",
+    "frontend": "yarn core-web && yarn minespace-web"
   },
   "lint-staged": {
     "*.{js,css,json,md,ts,tsx}": [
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -41,7 +41,10 @@
     "@syncfusion/ej2-pdfviewer": "31.2.5"
   },
   "scripts": {
-    "postinstall": "husky install"
+    "postinstall": "husky install",
```
matbusby-fw, line 44: Was trying out a different IDE and thought I'd throw in these for convenience

PR #3835, opened 2026-02-23
[MDS-6796] - NoW issue permit wrong status

matbusby-fw:
A number of notice of work applications have gotten stuck in a previous status rather than moving to approved when the permit is being issued.  There was a two part issue.

1. The function in the backend that handles the updating of the now_application status is set up in a way that it goes through a bunch of steps and saves data at multiple points throughout.  This isn't ideal because if it saves one thing, and then has an error later, the first thing remains updated, but it never gets to the rest.  I've put together a fix to make sure all of the changes are transactional (so if anything fails, everything gets rolled back to it's state before the change).  That's why we ended up with a status that didn't match where the application was intended to be.,
3. We have a function called find_current_appointments which runs during that process which should fetch the mine manager.  However, it has a filter applied to it so it only fetches appointments with no end date.  In instances where a mine manager has an end date in the future, the system thinks there is no mine manager because the end_date has a value.  So, the first function then thinks there is no mine manager, and it follows the logic in place for new_appt_needed and tries to create a new appointment.  But, when it tries to do that, it conflicts with a database constraint that doesn't allow overlapping appointment ranges and it fails (which leads us to issue #1).

Made the update function transactional, and updated the `find_current_appointment` function transactional to prevent partial updates on errors.

## Objective 

[MDS-6796](https://bcmines.atlassian.net/browse/MDS-6796)

### DIFF
--- a/services/core-api/app/api/now_applications/resources/now_application_status_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_status_resource.py
@@ -79,264 +79,267 @@ class NOWApplicationStatusResource(Resource, UserMixin):
         if now_application_status_code is None or current_status == now_application_status_code:
             return 200
 
-        # Handle approved status
-        if now_application_status_code == 'AIA':
-            permit = Permit.find_by_now_application_guid(application_guid)
-            if not permit:
-                raise NotFound('No permit found for this application.')
+        try:
+            # Handle approved status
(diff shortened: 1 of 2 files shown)


simensma-fresh (approved):

asinn134 (approved):
Nice digging 👍

PR #3838, opened 2026-02-24
[MDS-6108] Incidents file types

matbusby-fw:
Quick patch to allow common file types for internal ministry documents on incidents.  Also fixed a couple incidents of a type on the .heic file type.
(Technically the other file types were already allowed, this just updates the text)

## Objective 

[MDS-6108](https://bcmines.atlassian.net/browse/MDS-6108)


simensma-fresh (approved):
👀

asinn134 (approved):

PR #3839, opened 2026-02-25
[MDS-6728] Generate list of recuring crr reports

asinn134:
## Objective 

- Added CRR report generation for existing CRR recurring reports. Will run daily under the same cli command as the PRR report generation

- Added new mine_report_due_date_type called `YRL` to support the reports that have set due date on Jan 31st for the year.

- Made it so reports that are associated with a mine report definition that has `active_ind = false` to not show up in reports table.

- Mine report definitions with `active_ind = false` will no longer show up in the report name filter drop down and also the report name drop down when creating a report.

image

image

image

### DIFF
--- a/services/common/src/redux/slices/complianceReportsSlice.ts
+++ b/services/common/src/redux/slices/complianceReportsSlice.ts
@@ -74,7 +74,7 @@ const complianceReportSlice = createAppSlice({
         thunkApi.dispatch(showLoading());
         const resp = await CustomAxios({
           errorToastMessage: "Failed to load compliance reports",
-        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS(searchParams ?? {})}`, headers);
+        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS({ ...searchParams, active_ind: [true] })}`, headers);
         thunkApi.dispatch(hideLoading());
         return resp.data;
       },
--- a/services/core-api/app/api/mines/reports/models/mine_report.py
+++ b/services/core-api/app/api/mines/reports/models/mine_report.py
@@ -31,12 +31,13 @@ from app.config import Config
 from app.extensions import db
 from flask import current_app
 from pytz import timezone as pytz_timezone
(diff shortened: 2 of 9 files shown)


services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -74,7 +74,7 @@ const complianceReportSlice = createAppSlice({
         thunkApi.dispatch(showLoading());
         const resp = await CustomAxios({
           errorToastMessage: "Failed to load compliance reports",
-        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS(searchParams ?? {})}`, headers);
+        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS({ ...searchParams, active_ind: [true] })}`, headers);
```
asinn134, line 77: This was for making sure we don't display mine report definitions that are not active

services/core-api/app/api/mines/reports/resources/mine_reports.py
```diff
@@ -159,8 +159,19 @@ def get(self, mine_guid):
                 }
 
         # Base query; ordering is applied via ReportFilterHelper
-        query = MineReport.query.filter_by(mine_guid=mine_guid, deleted_ind=False)
-
+        query = (
+            MineReport.query
```
asinn134, line 163: This was just filtering out the inactive mine report definitions

services/core-api/app/api/mines/reports/models/mine_report.py
```diff
@@ -471,6 +472,58 @@ def create_from_permit_report_requirement(cls, requirement, due_date):
         
         return mine_report
 
+    @classmethod
+    def get_all_recurring_crr_reports(cls) -> list[Self]:
```
asinn134, line 476: This part searches for the CRR reports that have

1. A due date
2. A mine report definition that is active
3. Are of mine_report_due_date_type "FIS" or "YRL"
4. An active permit in the mine the report belongs to
5. A mine_report_category of "TSF" if the mine report definition is one related to TSFs or Dams

matbusby-fw (approved):
Looks great!

PR #3851, opened 2026-03-05
[MDS-6728] Changing crr report generation to use pagination

asinn134:
## Objective 

- Noticed in the logs that a SIGKILL error was showing up causing CRR report generation not to run
- Separated PRR and CRR report generation into their own tasks so they can run separately
- Adjusted CRR report generation query to use pagination


matbusby-fw (approved):
Nice!  That looks great.

PR #3853, opened 2026-03-05
[MDS-6728] Still show disabled definitions with existing reports

asinn134:
## Objective 

- Previously added some filtering around mine report definition where if the definition is not active we would filter it out. The problem with this is there are definitions that are not active but still reports that may have been submitted or requested in the past. We still want to allow the user to see/find these.

- Adjusted mine report definition filtering where if the definition is not active but there exists active mine reports associated with the definition we still show it in our report filter's report name dropdown option. 

image

image

### DIFF
--- a/services/common/src/redux/slices/complianceReportsSlice.ts
+++ b/services/common/src/redux/slices/complianceReportsSlice.ts
@@ -74,7 +74,7 @@ const complianceReportSlice = createAppSlice({
         thunkApi.dispatch(showLoading());
         const resp = await CustomAxios({
           errorToastMessage: "Failed to load compliance reports",
-        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS({ ...searchParams, active_ind: [true] })}`, headers);
+        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS(searchParams ?? {})}`, headers);
         thunkApi.dispatch(hideLoading());
         return resp.data;
       },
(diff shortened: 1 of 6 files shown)


services/common/src/redux/slices/complianceReportsSlice.ts
```diff
@@ -74,7 +74,7 @@ const complianceReportSlice = createAppSlice({
         thunkApi.dispatch(showLoading());
         const resp = await CustomAxios({
           errorToastMessage: "Failed to load compliance reports",
-        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS({ ...searchParams, active_ind: [true] })}`, headers);
+        }).get(`${ENVIRONMENT.apiUrl}${API.MINE_REPORT_DEFINITIONS(searchParams ?? {})}`, headers);
```
asinn134, line 77: This is changing it back to what it was before

services/core-api/app/api/mines/reports/resources/mine_reports.py
```diff
@@ -159,18 +159,7 @@ def get(self, mine_guid):
                 }
 
         # Base query; ordering is applied via ReportFilterHelper
-        query = (
```
asinn134, line 162: This is changing it back to what it was before

services/core-api/app/api/mines/reports/resources/reports_resource.py
```diff
@@ -65,17 +65,7 @@ def get(self):
             'region': request.args.getlist('region', type=str),
         }
 
-        query = (
```
asinn134, line 68: This is changing it back to what it was before

simensma-fresh (approved):

PR #3857, opened 2026-03-10
[MDS-6613] - Removed required tag from incidents number of fatalities

matbusby-fw:
## Objective 

Update to the incidents form to remove the required tag from the number of fatalities field.

[MDS-6613](https://bcmines.atlassian.net/browse/MDS-6613)


alva-fresh (approved):

alva-fresh:
Thank you for catching this!

PR #3858, opened 2026-03-11
[MDS-6801] - Introduce NoW Tier Tracking

matbusby-fw:
## Objective 
Implement the ability of Core users with the correct permissions to select tiers for NoWs and display.

- Database migration to add two tables
  - One to contain the tiers themselves
  - One to contain the tiers as connected to a notice of work application (this is the table that will eventually have a history)
- Api updates to handle NoW tiers
- Frontend updates for those tiers
- New feature flag for `notice_of_work_tier`

[MDS-6801](https://bcmines.atlassian.net/browse/MDS-6801)

image
image
image
image


simensma-fresh (changes requested):
A few SonarCloud issues it would be good to take a pass at (looks like mostly small little fixes).

simensma-fresh (approved):

PR #3869, opened 2026-03-25
[MDS-6835] Show tier allocation for mx and cx now apps

asinn134:
## Objective 

[MDS-6835](https://bcmines.atlassian.net/browse/MDS-6835)

-Add tier column with tool tip to minespace NoW table, setup behind tier feature flag. Made it so when a mine does not have a Mineral or Coal NoW application the Tier column will not show. When there is at least one Mineral or Coal NoW application Tier column will show and all other NoW application types will have "N/A" in the tier column.

-Add banner to minespace NoW view page, setup behind tier feature flag. Banner won't show until a tier is assigned to the NoW application

image

image

image

### DIFF
--- a/services/minespace-web/src/components/dashboard/mine/projects/NoticeOfWorkTable.tsx
+++ b/services/minespace-web/src/components/dashboard/mine/projects/NoticeOfWorkTable.tsx
@@ -92,7 +92,7 @@ export const NoticeOfWorkTable: FC<NoticeOfWorkTableProps> = ({ isLoaded, applic
             key: "now_application_tier_code",
             dataIndex: "now_application_tier_code",
             width: "120px",
-            sorter: dateSorter("now_application_tier_code"),
+            sorter: (a, b) => (a.now_application_tier_code > b.now_application_tier_code ? -1 : 1),
             defaultSortOrder: "descend" as SortOrder,
             render: (text, record) => {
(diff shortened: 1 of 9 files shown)


services/minespace-web/src/components/dashboard/mine/projects/NoticeOfWorkTable.tsx
```diff
@@ -64,6 +73,41 @@ export const NoticeOfWorkTable: FC<NoticeOfWorkTableProps> = ({ isLoaded, applic
...
+                >
+                  {" "}
+                  <InfoCircleOutlined className="info-tooltip icon-sm" />
+                </Tooltip>
+              </div>
+            ),
+            key: "now_application_tier_code",
+            dataIndex: "now_application_tier_code",
+            width: "120px",
+            sorter: dateSorter("now_application_tier_code"),
```
matbusby-fw: I'm trying to remember how this works.  Won't this try to run the date sorter operation on the tier code rather than a date?  I feel like that wouldn't work.

asinn134: Thanks for the catch! Yah that didn't make sense. This has been changed now.

matbusby-fw (approved):
Great!

PR #3875, opened 2026-03-31
[MDS-6802] - Implement tier history

matbusby-fw:
## Objective 

Introduce tracking of changes to now_application tiers.
- added history table using SQLAlchemy continuum
- Updated backend models including hybrid properties
- Added an resource to retrieve the history
- Created a  history modal accessible through the NoW page header tag
- Added logic to show the initial tier decision as the initial intake in various places

[MDS-6802](https://bcmines.atlassian.net/browse/MDS-6802)

image
image


asinn134 (approved):
Nice!

PR #3888, opened 2026-04-16
[MDS-6198] Separated docman and flower into separate containers when running app locally

simensma-fresh:
## Objective 

[MDS-6198](https://bcmines.atlassian.net/browse/MDS-6198)

Separated flower into a separate container when running the app locally. That way we'll be able to see the logs for both containers easily.

Also resolved a "No module named pkg_resources" error when building the docman image. Cause: the `opentelemetry` pulled in an outdated `setuptools` package

image

### DIFF
--- a/docker-compose.M1.yaml
+++ b/docker-compose.M1.yaml
@@ -56,7 +56,6 @@ services:
     build:
       context: services/postgres
       dockerfile: Dockerfile
-      platform: linux/amd64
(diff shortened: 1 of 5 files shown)


matbusby-fw (approved):
Sweet!

asinn134 (approved):
Thanks Simen!

PR #3889, opened 2026-04-17
[MDS-6892] Active Permit Count Bugfix

alazar-aot:
## Objective 

- Updated the logic that calculates a mine's active permit count to include a check for `permit_status = 'C'`, so the `active_permits` value accurately reflects both the number of non-draft permits, as well as non-closed permits
- Test file updated with the same logic

**_Note:_** Active permit filtering is currently split across two layers, `mine.mine_permit` excludes drafts (`permit_status == 'D'`), and the stats resource modified in this PR then excludes closed permits (`permit_status == 'C'`). Consolidating this logic into `mine.py` was considered but rejected, as `mine_permit` is used elsewhere in the codebase where closed permits should still be visible. The active permit definition was intentionally kept local to the stats calculation.  

[MDS-6892](https://bcmines.atlassian.net/browse/MDS-6892)

- It was noticed that the `active_permit` count for a mine did not accurately reflect how many open permits a mine had, instead it only filtered out the total number of non-draft permits (and kept included 'Closed' status permits)

### DIFF
--- a/services/core-api/app/api/mines/reports/resources/mine_report_stats.py
+++ b/services/core-api/app/api/mines/reports/resources/mine_report_stats.py
@@ -15,13 +15,12 @@ from flask_restx import Resource
 from pytz import timezone
 from werkzeug.exceptions import NotFound
 
-
 class MineReportStatsResource(Resource, UserMixin):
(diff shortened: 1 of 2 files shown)


matbusby-fw (approved):
Nice!  Looks solid!

PR #3897, opened 2026-05-04
[MDS-6883] - now_application document search

matbusby-fw:
## Objective 

[MDS-6883](https://bcmines.atlassian.net/browse/MDS-6883)

## Notice of Work Application Document Search

This branch introduces a document search feature for Notice of Work (NoW) applications. It allows reviewers to query the full text of all documents attached to a specific NoW application using natural language, with AI-generated summaries and highlighted excerpts. The feature is built on the same infrastructure as the existing Permit Condition Search, but adapts it at every layer to suit the different data model and operational constraints of NoW documents.

---

#### Step 1 — The reviewer triggers indexing from the UI

A new **Search Documents** tab is added to the NoW application page (`NowApplicationDocumentSearch.tsx`). When a reviewer clicks **Index Documents**, the frontend dispatches to the `nowApplicationSearchSlice`, which calls the core-api endpoint.

The core-api resource (`now_application_document_search_resource.py`) fetches the list of documents attached to the NoW application from the existing application model, then hands them to `NowApplicationSearchService` for processing.

> **Divergence from permit search:** Permit indexing is triggered automatically as part of the permit condition extraction pipeline. NoW indexing is user-initiated and can be re-triggered on demand, because application documents change over time as the review progresses.

---

#### Step 2 — Core-api downloads and forwards documents one at a time

`NowApplicationSearchService.index_documents()` loops over the document list and for each one:

1. Downloads the file from Document Manager into an **in-memory `BytesIO` buffer** (rather than a temp file on disk)
2. Posts the buffer as a single-file multipart upload to the permits service
3. Lets the buffer go out of scope so Python can free the memory before the next file is downloaded

This one-at-a-time approach is deliberate. NoW applications can have many documents and individual files can be up to 750 MB. Downloading all files before sending any of them would accumulate gigabytes in memory simultaneously.

If a document is already being indexed (the permits service returns `409`), it is skipped gracefully and the loop continues with the next document.

> **Divergence from permit search:** The permit pipeline sends a single pre-extracted text payload to the permits service. NoW sends raw binary files one at a time, because there is no equivalent pre-extraction step in core-api for application documents.

---

#### Step 3 — The permits service queues a Celery task per document

The permits service endpoint (`document_search_resource.py`) receives each file upload and:

1. Checks Redis for an existing active task for that specific document (keyed on `now_doc_index:{now_guid}:{doc_guid}`) — returns `409` if one is already running
2. Saves the file to a shared volume path accessible by the Celery worker
3. Launches a `run_now_document_indexing` Celery task and stores the task ID in Redis
4. Adds the task ID to a per-application Redis set (`now_doc_index_tasks:{now_guid}`) so that status aggregation and cancellation can work across all documents belonging to the same application

> **Divergence from permit search:** Permit condition extraction runs as a single monolithic Celery task per permit. NoW uses one task per document, tracked via a Redis task set, so that large applications with many documents can be processed concurrently and progress can be tracked and cancelled at the application level.

---

#### Step 4 — The Celery worker extracts, embeds, and indexes

Each `run_now_document_indexing` task progresses through four phases, reporting progress throughout:

| Phase | % range | What happens |
|---|---|---|
| Delete | 0–15% | All existing index chunks for this document are deleted from Azure Search first. This prevents orphaned stale chunks when a re-index produces fewer chunks than the previous run. |
| Extract | 15–30% | Azure Document Intelligence converts the raw PDF or Office file into structured text |
| Embed | 30–75% | The text is split into chunks by `DocumentChunker`, then batched through Azure OpenAI (`text-embedding-3-large`) to produce 3072-dimension vector embeddings |
| Push | 75–100% | Chunks are uploaded to Azure AI Search in batches of 100 |

Each chunk is assigned a **deterministic ID** based on `sha256(now_application_guid + document_manager_guid + chunk_index)`, so re-indexing is idempotent.

The `DocumentChunker` extracts the plain text `"text"` field from the JSON structure that `AzureDocumentIntelligenceConverter` produces internally — without this step the raw JSON string (including field names and metadata) would be stored verbatim in the search index.

> **Divergence from permit search:** The permit pipeline uses a push-based Azure Search indexer (blob → indexer → index). NoW uses a direct push approach (Celery worker → SDK upload), which gives finer progress visibility and avoids the need for blob storage as an intermediate. The delete-before-reindex step is also unique to NoW because the direct push approach does not have the automatic overwrite semantics of the Azure indexer.

---

#### Step 5 — Progress is aggregated across all documents

When the frontend polls for indexing status, the permits service fetches all task IDs from the application's Redis set, checks their Celery state, and computes a single aggregate progress percentage weighted by **chunk count**.

A large document that produces 800 chunks contributes proportionally more to the overall progress bar than a small document that produces 20 chunks. This prevents the bar from jumping to 50% the moment a trivial document finishes while a large one is still processing.

Once all tasks reach a terminal state (`SUCCESS`, `FAILURE`, or `REVOKED`), the overall status is reported as `success` or `error` accordingly.

---

#### Step 6 — The reviewer searches

When the reviewer types a query, the frontend dispatches to `nowApplicationSearchSlice`, which SSE-streams results from the permits service search endpoint. The search pipeline:

1. Embeds the query using the same `text-embedding-3-large` model
2. Runs a **hybrid search** (BM25 keyword + vector similarity) against the NoW-specific Azure Search index, with `now_application_guid` as a mandatory pre-filter so results are always scoped to the current application
3. Applies semantic re-ranking across the top candidates
4. Returns up to 25 results with highlighted excerpt fragments (where the keyword component matched) and computes facets on `document_type` and `submitted_date` for the filter UI

The AI-generated summary is produced in parallel: the top retrieved chunks are assembled into a Jinja2 prompt and sent to an Azure OpenAI chat completion. The response streams back over SSE and is rendered via `MarkdownViewer`. The LLM cites sources as `[doc:1]`, `[doc:2]` etc., which `MarkdownViewer` converts to clickable links that scroll to and briefly highlight the corresponding result card on the left.

> **Divergence from permit search:** The permit search index contains pre-structured condition hierarchy metadata (step paths, parent/child IDs). The NoW index schema is intentionally simpler — it stores `now_application_guid`, `document_name`, `document_type`, `submitted_date`, and `content`. The NoW search also uses a separate Redux slice and separate API endpoints so the two search contexts never share state.

---

### Key files added

| Layer | File |
|---|---|
| Frontend state | `nowApplicationSearchSlice.ts` |
| Frontend UI | `NowApplicationDocumentSearch.tsx`, `NowApplicationDocumentSearchSplashScreen.tsx`, `NowDocumentResultItem.tsx` |
| Core-api proxy | `now_application_document_search_resource.py`, `now_application_search_service.py` |
| Permits pipeline | `document_search_pipeline.py`, `document_search_resource.py`, `document_chunker.py`, `indexing.py`, `search_index_fields.py`, `create_search_index.py` |
| Celery task | `tasks.py` (`run_now_document_indexing`) |
| LLM prompt | `now_application_search_prompts.yaml` |

image

image


alazar-aot (dismissed):
Pulled and tested the changes locally - everything seems to be working well! 

Great work

asinn134 (commented):
Looks great! I think just have to resolve the merge conflicts and should be good to go.

alazar-aot (approved):
Re-Approved! Even better now

asinn134 (dismissed):

services/common/src/components/documents/DocumentTable.tsx
```diff
@@ -102,15 +102,6 @@ export const DocumentTable: FC<DocumentTableProps> = ({
     setDocuments(parseDocuments(props.documents ?? []));
   }, [props.documents]);
 
-  const normalizeSelectedRows = (selectedRows: MineDocument[]) => {
```
simensma-fresh, line 105: Were these removals intentional?

services/common/src/components/reports/ReportGetStarted.tsx
```diff
@@ -132,6 +134,24 @@ const ReportGetStarted: FC<ReportGetStartedProps> = ({
             },
           ]}
         />
+        {system !== SystemFlagEnum.core &&
```
simensma-fresh: Think this was intentionally removed in a previous PR. Think you might have to do a deeper look at the rebasing done :)

simensma-fresh (changes requested):

services/core-api/app/api/mines/reports/resources/mine_report_stats.py
```diff
@@ -29,7 +30,7 @@ def get(self, mine_guid):
             raise NotFound('Mine not found')
 
         # Active permits are those surfaced by Mine.mine_permit (excludes drafts)
-        active_permits = len([permit for permit in mine.mine_permit if permit.permit_status_code != 'C'])
+        active_permits = len(mine.mine_permit_numbers)
```
simensma-fresh: This was changed in a previous PR, will have to revert

services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
```diff
@@ -0,0 +1,144 @@
...
+from app.api.utils.resources_mixins import UserMixin
+from app.extensions import api
+from flask import Response, request, stream_with_context
+from flask_restx import Resource
+from werkzeug.exceptions import NotFound, ServiceUnavailable
+
+
+def _require_feature():
+    if not is_feature_enabled(Feature.NOW_APPLICATION_DOCUMENT_SEARCH):
+        raise ServiceUnavailable("NoW application document search is not enabled.")
```
simensma-fresh: Maybe let's use something like a `403 Forbidden` error here, seeing as a 503 Generally signals that the service is unavailable

services/permits/app/__init__.py
```diff
@@ -1,5 +1,35 @@
+import os
 from dotenv import find_dotenv, load_dotenv
 
 ENV_FILE = find_dotenv()
 if ENV_FILE:
     load_dotenv(ENV_FILE)
+
+# Set dummy environment variables to allow top-level imports and assertions to pass.
+# We apply these if the keys are missing OR if they are empty strings.
+dummy_env = {
```
simensma-fresh: I'm wondering if it would be better to warn the user / or fail hard if what you provided in the `.env` doesn't match. Just worried this might cause some extra confusion 😅

services/permits/requirements.txt
```diff
@@ -9,6 +9,7 @@ requests~=2.32.4
 python-dotenv==1.0.0
 boto3==1.34.139
 fastapi==0.115.6
+anyio>=4.0.0
```
simensma-fresh: Can we lock this one down with `==`

simensma-fresh (changes requested):

services/core-api/tests/reports/resource/test_mine_report_stats_resource.py
```diff
@@ -14,15 +14,13 @@
...
-def test_report_stats_active_permits_excludes_inactive_permits(test_client, db_session, auth_headers):
+def test_report_stats_active_permits_excludes_drafts(test_client, db_session, auth_headers):
     mine = MineFactory(mine_reports=0, minimal=True)
 
     p_draft = PermitFactory(permit_status_code='D')
     p_open = PermitFactory(permit_status_code='O')
-    p_closed = PermitFactory(permit_status_code='C') 
     MinePermitXrefFactory(permit=p_draft, mine=mine)
     MinePermitXrefFactory(permit=p_open, mine=mine)
-    MinePermitXrefFactory(permit=p_closed, mine=mine)
```
alazar-aot, line 25: These changes also need to be preserved

alazar-aot (changes requested):

services/common/src/components/documents/DocumentTable.tsx
asinn134: Oh nice catch Simen

simensma-fresh (dismissed):

simensma-fresh (dismissed):

alazar-aot (changes requested):
I've noticed 2 small issues:

1. OIDC config re-fetched on every request
- File: `services/core-api/app/api/search/search/now_application_search_service.py` (Same bug also exists in permit_search_service.py)
- Issue: `self._oidc_configuration = ...` writes to the _instance_, not the class. Since a new `NowApplicationSearchService()` is created per request, the class attribute stays `None` forever and the OIDC well-known config + OAuth2 token are fetched on every single request.
- Fix: I believe changing the assignment in the property from `self.oidc_configuration = requests.get...` to the class name (`NowApplicationSearchService._oidc_configuration = requests.get...` should fix the issue, though I haven't tested this myself

2. Status shows "Indexed" after partial cancellation
- `services/permits/app/pipelines/document_search/resources/document_search_resource.py`
- Issue: `cancel_indexing` deletes the Redis task set immediately after revoking tasks. The status endpoint then falls back to querying Azure Search, and if any chunks were already indexed before cancellation, the count is > 0 and it returns "success", showing the "`Indexed ✓`" badge even though the index is incomplete (misleading). 
- Fix: Write a "cancelled" marker key to Redis before deleting the task set, and check for it in `get_indexing_status` before falling through to the Azure Search count

alazar-aot (approved):
Looking good, thanks for making those changes!

PR #3904, opened 2026-05-11
[CHORE] - Fix NRIS Deploy

matbusby-fw:
## Objective 
PR to fix the deployment of NRIS after recent python upgrades

### DIFF
--- a/services/nris-api/backend/Dockerfile
+++ b/services/nris-api/backend/Dockerfile
@@ -29,7 +29,7 @@ WORKDIR /app
 
 # Install the requirements
 COPY requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt
+RUN pip install "setuptools<70.0.0" wheel && pip install --no-cache-dir --no-build-isolation -r requirements.txt
 
 # Oracle config files
 RUN mkdir -p /opt/app-root/src/wallet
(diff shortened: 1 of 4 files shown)


alazar-aot (dismissed):

simensma-fresh (changes requested):
If possible, let's pin the version in `requirements.txt` instead, and pin to an exact version 🙃

services/nris-api/backend/Dockerfile
```diff
@@ -29,6 +29,7 @@ WORKDIR /app
 
 # Install the requirements
 COPY requirements.txt .
+ENV PIP_CONSTRAINT=requirements.txt
```
matbusby-fw, line 32: setting PIP_CONSTRAINT to pip forces isolated build environments (which are used when building old packages like cx_Oracle) to honor the version limits defined in the requirements file

services/nris-api/backend/requirements.txt
```diff
@@ -1,5 +1,5 @@
 cached_property==1.5.1
-setuptools
+setuptools<70.0.0
```
simensma-fresh: Can we pin it to the version we want? `==`
There's been too many vulnerabilities lately that stem from dynamic resolution 😅

simensma-fresh (changes requested):

simensma-fresh (approved):

PR #3919, opened 2026-05-25
[CHORE] search task management

matbusby-fw:
## Objective 

The last fix didn't manage to clear the old tasks.  Added logic to check only the latest task for it's status and clear out older onces when status checks are made.  This should keep things a bit tidier and be more reliable.


alazar-aot (approved):
Looks good to me, other than I'm not sure if you want to update the SonarQube warnings about the logger message type (exception vs error). 

Approving but worth confirming the correct message type!

simensma-fresh (approved):

PR #3921, opened 2026-05-27
[MDS-6845] [Core] Permit-Search-Indicate-Multiple-Mine-Relationships

alazar-aot:
## Objective 

[MDS-6845](https://bcmines.atlassian.net/browse/MDS-6845)

With the recent upgrade to Global Search in Core, users can search directly for permit numbers.
Some permits are associated with multiple mines, but search results currently present the permit as if it is associated with a single mine, which can mislead users and result in incorrect assumptions.

The changes in this PR have been made so that:
1. [COMPLETE] If a permit is associated to one mine, display the global search results normally (no changes)
2. [COMPLETE] If a permit is associated to multiple mines, results explicitly indicate multiple associations, and these associations can be navigated to
3. [NOT STARTED] On the permit details page, show a dedicated “associated mines” section

I a requesting review and deployment into the Dev/Test env now, because I cannot easily test these changes locally (since the Permit Service doesn't deploy well locally). 

The changes for task 3 (Permit Details Page) will follow once I verify these are working as expected.

### DIFF
--- a/services/common/src/interfaces/search/searchResult.interface.ts
+++ b/services/common/src/interfaces/search/searchResult.interface.ts
@@ -16,6 +16,10 @@ export interface ISimpleSearchResult {
   description?: string;
   highlight?: string;
   mine_guid?: string;
+  mines?: {
+    mine_guid: string;
+    mine_name: string;
+  }[];
 }
 
 export interface IExplosivesPermitSearchResult {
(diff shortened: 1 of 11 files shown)


matbusby-fw (approved):
Nice!  Looks great!

PR #3923, opened 2026-05-29
[MDS-6914] Dynamically Managed Ministry Contact Distribution Lists 

matbusby-fw:
## Objective 

[MDS-6914](https://bcmines.atlassian.net/browse/MDS-6914)

## 🚀 Description

  Currently, whenever business stakeholders need to add or remove users from an internal email distribution list, they must wait for a developer to manually update hard-coded arrays within the codebase.
  This PR resolves this operational bottleneck by introducing user-managed Distribution Lists. Administrators with the  core_edit_emli_contacts  role can now configure which ministry contacts belong to which notification group (Distribution List) directly via the Admin UI. This allows
  for instant updates without any developer intervention or application deployments.
Added an optional utility container that can be used in testing email notifications 

  ## 🛠️ Technical Implementation

  ### 📬 Mailpit Integration for Local Email Testing

  As part of this PR, I've added an optional Mailpit container to improve the local development and testing experience for email notifications.

  • Purpose: Mailpit acts as a local SMTP server and web interface that intercepts all outgoing emails. This allows developers and testers to inspect email delivery, verify Jinja2 template rendering, and test the new distribution list logic without risking sending accidental test
  emails (or spam) to actual production ministry inboxes.
  • How it works: By setting  USE_LOCAL_MAILPIT=true  in your  .env  file, the  EmailService  automatically routes outgoing notifications through the Mailpit container rather than the CHES API.
  • Usage: You can quickly spin up the service by running  make mailpit  (which utilizes the new  docker-compose.mailpit.yaml ). Once running, emails can be reviewed in real-time by navigating to Mailpit's local web UI (typically on port  8025 ).

  ### Database Changes

  • New Tables: Created  distribution_list  and  distribution_list_user  tables.
  • Email Tracking: Added a  distribution_list_guid  column to  email_tracking  to track emails delivered through these dynamic lists.
  • Migration Script: Added a seed script mapping legacy hard-coded emails to our new contacts system, and dynamically seeding them to the relevant  distribution_list .

  ### Backend (Core API)

  • Refactored notification services across various features (Notice of Departure, TSFs, Incidents, Major Projects, Variances, Notice to Start/Stop Work, Report Submissions) to query lists via  DistributionList.find_by_name()  instead of static constants.
  • Offloaded email notifications to asynchronous Celery tasks ( send_email_task  &  send_template_email_task ).
  • Ensured any pre-existing dynamic business logic (such as assigning regional vs major mines or assigned inspectors) operates in addition to the new generic distribution lists.
  • Added  distribution_list_guid  tracing down through the email tasks to the  email_tracking  table.

  ### Frontend (Core Web)

  • Safely migrated  MineSpaceMinistryContactManagement.js  and  MinistryContactForm.js  to functional components utilizing TypeScript and  useAppDispatch  /  useAppSelector .
  • Implemented a multi-select field for "Notification Groups" in the Contact Edit Modal.
  • Updated the Admin Ministry Contacts page to showcase a tabbed view breaking down configured Notification Groups.
  ──────
  ## 📇 Notification Groups & Dynamic Contact Notes

  As a reference for QA and Reviewers, below is a breakdown of the specific workflows that have been transitioned, the pre-existing logic that remains, and the contacts that have been securely transitioned from hard-coded values into the initial DB seed:

  1. Notice of Departure

  • Notes: Replaces a series of hard-coded individual emails.
  • Seeded List:  mds@gov.bc.ca ,  sean.shaw@gov.bc.ca ,  Magda.Kingsley@gov.bc.ca ,  Kelsey.Norlund@gov.bc.ca ,  Kristy.Emery@gov.bc.ca ,  Samuel.Barnes@gov.bc.ca ,  Brent.Timmons@gov.bc.ca .

  2. TSFs (Tailings Storage Facilities)

  • Notes: Replaces hard-coded individual emails.
  • Seeded List:  permrecl@gov.bc.ca ,  mark.smith@gov.bc.ca ,  victor.marques@gov.bc.ca ,  mds@gov.bc.ca .

  3. Major Projects

  • Notes: Replaces hard-coded generic inboxes and individuals. Dynamic Assigned Inspector logic was intentionally preserved and acts in parallel.
  • Seeded List:  mds@gov.bc.ca ,  Magda.Kingsley@gov.bc.ca ,  Kelsey.Norlund@gov.bc.ca ,  Kristy.Emery@gov.bc.ca ,  Samuel.Barnes@gov.bc.ca .

  4. Variances
  • Notes: Replaces hard-coded generic inboxes and individuals.
  • Seeded List:  hermanus.henning@gov.bc.ca .
  5. Notice to Start/Stop Work

  • Notes: Replaces hard-coded generic inboxes. Business stakeholders explicitly requested the ability to add specific individuals to this over time.
  • Seeded List: Setup but left intentionally empty for Admins to customize via the UI.

  6. Incidents

  • Notes: Replaces hard-coded generic inbox. Dynamic Assigned Inspector logic was intentionally preserved and acts in parallel.
  • Seeded List:  mineincidents@gov.bc.ca .

  7. Report Submissions (Regional vs. Major Mines)

  • Notes: The logic split for identifying whether a report is intended for a Regional or Major Mine has been preserved. The codebase routes the email to one of two configured distribution lists.
  • Seeded List: Both the  Report Submission - Major Mines  and  Report Submission - Regional Mines  lists default seed to  mds@gov.bc.ca .

image
image
image

  ──────
  ## ✅ Acceptance Criteria Checklist

  [✓] Administrative users can enter "Edit" mode and select/deselect multiple notification groups for a contact.
  [✓] Editing rights require the  core_edit_emli_contacts  role.
  [✓] Selection of the group directly controls the list of internal ministry emails.
  [✓] Assigned inspector logic and Regional vs. Major mines logic remains unaffected and stacks gracefully with notification groups.
  [✓] Hard-coded business emails are fully eradicated, preventing the need for future developer ticket workflows.
  [✓] Asynchronous execution introduced via Celery tasks for sending notifications.
  [✓] New DB schema ( distribution_list  and  distribution_list_user ) successfully tracks the lists.
  [✓] The  email_tracking  table explicitly records the distribution list ID correctly.


alazar-aot:
Putting a note here on the issue I had with the Core-API Unit Tests,

As part of this ticket’s unit testing, there is a `test_post_now_application_happy_path` case in the `test_application_list_resouce.py` file.

When first pushed, this test was failing with a cryptic/generic “Assertion Error”:
```
>       assert post_resp.status_code == 200 or post_resp.status_code == 201, post_resp.response
E       AssertionError: [b'{
E             "status": 400,
E             "message": "Error occurred while processing data",
E             "trace_id": "302215506883090039991710025111897488818"
E         }
E         ']
E       assert (400 == 200 or 400 == 201)
E        +  where 400 = <WrapperTestResponse 136 bytes [400 BAD REQUEST]>.status_code
E        +  and   400 = <WrapperTestResponse 136 bytes [400 BAD REQUEST]>.status_code

tests/now_submissions/resources/test_application_list_resource.py:744: AssertionError
```

After _a lot_ of testing and debugging, I’ve determined the issue was due to the following:
- When the tests begin, the `setup_marshmallow()` function runs once at startup and stores `db.session` as `sqla_session` on every marshmallow schema
- Then, `db_session` (the test fixture) later replaces `db.session` with a connection-bound session for each test
- Because the marshmallow schemas cached the original session, they ended up pointing to a stale, closed connection from a previous test's fixture
  - The reason it only affected the `test_post_now_application_happy_path` test is because it’s the only test that exercises the full happy-path schema load (checking for 2xx status), meaning marshmallow’s `get_instance` queries the DB during the test - but that DB query would hit the stale connection and fail immediately
  - Technically speaking, when the POST handler called `Application._schema().load(request.json)`, marshmallow's `get_instance` tried to query the DB using the now-dead session, failing immediately, before `save()` was ever reached

To fix this, I made it so that the `db_session` fixture sets `db.session = sess`, and iterates over all the registered model classes and updates their schema’s `opts.sqla_session` to the fresh session. This made it so that marshmallow always queries through a live connection for the current test. 

Specifically, in `conftest.py`;
- Imported `Base` from `models_mixins`
- After `db.session = sess`, loop over all model classes and set `c._schema.opts.sqla_session = sess`

migrations/sql/V2026.05.25.13.29__create_distribution_lists.sql
```diff
@@ -0,0 +1,117 @@
...
+    ('Report Submission - Major Mines', 'Notification group for Report Submissions (Major Mines)', 'system', 'system'),
+    ('Report Submission - Regional Mines', 'Notification group for Report Submissions (Regional Mines)', 'system', 'system');
+
+-- Add GEN to emli_contact_type
+INSERT INTO emli_contact_type (emli_contact_type_code, description, display_order, create_user, update_user)
+VALUES ('GEN', 'General Contact', 100, 'system', 'system')
+ON CONFLICT DO NOTHING;
+
+-- Function to safely insert emli_contact and distribution_list_user
+CREATE OR REPLACE FUNCTION pg_temp.seed_distribution_user(
```
simensma-fresh: I'm wondering if we maybe should remove this function and add this as a task to do right after deploy through the new interface.

Seeing there's actual peoples email addresses here, I think it might be best to not have that propagate to dev / test envs or local if we can help it?

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -275,7 +275,12 @@ def validates_contact_method(self, key, value):
         return value
 
     def send_incidents_email(self):
-        ministry_recipients = [INCIDENTS_EMAIL]
+        from app.api.ministry_contacts.models.distribution_list import DistributionList
+        from app.api.email_tracking.email_status_tasks import send_template_email_task
+
+        dl = DistributionList.find_by_name('Incidents')
```
simensma-fresh: Can we change this so we don't look up a DL by its name? Generally preferred if we change the name of the DL (think enum that maps to an ID or slug)

`DistributionList.find_by_name(DistributionLists.INCIDENTS)`.

Same would apply to all the email sending we're doing

services/core-api/app/api/incidents/models/mine_incident.py
```diff
@@ -312,8 +317,27 @@ def send_incidents_email(self):
             },
             "minespace_incident_link": f'{Config.MINESPACE_PROD_URL}/mines/{self.mine.mine_guid}/incidents/{self.mine_incident_guid}',
         }
-        EmailService.send_template_email(subject, ministry_recipients, ministry_template, ministry_context, cc=cc)
-        EmailService.send_template_email(subject, minespace_recipients, minespace_template, minespace_context, cc=cc)
+        send_template_email_task.apply_async(kwargs={
```
simensma-fresh: I would recommend we abstract some of this to keep things simple. Instead of having to do this `apply_async` pattern with kwargs etc. every time we send an email, update `EmailService.send_template_email/send_email` to do this behind the scenes.

simensma-fresh: Another little quality of life we could do. Make the `EmailService.send_template_email/send_email` optionally take in a `DistributionList` name/slug. And have the lookup be done there. That way we don't have to do this pattern every time.

```
        dl = DistributionList.find_by_name('Incidents')
        ministry_recipients = dl.get_emails() if dl else []
        distribution_list_guid = str(dl.distribution_list_guid) if dl else None
```

=>

```
EmailService.send_email(..., distribution_list=DistributionLists.INCIDENTS)
```

services/core-api/app/api/ministry_contacts/resources/distribution_list_list.py
```diff
@@ -0,0 +1,13 @@
...
+from app.api.utils.resources_mixins import UserMixin
+from app.api.ministry_contacts.models.distribution_list import DistributionList
+from app.api.ministry_contacts.response_models import DISTRIBUTION_LIST_MODEL
+
+class DistributionListListResource(Resource, UserMixin):
+    @api.doc(description='Returns all Distribution Lists.')
+    @api.marshal_with(DISTRIBUTION_LIST_MODEL, code=201, envelope='records')
+    @requires_any_of([VIEW_ALL, MINESPACE_PROPONENT])
+    def get(self):
+        return DistributionList.get_all()
```
simensma-fresh: Would recommend making this paginated. It'll come back to bite us eventually 😅

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
```diff
@@ -42,9 +44,26 @@ def put(self, contact_guid):
 
         data = self.parser.parse_args()
 
+        distribution_list_guids = data.pop('distribution_list_guids', [])
+        
         for key, value in data.items():
             setattr(contact, key, value)
 
+        from app.api.ministry_contacts.models.distribution_list_user import DistributionListUser
```
simensma-fresh: Generally do imports at the top of the file

services/core-api/app/api/ministry_contacts/resources/ministry_contact_list.py
```diff
@@ -103,11 +105,21 @@ def post(self):
...
+            deleted_ind=data.get('deleted_ind', False),
+            add_to_session=False)
 
         if not contact:
             raise BadRequest('Error: Failed to create MCM contact.')
 
+        contact.save(commit=False)
+        db.session.flush()
+
+        from app.api.ministry_contacts.models.distribution_list_user import DistributionListUser
```
simensma-fresh: Generally do imports at the top of the file

services/core-api/app/api/notice_of_departure/models/notice_of_departure.py
```diff
@@ -231,11 +231,24 @@ def delete(self):
         super(NoticeOfDeparture, self).delete()
 
     def nod_submission_email(self):
-        recipients = MAJOR_MINES_NOD_NOTFICATION_EMAILS
-        cc = [MDS_EMAIL]
+        from app.api.ministry_contacts.models.distribution_list import DistributionList
+        from app.api.email_tracking.email_status_tasks import send_email_task
```
simensma-fresh: Generally do imports at top of the file

simensma-fresh (changes requested):
Just a left a few suggestions to make things a bit cleaner and less brittle for future use of the DL's / email sending

migrations/sql/V2026.05.25.13.29__create_distribution_lists.sql
alazar-aot: Agreed - I've removed the `CREATE OR REPLACE FUNCTION`, keep everything else as-is. We will need to manually update the emails in Prod after this is released

services/core-api/app/api/incidents/models/mine_incident.py
alazar-aot: Yes - I created an enum `DistributionListNames` and the call sites all now use that enum (eg `DistributionListNames.REPORT_SUBMISSION_MAJOR_MINES`, so if we ever update or add new DL names, only the enum needs to be updated!

services/core-api/app/api/ministry_contacts/resources/distribution_list_list.py
alazar-aot: Agreed - this has been updated to return paginated results, and the UI has also been updated to display the pagination!

services/core-api/app/api/incidents/models/mine_incident.py
alazar-aot: Good thinking - I updated this so that it does a lookup on the `DistributionList` model. Let me know what you think

simensma-fresh (approved):

PR #3929, opened 2026-06-05
[MDS-6911] improve speed when reviewing extracted permit conditions

asinn134:
## Objective 

[MDS-6911](https://bcmines.atlassian.net/browse/MDS-6911)

- Change to use new GET request endpoint in `refreshData` function in the permit condition page to retrieve the necessary data only and not the whole permit

- Adjusted permit condition form UI so when a user makes a change in the form (text edit, add new condition, delete condition, editing report requirement, complete review) it doesn't disable the whole form. Instead now when a condition has some change submitted for it, the condition and it's sub conditions are disabled with a load spinner indicating that part is loading. While that condition and it's sub conditions are disabled the user can potentially click on other conditions and start making edits on those.

- I did not make much change around when a user creates a new report requirement or when they delete a report requirement because a modal is used for both of these processes and the modal does not go away until submission is done loading.

- When it comes to moving a condition up or down the condition, it's sub conditions and it's sibling conditions within that specific condition group will be disabled


services/common/src/components/permits/PermitConditionLayer.tsx
```diff
@@ -135,13 +162,14 @@ const PermitConditionLayer: FC<PermitConditionLayerProps> = ({
           canEditPermitConditions={canEditPermitConditions}
           setEditingFormName={setEditingFormName}
           editingFormName={editingFormName}
-          moveUp={currentPosition > 0 ? moveUp : undefined}
-          moveDown={currentPosition < conditionCount - 1 ? moveDown : undefined}
+          moveUp={currentPosition > 0 ? (c) => handleMove(c, true) : undefined}
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -87,39 +98,37 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
...
+            dispatch(reset(editingFormName));
         }
+        handleEdit();
     };
 
     const cancelEdit = () => {
         setIsEditMode(false);
         setEditingFormName(null);
         setIsAddingListItem(false);
+        setActiveConditionId(null);
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/core-api/app/api/mines/permits/permit_amendment/resources/permit_conditions_data.py
```diff
@@ -0,0 +1,37 @@
...
+                noload("vc_credential_exch"),
+                noload("now_application_identity"),
+            )
+            .filter_by(permit_amendment_guid=permit_amendment_guid, deleted_ind=False)
+            .first()
+        )
+        if not permit_amendment:
+            raise NotFound('Permit Amendment not found.')
+        if not str(permit_amendment.mine_guid) == mine_guid:
+            raise BadRequest('Permits mine_guid and supplied mine_guid mismatch.')
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/ReportPermitRequirementForm.tsx
```diff
@@ -129,14 +164,19 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
...
     if (values.mine_report_permit_requirement_id) {
       resp = await dispatch(updateMineReportPermitRequirement({ mineGuid, values }));
     } else {
       resp = await dispatch(createMineReportPermitRequirement({ mineGuid, values }));
     }
     if (resp.payload) {
-      await refreshData(false);
+      await refreshData();
+      clearActiveConditionId(conditionId);
       // this is for an edge case where an existing report is edited and so it has a form on the page
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/ReportPermitRequirementForm.tsx
```diff
@@ -99,18 +124,28 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
...
     }
   }, [canEditPermitConditions]);
 
   const handleDeleteReportRequirement = async ({ mine_report_permit_requirement_id, permit_condition_ids }) => {
     const updateData = async () => {
-      await refreshData();
+      await refreshData(false);
       setIsEditMode(false);
+      clearActiveConditionId(conditionId);
     };
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -87,39 +98,37 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
...
+    const editingFormNameRef = useRef(editingFormName);
+    useEffect(() => {
+        editingFormNameRef.current = editingFormName;
+    }, [editingFormName]);
+
     const startEdit = () => {
         const handleEdit = () => {
             onEdit();
             setEditingFormName(formName);
             setIsEditMode(true);
```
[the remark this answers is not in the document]
asinn134: Ignoring this one

services/common/src/components/permits/PermitConditionViewEdit.tsx
```diff
@@ -42,8 +42,8 @@ interface PermitConditionViewEditProps {
   collapseCategories?: boolean;
   isExpanded?: boolean;
   setSelectedCondition?: (condition: IPermitCondition) => void;
-  editingFormName: string;
-  setEditingFormName: (formName: string) => void;
+  editingFormName: string | null;
+  setEditingFormName: React.Dispatch<React.SetStateAction<string | null>>;
   addingToCategoryCode: string;
   setAddingToCategoryCode: (categoryCode: string) => void;
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/PermitConditions.tsx
```diff
@@ -143,7 +143,34 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
+  };
+
+  // This keeps track of condition ids that are currently being submitted.
+  const [submittingConditionIds, setSubmittingConditionIds] = useState<number[]>([]);
+  const addSubmittingCondition = (id: number) => {
+    setSubmittingConditionIds(prev => [...prev, id]);
+  };
+  const removeSubmittingCondition = (id: number) => {
+    setSubmittingConditionIds(prev => prev.filter(existing => existing !== id));
+  };
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/PermitConditions.tsx
```diff
@@ -143,7 +143,34 @@ const PermitConditions: FC<PermitConditionProps> = ({
   const [selectedCondition, setSelectedCondition] = useState<IPermitCondition | null>(null);
   const [editingFormName, setEditingFormName] = useState<string>();
   const [addingToCategoryCode, setAddingToCategoryCode] = useState<string>();
```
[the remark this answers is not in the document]
asinn134: this has been adjusted

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -70,56 +72,61 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
...
-                    dispatch(reset(editingFormName));
-                    handleEdit()
-                },
-                cancelText: "Continue editing",
-                okText: "Discard"
-            })
-        } else {
-            handleEdit();
+        if (editingFormName) {
+            dispatch(reset(editingFormName));
```
alazar-aot: The removed code showed a Modal (`Modal.confirm`), but now it's calling the dispatch without any confirmation - is this intentional? Users can lose in-progress edits without any way to cancel

services/common/src/components/permits/PermitConditions.tsx
```diff
@@ -141,9 +141,44 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
-  const [addingToCategoryCode, setAddingToCategoryCode] = useState<string>();
-  const [loading, setLoading] = useState(false);
+  const [editingFormName, setEditingFormName] = useState<string | null>(null);
+  const [addingToCategoryCode, setAddingToCategoryCode] = useState<string | null>(null);
+  const [loadingCount, setLoadingCount] = useState(0);
+  const loading = loadingCount > 0;
+  // This keeps track of the number of in-flight loading states. We need this because multiple conditions can be in-flight at once.
+  const setLoading = (value: boolean) => {
+    setLoadingCount(prev => value ? prev + 1 : Math.max(0, prev - 1));
+  };
```
alazar-aot: I think this should be wrapped in a useCallback, right now every component calling usePermitConditions() will re-render whenever PermitConditions re-renders. 

Something like 
```
const setLoading = useCallback((value: boolean) => {
    setLoadingCount(prev => value ? prev + 1 : Math.max(0, prev - 1));
}, []);
```

services/common/src/components/permits/PermitConditions.tsx
```diff
@@ -141,9 +141,44 @@ const PermitConditions: FC<PermitConditionProps> = ({
...
+
+  // This keeps track of condition ids that are currently being submitted. 
+  const [submittingConditionIds, setSubmittingConditionIds] = useState<Record<number, number>>({});
+  const addSubmittingCondition = (id: number) => {
```
alazar-aot: Same with this (and all the others in this file)

services/common/src/components/permits/PermitConditionLayer.tsx
```diff
@@ -205,4 +261,17 @@ const PermitConditionLayer: FC<PermitConditionLayerProps> = ({
...
+  prev.isEditing === next.isEditing &&
+  prev.editingFormName === next.editingFormName &&
+  prev.refreshData === next.refreshData &&
+  prev.handleMoveCondition === next.handleMoveCondition &&
+  prev.isExpanded === next.isExpanded &&
+  prev.currentPosition === next.currentPosition &&
+  prev.conditionCount === next.conditionCount &&
+  prev.canEditPermitConditions === next.canEditPermitConditions &&
+  prev.isInsideSubmittingCondition === next.isInsideSubmittingCondition &&
+  prev.isInsideActiveCondition === next.isInsideActiveCondition
```
alazar-aot: The handleMove function defined above closes over `siblingIds` from the render closure, but this doesn't check `siblingIds`, so I think it can return true (skip re-render) even when `siblingIds` changes and should re-render.... meaning handleMove would end up with a stale copy of the sibling ID?

I think this should use a ref, like how we've done elsewhere in the code with `activeConditionId`.

services/common/src/components/permits/PermitConditionStatus.tsx
```diff
@@ -61,7 +88,22 @@ export const PermitConditionStatus: FC<PermitConditionStatusProps> = ({
         },
         width: 2048,
         content: (props) => {
-          const value = { mineGuid, permitGuid, latestAmendment, previousAmendment, currentAmendment, loading, setLoading, refreshData };
+          const value = {
```
alazar-aot, line 88: This is manually mirroring the `PermitConditionsContextType` interface, anytime we update the interface we'd need to remember to update here. Instead this should use a helper function or export/spread the data so it reflects the interface automatically

alazar-aot (changes requested):
Looks great overall, just a few things that caught my eye

services/common/src/components/permits/PermitConditionForm.tsx
asinn134: The modal.confirm has been added back for this now.

services/common/src/components/permits/PermitConditions.tsx
asinn134: This is now wrapped in userCallBack

services/common/src/components/permits/PermitConditions.tsx
asinn134: The others in this file are also wrapped in userCallBack now

services/common/src/components/permits/PermitConditionLayer.tsx
asinn134: Changed to have siblingId to use ref now

services/common/src/components/permits/PermitConditionStatus.tsx
asinn134: This makes use of spread now instead.

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -133,6 +172,8 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
 
     const handleSubmit = async (values) => {
```
alazar-aot, line 172: This should probably be in a try/finally, otherwise I think the condition form could end up being frozen if an error is thrown

services/common/src/components/permits/PermitConditionForm.tsx
```diff
@@ -150,22 +191,40 @@ const PermitConditionForm: FC<PermitConditionFormProps> = ({
...
     };
     const handleAddListItem = () => {
         setIsAddingListItem(true);
+        setActiveConditionId(condition.permit_condition_id);
     };
 
     const handleDelete = async () => {
         const title = "Delete Condition";
 
         const onSubmit = async () => {
```
alazar-aot, line 224: Also consider a try/finally here, similar issue

services/common/src/components/permits/PermitConditionLayer.tsx
```diff
@@ -99,14 +136,18 @@ const PermitConditionLayer: FC<PermitConditionLayerProps> = ({
   const handleSaveListItem = async () => {
     await refreshData();
```
alazar-aot, line 137: This should pass false (refreshData(false)) to avoid accidentally closing another conditions edit form, potential to lose unsaved changes without warning

services/common/src/components/permits/PermitConditionLayer.tsx
```diff
@@ -99,14 +136,18 @@ const PermitConditionLayer: FC<PermitConditionLayerProps> = ({
...
 
-  const moveUp = async (condition: IPermitCondition | IStandardPermitCondition) => {
-    await handleMoveCondition(condition, true);
-  };
+  const siblingIdsRef = useRef(siblingIds);
+  siblingIdsRef.current = siblingIds;
 
-  const moveDown = async (condition: IPermitCondition | IStandardPermitCondition) => {
-    await handleMoveCondition(condition, false);
+  const handleMove = async (condition: IPermitCondition | IStandardPermitCondition, isMoveUp: boolean) => {
```
alazar-aot, line 145: This should probably be in a try/finally block - if an unhandled rejection or error occurs the loadingCount will be incremented and never decremented, so loading would stay true forever

services/common/src/redux/reducers/permitReducer.ts
```diff
@@ -111,6 +128,29 @@ export const permitReducer = (state: PermitState = initialState, action) => {
...
+        return {
+          ...permitAmendment,
+          ...amendmentFields,
+          conditions: updateChangedConditions(amendmentFields.conditions ?? [],
+            permitAmendment.conditions ?? [])
+        };
+      };
+      return {
+        ...state,
+        permits: state.permits.map((permit) =>
```
alazar-aot, line 151: This is only updating `state.permits`, consider also updating `state.latestPermitAmendments` and `state.permitAmendments` like lines 120-124. I think it we don't update them all, then other selectors will serve stale data after this action fires

alazar-aot (changes requested):
A few really small tweaks and then I think this is gravy!

services/common/src/redux/reducers/permitReducer.ts
asinn134: This has been adjusted to update `state.latestPermitAmendments` and `state.permitAmendments` as well now.

services/common/src/components/permits/PermitConditionLayer.tsx
asinn134: This now has a try/finally block added to it.

services/common/src/components/permits/PermitConditionForm.tsx
asinn134: This now has a try/finally block added to it.

services/common/src/components/permits/PermitConditionForm.tsx
asinn134: This now has a try/finally block added to it.

services/common/src/components/permits/PermitConditionLayer.tsx
asinn134: I ended up leaving the false out of this `refreshData` as it was causing issues with the discard modal popping up. 

I tested the behaviour around this without the false in `refreshData` and found the discard modal will pop up when there are unsubmitted changes when closing the condition edit form to move to another condition so I think it should be ok.

simensma-fresh (approved):

services/common/src/components/permits/PermitConditionLayer.tsx
alazar-aot: Ok great - if that discard Modal is popping up then we're good to go! I'll mark it as resolved

alazar-aot (approved):
Great work!

PR #3933, opened 2026-06-08
[MDS-6921] Show tables and images in notice of work document search results

simensma-fresh:
## Summary
This PR adds artifact-first Notice of Work (NoW) document search results, focused on extracted figures, images, and tables.

At a high level:
- The indexing pipeline now extracts tables, figures and images from NoW documents. 
- Search results in the NoW UI now render artifact previews and formatted table content directly in the result card, instead of relying only on raw text snippets.
- An image of the artifact is pushed to S3 (through docman) to enable viewing of the artifact in search results
- Some transformations happen, e.g. we attempt to rotate images / tables that are vertical based on hint given by document intelligence

## What Changed
- Permit service: Added backend pipeline support to extract artifact regions (tables/figures), attach page and bounding-box context, and include artifact metadata in indexed chunks.
- Added artifact upload/registration flow so extracted preview images are stored and retrievable (docman)
- Updated NoW search result rendering to show:
  - Artifact image preview near the top of each result card.
  - Expandable formatted table section for table artifacts.
  - Caption for the artifact if Document Intelligence finds one. If not, we fall back to an AI generated caption
- AI generated description of the image / table gets indexed to enable images to show up in search results
- Added in a rough categorization piece to identify whether images are maps, a site Photo or something else. 

## Why Document Manager
To keep it as the central place to manage files in S3, and have a record of the artifact in our DB.

## Screenshots
image

image

image

image


asinn134 (dismissed):
Nice work Simen 👍

services/core-api/app/api/search/search/now_application_search_service.py
```diff
@@ -60,14 +71,106 @@ def search(self, now_application_guid: str, search_params: dict):
...
+            if not artifact_document_manager_guid:
+                continue
+
+            token = token_cache.get(artifact_document_manager_guid)
+            if token is None:
+                token = str(DocumentManagerService.create_download_token(artifact_document_manager_guid))
+                token_cache[artifact_document_manager_guid] = token
+
+            if self.docman_base_url:
+                meta["artifact_presigned_url"] = self._resolve_artifact_presigned_url(token)
```
alazar-aot, line 134: It looks like `_resolve_artifact_presigned_url` is called for every artifact hit, and has a 10 second timeout before the event is yielded - I'm wondering if multiple artifact hits in one event could stall the stream? 

Would it be ill-advised to add a pre-signed URL cache alongside the `token_cache`?

services/document-manager/backend/app/services/object_store_storage_service.py
```diff
@@ -101,6 +101,26 @@ def generate(result):
         except ClientError as e:
             raise Exception(f'Failed to download the file: {e}')
 
+    def generate_download_presigned_url(self, path, display_name, as_attachment, version_id=None, expires_in=300):
```
alazar-aot: I feel like the default TTL of 5 mins is too short - what happens if a user leaves the results page open for longer than that? Would the images break once the TTL is reached (forcing the user to re-search)?

services/permits/app/pipelines/document_search/artifact_registration_client.py
```diff
@@ -0,0 +1,317 @@
...
+        upload_info = self._initialize_upload(
+            token,
+            file_bytes,
+            file_name,
+            self._resolve_folder(upload_data, now_application_guid, source_document_manager_guid),
+        )
+        artifact_document_manager_guid, upload_id, upload_parts = self._parse_upload_info(upload_info)
+        completed_parts = self._upload_parts(file_bytes, upload_parts)
+        complete_info = self._complete_upload(token, artifact_document_manager_guid, upload_id, completed_parts)
+        logger.warning(
```
alazar-aot: I feel like this should be debug instead of warn (since it's the 'happy path')

services/permits/app/pipelines/document_search/document_manager_client.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import os
+from typing import Optional
+
+import requests
+from app.pipelines.document_search.artifact_registration_client import _build_oauth_session
```
alazar-aot: Does importing a private method cause any issues or couple things too tightly? I do see it used in a few other places in the codebase, but that's not something I'm used to seeing (would expect a utility module or something like that)

services/permits/app/pipelines/document_search/create_search_index.py
```diff
@@ -85,10 +88,26 @@
...
+        message = str(e)
+        if "CannotChangeExistingField" in message:
+            raise RuntimeError(
+                f"Index '{index.name}' requires an immutable schema change. "
+                "Delete and recreate the index manually before re-running this command."
+            ) from e
+        if (
+            "ResourceNameAlreadyInUse" in message
+            or "CannotCreateExistingIndex" in message
+            or "Not Found" in message
```
alazar-aot: I feel like "Not Found" being mapped to "Index already exists, skipping" is a bit odd, especially since we call `get_index()` right after (which I think would result in an error?)

alazar-aot (changes requested):
Noticed a few small things, nothing major though so great work Simen, was cool to see how you went about implementing this!

asinn134 (approved):

alazar-aot (approved):
Looks great - let's get this out there!!

PR #3935, opened 2026-06-11
[MDS-6283] Fix 'Form-Filled' PDF UI Bug

alazar-aot:
## Objective 
[MDS-6283](https://bcmines.atlassian.net/browse/MDS-6283)

It was noticed that when opening and viewing a form-filled PDF using the "View in Document Viewer" functionality (Filesystem-provider service), the form-filled fields were not displaying in the correct location:

image

This PR resolves that issue by overriding`position: static;` on `.e-pdfviewer-formFields`, so that it sits correctly within the Document Viewer. 

I believe this issue was introduced when we upgraded to `Syncfusion 31.2.5` in the backend, my theory is that upgrade changed the HTML structure for text input form fields in a way that exposed a pre-existing CSS conflict in the codebase - our global `input { position: relative }` rule in `themeOverride.scss` had previously been silently overridden by Syncfusion's old inline `position: absolute` style, but the new wrapper div structure removed that inline override, leading to the issue. 

Observe how the form-filled fields now sit correctly in the viewer:

image


services/common/src/components/syncfusion/DocumentViewer.tsx
```diff
@@ -121,7 +121,7 @@ export const ViewPdf: React.FC<ViewPDFProps> = ({
       serviceUrl={pdfViewerServiceUrl}
       documentPath={documentPath}
       ajaxRequestSettings={ajaxRequestSettings}
-      style={{ display: "block", height: "80vh" }}
+      // height="80vh"
```
simensma-fresh: Can probably remove instead of comment out if not needed

simensma-fresh (commented):
Woho! Good catch

simensma-fresh (dismissed):

services/common/src/components/syncfusion/DocumentViewer.tsx
alazar-aot: Shoot - that was actually my bad, it shouldn't be commented out. I've reverted it back - nice catch

asinn134 (dismissed):

asinn134 (approved):

asinn134:
Nice findings 👍

simensma-fresh (approved):

PR #3940, opened 2026-06-23
[MDS-6946] Adjusted major mine contact that is not general contact to not show in minespace

asinn134:
## Objective 

[MDS-6946](https://bcmines.atlassian.net/browse/MDS-6946)

- There was an issue where a Minespace MCM contact that was created as a major mine contact, but not a general contact, when there is no mine region selected for it, the contact would appear in the contact section for all Minespace major mines. The expected behaviour is when a Minespace MCM contact is not a general contact it shouldn't show up in Minespace at all.
- The fix was to add `is_general_contact=True` check in the filter of the `find_majoy_mine_office` function. From what I can tell this logic is only really called within Minespace.


alazar-aot (approved):
Looks good!

PR #3941, opened 2026-06-26
[MDS-6946] Adjustment to make general contact option determine if it shows in minespace

asinn134:
## Objective 

[MDS-6946](https://bcmines.atlassian.net/browse/MDS-6946)

- Adjusted the MCM contact creation modal to have general contact option be available even if user selects no to the major mine contact option.
- The general contact option is now the main indicator to determine if a contact shows up in minespace's frontend or not.
- Fixed a bug that occurs when a contact is major contact, a general contact, and region is selected, the contact will show up for all regional mines under the "General Ministry Contacts" section. It now should only show up for the regional mine that belongs in the selected region.

### DIFF
--- a/services/core-api/tests/ministry_contacts/models/test_ministry_contact_model.py
+++ b/services/core-api/tests/ministry_contacts/models/test_ministry_contact_model.py
@@ -50,19 +50,19 @@ def test_find_ministry_contacts_by_mine_region_includes_non_general_regional_con
 
 def test_find_ministry_contacts_by_mine_region_regional_mine_includes_general_contact_via_union(db_session):
     contact = MinistryContactFactory(is_major_mine=True, is_general_contact=True, mine_region_code=None)
-    results = MinistryContact.find_ministry_contacts_by_mine_region('NE', False)
+    results = MinistryContact.find_ministry_contacts_by_mine_region(contact.mine_region_code, False)
     assert contact.contact_guid in [c.contact_guid for c in results]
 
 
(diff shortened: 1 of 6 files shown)


services/core-api/tests/ministry_contacts/models/test_ministry_contact_model.py
```diff
@@ -49,8 +49,8 @@ def test_find_ministry_contacts_by_mine_region_includes_non_general_regional_con
 
 
 def test_find_ministry_contacts_by_mine_region_regional_mine_includes_general_contact_via_union(db_session):
-    contact = MinistryContactFactory(is_major_mine=True, is_general_contact=True)
-    results = MinistryContact.find_ministry_contacts_by_mine_region(contact.mine_region_code, False)
+    contact = MinistryContactFactory(is_major_mine=True, is_general_contact=True, mine_region_code=None)
+    results = MinistryContact.find_ministry_contacts_by_mine_region('NE', False)
```
alazar-aot: Approving, but do we want the region hardcoded to NE?

alazar-aot (dismissed):
Approved but pls note my comment! Not a blocker for merging, just want to make sure that was intentional

services/core-api/tests/ministry_contacts/models/test_ministry_contact_model.py
asinn134: This has been adjusted to use `contact.mine_region_code` now.

alazar-aot (approved):
Nice!

PR #3944, opened 2026-06-30
[MDS-6953] - MCM contact management validation

matbusby-fw:
## Objective 

[MDS-6953](https://bcmines.atlassian.net/browse/MDS-6953)

## Form validation preventing updating of various MCM contact types
- Discovered that the current form validation was preventing the submission of changes to various contact types.  This was particularly difficult for "Report Designated Contacts" that didn't have a region, but the region field was both required and disabled.  
  - Adjusted the validation logic to relax certain requirements based on discussion with Rebecca

### Context / Background

  Recently, EMLI contact management transitioned to using dynamically managed dynamic distribution lists (e.g.,  Incidents ,  Variances , etc.). However, after this change:

  1. On-call inspectors were not receiving incident notifications: This occurred because the new  Incidents  distribution list started out empty, and administrators were blocked from editing the on-call inspector contact (
  mineincidents@gov.bc.ca  — an  RDC  contact type) to add it to the distribution list.
  2. Mine Region was required and disabled on edit: Users were unable to save  RDC  contacts (which do not have/need a mine region) because the form validation required it, but the dropdown was disabled during editing. Additionally, the
  backend  PUT  endpoint did not support updating the  mine_region_code , which is why the field was disabled in the first place.

### DIFF
--- a/services/core-api/app/api/ministry_contacts/models/ministry_contact_type.py
+++ b/services/core-api/app/api/ministry_contacts/models/ministry_contact_type.py
@@ -15,7 +15,7 @@ class MinistryContactType(AuditMixin, Base):
 
     @classmethod
     def get_all(cls):
-        return cls.query.filter_by(active_ind=True).all()
+        return cls.query.all()
 
     @classmethod
     def get_active(cls):
--- a/services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
+++ b/services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
@@ -15,11 +15,11 @@ class MinistryContactResource(Resource, UserMixin):
     parser = reqparse.RequestParser()
     parser.add_argument('first_name', type=str, trim=True, help='MCM First name', location='json')
     parser.add_argument('last_name', type=str, trim=True, help='MCM Last name.', location='json')
-    parser.add_argument('email', type=str, help='MCM email.', required=True, location='json')
+    parser.add_argument('email', type=str, help='MCM email.', required=False, location='json')
     parser.add_argument(
         'is_general_contact', type=bool, help='is is_general_contact? true/false', location='json')
     parser.add_argument(
-        'phone_number', type=str, help='MCM phone number', required=True, location='json')
+        'phone_number', type=str, help='MCM phone number', required=False, location='json')
     parser.add_argument(
         'fax_number', type=str, help='MCM Regional Office fax number', location='json')
     parser.add_argument(
@@ -33,6 +33,8 @@ class MinistryContactResource(Resource, UserMixin):
         help='MCM Regional Office mailing address line 2',
         location='json')
     parser.add_argument(
+        'mine_region_code', type=str, trim=True, help='Mine region code', location='json')
+    parser.add_argument(
         'distribution_list_guids', type=list, location='json', default=[])
 
     @api.doc(description='Update an existing MCM contact.')
@@ -45,8 +47,18 @@ class MinistryContactResource(Resource, UserMixin):
 
         data = self.parser.parse_args()
 
+        if not data.get('email') and not data.get('phone_number'):
+            raise BadRequest('Either email or phone number is required.')
+
         distribution_list_guids = data.pop('distribution_list_guids', [])
         
+        new_region_code = data.get('mine_region_code')
+        if new_region_code and new_region_code != contact.mine_region_code:
+            if contact.emli_contact_type_code == 'ROE':
+                existing_roe = MinistryContact.find_ministry_contact('ROE', new_region_code)
+                if existing_roe and existing_roe.contact_guid != contact.contact_guid:
+                    raise BadRequest('Error: Restricted to one Regional Office contact by mine region.')
+
         for key, value in data.items():
             setattr(contact, key, value)
 
--- a/services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
+++ b/services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
@@ -47,12 +47,12 @@ class MinistryContactResource(Resource, UserMixin):
 
         data = self.parser.parse_args()
 
-        if contact.emli_contact_type_code == 'RDC':
-            if not data.get('email'):
-                raise BadRequest('Email is required.')
-        else:
-            if not data.get('email') or not data.get('phone_number'):
-                raise BadRequest('Both email and phone number are required.')
+        if not data.get('email'):
+            raise BadRequest('Email is required.')
+
+        if contact.emli_contact_type_code != 'RDC' and not data.get('is_general_contact'):
+            if not data.get('phone_number'):
+                raise BadRequest('Phone number is required.')
 
         distribution_list_guids = data.pop('distribution_list_guids', [])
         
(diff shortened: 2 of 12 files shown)


services/core-web/src/components/Forms/MinistryContacts/MinistryContactForm.tsx
```diff
@@ -89,6 +93,26 @@ export const MinistryContactForm: FC<MinistryContactFormProps> = (props) => {
         reduxFormConfig={{
           touchOnBlur: false,
           enableReinitialize: true,
+          validate: (values) => {
+            const errors: any = {};
+            const isMajorOrGeneral = values.is_major_mine || values.is_general_contact;
```
alazar-aot: This requires `first_name` and `last_name` whenever `is_major_mine` or `is_general_contact` is set, but those fields are unmounted for `ROE` and `MMO` contact types - what happens if a user picks one of those two contact types and also checks "Major Mine Contact" or "General Contact"? 

I think the form could get stuck with errors on fields the user can't see or fix, and saving would do nothing?

Maybe I'm understanding it wrong, but I think the ROE and MMO contact types should be excluded from the `isMajorOrGeneral` check

services/core-api/app/api/ministry_contacts/models/ministry_contact_type.py
```diff
@@ -15,7 +15,7 @@ def __repr__(self):
 
     @classmethod
     def get_all(cls):
-        return cls.query.filter_by(active_ind=True).all()
+        return cls.query.all()
```
alazar-aot, line 18: I'm wondering - does this mean the shared static-content payload (`core_static_content_resource.py`) that's loaded by every Core/MineSpace session now always includes inactive contact types? If so, do we want that or does that have any unintended impacts?

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
```diff
@@ -15,11 +15,11 @@ class MinistryContactResource(Resource, UserMixin):
     parser = reqparse.RequestParser()
     parser.add_argument('first_name', type=str, trim=True, help='MCM First name', location='json')
     parser.add_argument('last_name', type=str, trim=True, help='MCM Last name.', location='json')
-    parser.add_argument('email', type=str, help='MCM email.', required=True, location='json')
+    parser.add_argument('email', type=str, help='MCM email.', required=False, location='json')
```
alazar-aot: If we make email optional here, there are several reads in `mine_report.py` and `mine_report_notification.py` that assume `.email` is always set and feed that value straight into the report-submission notification recipient lists... I think having a phone-only ROE contact will break or error PRR/CRR submission emails for that region. 

We probably want to guard those for the case when email is None - or at least test out that behaviour

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
```diff
@@ -15,11 +15,11 @@ class MinistryContactResource(Resource, UserMixin):
...
     parser.add_argument(
         'is_general_contact', type=bool, help='is is_general_contact? true/false', location='json')
     parser.add_argument(
-        'phone_number', type=str, help='MCM phone number', required=True, location='json')
+        'phone_number', type=str, help='MCM phone number', required=False, location='json')
```
alazar-aot, line 22: Same issue as line 18, worth testing to see if this causes any errors

services/core-web/src/components/Forms/MinistryContacts/MinistryContactForm.tsx
```diff
@@ -40,12 +40,13 @@ const regionalOfficeCode = "ROE";
 const majorMineOfficeCode = "MMO";
 const chiefPermittingCode = "CHP";
 const chiefInspectorCode = "CHI";
+const reportDesignatedContactCode = "RDC";
 const officeCodes = [regionalOfficeCode, majorMineOfficeCode];
```
alazar-aot: These codes are defined here as well as in `MineSpaceMinistryContactManagement.tsx:88`, meaning we'd need to remember to update in both places in the event they ever change. 

If we don't think these are likely to change then this should be fine, otherwise I think it's worth putting them in a shared constant so any future changes don't need updating in two places!

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
```diff
@@ -32,6 +32,8 @@ class MinistryContactResource(Resource, UserMixin):
         type=str,
         help='MCM Regional Office mailing address line 2',
         location='json')
+    parser.add_argument(
+        'mine_region_code', type=str, trim=True, help='Mine region code', location='json')
```
alazar-aot, line 36: It looks like this field is optional - what happens in the `setattr` loop below if a value is omitted for this argument? 

It looks like the frontend always sends a value, but if for some reason this argument was missing, it looks like we could null the contact's existing region because there's no server-side guard... probably worth ensuring we preserve the current value for region in that scenario ?

alazar-aot (changes requested):
I think this is good overall, some of the below comments might be overly cautious but I know this email functionality is a bit tricky so wanted to raise what I saw just in case

services/core-web/src/components/Forms/MinistryContacts/MinistryContactForm.tsx
asinn134: Looks like Mat addressed this.

services/core-api/app/api/ministry_contacts/models/ministry_contact_type.py
asinn134: This should be fine, the frontend function `filteredContactTypes` in `MinistryContactForm.tsx` filters out all inactive contact types.

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
asinn134: Mat reverted this change

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
asinn134: This looks fine. The notification code only ever reads .email to build recipient lists — it never touches phone_number at all, so making it optional won't break anything there.

services/core-web/src/components/Forms/MinistryContacts/MinistryContactForm.tsx
asinn134: This has been adjusted to include a constant

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
asinn134: This should be fine. The backend PUT endpoint to update a MCM contact has this

```
        for key, value in data.items():
            if key == 'mine_region_code' and 'mine_region_code' not in (request.json or {}):
                continue
            setattr(contact, key, value)
```

So when a mine_region_code is not provided from the front end it doesn't touch the existing contact's region.

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
alazar-aot: Awesome!

services/core-api/app/api/ministry_contacts/resources/ministry_contact.py
alazar-aot: Sounds good

alazar-aot (approved):
Thanks for making the changes/addressing my comments, should be good to go!

PR #3946, opened 2026-07-03
[MDS-6932] Fix for placeholder text showing up in issued permits

asinn134:
## Objective 

[MDS-6932](https://bcmines.atlassian.net/browse/MDS-6932)

- There was a bug where core generated permits that are issued would continue to show the condition data variables placeholder text in the permit condition page.

- The changes in this PR makes use of some existing functions that change the placeholder text of condition data variables to it's actual values. So now when the permit is issued in Core the placeholder text will be changed to the actual values. The existing functions were being used in the workflow that occurs when users download the draft permit document.

### DIFF
--- a/services/core-api/app/api/now_applications/resources/now_application_status_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_status_resource.py
@@ -20,6 +20,8 @@ from app.api.services.issue_to_orgbook_service import OrgBookIssuerService
 from app.api.mines.mine.models.mine_type import MineType
 from app.api.mines.mine.models.mine_type_detail import MineTypeDetail
 from app.api.mines.permits.permit_conditions.tasks import export_and_index_permit_amendments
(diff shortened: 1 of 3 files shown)


services/core-api/app/api/now_applications/resources/now_application_status_resource.py
```diff
@@ -20,6 +20,8 @@
 from app.api.mines.mine.models.mine_type import MineType
 from app.api.mines.mine.models.mine_type_detail import MineTypeDetail
 from app.api.mines.permits.permit_conditions.tasks import export_and_index_permit_amendments
+from app.api.now_applications.now_template_transformer import (
+    calculate_liability, transform_variables_to_data, replace_condition_value_with_data)
```
asinn134, line 24: These imported functions are the existing functions

matbusby-fw (approved):
Looks great!

PR #3952, opened 2026-07-13
[MDS-6956] - File Viewer Redis Cache Migration & S3 Enhancements

matbusby-fw:
## Description
This PR addresses the intermittent failures and pod restarts associated with the Core file viewer. Previously, the file viewer used `IMemoryCache` (in-memory caching of loaded files), which caused serving pods to exceed their memory limits and get OOM-killed (restarted) under heavy load.

To resolve this:
1. migrated the cache layer to use a distributed Redis cache (`IDistributedCache`), which offloads the RAM overhead to a dedicated Redis instance and supports state sharing across scaled pods.
2. upgraded the `AWSSDK.S3` dependency to `4.0.17` and implemented null-safety adjustments required by the newer SDK types.
3. resolved a concurrency issue during folder downloads by using request-scoped unique temp directories (`Guid`).
4. added elegant exception handling for missing S3 objects (404/NoSuchKey) to return a graceful `NotFound` result instead of an unhandled `500` error.

## Related Tickets & PRs
- GitOps PR: (Insert GitOps PR Link here)

## Acceptance Criteria Checklist
- [x] Memory cache is moved to Redis from internal pod memory.
- [x] Safe fallback to local in-memory caching is maintained for local development.
- [x] Pods can scale seamlessly without cache inconsistency.

## Changes Made
- **[Startup.cs](file:///home/mat/Code/mds/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Startup.cs)**: Updated service registration to use StackExchange Redis Cache when `CACHE_REDIS_HOST` is present; falls back to `AddDistributedMemoryCache` for local development.
- **[PdfViewerController.cs](file:///home/mat/Code/mds/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs)**: Swapped `IMemoryCache` injection for `IDistributedCache`. Modified `ImportAnnotations` to fetch files directly from S3 using the file provider client, removing state reliance on local pod filesystems.
- **[AmazonS3FileProvider.cs](file:///home/mat/Code/mds/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Models/AmazonS3FileProvider.cs)**:
  - Configured `TransferUtility` with `ConcurrentServiceRequests = 10` for optimized parallel uploads and directory downloads.
  - Implemented request-scoped temporary directories using `Guid` to prevent parallel download conflict errors.
  - Added null safety checks for metadata and size properties required by `AWSSDK.S3` version 4.
  - Handled `AmazonS3Exception` (404/NoSuchKey) to return `null` instead of throwing unhandled exceptions.
- **[EJ2AmazonS3ASPCoreFileProvider.csproj](file:///home/mat/Code/mds/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/EJ2AmazonS3ASPCoreFileProvider.csproj)**: Upgraded `AWSSDK.S3` package to `4.0.17` and added `Microsoft.Extensions.Caching.StackExchangeRedis` version `7.0.7`.

## How to Test
1. Run the application locally (without `CACHE_REDIS_HOST` env set) and verify the file viewer functions correctly using the local in-memory fallback.
2. Run with a local Redis connection and inspect keys starting with `FilesystemProvider:` as documents are viewed.
3. Test loading a non-existent document key and verify the response is a graceful HTTP `404` instead of a `500` server exception.
4. Execute concurrent folder downloads to verify that unique Guid-based temp directories are created and cleaned up without locking errors.

### DIFF
--- a/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs
+++ b/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs
@@ -70,7 +70,10 @@ namespace EJ2AmazonS3ASPCoreFileProvider.Controllers
                         {
                             return NotFound(jsonObject["document"] + " is not found");
                         }
-                        fsr.FileStream.CopyTo(stream);
+                        using (fsr.FileStream)
+                        {
+                            fsr.FileStream.CopyTo(stream);
+                        }
                     }
                     else if (jsonObject.TryGetValue("document", out var base64))
                     {
@@ -230,6 +233,7 @@ namespace EJ2AmazonS3ASPCoreFileProvider.Controllers
                         return NotFound(jsonObject["importedData"] + " is not found");
                     }
 
+                    using (fsr.FileStream)
                     using (MemoryStream ms = new MemoryStream())
                     {
                         fsr.FileStream.CopyTo(ms);
--- a/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs
+++ b/services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs
@@ -191,22 +191,24 @@ namespace EJ2AmazonS3ASPCoreFileProvider.Controllers
         [HttpPost]
         [Route("ImportAnnotations")]
         [Authorize("View")]
-        // NOTE: This is not implemented properly as it will need to get the document from the S3 bucket.
         public IActionResult ImportAnnotations([FromBody] Dictionary<string, string> jsonObject)
         {
             PdfRenderer pdfviewer = new PdfRenderer(_cache);
-            string jsonResult = string.Empty;
-            object JsonResult;
+
             if (jsonObject != null && jsonObject.ContainsKey("fileName"))
             {
-                string documentPath = GetDocumentPath(jsonObject["fileName"]);
-                if (!string.IsNullOrEmpty(documentPath))
+                string path = Path.GetDirectoryName(jsonObject["fileName"]) + "/";
+                string filename = Path.GetFileName(jsonObject["fileName"]);
+                FileStreamResult fsr = this.operation.Download(path, new string[] { filename });
+
+                if (fsr == null)
                 {
-                    jsonResult = System.IO.File.ReadAllText(documentPath);
+                    return NotFound(jsonObject["fileName"] + " is not found");
                 }
-                else
+
+                using (StreamReader reader = new StreamReader(fsr.FileStream))
                 {
-                    return this.Content(jsonObject["document"] + " is not found");
+                    return Content(reader.ReadToEnd());
                 }
             }
             else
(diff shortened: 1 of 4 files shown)


services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Controllers/PdfViewerController.cs
```diff
@@ -191,50 +191,55 @@ public IActionResult ExportAnnotations([FromBody] Dictionary<string, string> jso
...
-                        JsonResult = pdfviewer.ImportAnnotation(jsonObject);
-                        return Content(JsonConvert.SerializeObject(JsonResult));
+                        return NotFound(jsonObject["importedData"] + " is not found");
                     }
-                    else
+
+                    using (MemoryStream ms = new MemoryStream())
                     {
-                        return this.Content(jsonObject["document"] + " is not found");
+                        fsr.FileStream.CopyTo(ms);
```
alazar-aot, line 239: The FileStream isn't closed or disposed of, meaning each .xfdf import through this path would leak the S3 stream. Should be wrapped in `using`, (eg `using (fsr.FileStream)`) like how it's done lines 202-209

services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Models/AmazonS3FileProvider.cs
```diff
@@ -696,23 +731,38 @@ public virtual async Task<FileStreamResult> DownloadAsync(string path, string[]
...
                 {
                     GetBucketList();
                     ListingObjectsAsync("/", RootName.Replace("/", "") + path, false).Wait();
-                    Stream stream = await fileTransferUtility.OpenStreamAsync(bucketName, RootName.Replace("/", "") + path + Names[0]);
+                    var streamRequest = new TransferUtilityOpenStreamRequest
+                    {
+                        BucketName = bucketName,
+                        Key = RootName.Replace("/", "") + path + Names[0]
+                    };
+                    var streamResponse = await fileTransferUtility.OpenStreamWithResponseAsync(streamRequest);
```
alazar-aot, line 746: I had Claude review the new S3 version we're switching to and it flagged this line, as well as line 712 (`GetImage`) as potential issues,

In the new version, `OpenStreamWithResponseAsync` returns a `TransferUtilityOpenStreamResponse`, which implements `IDisposable`. We only keep `.ResponseStream`, the `streamResponse` itself is never disposed of. 

It sounds like this could keep the network connection open, and risk causing issues under high-traffic. 

I think it would just be a matter of wrapping these in `using`, to make sure the outer `streamResponse` gets dealt with.

alazar-aot (changes requested):
Noticed two potential issues with how we handle open streams, otherwise this looks great!

services/filesystem-provider/ej2-amazon-s3-aspcore-file-provider/Models/AmazonS3FileProvider.cs
matbusby-fw: Ah, good call!

alazar-aot (approved):
Fabulous 💅

PR #3953, opened 2026-07-16
[MDS-6974] - Strongly type existing orgbook search api gateway endpoints

Jsyro:
Strongly type whats already there so that we can change data source to new Registries api.


services/core-api/app/api/orgbook/resources/orgbook_resources.py
```diff
@@ -1,54 +1,44 @@
...
+from app.api.orgbook.response_models import ORGBOOK_SEARCH_RESULT, ORGBOOK_CREDENTIAL, ORGBOOK_VERIFICATION_RESPONSE
 
 
 class SearchResource(Resource):
+
     @api.doc(
         description='Search OrgBook.',
         params={'search': 'The search term to use when searching OrgBook.'})
-    @requires_role_view_all
+    @api.marshal_with(ORGBOOK_SEARCH_RESULT, code=200)
```
matbusby: Wont the .search() method return a list?  If that's the case, I think we'd need an `as_list=True` here, no?

services/core-api/app/api/services/orgbook_service.py
```diff
@@ -1,18 +1,87 @@
...
+    BusinessSearchResult,
+    OrgBookCredential,
+    OrgBookSearchResult,
+    OrgBookVerificationResponse,
+)
+
+
+class BCRegistriesService():
+
+    def search(self, search: str) -> List[BusinessSearchResult]:
```
matbusby, line 17: Looks like the search here got missed being added to the payload.

services/core-api/app/api/services/orgbook_service.py
```diff
@@ -1,18 +1,87 @@
...
+        BC_REGISTRIES_API_URL = "https://sandbox.api.connect.gov.bc.ca/registry-search"
```
matbusby: Shall we add these as env variables on OpenShift?

matbusby (changes requested):
Mostly looking great!  Just noticed a couple things that might need some attention.

alazar-aot (dismissed):
Approving, but noting that:
1. With these changes the BC Reg lookup/API call is made every OrgBook search (so an entire extra call), I know we need this for testing but this should be gated behind a flag/env var to avoid extra traffic in Prod
2. The code currently silences any errors regarding missing `BC_REGISTRIES_SECRET_TOKEN` - fine for testing but needs to be updated if this is going to be kept
3. The SonarQube warnings should be addressed when this work is complete (if being kept)

alazar-aot (approved):
Approved again!

PR #3956, opened 2026-07-21
[MDS-6930] fix for merge contacts address not showing

asinn134:
## Objective 

[MDS-6930](https://bcmines.atlassian.net/browse/MDS-6930)

- There was an issue on the page to merge contacts in Core where the address of all contacts would not show up even when they do have an address.
- The fix was mainly setting up the contact address in pgsync, and adjusting the front and backend to retrieve and display the data properly

### DIFF
--- a/services/core-api/app/api/search/search/search_transformers.py
+++ b/services/core-api/app/api/search/search/search_transformers.py
@@ -66,12 +66,16 @@ def prepare_party_source(source):
                 'permit_no': None,
             })
     
+    address = source.get('address', [])
+    if address and not isinstance(address, list):
+        address = [address]
+
     prepared = dict(source)
     prepared['name'] = f"{first_name} {party_name}".strip() if first_name else party_name
     prepared['party_orgbook_entity'] = source.get('party_orgbook_entity') #.get() returns None by default
     prepared['business_role_appts'] = []
     prepared['mine_party_appt'] = transformed_appts
-    prepared['address'] = source.get('address', [])
+    prepared['address'] = address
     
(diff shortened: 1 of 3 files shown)


services/core-api/app/api/search/search/search_transformers.py
```diff
@@ -71,7 +71,7 @@ def prepare_party_source(source):
     prepared['party_orgbook_entity'] = source.get('party_orgbook_entity') #.get() returns None by default
     prepared['business_role_appts'] = []
     prepared['mine_party_appt'] = transformed_appts
-    prepared['address'] = []
+    prepared['address'] = source.get('address', [])
```
alazar-aot: What happens if a party is only a single address? I think this might need to be done using the `if not isinstance(x, list): x = [x]` pattern (done for the other pgsync fields), otherwise on the frontend `data?.address?.[0]` would be undefined, and end up with the "address not showing" bug

alazar-aot (changes requested):
One comment/question, otherwise looks great!

services/core-api/app/api/search/search/search_transformers.py
asinn134: Nice catch! This has been adjusted now.

alazar-aot (changes requested):
Ready to Rock!

asinn134:
> Ready to Rock!

Should be good for re review now.

simensma-fresh (approved):

alazar-aot (approved):

PR #3958, opened 2026-07-23
[MDS-6974] Feature/bcregistries api

Jsyro:
## Objective 

[MDS-6974](https://bcmines.atlassian.net/browse/MDS-6974)


services/common/src/components/projectSummary/Agent.tsx
```diff
@@ -95,16 +94,14 @@ export const Agent: FC<ProjectSummaryFormComponentProps> = ({ fieldsDisabled })
...
-        setVerified(response.success);
-        setCheckingStatus(false);
-        const payload = {
-          businessNumber: credential.topic.source_id || "-",
-          registrationStatus: credential.inactive ? "Inactive" : "Active",
-          registriesId: credential.id,
-        };
-        setVerifiedCredential(payload);
-      });
+      setVerified(true);
```
alazar-aot, line 97: Looks like we're no longer verifying the credentials, and always telling the UI that the business is verified - is this intentional?

services/common/src/components/projectSummary/Applicant.tsx
```diff
@@ -140,14 +127,14 @@ const Applicant: FC<ProjectSummaryFormComponentProps> = ({ fieldsDisabled }) =>
     let color = undefined;
     let text = undefined;
 
-    if ((!checkingStatus && verified) || verifiedCredential) {
+    if ((!checkingStatus) || verifiedCredential) {
```
alazar-aot, line 130: Similar change as Agent.tsx - no longer verifying the credentials, so the UI will always show the business is verified - is this intentional?

services/core-web/src/components/Forms/parties/PartyOrgBookForm.tsx
```diff
@@ -81,7 +84,7 @@ export const PartyOrgBookForm: FC<PartyOrgBookFormProps> = ({ party }) => {
           <Button
             type="primary"
             className="full-mobile"
-            disabled={!isAssociated ? !hasOrgBookCredential : false} //Admin is allowed to disassociate
+            // disabled={!isAssociated} //Admin is allowed to disassociate
```
alazar-aot: Removing `disabled` completely means the button will always be visible (even when nothing has been searched/selected), is that what we want? Or should this be updated to use the new `registrationId`?

services/core-api/app/api/services/orgbook_service.py
```diff
@@ -49,51 +47,48 @@ def search(self, search: str) -> List[BusinessSearchResult]:
         except (ValueError, TypeError) as e:
             raise BadGateway('BC Registries API responded with unexpected data.') from e
 
-        return response.get('results', [])
+        return [], 200
```
alazar-aot, line 50: I assume we're discarding the response and returning `[], 200` as a temporary flow? Just want to confirm that eventually we return the actual response

services/core-api/app/api/verifiable_credentials/manager.py
```diff
@@ -621,28 +623,28 @@ def ensure_start_date_type(d) -> date:
...
             current_app.logger.warning(
                 f"No valid permittee appointments found for permit_amendment_guid={permit_amendment.permit_amendment_guid}, cannot produce Mines Act Permit UNTP CC"
             )
             return None
-        
+
+
 #
-        orgbook_entity = curr_appt.party.party_orgbook_entity
+        orgbook_entity = curr_appt.party.party_bc_registration
```
alazar-aot, line 647: This column is nullable for BC Registries sourced `party_bc_registration` rows, so as-is this could produce a URL like `.../credential/None` I think? 

It looks like it would only be an issue for the shadow/comparison path, but once this gets finished (per the "replace code above" comment) we should have a plan for parties that have no credential id, else we might end up emitting broken URL's once this goes live. 

I might not fully be understanding this piece though, happy to chat about it if we don't think it's an issue

alazar-aot (changes requested):
Looks really good overall man, I left a few questions about things I noticed. Nothing major, I'm happy to switch this to approved if the items I flagged are just temporary as we work through this refactor!

services/common/src/components/projectSummary/Agent.tsx
Jsyro: The existence of the record in the source system now has to be considered 'verification', cryptographically verifying the actual credentials is an orgbook specific thing that goes beyond the new criteria. 

Except that apparently BC Registries accept self-submitted data without verification somewhere.

services/core-api/app/api/services/orgbook_service.py
Jsyro: Yes we will return data once we get a sandbox api key. 

The linter was flagging that return value as not compatible with the typehint. so i did this instead for now.

services/core-web/src/components/Forms/parties/PartyOrgBookForm.tsx
Jsyro: I'll look at this. this doesn't look right.

services/core-api/app/api/verifiable_credentials/manager.py
Jsyro: party_bc_registration is the object. some of the values inside were previously not nullable, but are now nullable. 

Yes you are correct that could be null, but we don't have a new place to link to at the moment. I'll link to the party page but not the credential directly.

alazar-aot (approved):
Approving, and noting that a few items are still works in progress but that's due to issues out of our control (eg no API Key yet, no public BC Reg page to link to yet). 

This is ready to rock as far as I'm concerned, good work!

services/core-api/app/api/verifiable_credentials/manager.py
alazar-aot: Awesome - thanks for keeping the link to Orgbook, this is good for now. If we get a BC Reg public URL at some point we can update to point to that, else we keep pointing to Orgbook until it's dead (and if that happens without a BC Reg URL, then we can scrap the URL generation entirely and just display the BC Reg API/JSON data in Core)

services/core-api/app/api/services/orgbook_service.py
alazar-aot: Totally fair for now, will resolve

services/core-web/src/components/Forms/parties/PartyOrgBookForm.tsx
alazar-aot: Thanks for making the fix, this is resolved

services/common/src/components/projectSummary/Agent.tsx
alazar-aot: Understood, good to know that moving forward the existence of a BC Reg record counts as 'verification' for this functionality. This is good to go as-is!

PR #3959, opened 2026-07-24
[MDS-6962] - Update formatting of shared s3 endpoint variable in tusd

matbusby-fw:
[MDS-6962] - (https://bcmines.atlassian.net/browse/MDS-6962)

### Summary

Following the tusd `v1.3.0` → `v2.10.0` upgrade, file uploads failed in production with `Attempt to retrieve create file url with status 502` (surfaced to users as a generic error during Notice of Work file generation). Investigation traced this to three separate breaking changes introduced by tusd's move to `aws-sdk-go-v2`, plus one latent bug in the custom `post-finish` hook script that the same migration exposed. Production has been rolled back to `v1.3.0` in the meantime; this PR contains the fixes needed to re-attempt the `v2.10.0` upgrade.

### Root causes & fixes

**1. Invalid S3 endpoint URI** (`services/tusd/init.sh`)
`aws-sdk-go-v2` requires `-s3-endpoint` to be a full URI including scheme. `S3_ENDPOINT` (from the shared `object-store-credentials` secret) is a bare hostname (`nrs.objectstore.gov.bc.ca`), which every other consumer of that secret (docman, docman-worker, filesystem-provider, the permits import script) already handles by prepending `https://` in their own code. tusd was the outlier, passing the bare value straight through.
- Fix: normalize `S3_ENDPOINT` to `https://<host>` in `init.sh` only if it doesn't already include a scheme — no change to the shared secret or any other consumer.

**2. `XAmzContentSHA256Mismatch` on every S3 write** (`services/tusd/init.sh`)
`aws-sdk-go-v2` now attaches a CRC32 request checksum by default on S3 API calls. BC gov's NRS object store (a non-AWS S3-compatible backend) doesn't validate it the way AWS does, so every `PutObject`/`CreateMultipartUpload` failed.
- Fix: export the standard SDK-recognized env vars `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` and `AWS_RESPONSE_CHECKSUM_VALIDATION=when_required`, opting back into pre-2024 checksum behavior. No code/binary changes required — these are read automatically by the SDK's default config loader. Also added `-s3-disable-content-hashes` as defense-in-depth (tusd's own content-hash flag).

**3. Corrupted `post-finish` webhook request** (`services/tusd/hooks/post-finish`)
The `v2.10.0` migration switched from tusd's built-in `--hooks-http` mechanism (which only forwarded the `Authorization` header via `-hooks-http-forward-headers=Authorization`) to a custom file-based hook script that forwarded **all** of the original upload request's headers — including `Host` — into a single malformed `curl -H` argument. This corrupted the outgoing request to docman's `/tusd-hooks` endpoint, which crashed with `ValueError: Port could not be cast to integer value as '5001,tusd:1080'` on every retry, indefinitely. Since `/tusd-hooks` never returned success, the document's status stayed "In Progress" forever, and core-api's client-side poll timed out after 60s with `Timed out while waiting for file upload to finish`.
- Fix: forward only the `Authorization` header, matching what the endpoint actually requires (`@requires_any_of` auth decorator; it reads nothing else from headers) and restoring the pre-upgrade behavior.

**4. False-alarm `HookInvocationError` on successful uploads** (`services/tusd/hooks/post-finish`)
tusd reads the hook script's stdout and tries to parse it as a JSON `HookResponse`. The script's own diagnostic `echo` statements and curl's unsuppressed progress meter/response body were all going to stdout, so tusd failed to parse it and logged an `ERROR`-level `HookInvocationError` on **every successful upload** — noisy and indistinguishable from a real failure in logs/monitoring.
- Fix: redirect all diagnostic output to stderr and run curl with `-s -o /dev/null -w "..."` so stdout stays empty, which tusd accepts as a no-op response.

### Testing

All four fixes were verified locally via `docker compose` against a real S3-backed upload flow (tusd → docman → core-api), confirming:
- File creation (`POST`) and chunk upload (`PATCH`) both succeed without S3 errors
- The `post-finish` webhook reaches docman successfully (`204`) and completes with a clean `HookInvocationFinish`, no `HookInvocationError`
- End-to-end upload through the actual app (real JWT) completes without timing out, with document status correctly reaching "Success"

### Files changed
- `services/tusd/init.sh`
- `services/tusd/hooks/post-finish`

### Deployment notes
- No changes to the `object-store-credentials` secret or any other service's environment — this PR is scoped entirely to tusd's own script layer.
- Recommend re-promoting through `dev` → `test` → `prod` with the same verification steps used locally, given the prior incident.

### DIFF
--- a/services/tusd/hooks/post-finish
+++ b/services/tusd/hooks/post-finish
@@ -1,7 +1,11 @@
 #!/bin/sh
 
+# tusd reads this script's stdout and tries to parse it as a JSON
+# HookResponse; anything else logs as a HookInvocationError even on success.
+# Keep all diagnostic output on stderr so stdout stays empty.
+
 # print a message to the tusd log
-echo "post-finish hook called"
+echo "post-finish hook called" >&2
 
 # Set the AWS credentials
 export AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
@@ -17,14 +21,17 @@ upload=$(echo "$stdin_data" | jq -r '.Event.Upload')
 # check if the upload object has version_guid in MetaData
 version_guid=$(echo "$upload" | jq -r '.MetaData."version_guid"')
 
-# Extract all the HTTP headers from the JSON-encoded object
-headers=$(echo "$stdin_data" | jq -r '.Event.HTTPRequest.Header | to_entries | map("\(.key): \(.value | join(","))") | join("\n")')
+# Only forward the Authorization header from the original request - the
+# tusd-hooks endpoint requires it for auth and reads nothing else. Forwarding
+# the full raw header set (previous behavior) included Host, which corrupts
+# the outgoing request to document-manager.
+authorization=$(echo "$stdin_data" | jq -r '.Event.HTTPRequest.Header.Authorization[0] // empty')
 
 # Extract the ID and Size properties from the Upload object
 id=$(echo "$upload" | jq -r '.ID')
 size=$(echo "$upload" | jq -r '.Size')
 
-echo "Upload ID $id ($size bytes) finished"
+echo "Upload ID $id ($size bytes) finished" >&2
 
 S3_PREFIX_CLEAN=$(echo $S3_PREFIX | sed 's/\/$//')
 storageKey="$S3_PREFIX_CLEAN$(echo "$upload" | jq -r '.MetaData.path')"
@@ -35,7 +42,7 @@ if [ $version_guid != "null" ]; then
 
   aws_exit_code=$?
   if [ $aws_exit_code -ne 0 ]; then
-    echo "aws command failed with exit code $aws_exit_code"
+    echo "aws command failed with exit code $aws_exit_code" >&2
   fi
 
   # Forward the request through the hooks-http option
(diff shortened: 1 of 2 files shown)


services/tusd/hooks/post-finish
```diff
@@ -44,4 +51,4 @@ else
   json_data=$(echo "{\"Upload\": $upload}" | tr -d '\r\n' | tr -d '\0')
 fi
 
-curl --retry 5 --retry-delay 2 --retry-all-errors -X POST -H "Content-Type: application/json" -H "$headers" -H "Hook-Name: post-finish" -d "$json_data" $DOCUMENT_MANAGER_URL/tusd-hooks
\ No newline at end of file
+curl -s -o /dev/null -w "tusd-hooks response status: %{http_code}\n" --retry 5 --retry-delay 2 --retry-all-errors -X POST -H "Content-Type: application/json" -H "Authorization: $authorization" -H "Hook-Name: post-finish" -d "$json_data" $DOCUMENT_MANAGER_URL/tusd-hooks >&2
```
alazar-aot: Maybe I'm over thinking this - if Docman fails to record a record as finalized (eg DB or auth issue), this curl  only logs the HTTP status code and discards the response body - I believe curl considers this a successful response even when Docman returns a 4xx/5xx code?

Essentially meaning a rejected finalize call would fail silently (where the file is in S3 but never gets recorded as finalized), the only indicator something failed would be a status code in the logs that we would have to look for/be told about. 

It would give us better visibility by having curl exit and log the response body on failure (4xx/5xx) - not strictly required but might save us some headache in the future

matbusby-fw: Ah good point.  I quieted that because the way tusd tries to parse the response it was showing an error for every success as well.  I'll add something a bit more refined so it quiets the success responses, but still successfully reports the errors.

alazar-aot: Ahh friggin tusd! If parsing the response is complex/adds a bunch of logic this can go as-is, there is _something_ in the logs so it's not like we have zero visiblity

alazar-aot (approved):
Nice!

PR #3961, opened 2026-07-30
[MDS-6968] Update NoW to reflect new/removed FCBC fields

alazar-aot:
## Objective 

[MDS-6968](https://bcmines.atlassian.net/browse/MDS-6968)

FCBC has updated some of the sections of a Notice of Work (NoW) Application, in the form of deprecated (always `null`) and new fields on incoming NoW applications. 

This PR adds support for the new fields, and removes the unused/no longer required fields. 

Note that there is an existing (legacy) section for "Fuel/Lubricant Storage", where moving forward all related fields will be `null`. Ideally this entire section would be removed, however we cannot currently do this as there are existing, non-approved applications in Core that require this data to be visible. Thus, those fields are conditionally rendered _only_ if there is a value for any of them (if they are all `null`, as will be the case for applications moving forward, then they are not rendered).

**Sumary of Change Made:**

- Added new field (`work_year_info`) for 'Time of Proposed Activities' - new DB column and API mapping, exposed field in UI
- Added new fields (`has_more_than_25_unreclaimed_drill_sites`, `num_unreclaimed_drill_sites`) for 'Exploration Surface Drilling' - new DB Columns, API Mapping and exposed fields in UI
- Removed old Indigenous/First Nation Consultation sections, added new Cultural Heritage section and hooked up the new `protectionofculturalheritageresources` field to this section (new DB column, API mapping) and unit tests
- Added new fuel values (`fueltype`, `fuelrelatedactivity`, `estimatedfuelvolume`, `descriptionoffuelrelatedactivity`, `descriptionofprecautionarymeasures`), added conditional rendering of legacy fuel fields in the UI so they only appear if there's an associated value

All unit tests / code coverage has been updated as well


matbusby-fw:
I think we'll need to update the response models in `services/core-api/app/api/now_applications/response_models.py`.  I'm pretty sure we'll need to update the `NOW_APPLICATION_CAMP_DETAIL` response model with a new custom `NOW_APPLICATION_FUEL_DETAIL` since `fuel_details` was added to `camp` in the `transmogrify_now.py` (or re-use the one in the other `response_models` file if viable.

services/core-api/app/api/now_applications/models/activity_detail/activity_detail_base.py
```diff
@@ -77,10 +79,16 @@ class _ModelSchema(Base._ModelSchema):
...
+            db.select([text("'fuel'")]).where(
+                and_(
+                    ActivitySummaryFuelDetailXref.activity_summary_id ==
+                    ActivitySummaryBase.activity_summary_id,
+                    ActivitySummaryFuelDetailXref.activity_detail_id == activity_detail_id,
                     ActivitySummaryBase.activity_type_code == 'camp')).limit(1).as_scalar()))
 
     activity_type_code = db.column_property(
-        case([(_activity_type_code_identity in ['building', 'staging_area'], 'camp')], else_=_activity_type_code_identity))
+        case([(_activity_type_code_identity in ['building', 'staging_area', 'fuel'], 'camp')], else_=_activity_type_code_identity))
```
matbusby-fw: I think this is using the python `in` operator instead of the SQLAlchemy one.  I think you'd want to use the dot function on the _activity_type_code_identity instead.
ie:
``````
case([(_activity_type_code_identity.in_(['building', 'staging_area', 'fuel']), 'camp')]

File: migrations/sql/V2026.07.28.16.59__add_fuel_transportation_and_storage_tables.sql [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+
+CREATE TABLE IF NOT EXISTS activity_summary_fuel_detail_xref (
+    activity_summary_id INTEGER NOT NULL REFERENCES activity_summary(activity_summary_id),
+    activity_detail_id INTEGER NOT NULL REFERENCES activity_detail(activity_detail_id),
+
+    PRIMARY KEY(activity_summary_id, activity_detail_id)
+);
+ALTER TABLE activity_summary_fuel_detail_xref OWNER TO mds;
+
+CREATE TABLE IF NOT EXISTS fuel_detail (
```
matbusby-fw, line 20: Might want to add the `ON DELETE CASCADE` here to match with the python model.

services/core-api/app/api/now_applications/transmogrify_now.py
```diff
@@ -374,6 +377,15 @@ def _transmogrify_camp_activities(now_app, now_sub, mms_now_sub):
                 timber_volume=detail.timbervolume)
             camp.staging_area_details.append(staging_area_detail)
 
+        for detail in now_sub.fuel:
```
matbusby-fw, line 380: Not sure if they'll be passing an empty array or a none type if this isn't present, but might be good to do a null check here?

matbusby-fw (changes requested):
Noticed a couple things that might need looking at, but overall it's looking really great!

alazar-aot:
> I think we'll need to update the response models in `services/core-api/app/api/now_applications/response_models.py`. I'm pretty sure we'll need to update the `NOW_APPLICATION_CAMP_DETAIL` response model with a new custom `NOW_APPLICATION_FUEL_DETAIL` since `fuel_details` was added to `camp` in the `transmogrify_now.py` (or re-use the one in the other `response_models` file if viable.

@matbusby-fw I believe this is already handled, in `response_models.py` lines 82-89 `NOW_APPLICATION_FUEL_DETAIL` is defined, and lines 112-113 nests `NOW_APPLICATION_CAMP.fuel_details`. If I'm understanding that wrong let me know, happy to review it again or chat about it!

services/core-api/app/api/now_applications/transmogrify_now.py
alazar-aot: I think since `fuel` is is an SQLAlchemy relationship and not a raw JSON field it's not needed... based on the fact that the `now_sub.camps`, `now_sub.buildings`, `now_sub.stagingareas` loops above this line don't implement any null checks. 

The google AI summary says "for a one-to-many relationship SQLAlchemy always returns an (possibly empty) list, never `None` or a non-iterable shape" - so I think this is fine as-is. Let me know what you think

migrations/sql/V2026.07.28.16.59__add_fuel_transportation_and_storage_tables.sql
alazar-aot: Good catch, I've added it to this migration but note that the `ON DELETE CASCADE` clause is missing from the older `building`/`staging_area_xref` tables as well. I don't think I should update those old ones now (they've been like that since 2021, and would require re-running those migrations), let me know if you think we should deal with the old ones @matbusby-fw

services/core-api/app/api/now_applications/models/activity_detail/activity_detail_base.py
alazar-aot: Yep, that's a bug! Switched to using `.expression.in_([...])`, as `_activity_type_code_identity`  is a columnProperty at this point in the code execution, so `.expression` exposes the underlying SQL expression that `.in_()` is defined on. 

Tested and verified to be working

matbusby-fw (approved):
Saweeeet!

PR #3962, opened 2026-08-05
[MDS-6949] - Separate template report requirements

matbusby-fw:
## Objective 

[MDS-6949](https://bcmines.atlassian.net/browse/MDS-6949)

## Summary

Fixes [MDS-6949]: report requirement names on permit condition templates were unique **globally** across every template, so adding a report (e.g. "Annual Reclamation") to one template (e.g. Sand and Gravel) blocked adding a report with the same name to any other template (e.g. MX), even though each template's report requirements should be entirely independent.

Report name uniqueness is now scoped to the template (`notice_of_work_type`), matching how the rest of the permit condition template system is scoped.

## Changes

**Database**
- Added `condition_category_code` and `notice_of_work_type` columns to `mine_report_permit_requirement` (and `mine_report_permit_requirement_version`), populated only for standard (template) report requirements.
- Backfilled existing standard report requirements from their linked `standard_permit_conditions`.
- Replaced the global `unique_standard_report_name_idx` (`UNIQUE (report_name) WHERE is_standard = TRUE`) with one scoped to the template: `UNIQUE (report_name, notice_of_work_type) WHERE is_standard = TRUE`.

**Backend**
- `StandardReportPermitRequirement.create()` now accepts and persists `condition_category_code`/`notice_of_work_type`.
- `StandardReportPermitRequirementResource` (`post`/`put`) derives these values from the linked `StandardPermitConditions` and passes them through, so every standard report requirement is tagged with its template on create/update.
- Exposed both fields on the `MINE_REPORT_PERMIT_REQUIREMENT` API response model.

**Frontend**
- `ReportPermitRequirementForm`: the "Select an Existing Report" dropdown and the client-side duplicate-name validator are now scoped to the current template's `notice_of_work_type` (falling back to a `conditionMap` lookup for older records without the field), instead of comparing against every template's reports.
- Fixed a UI race where, right after creating/updating a report requirement, the just-saved record (already present in the Redux store) would briefly fail the duplicate-name validation against itself before the form closed — `selectedRequirement` is now synced immediately on a successful save.

**Tests**
- Added/updated backend resource tests covering:
  - Same report name can be reused across two different templates (different `notice_of_work_type`) — now succeeds.
  - Same report name within the same template but a different condition category — still correctly rejected.
  - Existing duplicate-name-within-a-template test still passes.

## Expected result (from the ticket)

- ✅ Templates have their own distinct report requirements, independent of other templates.
- ✅ The same report name can be reused across different template types.
- ✅ "Select an Existing Report" is scoped to the current template, not all templates.
- Submitted reports were already correctly associated to a single template/permit and required no change (verified only).

### DIFF
--- a/migrations/sql/V2026.08.05.10.00__scope_standard_report_name_uniqueness_by_template.sql
+++ b/migrations/sql/V2026.08.05.10.00__scope_standard_report_name_uniqueness_by_template.sql
@@ -1,0 +1,36 @@
+-- Report requirement names for standard/template conditions were only unique globally
+-- (report_name), preventing the same report name from being used across different
+-- permit condition templates. Scope uniqueness to the template instead, where a
+-- template is identified by notice_of_work_type.
+
+ALTER TABLE mine_report_permit_requirement
+    ADD COLUMN condition_category_code VARCHAR REFERENCES permit_condition_category(condition_category_code),
+    ADD COLUMN notice_of_work_type VARCHAR REFERENCES notice_of_work_type(notice_of_work_type_code);
+
+ALTER TABLE mine_report_permit_requirement_version
+    ADD COLUMN condition_category_code VARCHAR,
+    ADD COLUMN notice_of_work_type VARCHAR;
+
+-- Backfill existing standard report requirements from one of their linked standard conditions
+UPDATE mine_report_permit_requirement mrpr
+SET condition_category_code = linked.condition_category_code,
+    notice_of_work_type = linked.notice_of_work_type
+FROM (
+    SELECT DISTINCT ON (xref.mine_report_permit_requirement_id)
+        xref.mine_report_permit_requirement_id,
+        spc.condition_category_code,
+        spc.notice_of_work_type
+    FROM mine_report_req_permit_condition_xref xref
+    JOIN standard_permit_conditions spc
+        ON spc.standard_permit_condition_id = xref.standard_permit_condition_id
+    WHERE xref.is_standard = TRUE AND xref.deleted_ind = FALSE
+) linked
+WHERE mrpr.mine_report_permit_requirement_id = linked.mine_report_permit_requirement_id
+    AND mrpr.is_standard = TRUE;
+
+DROP INDEX unique_standard_report_name_idx;
+
+CREATE UNIQUE INDEX unique_standard_report_name_idx
+ON mine_report_permit_requirement (report_name, notice_of_work_type)
+WHERE is_standard = TRUE AND report_name IS NOT NULL;
+
--- a/migrations/sql/V2026.08.05.10.00__scope_standard_report_name_uniqueness_by_template.sql
+++ b/migrations/sql/V2026.08.05.10.00__scope_standard_report_name_uniqueness_by_template.sql
@@ -24,6 +24,7 @@ FROM (
     JOIN standard_permit_conditions spc
         ON spc.standard_permit_condition_id = xref.standard_permit_condition_id
     WHERE xref.is_standard = TRUE AND xref.deleted_ind = FALSE
+    ORDER BY xref.mine_report_permit_requirement_id, spc.standard_permit_condition_id ASC
 ) linked
 WHERE mrpr.mine_report_permit_requirement_id = linked.mine_report_permit_requirement_id
     AND mrpr.is_standard = TRUE;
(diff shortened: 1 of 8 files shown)


services/core-api/app/api/mines/permits/permit_conditions/resources/standard_report_permit_requirement_resource.py
```diff
@@ -55,7 +55,9 @@ def post(self):
                 due_date_period_months=data.get("due_date_period_months"),
                 cim_or_cpo=cim_or_cpo,
                 ministry_recipient=data.get("ministry_recipient"),
-                permit_condition_ids=permit_condition_ids
+                permit_condition_ids=permit_condition_ids,
+                condition_category_code=permit_conditions[0].condition_category_code,
```
alazar-aot: It looks like `StandardPermitConditions.find_many_by_permit_condition_ids` (line 44) doesn't use an ORDER_BY, which means that accessing the first element (`permit_conditions[0]`) doesn't guarantee that we get the first one the client submitted. Maybe we could sort by the submitted id's before indexing [0]? Or if possible better validate/derive the correct value rather than just taking the first?

Or maybe this isn't an issue, happy to discuss further!

migrations/sql/V2026.08.05.10.00__scope_standard_report_name_uniqueness_by_template.sql
```diff
@@ -0,0 +1,36 @@
...
+ALTER TABLE mine_report_permit_requirement_version
+    ADD COLUMN condition_category_code VARCHAR,
+    ADD COLUMN notice_of_work_type VARCHAR;
+
+-- Backfill existing standard report requirements from one of their linked standard conditions
+UPDATE mine_report_permit_requirement mrpr
+SET condition_category_code = linked.condition_category_code,
+    notice_of_work_type = linked.notice_of_work_type
+FROM (
+    SELECT DISTINCT ON (xref.mine_report_permit_requirement_id)
```
alazar-aot, line 19: We probably want to add an ORDER BY here, otherwise for any pre-existing report requirement that's linked to conditions from more than one category it's arbitrary which one gets backfilled (eg we wont get consistent results)

services/core-api/app/api/mines/permits/permit_conditions/resources/standard_report_permit_requirement_resource.py
alazar-aot, line 40: Question - could this ever operate on `permit_condition_ids` that come from different NoW Types?

If so, both this and the PUT on line 72 don't have a check that all `permit_condition_ids` within a single request share the same NoW Type - I think having an explicit check here is a good idea so that this doesn't potentially corrupt the scoping if this is ever called for id's with different NoW types.

services/common/src/components/permits/ReportPermitRequirementForm.tsx
```diff
@@ -99,6 +94,24 @@ export const ReportPermitRequirementForm: FC<ReportPermitRequirementProps> = ({
 
   const { conditionMap, categoriesWithConditions } = conditions;
 
+  // Standard report requirements aren't filtered by template on the GET endpoint, so
+  // filter to only those requirements matching the notice_of_work_type (template)
+  // as the condition currently being edited.
+  const currentTemplateCondition = condition as unknown as IStandardPermitCondition;
```
alazar-aot, line 100: This is something Claude flagged:

The `currentTemplateCondition` prop could potentially come from the `PermitConditionReportRequirements` → `PermitConditionLayer` flow (when editing a requirement from the "Report Requirements" list rather than a specific conditions inline form). 

What this means is that in that flow, `currentTemplateCondition` is `undefined`, so the check on line 104 below would filter `standardRequirementsForTemplate` down to empty. This effectively disables the client-side duplicate name validation for that flow. 

Suggestion: guard this filter so it only applies when `condition` is actually present, and fall back to the unfiltered list if it's not present. 

I traced this quickly myself and it does seem like we should guard against `condition` being `None`/`undefined`, tho I don't fully understand the flow that leads to that possibility. I'm trusting the robot on this one but happy to chat or dig further if you don't think it's a concern

alazar-aot (changes requested):
Noticed a few potential gotchas, let me know if you've got any questions

alazar-aot (approved):
GAW-GEOUS

PR #3963, opened 2026-08-05
[MDS-6990] - Update API to align with 0.7 untp publisher

Jsyro:
HAS ENVIRONMENT VARIABLE CHANGES.

## Objective 

[MDS-6990](https://bcmines.atlassian.net/browse/MDS-6990)

Publisher service has moved to a new URL, and is now UNTP specific and self-hosts credentials rather than orgbook built as a packaging service. 

Update the endpoint shapes and remove old self-sign and push to orgbook. 

Localhost batch job pushed successfully to dev untp-publisher. 
https://untp-publisher-api-dev.apps.gold.devops.gov.bc.ca/discovery


services/core-api/app/api/parties/party/models/party.py
```diff
@@ -68,8 +72,8 @@ class Party(SoftDeleteMixin, AuditMixin, Base):
         'and_(PartyBusinessRoleAppointment.party_guid == Party.party_guid, PartyBusinessRoleAppointment.deleted_ind==False)',
         back_populates='party')
 
-    party_bc_registration = db.relationship(
-        'PartyBCRegistration', backref='party', uselist=False, lazy='select')
+    party_bc_registration: Mapped["PartyBCRegistration | None"] = db.relationship(
+        'PartyBCRegistration', back_populates='party', uselist=False, lazy='select')
```
Jsyro, line 76: This allows pylance to correctly type check across relationships (otherwise it checks the relationship object, not the target class).

PatStLouis:
Reviewed alongside Copilot’s threads — not repeating those (single-issue `w3c_map` path, nullable/collision handling, `find_by_party_guid`, `permittee_appt[0]`, etc.). A few additional items:

### Blocker — integration tests

`tests-integration-backend` is red on 3 tests in `tests/verifiable_credentials/test_manager.py`. All fail with:

```
NameError: name 'VerifiableCredentialManager' is not defined
```

The import was updated to `AnonCredCredentialManager`, but the test bodies still call the removed `VerifiableCredentialManager`:

- `test_collect_attributes_for_mines_act_permit_111` → `AnonCredCredentialManager.collect_attributes_for_mines_act_permit_111`
- `test_produce_untp_cc_map_payload_happy` / `_null_if_no_orgbook` → `UNTPCredentialManager.prepare_permit_amendment_untp_credential_without_id` (assertions need updating for the new `dict` payload shape)

### Should fix — token refresh

In `untp_publisher.py`, the 403 retry path calls `self.get_new_token()` but never assigns `self.token`, so the retry still sends the expired Bearer token. Same pattern as the old OrgBook client:

```python
self.token = self.get_new_token()
```

Publisher returns 403 on invalid/expired JWT for publish, so this path is reachable.

### Breaking change — GET credential by hash

`develop` registers `W3CCredentialResource` at `/credentials/<vc_unsigned_hash>`. This PR removes that route from `namespace.py` (only `/credentials/issue` remains). `W3CCredentialResource` is now dead code in `w3c_map_credential_resource.py`.

Worth calling out in the PR/deploy notes for any consumer that fetched VCs by hash — replacement is publisher-side retrieval (Discovery / `GET /credentials/{id}` on the UNTP publisher).

### Nit

`untp_manager.py` still has comments referring to pushing to OrgBook (lines ~92–94); worth updating for the UNTP-only flow.

---

`client_id` / issuer auth: confirmed we validate against the provisioned `did:web` issuer id — MDS sending `CHIEF_PERMITTING_OFFICER_DID_WEB` aligns with publisher `IssuerInstanceRecord.id` after yaml DID expansion. No concern there.

Batch publish against dev looks like the right direction once the above is addressed and CI is green.

esune (approved):

esune:
Did a cursory pass, I think the changes look good from a functional standpoint - will defer to code maintainers final approval.

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -44,50 +45,89 @@ class W3CCredentialIssueResource(Resource, UserMixin):
...
+            permit_amendment_guid=permit_amendment_guid,
+            party_guid=payload["data"]["permittee"]["party_guid"],
+            signed_credential='Produced by publisher',
+            publish_state=None,
+            permit_number=payload["data"]["permit"]["identifier"],
+            orgbook_entity_id=payload["data"]["permittee"]["identifier"],
+            orgbook_credential_id=None,
+            error_msg=None)
+        try:
+            collision = True
```
alazar-aot: This and line 105 seem backwards to me, wouldn't `collision = True` be when there _is_ an error? (so this line would be `collision = False`, line 105 would be true)

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -44,50 +45,89 @@ class W3CCredentialIssueResource(Resource, UserMixin):
...
+
+            post_resp = publisher_service.publish_cred(payload)
+            #save success
+            publish_record.publish_state = post_resp.ok
+            publish_record.error_msg = post_resp.text if not post_resp.ok else None
+            publish_record.orgbook_credential_id = post_resp.json().get("credential_id")
```
alazar-aot: Here we're using `credential_id` but `untp_manager.py` line 165 uses `credentialId` - one of these must be wrong and thus would store `None`. Whichever is wrong should be updated to use the correct key

services/core-api/tests/verifiable_credentials/test_manager.py
```diff
@@ -67,30 +68,3 @@ def test_collect_attributes_for_mines_act_permit_111(self, test_client, db_sessi
         assert attributes["latitude"] == str(pa.mine.latitude)
         assert attributes["longitude"] == str(pa.mine.longitude)
         assert attributes["bond_total"] == str(pa.permit.active_bond_total)
-
-    def test_produce_untp_cc_map_payload_happy(self, db_session):
```
alazar-aot, line 71: This test and the `test_produce_untp_cc_map_payload_null_if_no_orgbook` don't seem to have a replacement, just wondering if we should be adding coverage for`prepare_permit_amendment_untp_credential_without_id`? Feels like we should, given that that function handles quite a bit of orchestration

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -44,50 +45,89 @@ class W3CCredentialIssueResource(Resource, UserMixin):
...
+
+        existing: bool = PermitAmendmentOrgBookPublish.find_by_unsigned_payload_hash(
+            payload_hash) is not None
+
+        publisher_service = UNTPPublisherService()
+        post_resp = None
+        publish_record = PermitAmendmentOrgBookPublish(
+            unsigned_payload_hash=payload_hash,
+            permit_amendment_guid=permit_amendment_guid,
+            party_guid=payload["data"]["permittee"]["party_guid"],
```
alazar-aot: In `untp_manager.py` there's no `party_guid` key defined - so using it here would raise a `KeyError` on every call. I think this needs to come from somewhere else in the manager or be passed through to the payload separately?

services/core-api/app/api/verifiable_credentials/untp_manager.py
```diff
@@ -0,0 +1,284 @@
...
+            current_app.logger.warning(
+                f"Permit Amendment or mine not found for permit_amendment_guid={permit_amendment_guid}"
+            )
+            return
+        #get other permit_amendments
+        pa.permit._context_mine = mine
+        pmt_appt_list = pa.permit.permit_amendments
+        pos = pmt_appt_list.index(pa)
+
+        permittee_appt = [
```
alazar-aot: The old `manager.py` had this wrapped in a try/except, with this version not having any guards I think there's a possibility that if there's no permittee appointment it will raise an unhandled error and abort the whole `push_untp_map_data_to_publisher` batch loop. 

If we can add the try/except (so that the issue is logged and skipped) that would ensure that the whole process doesn't get nuked!

alazar-aot (changes requested):
Noticed a few things, most look like easy fixes but let me know if you have any questions

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: You right, good catch.

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: Yes, the UNTP publisher responds with `credentialId` so another good catch. thanks.

services/core-api/tests/verifiable_credentials/test_manager.py
Jsyro: It is more than a simple transformation. I'll admit i thought these tests were for the old self-sign and push to orgbook. but it's still relevant.

services/core-api/app/api/verifiable_credentials/untp_manager.py
Jsyro: added

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: centralized the one method into two and resolved this issue

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
```diff
@@ -44,50 +44,61 @@ class W3CCredentialIssueResource(Resource, UserMixin):
...
-        permit_amendment = PermitAmendment.find_by_permit_amendment_guid(
-            data["permit_amendment_guid"])
-        
-        payload = VerifiableCredentialManager.prepare_permit_amendment_untp_credential_without_id( data["permit_amendment_guid"])
+        data = self.query_parser.parse_args()
+        permit_amendment_guid = data["permit_amendment_guid"]
+
+        permit_amendment = PermitAmendment.find_by_permit_amendment_guid(permit_amendment_guid)
+
+        payload = UNTPCredentialManager.prepare_permit_amendment_untp_credential_without_id(
```
alazar-aot: I think this and the two lines below need to changed to handle the updated return type from `prepare_permit_amendment_untp_credential_without_id`, as currently they still expect a dict, but get a tuple

alazar-aot (changes requested):
One last tiny change to handle the new return type - otherwise this is looking solid!

services/core-api/app/api/verifiable_credentials/resources/w3c_map_credential_resource.py
Jsyro: fixed

alazar-aot (approved):
Bellissimo!

PR #3964, opened 2026-08-06
[MDS-6981] Adding system generated doc to NoW permit package table's 1.1 row

asinn134:
## Objective 

[MDS-6981](https://bcmines.atlassian.net/browse/MDS-6981)

- Made changes to include the latest system generated notice of work application document to show up in the permit package table as the row with order# 1.1. The 1.1 row is setup so it can't be reordered, it can't be removed from the permit package manually by a user, and the user also cannot edit it.

image

- The 1.1 row record as confirmed with Rebecca will have some set text in it's column values, please refer to image below for the design:
image

- Also Made sure the system generated document that is our 1.1 file shows up in the permit package column in the Minespace permit table

image

- Also made sure the issued permit document, downloadable draft permit, and downloadable "Edited Application" document doesn't have any changes, so the 1.1 notice of work application document doesn't show up in these.


matbusby-fw (commented):
Looks really good.  The only small thing I'd add would be to add checks on the backend to prevent edits on the system generated docs.  Would be good to add some protection here `services/core-api/app/api/now_applications/resources/now_application_document_resource.py` for the put command.
And here services/core-api/app/api/now_applications/resources/now_application_document_resource.py (NOWApplicationDocumentSortResource.put) will set an arbitrary final_package_order on it — there's no filtering by is_system_generated.

Since I'm gone next week I've just marked these as comments rather than 'Request Changes'.

services/core-web/src/components/modalContent/EditFinalPermitDocumentPackage.js
```diff
@@ -22,12 +22,40 @@ const defaultProps = {
 };
 
 export const EditFinalPermitDocumentPackage = (props) => {
+  const applicationFilesTypes = ["AAF", "AEF", "MDO", "SDO"];
+  const systemGeneratedNtrDoc = props.noticeOfWork.documents.find(
+    (doc) =>
+      doc.now_application_document_type_code === "NTR" &&
```
matbusby-fw: Looks like we have triplicated "find the locked NTR" logic, with some inconsistency.
The same filter+sort-by-upload_date logic exists in three places:

getLockedSystemNtrGuid (JS, helpers.ts:114-129)
an inline copy in FinalPermitDocuments.js:524-535

Then this file has a third, looser variant that finds the "locked" doc by now_application_document_type_code === "NTR" && doc.is_final_package — without checking is_system_generated. Under current backend invariants (only add_now_form_to_fap ever sets is_final_package=True on an NTR doc) this happens to coincide with "system-generated," but it's an inconsistent criterion sitting right next to three other implementations that all explicitly check is_system_generated. Worth consolidating into one shared helper that returns the record (not just the GUID) so FinalPermitDocuments.js doesn't need its own copy, and using it consistently in EditFinalPermitDocumentPackage.js too.

asinn134: Nice, thanks for the catch. Removed the inconsistent duplicates, and now the retrieval of the guid is done by a helper function that makes use of the getLockedSystemNtrDoc in the helper.ts file.

asinn134:
> Looks really good. The only small thing I'd add would be to add checks on the backend to prevent edits on the system generated docs. Would be good to add some protection here `services/core-api/app/api/now_applications/resources/now_application_document_resource.py` for the put command. And here services/core-api/app/api/now_applications/resources/now_application_document_resource.py (NOWApplicationDocumentSortResource.put) will set an arbitrary final_package_order on it — there's no filtering by is_system_generated.
> 
> Since I'm gone next week I've just marked these as comments rather than 'Request Changes'.

Thanks, backend guards have been set up now for this.

services/common/src/utils/helpers.ts
```diff
@@ -165,4 +165,25 @@ export const inspectionOrderNumberSorter = (a, b, dataIndex) => {
...
+      doc.now_application_document_type_code === "NTR" &&
+      doc.is_final_package &&
+      doc.is_system_generated &&
+      !doc.deleted_ind
+  );
+  if (!qualifying.length) return null;
+  const sorted = [...qualifying].sort((a, b) => {
+    const aDate = a.mine_document?.upload_date || "";
+    const bDate = b.mine_document?.upload_date || "";
+    return bDate > aDate ? 1 : -1;
```
alazar-aot: It looks like `mine_document.upload_date` is a `db.Date` (meaning day only, no hours/mins) - I think it's possible that we end up with two system-generated NTR's created on the same day (eg a Type-of-NoW change and Technical Review completion), in which case the code won't know which of those is the "latest", since they share the exact same date, with no time identifiers. 

I also noticed that the two python files that handle this (`now_application_document_resource.py:32` and `now_application_export_resource.py:254`) each have different implementations for how to handle a "tie" - so the behaviour in this scenario won't be consistent, and I'm not sure what impact that would have overall (likely locked documents that should be editable). 

I'm wondering if it's possible to sort on the `create_timestamp` instead (for all 3 files)?

services/core-api/app/api/now_applications/response_models.py
```diff
@@ -344,6 +344,7 @@ def format(self, value):
         'now_application_document_sub_type_code': fields.String,
         'description': fields.String,
         'is_final_package': fields.Boolean,
+        'is_system_generated': fields.Boolean,
```
alazar-aot, line 347: Technically this field is editable by the client... since `is_system_generated` is _not_ a PK column, `deep_update_from_dict` can re-write this value to false and bypass all the protections in place (400 on edit, the sort skip, the export exclusion).

I don't think it's likely, but perhaps this field should be excluded from the `deep_update_from_dict` path to prevent this from being a possibility?

services/common/src/utils/helpers.ts
```diff
@@ -165,4 +165,25 @@ export const inspectionOrderNumberSorter = (a, b, dataIndex) => {
 
   if (mainA !== mainB) return mainA - mainB;
   return subA - subB;
+};
+
+export const getLockedSystemNtrDoc = (documents) => {
```
alazar-aot: I noticed that there's 4 places where we're determining which is the "locked" 1.1 document (this file, as well as `now_application_document_resource.py:22`, `now_application_export_resource.py:243`, and `FinalPermitDocuments.js:84`). It looks like the logic to determine "locked" is also slightly different across the files, which makes updating or maintaining this a bit tricky. 

I know this was a pretty complicated piece and we don't want to add any flags to the DB, but I am wondering if perhaps we could determined the "locked" value once, server-side and expose it 
as a field on `NOW_APPLICATION_DOCUMENT`, then have the UI read from that value (and the backend code reference/update that value as needed when which document is "locked" changes)? 

Happy to discuss this or think it through further, if we do go with the code as-is I think adding a comment to help clarify the differences (and where else this logic is determined) will help us maintain in the long run

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
```diff
@@ -47,6 +48,76 @@ export class FinalPermitDocuments extends Component {
         ({ is_final_package }) => is_final_package
       );
 
+    const isNoWApplication = this.props.noticeOfWork.application_type_code === "NOW";
```
alazar-aot: Here to line 120 is doing a lot of heavy lifting... maybe we should extract this to a`getNowApplicationDocument(noticeOfWork, progress)` helper/utility function? 

If we do it that way, then the spec that already uses `describe("getNowApplicationDocument")` in `FinalPermitDocumentsSpec` wouldn't need to render the entire component, it could just call the helper function and assert on the output of that function

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
```diff
@@ -47,6 +48,76 @@ export class FinalPermitDocuments extends Component {
...
+    let nowApplicationDocument = null;
+    let lockedNtrGuid = null;
+    if (isNoWApplication && this.props.showInUnifiedView) {
+      // Shared fields for all locked 1.1 row variants.
+      const LOCKED_ROW_BASE = {
+        now_application_document_xref_guid: "application-form-1.1",
```
alazar-aot: It looks to me like this becomes the table key via `NOWDocuments.js:96`, but it feels wrong to be putting a non-GUID into the same field where all the other rows will be a GUID... that said I don't think this causes any issues today as this table doesn't have any SELECTs happening, but if that changes this could be a problem. 

Since the real NTR is already filtered out via `lockedNtrGuid`, could we keep the real GUID and just rely on `isLockedApplicationForm`?

services/common/src/utils/helpers.ts
alazar-aot: Nitpicking on this one, but it looks like you added proper typing to `INoWApplicationForm["documents"]`, so we could type the `documents` parameter here instead of leaving it implicitly `any`

services/core-web/src/components/noticeOfWork/applications/NOWSubmissionDocuments.js
```diff
@@ -679,7 +681,9 @@ export const NOWSubmissionDocuments = (props) => {
                 props.selectedRows.setSelectedSubmissionRows(selectedRowKeys);
               },
               getCheckboxProps: (record) => ({
-                disabled: record && !record.mine_document_guid,
+                disabled:
+                  (record && !record.mine_document_guid) ||
+                  props.lockedRowKeys?.includes(record.mine_document_guid),
```
alazar-aot, line 686: If `record` is null, then `record && !record.mine_document_guid` would be false, and the code would run `props.lockedRowKeys?.includes(record.mine_document_guid)` - but since`record` is null, this would throw an error I believe? 

Should just be a matter of doing `record?.mine_document_guid` to safely try and access the `mine_document_guid`

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -145,10 +147,24 @@ export class NOWDocuments extends Component {
 
   onSortEnd = ({ oldIndex, newIndex }) => {
     if (oldIndex !== newIndex) {
-      const newData = arrayMove([].concat(this.state.dataSource), oldIndex, newIndex);
-      newData.map((doc, index) => ((doc.index = index), (doc.final_package_order = index)));
-      this.setState({ dataSource: newData });
-      this.handleSortDocument(newData);
+      const hasLockedRow = this.state.dataSource?.some((d) => d.isLockedApplicationForm);
+      if (hasLockedRow) {
```
alazar-aot: Both this and the `else` block are almost identical, do they need to be separated like this? Thinking that there could be just one path the code takes and uses `hasLockedRow` as a modifier to determine whether to add the target index/renumber skip?

alazar-aot (changes requested):
This is super solid - great work on a complex refactor! 

I did notice a few things - happy to connect and chat them over as I might be being overly cautious with some of my concerns

services/common/src/utils/helpers.ts
asinn134: This should be updated now to use create_timestamp instead

services/core-api/app/api/now_applications/response_models.py
asinn134: This has been updated now so `is_system_generated` is excluded from `deep_update_from_dict` for notice of work application document

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
asinn134: Nice suggestion, this should be extracted now

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
asinn134: I ended up removing this from the LOCKED_ROW_BASE as the actual xref guid is already being used. Instead I left the mock guid string in the NA_ROW (not applicable row)

services/common/src/utils/helpers.ts
asinn134: Nice catch, the type for this has been added now

services/core-web/src/components/noticeOfWork/applications/NOWSubmissionDocuments.js
asinn134: This has been adjusted now to use  `record?.mine_document_guid`

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
asinn134: This if else block has been simplified now.

services/common/src/utils/helpers.ts
asinn134: This has been adjusted to setup the locked logic in the backend as the one source of the data to be used in the front end. The backend file `now_application.py` contains the l`ocked_ntr_guid` function to determine which ntr guid is locked. The front end will now make use of the `locked_ntr_guid` from the backend response.

alazar-aot: Awesome, great work getting that switched around!

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
```diff
@@ -35,8 +36,88 @@ const defaultProps = {
...
+  },
+};
+
+const TECHNICAL_REVIEW_NTR_DESCRIPTION =
+  "This document was automatically created when Technical Review was completed.";
+
+export const getNowApplicationDocument = (noticeOfWork, progress, showInUnifiedView) => {
+  const nullResult = { nowApplicationDocument: null, lockedNtrGuid: null };
+
+  if (noticeOfWork.application_type_code !== "NOW" || !showInUnifiedView) {
```
alazar-aot: Just thinking - should the gate on this row be only `application_type_code === "NOW"` (dropping `showInUnifiedView`)? 

The exclusions in `now_application_export_resource.py:420`, `ProcessPermit.js:400` and `NOWPermitGeneration.tsx:251` are unconditional and don't rely on `showInUnifiedView` (as they should be), so my thinking is that this also wouldn't need it (`showInUnifiedView` still makes sense for the other cosmetic differences on lines 166/223/229, just not for whether the user finds out why a document vanished).

If you disagree and think this makes more sense I'm happy to go with it (so I'll approve instead of request changes)

alazar-aot (dismissed):
Left one more question, but I'm going to approve the PR as I don't think it's critical to address it if you don't feel like it's worth it. 

Really good work on this one dude, you crushed it.

services/core-web/src/components/noticeOfWork/applications/FinalPermitDocuments.js
asinn134: `showInUnifiedView` is removed from here now

alazar-aot (approved):

PR #3966, opened 2026-08-12
[CHORE] [BDC-11] Add untp credential links to minespace

Jsyro:
Add links to UNTP CC on Publisher to minespace permit table. 

image

The link would go to this for example. 

https://untp-publisher-api-dev.apps.gold.devops.gov.bc.ca/view?url=https://untp-publisher-api-dev.apps.gold.devops.gov.bc.ca/credentials/7213beca-cb54-48af-941f-b282041b814e

### DIFF
--- a/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
+++ b/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
@@ -121,7 +121,7 @@ class PermitAmendment(SoftDeleteMixin, AuditMixin, Base):
         order_by='desc(PartyVerifiableCredentialMinesActPermit.update_timestamp)')
     orgbook_publish_status_records: list["PermitAmendmentOrgBookPublish"] = db.relationship(
         'PermitAmendmentOrgBookPublish',
-        lazy='select',
+        lazy='selectin',
         order_by='desc(PermitAmendmentOrgBookPublish.update_timestamp)')                                   #type: ignore[reportAssignmentType]
     mines_act_permit_vc_locked = association_proxy("permit", 'mines_act_permit_vc_locked')
 
--- a/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
+++ b/services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
@@ -1,6 +1,6 @@
 import uuid
 from datetime import date, datetime
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
 from app.api.constants import *
 from app.api.mines.permits.permit_amendment.models.permit_amendment_document import (
@@ -20,6 +20,9 @@ from sqlalchemy.orm import validates
 from sqlalchemy.schema import FetchedValue
 
 from . import permit_amendment_status_code, permit_amendment_type_code
+
+if TYPE_CHECKING:
+    from app.api.verifiable_credentials.models.orgbook_publish_status import PermitAmendmentOrgBookPublish
 
 
 class PermitAmendment(SoftDeleteMixin, AuditMixin, Base):
@@ -116,6 +119,10 @@ class PermitAmendment(SoftDeleteMixin, AuditMixin, Base):
         'PartyVerifiableCredentialMinesActPermit',
         lazy='selectin',
         order_by='desc(PartyVerifiableCredentialMinesActPermit.update_timestamp)')
+    orgbook_publish_status_records: list["PermitAmendmentOrgBookPublish"] = db.relationship(
+        'PermitAmendmentOrgBookPublish',
+        lazy='select',
+        order_by='desc(PermitAmendmentOrgBookPublish.update_timestamp)')                                   #type: ignore[reportAssignmentType]
     mines_act_permit_vc_locked = association_proxy("permit", 'mines_act_permit_vc_locked')
 
     # Note: This relationship is lazy loaded on purpose to avoid being loaded unless absolutely necessary
@@ -190,6 +197,18 @@ class PermitAmendment(SoftDeleteMixin, AuditMixin, Base):
         else:
             return self.vc_credential_exch[0].cred_exch_state if len(
                 self.vc_credential_exch) > 0 else None
+
+    @hybrid_property
+    def active_orgbook_publish_status(self):
+        active = [
+            x for x in self.orgbook_publish_status_records if x.publish_state is True
+        ]
+
+        if active:
+            return active[0]
+
+        return self.orgbook_publish_status_records[0] if len(
+            self.orgbook_publish_status_records) > 0 else None
 
     def __repr__(self):
         return '<PermitAmendment %r, %r>' % (self.mine_guid, self.permit_id)
--- a/services/core-api/app/api/mines/response_models.py
+++ b/services/core-api/app/api/mines/response_models.py
@@ -207,6 +207,20 @@ PERMIT_AMENDMENT_DOCUMENT_MODEL = api.model(
         'create_timestamp': fields.DateTime
     })
 
+PERMIT_AMENDMENT_ORGBOOK_PUBLISH_STATUS_MODEL = api.model(
+    'PermitAmendmentOrgBookPublishStatus', {
+        'unsigned_payload_hash': fields.String,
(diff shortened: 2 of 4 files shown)


services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -116,6 +119,10 @@ class PermitAmendment(SoftDeleteMixin, AuditMixin, Base):
         'PartyVerifiableCredentialMinesActPermit',
         lazy='selectin',
         order_by='desc(PartyVerifiableCredentialMinesActPermit.update_timestamp)')
+    orgbook_publish_status_records: list["PermitAmendmentOrgBookPublish"] = db.relationship(
+        'PermitAmendmentOrgBookPublish',
+        lazy='select',
```
alazar-aot: We should probably use `selectin` here (like how we did for `vc_credential_exch` on line 120)  and add an index for `permit_amendment_orgbook_publish_status.permit_amendment_guid`, since `PERMIT_AMENDMENT_MODEL` is nested in `PERMIT_MODEL`, each amendment would mean doing an extra SELECT

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.tsx
```diff
@@ -180,6 +180,38 @@ export const PermitsTable: FC<PermitsTableProps> = (props) => {
     rowData = permitRowData;
   }
 
+  const hasCredentialLinks = rowData.some((record) =>
+    record.permit_amendments?.some((amendment) =>
+      Boolean(amendment.active_orgbook_publish_status?.orgbook_credential_id)
+    )
+  );
```
alazar-aot, line 187: If `orgbook_credential_id` comes back as a UUID instead of a URL (which `untp_manager.py:187` is doing), it looks to me like the column will render will all fields showing as `Strings.EMPTY_FIELD` - is that what we want?

services/core-api/app/api/mines/response_models.py
```diff
@@ -207,6 +207,20 @@ def format(self, value):
...
+PERMIT_AMENDMENT_ORGBOOK_PUBLISH_STATUS_MODEL = api.model(
+    'PermitAmendmentOrgBookPublishStatus', {
+        'unsigned_payload_hash': fields.String,
+        'permit_amendment_guid': fields.String,
+        'party_guid': fields.String,
+        'publish_state': fields.Boolean,
+        'permit_number': fields.String,
+        'orgbook_entity_id': fields.String,
+        'orgbook_credential_id': fields.String,
+        'error_msg': fields.String,
```
alazar-aot: Since this model is nested within `PERMIT_AMENDMENT_MODEL`, all these values are exposed in the JSON response to MineSpace - so anyone in MineSpace could theoretically go into DevTools and see the `error_msg` field (along with the `unsigned_payload_hash`, `party_guid`) - which I don't think we want to be exposing. 

Do all these fields need to be on the model? Or can we cut only the ones we need, and remove the rest?

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
```diff
@@ -191,6 +198,18 @@ def vc_credential_exch_state(self):
...
+    def active_orgbook_publish_status(self):
+        active = [
+            x for x in self.orgbook_publish_status_records if x.publish_state is True
+        ]
+
+        if active:
+            return active[0]
+
+        return self.orgbook_publish_status_records[0] if len(
+            self.orgbook_publish_status_records) > 0 else None
```
alazar-aot, line 211: This returns the most recent record regardless of the `publish_state`, to me it seems like we could be returning a failed record from something named `active_` which is misleading/makes it sound like return record is active/published. Could this return `None` instead, or renamed to something like "latest orgbook publish status" to avoid it sounding like something is active when it might not be?

Let me know if I'm misunderstanding the intention of this

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.tsx
```diff
@@ -180,6 +180,38 @@ export const PermitsTable: FC<PermitsTableProps> = (props) => {
...
+    render: (status) => {
+      const credentialUrl = status?.orgbook_credential_id;
+
+      if (!credentialUrl) {
+        return Strings.EMPTY_FIELD;
+      }
+
+      try {
+        const publisherHostname = new URL(credentialUrl).hostname;
+        const linkUrl = `https://${publisherHostname}/view?url=${credentialUrl}`;
```
alazar-aot: Is it possible that the `credentialUrl` could contain symbols like  `&`, `#`, or `?` in it? 
If so, I believe we could end up truncating or accidentally injecting query params into `linkUrl` - so we may want to do something like `?url=${encodeURIComponent(credentialUrl)}` to handle that.  

But thinking about this, if `publisherHostname` is being taken from whatever the publisher returned, and the publisher returns a UUID (not a URL), `new URL` would throw an error and the link wouldn't be rendered at all.... would it be safer to build the full `linkUrl` on the backend and just expose it? That way the host could come from our config rather than from the response data, the encoding could be handled server-side, and the `hasCredentialLinks` check on line 183 would be the same the renderer uses (right now the renderer requires `new URL`, where the 183 check doesn't

alazar-aot (changes requested):
Left a few mostly questions, some of it I might just not be fully understanding so happy to chat about any of it!

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
Jsyro: Copliot suggested this too. 

I didn't add an index, is that still needed?

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.tsx
Jsyro: Orgbook credential id is only set after the publisher responds successfully, so it should always be a URL with the new publisher. 

I should double check was was being saved for the previously published records to the old publishers.... but those should probably be destroyed anyways as that service no longer exists.

services/core-api/app/api/mines/response_models.py
Jsyro: Good eye, removed some of the unnecessary fields.

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
Jsyro: See line 204, the list comprehension filters by publish_state.

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.tsx
Jsyro: Copilot suggested a more native construction of the URL. I think it's safe enough to handle this, but let me take a closer look to make sure these cases aren't an issue.

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
alazar-aot: I guess the index isn't critical right now as this table will be fairly small, as it grows that could be something we want to do to help performance. 

If you want to add it now that'd be awesome, but it's not a blocker for the PR

services/core-api/app/api/mines/permits/permit_amendment/models/permit_amendment.py
alazar-aot: Fair, and with the error_msg being removed from the reponse we're less likely to leak data - I'm fine with it

services/minespace-web/src/components/dashboard/mine/permits/PermitsTable.tsx
alazar-aot: Ok, if those old records are getting destroyed then I'm not too worried - worse case scenario we end up rendering a column full of `Strings.EMPTY_FIELD` which isn't the worst impact, I'm fine to leave this as-is

alazar-aot (approved):
Approved 🫡

PR #3967, opened 2026-08-12
[MDS-6974] Connect to real sandbox API

Jsyro:
Sandbox API key and URL. 

Only logs responses, not included in endpoint response.


services/core-api/app/api/services/orgbook_service.py
```diff
@@ -10,6 +10,46 @@
 )
 
 
+class BCRegistriesService():
```
alazar-aot: Is this just a duplicate of the new `bcregistries_api_service.py:13` for historical reasons? 

Noticed that this is returning a tuple (`[], 200`), which seems to go against the return type `List[OrgBookSearchResultItem]:` - and if this is also already defined in `bcregistries_api_service.py` do we need this at all?

services/core-api/app/api/services/bcregistries_api_service.py
```diff
@@ -23,28 +23,35 @@ def search(self, search: str) -> List[OrgBookSearchResultItem]:
...
-            headers={"X-Apikey": BC_REGISTRIES_SECRET_TOKEN},
+            headers={
+                "x-apikey": BC_REGISTRIES_SECRET_TOKEN,
+                "account-id": ""
+            },
             timeout=10,
         )
 
         if resp.status_code != requests.codes.ok:
+            print(resp.text)
```
alazar-aot: For testing this is fine, but it'd be best to use `current_app.logger`

alazar-aot (changes requested):
Noticed 2 small things I'm curious about - nothing major so happy to approve if they're not an issue

services/core-api/app/api/services/orgbook_service.py
Jsyro: Good catch again, i separated it but a merge must have put it back in. removed. separate file is the one in use.

services/core-api/app/api/services/bcregistries_api_service.py
Jsyro: changed.

alazar-aot (approved):
Noiiice!

PR #3970, opened 2026-08-14
[MDS-6973] Adding missing text

asinn134:
## Objective 

[MDS-6973](https://bcmines.atlassian.net/browse/MDS-6973)

- Adding some missing text in the Minespace major mine application documents tab.

image


services/common/src/components/projects/ProjectDocumentsTab.tsx
```diff
@@ -327,7 +327,7 @@ const ProjectDocumentsTab: FC<ProjectDocumentsTabProps> = ({ project }) => {
         <>
           <Typography.Title level={3}>Application</Typography.Title>
           <Typography.Paragraph>
-            Below are the documents submitted as part of the Major Mine Application.
```
asinn134, line 330: This is the only change

alazar-aot (approved):

PR #3972, opened 2026-08-19
[MDS-6992] NoW verify contact, check existance of address to avoid error

asinn134:
## Objective 

[MDS-6992](https://bcmines.atlassian.net/browse/MDS-6992)

- In the verification stage for a NoW application when the user clicks on a checkbox of a matching contact a white screen with a console error appears for contacts that do not have an address. The fix was to add optional chaining for address to take into account the contacts that do not have an address.

### DIFF
--- a/services/core-web/src/tests/components/Forms/noticeOfWork/VerifyNoWContacts.spec.tsx
+++ b/services/core-web/src/tests/components/Forms/noticeOfWork/VerifyNoWContacts.spec.tsx
@@ -1,11 +1,13 @@
 import React from "react";
 import { render, screen, fireEvent, waitFor, within, act } from "@testing-library/react";
+import { http, HttpResponse } from "msw";
 import VerifyNoWContacts from "@/components/Forms/noticeOfWork/VerifyNoWContacts";
 import * as FORM from "@/constants/forms";
 import { ReduxWrapper } from "@/tests/utils/ReduxWrapper";
 import FormWrapper from "@mds/common/components/forms/FormWrapper";
 import { STATIC_CONTENT } from "@mds/common/constants/reducerTypes";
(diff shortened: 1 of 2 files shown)


services/core-web/src/tests/components/Forms/noticeOfWork/VerifyNoWContacts.spec.tsx
```diff
@@ -152,3 +154,48 @@ describe("VerifyNoWContacts (add new contact flow)", () => {
...
+                        ],
+                    },
+                });
+            })
+        );
+
+        renderWithStore();
+
+        fireEvent.click(screen.getAllByRole("button", { name: /Search Contact/i })[0]);
+        await waitFor(() => expect(screen.getByText(/Add New Core Contact/i)).toBeInTheDocument());
```
matbusby-fw: sonarcloud makes a good point here.  Using screen.findByText is a bit cleaner since it's the async alternative to getByText.
Could do something like:
```
const addCoreContactButton = await screen.findByText(/Add New Core Contact/i)
expect(addCoreContactButton).toBeInTheDocument();

matbusby-fw (dismissed):
Looks good.  I made a small comment agreeing with SonarCloud about a test await call, but it's not really a blocker, so I'll leave it up to you as to whether you want to address it.

services/core-web/src/tests/components/Forms/noticeOfWork/VerifyNoWContacts.spec.tsx
asinn134: That makes sense, this is changed now.

alazar-aot (approved):
Looks great to me!

PR #3973, opened 2026-08-20
[MDS-6995] Define NoW Document Types (Figures, Documents)

alazar-aot:
## Objective 

[MDS-6995](https://bcmines.atlassian.net/browse/MDS-6995)

To better support NoW workflows, we want permitting inspectors to be able to refer to specific types of permit package files so that they don’t have to make changes to the issued permit PDF to make it match the list of figures they want to attach. In order to do this, we need to tag/categories files added to the Permit Package as either 'Figures' or 'Documents', so they can be easily referred to later.

This PR adds support for this by letting users define the type of a NoW file (Document, Figure) when it's being uploaded to Core in the NoW flow. 

**NOTE:** This functionality is gated behind the `inspector_permit_package_type_selector` Flagsmith Flag (Currently enabled in Dev/Test, but NOT Prod)

**NOTE:** Currently, the Title, Date, Author of a permit package file is _not_ mandatory - the Jira ticket makes note that this will likely need to change, but that some conversations need to be had first.

Specifically:
- Added backend support for NoW File Type (Figure, Document) - including migration and models for new DB column, updated API models to handle new fields, updated unit tests as necessary
- Added frontend support for ability to define a NoW file type - added dropdown for selecting the file type, include this value in API calls to backend, added unit tests and updated relevant snapshots

### DIFF
--- a/services/core-api/app/api/now_applications/resources/now_application_document_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_document_resource.py
@@ -163,7 +163,9 @@ class NOWApplicationDocumentResource(Resource, UserMixin):
             xref.preamble_title = data.get('preamble_title')
             xref.preamble_author = data.get('preamble_author')
             xref.preamble_date = data.get('preamble_date')
-            xref.permit_package_document_type_code = data.get('permit_package_document_type_code')
+            permit_package_document_type_code = data.get('permit_package_document_type_code')
+            if permit_package_document_type_code is not None:
+                xref.permit_package_document_type_code = permit_package_document_type_code
 
             now_application = NOWApplication.find_by_application_guid(application_guid)
             if not now_application:
--- a/services/core-api/app/api/now_applications/resources/now_application_document_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_document_resource.py
@@ -86,6 +86,7 @@ class NOWApplicationDocumentResource(Resource, UserMixin):
     parser.add_argument('preamble_author', type=str, required=False)
     parser.add_argument('preamble_date', type=str, required=False)
     parser.add_argument('is_final_package', type=bool, required=False)
+    parser.add_argument('permit_package_document_type_code', type=str, required=False)
     parser.add_argument('final_package_order', type=int, required=False)
     parser.add_argument('description', type=str, required=False)
 
@@ -162,6 +163,7 @@ class NOWApplicationDocumentResource(Resource, UserMixin):
             xref.preamble_title = data.get('preamble_title')
             xref.preamble_author = data.get('preamble_author')
             xref.preamble_date = data.get('preamble_date')
+            xref.permit_package_document_type_code = data.get('permit_package_document_type_code')
 
             now_application = NOWApplication.find_by_application_guid(application_guid)
             if not now_application:
@@ -173,6 +175,7 @@ class NOWApplicationDocumentResource(Resource, UserMixin):
             xref.final_package_order = final_package_order
         else:
             xref.final_package_order = None
+            xref.permit_package_document_type_code = None
 
         xref.save()
         mine_document.save()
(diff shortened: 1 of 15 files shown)


services/core-api/app/api/now_applications/resources/now_application_document_resource.py
```diff
@@ -162,6 +163,7 @@ def put(self, application_guid, mine_document_guid):
             xref.preamble_title = data.get('preamble_title')
             xref.preamble_author = data.get('preamble_author')
             xref.preamble_date = data.get('preamble_date')
+            xref.permit_package_document_type_code = data.get('permit_package_document_type_code')
```
matbusby-fw: Can we add a `None` check before making this assignment?  I think there may be scenarios where the user is just updating the description or something which could lead to this value be nulled when it previously had a value.

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
```diff
@@ -109,15 +127,35 @@ const EditNoticeOfWorkDocumentForm: FC<EditNoticeOfWorkDocumentFormProps> = ({
...
+                  id="is_final_package"
+                  name="is_final_package"
+                  label="Part of permit package"
+                  type="checkbox"
+                  component={RenderCheckbox}
+                />
+              </Col>
+            )}
+            {is_final_package && isFeatureEnabled(Feature.INSPECTOR_PERMIT_PACKAGE_TYPE_SELECTOR) && (
+              <Col md={12} xs={24}>
```
matbusby-fw: If `isInCompleteStatus` is true and `is_final_package` is true, the checkbox for the final package would disapear, right?  In that case, we might want to make the md sizing of the  `permit_package_document_type_code` dynamic based on that.

```suggestion
              <Col md={isInCompleteStatus ? 24 : 12} xs={24}>
```

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
```diff
@@ -17,6 +17,15 @@ import RenderField from "@mds/common/components/forms/RenderField";
 import RenderDate from "@mds/common/components/forms/RenderDate";
 import RenderFileUpload from "@mds/common/components/forms/RenderFileUpload";
 import { IMineDocument, INoWDocument } from "@mds/common/interfaces";
+import { useFeatureFlag } from "@mds/common/providers/featureFlags/useFeatureFlag";
+import { Feature } from "@mds/common/utils/featureFlag";
+import AuthorizationWrapper from "@mds/common/wrappers/AuthorizationWrapper";
+import * as Permission from "@/constants/permissions";
+
+const PERMIT_PACKAGE_DOCUMENT_TYPE_OPTIONS = [
```
matbusby-fw, line 25: I'm on the fence for this, but since we've added these values to the database, it might be good to fetch them dynamically rather than using a const.  It would make maintenance more straight forward.  The traditional way we've handled that was to add things like this to the static content that is served, but that's become enormous over time and I'm always a bit wary of adding to that.  Could be it's own network call as well I suppose and set up a redux selector for it.

matbusby-fw (changes requested):
Looks great.  Just left a couple comments/suggestions.

services/core-api/app/api/now_applications/resources/now_application_document_resource.py
alazar-aot: Yep, great shout! I've added the `None` check

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
alazar-aot: Good catch, I've updated it to use your suggestion

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
alazar-aot: Hmm... I see your point - I'm also on the fence. 

On one hand, I feel like since it's just 2 static values that map 1:1 to the lookup table (and aren't likely to change often, if ever) that the work to refactor it into a new endpoint/Redux selector + the loading/error handling is a bit much to render a dropdown with 2 options. Even if we added a third type, it would be a pretty minor change.

On the other hand, I do see that if the type list grows any larger or if we need more granular configuring options it would be worth the refactor to keep things manageable. 

I'm inclined to say for now this is fine, but if at any point the use-case for this changes, it'd be worth doing. That said, if you/the other devs feel it's worth doing the refactor now, I'm happy to.

matbusby-fw (approved):
Nice!

PR #3974, opened 2026-08-20
[CHORE] Adjustment to final_package_order value being saved for system generated documents

asinn134:
## Objective 

[MDS-6981](https://bcmines.atlassian.net/browse/MDS-6981)

- I noticed in the now_application_document_xref database table the final_package_order for system generated documents don't really accurate represent when a document is the latest system generated document (would show up as 1.1 order in the permit package table on the front end). Instead we would have for example a document that is suppose to be a value to represent 1.1 but the final_package_order in the database might show it as 5.

- This pull request is purely for data correctness, for the future in case someone is working around the order of the permit package table and they want to make use of final_package_order to do something

### DIFF
--- a/services/core-api/app/api/now_applications/models/now_application.py
+++ b/services/core-api/app/api/now_applications/models/now_application.py
@@ -21,7 +21,7 @@ from app.api.now_submissions.models.document import Document
 from app.api.mines.permits.permit_amendment.models.permit_amendment import PermitAmendment
 from app.api.mines.mine.models.mine_type import MineType
 from app.api.mines.permits.permit_conditions.models.permit_conditions import PermitConditions
-from flask import current_app
+LOCKED_NTR_FINAL_PACKAGE_ORDER = -1
 
 
 class NOWApplication(Base, AuditMixin):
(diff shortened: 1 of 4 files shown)


services/core-api/app/api/now_applications/models/now_application.py
```diff
@@ -21,7 +21,7 @@
 from app.api.mines.permits.permit_amendment.models.permit_amendment import PermitAmendment
 from app.api.mines.mine.models.mine_type import MineType
 from app.api.mines.permits.permit_conditions.models.permit_conditions import PermitConditions
-from flask import current_app
```
alazar-aot, line 24: Just want to check - was removing this intentional? Doesn't look like it's being used anywhere so won't cause issues

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
```diff
@@ -92,7 +92,11 @@ const transformDocuments = (
 ) =>
   documents &&
   documents
-    .sort((a, b) => a.final_package_order - b.final_package_order)
+    .sort((a, b) => {
+      if (a.isLockedApplicationForm) return -1;
+      if (b.isLockedApplicationForm) return 1;
+      return a.final_package_order - b.final_package_order;
```
alazar-aot, line 99: If for some reason these both had `isLockedApplicationForm` true then it would return `-1` for both (I think that would cause issues), though it doesn't look like it's currently possible for them to both be `-1` so not an immediate concern, but it's probably worth adding a `return 0` case to be safe!

alazar-aot (dismissed):
Looks great and good catch, approving as neither of my comments are regarding blocking issues, just small improvements!

services/core-web/src/components/noticeOfWork/applications/NOWDocuments.js
asinn134: This has been adjusted now

services/core-api/app/api/now_applications/models/now_application.py
asinn134: Yah this was intentional. I think I forgot to remove it at one point.

alazar-aot (approved):
Magnifique!

PR #3975, opened 2026-08-21
[MDS-6943] - Save "Last Indexed" now document search time

matbusby-fw:
[MDS-6943](https://bcmines.atlassian.net/browse/MDS-6943)

### Summary

Users indexing a Notice of Work (NoW) application's documents for search had no way to tell when it was last indexed, since nothing on the backend actually persisted that timestamp. The frontend already had a "Last run" badge wired up, but it read fields (`last_run_start`/`last_run_end`) that the permits service always returned as `None` — it only tracks in-flight indexing state in Redis with a 7-day TTL, and has no database of its own.

This PR adds durable persistence for indexing run history in core-api and surfaces it in the UI.

image

### Backend changes (`core-api`)

- New table `now_application_document_index_run` (Flyway migration) + model `NowApplicationDocumentIndexRun` — one row per index run for a NoW application, storing status, document count, timestamps, and error info. Modeled on the existing `PermitExtractionTask` pattern for "async job runs in an external service, core-api tracks it durably."
- New Celery task `poll_update_now_application_document_index_status` — polls the permits service's existing status endpoint every 10s until the run reaches a terminal state, then persists the result. Failure handling records status via the task's own run-id argument (not a lookup by Celery task id, which can race with the id being written back to the row).
- `now_application_document_search_resource.py`:
  - `POST .../index` creates a run row and kicks off the poll task.
  - `DELETE .../index` marks an in-flight run `cancelled`.
  - `GET .../index/status` overlays persisted `last_run_start`/`last_run_end`/`document_count` onto the live permits response, and only falls back to the persisted `status`/counts once *our* run row has reached a terminal state — so a freshly-completed job's live status is never shadowed by a stale "running" row.
- Timestamps are serialized with an explicit UTC (`Z`) suffix so the frontend's `new Date(...)` doesn't misinterpret them as local time.

### Frontend changes (`core-web` / `common`)

- Added `document_count` to `NowIndexerStatus` — the "Last run" label now shows the number of whole **documents** processed (e.g. "3 documents") instead of Azure Search chunk counts, which had no meaningful interpretation for users.
- Moved the "Last indexed" label to sit directly under the Index/Re-Index Documents button rather than inline beside it.
- Fixed indexing-status polling: it was continuing to poll every 10s indefinitely after a run completed (`success`/`error`/`never_run`, etc.) instead of stopping once settled. Polling now only runs while a job is actively `running`.

### Testing

- Backend: added/updated unit tests for the new model, Celery task (success/error/retry/on-failure), and resource endpoints
- Frontend: added coverage for the "Last indexed" label (date + document count, singular/plural, hidden while running or never-run) and for the polling fix (asserts polling stops after a terminal status and continues while running), plus updated stale snapshots from the layout change.

### DIFF
--- a/services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
+++ b/services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from app.api.mines.response_models import NOW_DOCUMENT_SEARCH_MODEL
 from app.api.now_applications.models.now_application_document_index_run import (
@@ -87,14 +87,22 @@ class NOWApplicationDocumentIndexResource(Resource, UserMixin):
         if not documents:
             return {'message': 'No indexable documents found for this application.'}, 200
 
-        result = NowApplicationSearchService().index_documents(now_application_guid, documents)
-
         run = NowApplicationDocumentIndexRun.create(
             now_application_guid=now_application_guid,
             status='running',
             document_count=len(documents),
             last_run_start=datetime.utcnow(),
         )
+
+        try:
+            result = NowApplicationSearchService().index_documents(now_application_guid, documents)
+        except Exception as exc:
+            run.status = 'error'
+            run.error_message = str(exc)
+            run.last_run_end = datetime.utcnow()
+            run.save()
+            raise
+
         poll_task = poll_update_now_application_document_index_status.delay(
             str(run.now_application_document_index_run_id))
         run.core_status_task_id = poll_task.id
@@ -134,6 +142,14 @@ class NOWApplicationDocumentIndexStatusResource(Resource, UserMixin):
 
         run = NowApplicationDocumentIndexRun.get_latest_by_now_application_guid(now_application_guid)
         if run:
+            if run.status == 'running' and run.last_run_start and (
+                datetime.utcnow() - run.last_run_start
+            ) > timedelta(minutes=STALE_RUNNING_THRESHOLD_MINUTES):
+                run.status = 'error'
+                run.error_message = 'Indexing run timed out with no status update from the search service.'
+                run.last_run_end = datetime.utcnow()
+                run.save()
+
             status['last_run_start'] = to_utc_isoformat(run.last_run_start)
             status['last_run_end'] = to_utc_isoformat(run.last_run_end)
             status['document_count'] = run.document_count
@@ -151,6 +167,12 @@ class NOWApplicationDocumentIndexStatusResource(Resource, UserMixin):
 # ---------------------------------------------------------------------------
 
 SPATIAL_EXTENSIONS = {'.shp', '.shx', '.dbf', '.prj', '.kml', '.kmz', '.gdb', '.gpx', '.geojson'}
+
+# The polling task retries every 10s for up to 360 attempts (~1 hour) before giving up and
+# marking the run as errored. If the celery worker is restarted mid-poll, the task is lost
+# entirely and on_failure never fires, so this is a fallback so a run doesn't display as
+# "running" forever in that case.
+STALE_RUNNING_THRESHOLD_MINUTES = 90
 
 
 def _collect_indexable_documents(now_application) -> list:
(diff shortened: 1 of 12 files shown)


services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
```diff
@@ -101,7 +130,20 @@ def get(self, now_application_guid):
...
 
-        return NowApplicationSearchService().get_index_status(now_application_guid), 200
+        status = NowApplicationSearchService().get_index_status(now_application_guid)
+
+        run = NowApplicationDocumentIndexRun.get_latest_by_now_application_guid(now_application_guid)
+        if run:
+            status['last_run_start'] = to_utc_isoformat(run.last_run_start)
+            status['last_run_end'] = to_utc_isoformat(run.last_run_end)
+            status['document_count'] = run.document_count
+            if run.status != 'running':
```
alazar-aot, line 156: If the status fails to update for some reason (eg celery hiccup or restart) then this would be stuck 'running' forever - might be worth adding a failsafe/fallback - basically treat 'running' as stale if it's been a certain amount of time?

services/core-api/app/api/now_applications/tasks.py
```diff
@@ -0,0 +1,67 @@
...
+            return Task.__call__(self, *args, **kwargs)
+
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
+        from app.tasks.celery_entrypoint import celery_app
+        with celery_app.app_context():
+            now_application_document_index_run_id = args[0]
+            run = NowApplicationDocumentIndexRun.query.get(now_application_document_index_run_id)
+
+            if not run:
+                raise InternalServerError('NowApplicationDocumentIndexRun not found')
```
alazar-aot: Could we log `current_app.logger.error(...)` and return instead of raise here? That way in the (unlikely) event that `run` isn't available we don't loose the error message?

services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
```diff
@@ -79,6 +88,18 @@ def post(self, now_application_guid):
             return {'message': 'No indexable documents found for this application.'}, 200
 
         result = NowApplicationSearchService().index_documents(now_application_guid, documents)
+
+        run = NowApplicationDocumentIndexRun.create(
```
alazar-aot: It looks like we're starting the document indexing _before_ we create a record of it - what happens if indexing fails for some reason? Would that index run be cancelled, or would it keep going in the background but appear to have failed? Wondering if the order should be flipped (create the index record, then start the indexing)?

alazar-aot (changes requested):
Looks good overall, I left a few small suggestions for potential edge cases/error handling mostly!

asinn134 (dismissed):

services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
```diff
@@ -89,7 +118,15 @@ def delete(self, now_application_guid):
         if not now_application_identity:
             raise NotFound('Notice of Work application not found.')
 
-        return NowApplicationSearchService().cancel_indexing(now_application_guid), 200
+        result = NowApplicationSearchService().cancel_indexing(now_application_guid)
+
+        run = NowApplicationDocumentIndexRun.get_latest_by_now_application_guid(now_application_guid)
+        if run and run.status == 'running':
+            run.status = 'cancelled'
```
alazar-aot, line 125: Does the frontend have `cancelled` as an enum/mappable status? I did a quick grep but didn't see anything - if it doesn't, then I'm not sure what would be displayed if we set the status to cancelled

alazar-aot (commented):
Left one final question/comment!

services/core-api/app/api/now_applications/resources/now_application_document_search_resource.py
matbusby: Good catch — it wasn't mapped. Added `cancelled` to `NowIndexerStatusValue` and to the frontend `configs` map (shows as a gray "Index cancelled" tag), so it no longer falls through to the "Never indexed" default.

matbusby-fw: Oh, good catch!

alazar-aot (approved):
Vunderbar!

PR #3976, opened 2026-08-24
[MDS-6952] - NoD restricted to major mines

matbusby-fw:
## Restrict Notice of Departure (NoD) to Major Mines

[MDS-6952](https://bcmines.atlassian.net/browse/MDS-6952)

### Summary

NoD (Notice of Departure) functionality is intended for Major Mines permitting processes only, but was previously available for Regional Mines as well. This PR restricts NoD access to Major Mines in both MineSpace and Core.

### Changes

**MineSpace**
- The "Notices of Departure" sidebar tab is now only shown for Major Mines (`mine.major_mine_ind`), following the same pattern already used for the "Applications" tab.
- Since sidebar routes are generated from the same `items` list used to render the nav, hiding the tab also removes the underlying route for Regional Mines — direct navigation to the NoD URL is no longer possible.

**Core**
- The "Notices of Departure" item in the Permits & Approvals nav menu is now gated by `isMajorMine`, matching the existing pattern used for "Major Projects".
- Added a page-level guard on `MineNoticeOfDeparture` so that Regional Mines render `PageNotFound` even if the page is reached directly (e.g. via a stale notification link or search result), rather than relying solely on the nav item being hidden.

### Test plan

- [ ] Manually verify in MineSpace: NoD tab hidden for a Regional Mine, visible for a Major Mine
- [ ] Manually verify in Core: NoD nav item hidden for a Regional Mine, direct URL to NoD page shows 404 for a Regional Mine, unaffected for a Major Mine

image
image

### DIFF
--- a/services/core-web/src/components/mine/MineNavigation.tsx
+++ b/services/core-web/src/components/mine/MineNavigation.tsx
@@ -71,10 +71,10 @@ const MineNavigation: FC<MineNavigationProps> = ({ mine, activeButton, openSubMe
           key: "external-authorizations",
           label: <Link to={routes.MINE_EXTERNAL_AUTHORIZATIONS.dynamicRoute(mine.mine_guid)}>Other Ministry Applications and Authorizations</Link>
         },
-        {
+        ...(isMajorMine ? [{
           key: "nods",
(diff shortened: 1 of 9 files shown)


asinn134 (approved):
Looks good!

simensma-fresh (approved):

alazar-aot (approved):
Magnifique 👌

PR #3978, opened 2026-08-25
[MDS-7007] [Bug Fix] Ensure 'system-generated documents' table always displays in NoW (for Administrative Amendments)

alazar-aot:
## Objective 

[MDS-7007](https://bcmines.atlassian.net/browse/MDS-7007)

In QA it was found that if there were no "System-generated Documents" present within an 'Administrative Amendment' NoW application, then the "System-generated Documents" table was not rendered/displayed, despite the Nav link for it being present. 

This PR fixes that issue - the "System-generated Documents" table (and Nav link) are always shown, regardless of whether or not there are any documents to display:

image

I also fixed an unrelated logic bug that made it so that the ability to upload docs was **disabled** when in edit mode - it should have been **enabled**. Moving forward you can only upload documents when it edit mode.


matbusby-fw (approved):
🚀

PR #3979, opened 2026-08-26
[MDS-6989] Fix for new document headers showing up in conditions during extraction

asinn134:
## Objective 

[MDS-6989](https://bcmines.atlassian.net/browse/MDS-6989)

-The issue was the document header is tagged by Azure as a "page header," so the pipeline already knew to skip it. The new template's header is laid out as a table instead, and Azure doesn't tag it that way — it just comes back as regular paragraphs. With no tag to catch it the header text was getting pulled in as conditions.

-Added a fall back where when no pageHeader tag is detected we determine the header block ourselves by identifying text that repeats at the top of a certain number of pages.

### DIFF
--- a/services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
+++ b/services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import re
 from typing import Any, Dict, List, Optional
 
 from app.common.types.context import context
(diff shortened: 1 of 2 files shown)


simensma-fresh (approved):

services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
```diff
@@ -225,4 +226,82 @@ def _identify_bottom_of_first_page_header(paragraphs):
...
+            text_repeats = (
+                _normalize_header_text(next_p.content.get("text")) in repeated_texts
+            )
+            touches_header_block = (
+                next_p.meta["bounding_box"]["top"] - header_bottom
+                <= HEADER_ADJACENCY_TOLERANCE
+            )
+
+            if not text_repeats and not touches_header_block:
+                break
```
alazar-aot, line 300: Question - is it certain that no actual condition text will ever be within 0.05 of the header? This is a super clever approach (nice work), my only concern is that this could pull condition paragraphs into the header, if they happen to be within 0.05 of the `header_bottom`.

If you don't think that's realistic then I'd say this is a non-issue and good to go!

services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
```diff
@@ -225,4 +226,82 @@ def _identify_bottom_of_first_page_header(paragraphs):
...
+    """
+    text_to_pages = {}
+    for p in paragraphs:
+        page = p.meta.get("page")
+        text = _normalize_header_text(p.content.get("text"))
+        if not text or page is None:
+            continue
+        text_to_pages.setdefault(text, set()).add(page)
+
+    return {text for text, pages in text_to_pages.items() if len(pages) >= min_pages}
```
alazar-aot, line 257: Anotha one (question) - to me it looks like `min_pages=3` and `len(pages) >= min_pages` means the fallback won't work for 1 or 2 page permits, is that correct/intentional?

asinn134: Yes that should be correct. I think this should be okay as there hasn't been any permit documents that are just 1 or 2 pages.

services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
asinn134: Good point, its not 100% certain, but given the new header and the sample docs there hasn't been any condition text within 0.05 of the header.

I think it should be fine, and easy to revisit if we ever see it cause a problem.

alazar-aot: Ok perfect - I miiiight be worth adding logging here so we can distinguish between "excluded because it's a confirmed repeating header line" and "excluded because it was merely close to one", something like this:
```
  if not text_repeats and touches_header_block:
      logger.info(
          f"Extending header block via adjacency (not text match): {next_p}, "
          f"gap={next_p.meta['bounding_box']['top'] - header_bottom}"
      )
```
But that's not critical, so I've approved the PR!

services/permits/app/pipelines/permit_condition_extraction/components/filter_conditions_paragraphs.py
alazar-aot: Sounds good!

alazar-aot (approved):

PR #3984, opened 2026-08-27
[MDS-6985] Validate spatial files during import of NoW documents

simensma-fresh:
## Objective 

[MDS-6985](https://bcmines.atlassian.net/browse/MDS-6985)

Added validation of spatial files to NoW file import process (VFCBC). 

- At time of import, detects shapefiles (based on extension), creates MineDocumentBundle for them, and runs validation of the files using the geomark web service
- Added in a spatial files section in core that summarizes all the spatial files detected
- Added ability to mark whether it's a mine boundary - did this a bit more generic, to allow for future expansion to more types
- Added preview (drawer) of a shapefile bundle

image
image
image
image
image


services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -375,49 +375,16 @@ def download_kml_kmz_files(cls, doc, file_path):
...
-
-        current_app.logger.info(f'Completed bundle upload: {bundle.geomark_id}')
-
-        return {'geomark_id': bundle.geomark_id, 'docman_bundle_guid': str(bundle.bundle_guid)}
+        from app.docman.utils.spatial_bundle_service import SpatialBundleService
+
+        result = SpatialBundleService.process_document_guids(
+            bundle_document_guids, name=name, blocking=True)
+
+        current_app.logger.info(f'Completed bundle upload: {result.get("geomark_id")}')
```
matbusby: I see you removed the validate_bundle call here.  Is there a reason we don't need that here anymore?

services/common/src/components/documents/spatial/ViewSpatialDetail.tsx
```diff
@@ -33,7 +96,14 @@ const ViewSpatialDetail: FC<ViewSpatialDetailProps> = ({ spatialDocuments }) =>
   const handleGetSpatialBundles = async () => {
     if (!spatialDocuments[0].geomark_id) {
       const spatialBundles = groupSpatialBundles(spatialDocuments);
-      await dispatch(fetchSpatialBundle(spatialBundles[0].bundle_id));
+      if (spatialDocuments[0].mine_guid && spatialBundles[0]?.bundle_id) {
+        await dispatch(
```
matbusby: If this dispatch fails, we're still running `setBundleNotYetCreate(false)`.  Can we make that conditional on the fetch request result?

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -12,27 +21,208 @@ class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
...
+        docs = cls._fetch_documents_by_manager_guid(document_manager_guids)
+        cls._assert_documents_share_one_mine(docs, mine_guid)
+        if existing:
+            existing._assert_documents_match_bundle_mine(cls._mine_guids_of(docs), mine_guid)
+        return docs
+
+    def _apply_spatial_result(self, name, docman_bundle_guid, geomark_id, validation_status,
+                              validation_error, validation_checks):
+        self.name = name
+        if geomark_id:
```
matbusby: If Re-validation fails before Geomark creates a mark I think we'd end up with an invalid status attached to the previous geomark.  So that would show the previous geomark preview with the invalid status of the new one.  I think I'm reading that correctly.

services/core-api/app/api/mines/documents/resources/mine_document_bundle.py
```diff
@@ -1,18 +1,118 @@
...
+    def patch(self, mine_guid, mine_document_bundle_id):
+        mine_document_bundle = _get_bundle_for_mine(mine_guid, mine_document_bundle_id)
+
+        data = self.parser.parse_args()
+        if 'purpose_codes' not in data:
+            raise BadRequest('purpose_codes is required')
 
-        return mine_document_bundle.json()
\ No newline at end of file
+        mine_document_bundle.set_purpose_codes(data.get('purpose_codes') or [])
```
alazar-aot: If the `or []` fires it would wipe all the existing purpose codes - instead of checking `if 'purpose_codes' not in data` on line 113 could we do an explicit `data.get('purpose_codes') is None` check, since that's what we're here?

services/core-api/app/api/mines/documents/resources/mine_document_bundle.py
```diff
@@ -1,18 +1,118 @@
...
+    if not mine:
+        raise NotFound('Mine not found.')
+    return mine
+
+
+def _assert_documents_belong_to_mine(documents, mine_guid):
+    mine_guid_str = str(mine_guid)
+    for doc in documents or []:
+        if getattr(doc, 'deleted_ind', False):
+            continue
```
alazar-aot: If documents is empty then we skip this `deleted_ind` check, which means this defaults to 'allow'... in theory allowing one mine to use another mine's bundle id and get GET/PATCH access to it?

Might not be understanding this entirely, but it seems like the if documents is empty we should reject instead of allow

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -12,27 +21,208 @@ class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
...
+            raise BadRequest('Spatial documents must belong to a single mine')
+
+    def _assert_documents_match_bundle_mine(self, document_mine_guids, mine_guid):
+        existing_mines = {
+            str(doc.mine_guid)
+            for doc in (self.bundle_documents or [])
+            if doc.mine_guid and not getattr(doc, 'deleted_ind', False)
+        }
+        if not existing_mines:
+            return
```
alazar-aot: Similar concern as `.../resources/mine_document_bundle.py` - the `if not existing_mines: return` check returns before the mine mismatch check, so theoretically possible for one mine to POST to another

services/document-manager/backend/app/docman/utils/spatial_bundle_service.py
```diff
@@ -0,0 +1,380 @@
...
+        if len(bundle_documents) == 1 and extensions.issubset(SINGLE_FILE_EXTENSIONS):
+            return {
+                'name': names[0],
+                'is_single_file': True,
+                'missing_extensions': [],
+                'found_extensions': sorted(extensions),
+                'complete': True,
+            }
+
+        missing = sorted(REQUIRED_SHAPEFILE_EXTENSIONS - extensions)
```
alazar-aot, line 122: I don't see any checks for 'extra' files in the bundle (just this check for the required files) - what happens if there's a PDF or non-spatial file in the bundle? Looks like it would get included, does that cause issues?

services/document-manager/backend/app/docman/utils/geomark_helper.py
```diff
@@ -70,44 +70,40 @@ def add_geomark_to_group(self, geomark_id, group_id):
...
-            os.remove(file_path)
+            )
+            headers = {
+                'Content-Type': multipart_data.content_type
+            }
+            response = requests.post(
+                url,
+                data=multipart_data,
+                headers=headers,
+                timeout=60,
```
alazar-aot: Just want to check - is this enough time to upload large bundles? If not then a 'valid' bundle that takes longer than 60s would be failed

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
matbusby-fw: If a bundle fails revalidation I think it would mark the validation as failed, but keep the previous geomark id.  I think that would propogate to the frontend as well since that geomark_id is what we use to fetch the preview

services/document-manager/backend/app/docman/utils/document_upload_helper.py
```diff
@@ -375,49 +375,16 @@ def download_kml_kmz_files(cls, doc, file_path):
...
-        db.session.add(bundle)
-        db.session.commit()
-
-        if bundle.error:
-            raise BadRequest(bundle.error)
-
-        current_app.logger.info(f'Completed bundle upload: {bundle.geomark_id}')
-
-        return {'geomark_id': bundle.geomark_id, 'docman_bundle_guid': str(bundle.bundle_guid)}
+        from app.docman.utils.spatial_bundle_service import SpatialBundleService
```
matbusby-fw, line 349: Was dropping the `cls.validate_bundle(bundle_documents)` intentional?  Feels like it's still a useful safety check.

services/common/src/components/documents/spatial/ViewSpatialDetail.tsx
```diff
@@ -33,7 +96,14 @@ const ViewSpatialDetail: FC<ViewSpatialDetailProps> = ({ spatialDocuments }) =>
...
+          fetchSpatialBundle({
+            mineGuid: spatialDocuments[0].mine_guid,
+            mine_document_bundle_id: spatialBundles[0].bundle_id,
+          })
+        );
+      }
```
matbusby-fw: If the dispatch fails, we're still calling `setBundleNotYetCreated(false)`.  Can we make the call conditional based on the dispatch result?

matbusby-fw (changes requested):
Looking great.  Just a couple small things I flagged.

alazar-aot (changes requested):
Phew, she's a big one - nice work though! Looks great overall, I did leave a few comments.

services/document-manager/backend/app/docman/utils/document_upload_helper.py
simensma-fresh: Wops, forgot to use the shared validation logic - fixed

services/core-api/app/api/mines/documents/resources/mine_document_bundle.py
simensma-fresh: Updated to 
- Clear purpose codes with `[]` (intentional)
- Not allow nulls

services/document-manager/backend/app/docman/utils/geomark_helper.py
simensma-fresh: Updated

asinn134 (dismissed):

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -12,27 +20,212 @@ class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
...
+                                   validation_status=None,
+                                   validation_error=None,
+                                   validation_checks=None,
+                                   preserve_purposes=True,
+                                   mine_guid=None):
+        """Create or update a Core bundle and link MineDocuments by document_manager_guid."""
+        existing = cls._find_bundle_for_spatial_result(name, docman_bundle_guid,
+                                                       document_manager_guids)
+        docs = cls._resolve_documents_for_spatial_result(document_manager_guids, mine_guid, existing)
+        preserved_purposes = existing.purpose_codes if (existing and preserve_purposes) else []
```
alazar-aot: To me it looks like if `preserved_purpose=false` then this evaluates to `[]`, which means line 223 (`if preserved_purposes:`) would be false, and it would skip the `bundle.set_purpose_codes(preserved_purposes)` line and not clear the codes? 

It doesn't look like anywhere in the code currently sets `preserved_purpose=false` so it's not actively an issue - maybe I'm not understanding what False is meant to represent here? (eg does it mean "clear the codes" or "leave them alone")

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
```diff
@@ -12,27 +20,212 @@ class MineDocumentBundle(SoftDeleteMixin, AuditMixin, Base):
...
+
+    def _apply_spatial_result(self, name, docman_bundle_guid, geomark_id, validation_status,
+                              validation_error, validation_checks):
+        self.name = name
+        self.geomark_id = geomark_id
+
+        if docman_bundle_guid:
+            self.docman_bundle_guid = docman_bundle_guid
+        if validation_status is not None:
+            self.validation_status = validation_status
```
alazar-aot: Not a blocking issue but felt it's worth flagging - if a caller ever sends `validation_status` without `validation_error`, the flask/restx parser will set `validation_error` to `None` and wipe any captured codes (since the parser has `default=None`/`store_missing=True`).

Currently the caller sends both, but if that ever changes we'd loose the `validation_error`

services/document-manager/backend/app/docman/utils/spatial_bundle_service.py
```diff
@@ -0,0 +1,392 @@
...
+            bundle.geomark_id = result['geomark_id']
+        cls._link_documents(bundle, bundle_documents, set_upload_completed=set_upload_completed)
+        result['docman_bundle_guid'] = str(bundle.bundle_guid)
+        return result
+
+    @classmethod
+    def _result_from_existing_bundle(cls, bundle, documents):
+        """Rebuild a Core-sync payload without calling Geomark again."""
+        document_guids = [str(d.document_guid) for d in documents]
+        if bundle.geomark_id:
```
alazar-aot: If we check this before checking for `bundle.error` it's possible that we set a previously invalid bundle to valid - suggest also checking for `bundle.error` here or (or check `bundle.error` first, then `bundle.geomark_id`)

alazar-aot (changes requested):
Last few things, thanks for updating the other items 👊

matbusby-fw (dismissed):

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
simensma-fresh: Agreed, that is confusing. I removed that `preserve_purposes` flag

services/core-api/app/api/mines/documents/models/mine_document_bundle.py
simensma-fresh: Updated this logic a bit.
Made `validation_status` required at the API level, and skip all of these None checks. In the end, a new run should reset all the fields

services/document-manager/backend/app/docman/utils/spatial_bundle_service.py
simensma-fresh: Simplified the logic here a bit, call Geomark for existing bundles as well

alazar-aot (approved):
I think she's ready for the world...

PR #3987, opened 2026-09-01
[CHORE] - Fix flaky test failures

matbusby-fw:
## Objective 

Some verifiable credential based tests were showing intermittent failures in the CI pipeline.  This was tied to the fact that contstraints in the downstream logic specifically do not allow exploration permits and the previous setup would randomly choose a prefix from this list: `['C-', 'CX-', 'M-', 'M-', 'P-', 'PX-', 'G-', 'Q-']`.  When a 'CX' or 'PX' was chosen, this would cause a failure in the test.  Fixed by manually setting a permit number when generating.


asinn134 (approved):

alazar-aot (approved):
Lovely, thank you!

PR #3989, opened 2026-09-03
[MDS-6985] Revalidate all spatial files during nNoW reimport

simensma-fresh:
## Objective 

[MDS-6985](https://bcmines.atlassian.net/browse/MDS-6985)

Found an issue with the NoW spatial file logic, where it would sometimes skip some files.

Issue: It would revalidate spatial files from the last import job, ignoring any files that were previously imported
Solution: Do the spatial file validation on all the imported fields of a Now application

### DIFF
--- a/services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
+++ b/services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
@@ -32,6 +32,20 @@ def mine_guid_from_documents(documents):
     return None
 
 
+def imported_documents_for_application(now_application_guid):
+    """Every document imported for a NoW, across all of its import jobs.
+
+    An import job only ever contains the documents that still needed importing, so scoping
+    spatial processing to one job would leave previously imported files unvalidated.
+    """
+    documents = {}
+    for job in ImportNowSubmissionDocumentsJob.find_by_now_application_guid(now_application_guid):
+        for import_document in job.import_now_submission_documents:
+            if import_document.document:
+                documents[import_document.document_id] = import_document.document
+    return list(documents.values())
+
+
 def sync_bundle_to_core(result, authorization_token, mine_guid):
     """Persist spatial bundle validation result on Core MineDocumentBundle."""
     if not mine_guid:
@@ -107,11 +121,8 @@ def process_now_spatial_bundles(self, import_now_submission_documents_job_id, mi
         )
         return {'success': False, 'message': 'Import job not found'}
 
-    # Collect imported Document records for this job
-    documents = []
-    for import_doc in import_job.import_now_submission_documents:
-        if import_doc.document_id and import_doc.document:
-            documents.append(import_doc.document)
+    # Bundle and validate every document imported for this NoW, not just this job's.
+    documents = imported_documents_for_application(import_job.now_application_guid)
 
     if not documents:
         logger.info(f'No documents to process for job {import_now_submission_documents_job_id}')


services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
```diff
@@ -107,11 +121,8 @@ def process_now_spatial_bundles(self, import_now_submission_documents_job_id, mi
         )
         return {'success': False, 'message': 'Import job not found'}
 
-    # Collect imported Document records for this job
-    documents = []
-    for import_doc in import_job.import_now_submission_documents:
-        if import_doc.document_id and import_doc.document:
-            documents.append(import_doc.document)
+    # Bundle and validate every document imported for this NoW, not just this job's.
+    documents = imported_documents_for_application(import_job.now_application_guid)
```
alazar-aot, line 125: Looks like this re-validates every spatial bundle from every prior import job each time a new job finishes (not just the newly imported files), might not be a major issue but it does seem like we'll be generating a lot of redundant Geomark traffic as a NoW application accumulates import jobs. 

Is there an (easy) way to shortcut/cache already-validated files to avoid re-validating them each time? If that would be too complex then I'm OK to move forward as-is

services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
```diff
@@ -32,6 +32,20 @@ def mine_guid_from_documents(documents):
...
 
 
+def imported_documents_for_application(now_application_guid):
+    """Every document imported for a NoW, across all of its import jobs.
+
+    An import job only ever contains the documents that still needed importing, so scoping
+    spatial processing to one job would leave previously imported files unvalidated.
+    """
+    documents = {}
+    for job in ImportNowSubmissionDocumentsJob.find_by_now_application_guid(now_application_guid):
```
alazar-aot, line 42: `find_by_now_application_guid` returns all jobs regardless of their status, should there be a status check here to ensure that we only bundle documents with a "valid to bundle" status? Just want to ensure we don't end up bundling any files that shouldn't be

alazar-aot (changes requested):
Left two comments/questions, I don't think either is critical so happy to re-review and approve if you don't think these are PR-blocking

services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
simensma-fresh: Yup, this was intentional doing a full re-import every time. In the end this is a low traffic part of the app, with a limited amount of files, and an action that is likely only ever run once, so I'm not worried about it. Which makes it a whole lot easier to test, and gives us the ability to reimport things if we want to capture more data from Geomark.

services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
simensma-fresh: Did it this way to capture documents that were successfully imported even if it failed for some, which can happen.

services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
alazar-aot: Fair enough, that sounds good to me!

services/document-manager/backend/app/tasks/process_now_spatial_bundles.py
alazar-aot: Gotcha, sounds good. In that case consider this resolved

alazar-aot (approved):

PR #3991, opened 2026-09-03
[MDS-7015] red boundary box not showing in pdf viewer for permit conditions

asinn134:
## Objective 

[MDS-7015](https://bcmines.atlassian.net/browse/MDS-7015)

- The "Open Permit in Document Viewer" side-by-side PDF view in the permit conditions page wasn't scrolling to or highlighting the selected condition. The cause of this was the highlight/scroll logic only ran inside the PDF's one-time `documentLoad` event, so it fired at most once per viewer open and never updated as you clicked through different conditions. It also only worked if you had selected a condition before you clicked "Open Permit In Document Viewer" button.

- The highlight/page-navigation now re-runs whenever the selected condition's page or bounding box actually changes, not just on the PDF's initial load. 

- The change also always clears the previous highlight before drawing the next one so there aren't any leftover bounding boxes just sticking around

- Also locked the bounding boxes so the user can't drag them boxes around

### DIFF
--- a/services/common/src/components/syncfusion/DocumentViewer.tsx
+++ b/services/common/src/components/syncfusion/DocumentViewer.tsx
@@ -127,15 +127,32 @@ export const ViewPdf: React.FC<ViewPDFProps> = ({
   onInit = null
 }) => {
   const pdfViewerRef = useRef<any>(null);
+  const [isDocumentLoaded, setIsDocumentLoaded] = useState(false);
+  const hasDrawnAnnotationRef = useRef(false);
+
+  useEffect(() => {
+    setIsDocumentLoaded(false);
+    hasDrawnAnnotationRef.current = false;
+  }, [documentPath]);
+
+  const { pageNumber, boundingBox } = annotationLocation ?? {};
+  const { top, right, bottom, left } = boundingBox ?? {};
+
+  useEffect(() => {
+    if (!isDocumentLoaded || !pdfViewerRef.current) {
+      return;
+    }
+
+    if (annotationLocation) {
+      hasDrawnAnnotationRef.current = true;
+      addAnnotationToPDFViewer(pdfViewerRef.current, pageNumber, boundingBox);
+    } else if (hasDrawnAnnotationRef.current) {
+      pdfViewerRef.current.annotation.clear();
+    }
+  }, [isDocumentLoaded, pageNumber, top, right, bottom, left]);
 
   const handleDocumentLoaded = () => {
-    if (pdfViewerRef.current && annotationLocation) {
-      addAnnotationToPDFViewer(
-        pdfViewerRef.current,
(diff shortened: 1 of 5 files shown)


simensma (approved):

simensma-fresh (dismissed):

services/common/src/components/syncfusion/DocumentViewer.tsx
```diff
@@ -127,15 +127,32 @@ export const ViewPdf: React.FC<ViewPDFProps> = ({
...
+      return;
     }
+
+    if (annotationLocation) {
+      hasDrawnAnnotationRef.current = true;
+      addAnnotationToPDFViewer(pdfViewerRef.current, pageNumber, boundingBox);
+    } else if (hasDrawnAnnotationRef.current) {
+      pdfViewerRef.current.annotation.clear();
+    }
+  }, [isDocumentLoaded, pageNumber, top, right, bottom, left]);
```
alazar-aot, line 152: If `documentPath` and `annotationLocation` both change in the same render then I think it's possible that we end up drawing the annotation on the old/reloading PDF. Might be better to make `isDocumentLoaded` only true when `pdfViewerRef.current` corresponds to the currently loaded `documentPath`, or add `documentPath` to the annotation's `useEffect` dependency array directly (and re-check load state against it before calling `addAnnotationToPDFViewer`)

services/common/src/components/syncfusion/pdfViewerAnnotations.ts
```diff
@@ -15,14 +15,18 @@ export const addAnnotationToPDFViewer = (
   page?: number,
   boundingBox?: IPdfViewerBoundingBox
 ) => {
-  pdfViewer.navigation.goToPage(page || 0);
+  pdfViewer.annotation.clear();
+
+  if (page) {
+    pdfViewer.navigation.goToPage(page);
+  }
+
```
alazar-aot, line 23: Since we fall back to `0` on line 32 if `page` is unknown, if we skip navigation here in the same scenario (page unknown) we'll draw the annotation on a page the viewer never scrolls to. 

Probably better to resolve `page` once, and use that value for both places so they can't disagree/differ

alazar-aot (changes requested):
Requesting a few small tweaks to make things a bit tighter, but great work overall!

services/common/src/components/syncfusion/pdfViewerAnnotations.ts
asinn134: This has been changed now

services/common/src/components/syncfusion/DocumentViewer.tsx
asinn134: This has been changed as well now

alazar-aot (approved):
Looks great, thank you!

PR #3992, opened 2026-09-09
[MDS-6987] Add Figures and Documents to Condition Data Variable Menu

alazar-aot:
## Objective 

[MDS-6987](https://bcmines.atlassian.net/browse/MDS-6987)

To allow permitting inspectors to more easily work with permit package references, the Condition Data Variable (CDV) Menu is to be updated to allow inspectors to insert permit package references into condition text when drafting the permit, so that the references remain accurate when permit package files are reordered or replaced. 

- [Commit 1] Removed old CDV menu code and Flagsmith gating - moving forward we only use the 'new' CDV picker without any Flagsmith gate (Cleanup).
- [Commit 2] Add new Flagsmith Flag for this functionality.
- [Commit 3] Moved logic for determining permit-package ordering into single helper/utility function, updated unit tests accordingly and pointed all files that need this functionality to use the new helper. Added 'Permit Package files -> Figures/Documents' to CDV menu, gated by feature flag (these selections insert 'tokens' for the permit package files into the condition text)
- [Commit 4] Replaced the original Permit Condition 'plain text area' with a react-quill editor for the PermitConditionForm's condition filed (this is UI only), which includes a custom Quill embed blot that renders the 'label' (green or red 'pill' of Doc/Fig index and title) for an inserted Figure/Document reference. Small fix to preamble text area (keep using original textarea, not rich-text Quill editor) and run a WGAC pass on the changes using Claude
- [Commit 5] Added comment regarding preservation of GUID, for when 'Archive/Replace' functionality is implented (risk of breaking the CDV references if guid is not maintained)
- [Commit 6] Added helper/utility function that checks if a Figure/Document is actually being referenced in any permit condition, used to warn users if they try to remove a file from the permit package that it will break the reference - also added unit tests for this new functionality
- [Commit 7 - Bug Fix] Fixed bug where copying and pasting condition text introduced a series of errors - related to not having a unique CSS/className attribute on the 'pill' elements
- [Commit 8 - UX Improvement] Updated 'this figure/doc is referenced' warning to fire immediately after trying to de-select the 'part of permit package'checkbox, instead of before trying to save
- [Commit 9] Ensured that issued permits correctly resolve Figure/Document references - they appear as the literal index (1.2, 1.3, etc) and title - not in a 'pill'. If a removed reference is included, it gets the literal text of 'Reference unavailable'.
- [Commit 10] Minor patches for es-lint warnings and overall code improvements + cleanup
- [Commit 11] Claude Pass: Refactor `@lru_cache` to instead determine locked doc value once, and pass through rather than global cache
- [Commit 12] Fix failing unit tests


services/core-api/app/api/now_applications/now_template_transformer.py
```diff
@@ -61,7 +86,95 @@ def transform_variables_to_data(now_application, permit_amendment, mine, total_l
...
+        now_application.application_type_code == 'NOW'
+        and _technical_review_ever_completed(now_application))
+
+    locked = None
+    reals = []
+
+    for doc in now_application.documents:
+        if not doc.is_final_package:
+            continue
+        if is_locked_row_eligible and doc.final_package_order == LOCKED_NTR_FINAL_PACKAGE_ORDER:
```
matbusby-fw: I don't think this quite matches the check on the frontend.  I'm not sure how much danger there is of this, but if there were more than one match, the latest one would overright any previous ones.  What if we only treated the doc as locked if it's identified by the `locked_ntr_guid` (or maybe the newest NTR?)

services/core-api/app/api/now_applications/now_template_transformer.py
```diff
@@ -61,7 +86,95 @@ def transform_variables_to_data(now_application, permit_amendment, mine, total_l
...
+
+    def _replace(match):
+        label = resolve_permit_package_file_reference(match.group(1), label_map)
+        return label if label is not None else BROKEN_PERMIT_PACKAGE_FILE_REFERENCE_TEXT
+
+    return PERMIT_PACKAGE_FILE_TOKEN_PATTERN.sub(_replace, text)
+
+
+def replace_condition_value_with_data(condition, condition_var, label_map=None):
+    condition = resolve_permit_package_file_tokens(condition, label_map)
```
matbusby-fw: I think there's an edge case scenario here where the titles created in this step end up matching variable names that would then get substituted even though they were meant to be free text.  (ie. a document containing `Total_Liability` in the name would resolve to that variable's value).  I think we can avoid that by moving this step after the re.sub step.  What do you think?

services/common/src/components/permits/ConditionRichTextEditor.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return;
+    }
+    const currentText = quillDeltaToConditionText(editor.getContents() as unknown as ConditionDelta);
+    const nextText = input.value ?? "";
+    if (currentText !== nextText) {
+      const delta = conditionTextToResolvedQuillDelta(nextText, resolveReference);
+      editor.setContents(delta as any);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [input.value]);
```
matbusby-fw, line 117: Since we aren't re-running this when `noticeOfWork` or `nowProgress` change, maybe we can add another useEffect with those for dependencies and have it re-run `conditionTextToResolvedQuillDelta` and then `setContents` on resolve?  I'm thinking if another component makes an update to the notice of work it could create stale data, but maybe that's not an issue here.

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
```diff
@@ -95,6 +102,23 @@ const EditNoticeOfWorkDocumentForm: FC<EditNoticeOfWorkDocumentFormProps> = ({
     setUploadedFiles(remainingFiles);
   };
 
+  // If a user unchecks the "part of permit package" checkbox for a file that's being referenced in a permit condition, it triggers a warning popup.
+  // This prevents a user from accidentally removing a referenced permit pacakge file, which breaks the reference.
```
matbusby-fw: Small typo on `pacakge`

services/common/src/utils/permitPackageDocuments.ts
```diff
@@ -0,0 +1,199 @@
...
+    orderLabel: getPermitPackageOrderLabel(index, hasLockedRow, !!doc.isLockedApplicationForm),
+  }));
+};
+
+/**
+ * The core-uploaded (non-submission) permit package documents tagged as the given Figure/Document type, each with its live display order label.
+ * This is the data source for the Condition Data Variable picker's "Permit Package Files" menu.
+ */
+export const getPermitPackageFilesByType = (
+  noticeOfWork: any,
```
matbusby-fw: Is there a reason we can't use an interface here?  I always prefer proper typing where possible.

matbusby-fw (changes requested):
Looks great!  Just had a few comments/suggestions.

services/core-api/app/api/now_applications/now_template_transformer.py
alazar-aot: Good catch - it was the case that if there was more than one '1.1' document, the latest would override the previous. 

I've updated it so that the locked row is identified via `now_application.locked_ntr_guid` (same lookup the frontend uses) instead of on `final_package_order == LOCKED_NTR_FINAL_PACKAGE_ORDER`. Unit tests have also been updated to reflect this!

services/core-api/app/api/now_applications/now_template_transformer.py
alazar-aot: Another good catch! There would have been issues if the file name matched an existing variable, I've updated it so that the variable substitution resolves _before_ extracting the `{permit_package_file:<guid>}` tokens, so a resolved document title can never get re-matched and clobbered by the variable regex

services/core-web/src/components/Forms/noticeOfWork/EditNoticeOfWorkDocumentForm.tsx
alazar-aot: Me fingers be moving too fast! I've got this fixed now

services/common/src/utils/permitPackageDocuments.ts
alazar-aot: No valid reason (other than me being a bit lazy) - a proper interface is definitely the right way to go so I've updated this to use one!

services/common/src/components/permits/ConditionRichTextEditor.tsx
alazar-aot: Hmm... currently there's no code path where `noticeOfWork` or `nowProgress` can change and cause the Quill data to become stale (the only place the `ConditionRichTextEditor` is used is within the Permit Conditions, and neither `noticeOfWork` or `nowProgress` can be modified from there) - I also believe that if I was to add the new useEffects it will cause issues due to how the rich text editor handles user typing (the new useEffects firing would jump the cursor position to the start of the doc while the user types - and fixing that looks a decent bit more complicated). 

Given that there's no way for this to currently be reached, and that the only way this will ever be a real issue is if the Documents-tab and Conditions-tab get mounted to the same view, I'm inclined to leave it as-is.

Let me know what you think @matbusby-fw, if we do leave it as-is I'm happy to add a quick code comment explaining that there's a possibility for the beast to awaken in the future if the Doc + Conditions are ever combined in a single view.

matbusby-fw (approved):
Rock on till the end of time!

PR #3993, opened 2026-09-10
[MDS-7009] - Enhance/update error reporting features

matbusby-fw:
## Objective 

[MDS-7009](https://bcmines.atlassian.net/browse/MDS-7009)

Implemented enhanced error reporting features. 
- detailed error context
  - severity
  - description
  - recurrence information
- integrating these details into emails, modals, and logs for improved troubleshooting and monitoring.
- Changed log linking to Openshift/Loki in accordance with our newer architectural patterns

Here is a concise summary you can paste into your PR description:

---

### **Postgres Base Image Upgrade to Debian 12 (Bookworm)**

* **Issue**: Integration test workflows in CI started failing during the `postgres` Docker build because Debian 11 (Bullseye) reached End-of-Life (EOL) on September 7, 2026. Upstream security repositories for Bullseye expired, causing `apt-get update` to fail with exit code 100. The previous base image (`postgis/postgis:16-3.5`) was built on Debian 11 Bullseye and has not received an upstream Debian 12 release.
* **Fix**:
  * Switched the base image in [`services/postgres/Dockerfile`](file:///home/mat/Code/Work/mds/services/postgres/Dockerfile) to the official `postgres:16-bookworm` (Debian 12).
  * Installed `postgresql-16-postgis-3` and `postgresql-16-postgis-3-scripts` via `apt` directly alongside build dependencies, eliminating the need for repository snapshot pins or expired archive workarounds.
  * Modernized Dockerfile `ENV` syntax and cleaned apt package caches.
* **Validation**:
  * Successfully built the image locally via `docker compose -f docker-compose.ci.yaml build postgres`.
  * Verified database startup, database initialization, and extension availability (`PostGIS 3.6` and `oracle_fdw 2.6.0` with Oracle Instant Client 21.6) inside a fresh container.

image

image

### DIFF
--- a/services/core-api/app/api/report_error/resources/report_error_resource.py
+++ b/services/core-api/app/api/report_error/resources/report_error_resource.py
@@ -1,6 +1,8 @@
 from app.api.services.email_service import EmailService
 from flask_restx import Resource
 from flask import request
+from markupsafe import Markup, escape
+from urllib.parse import quote
 
 from app.api.utils.access_decorators import (requires_any_of, VIEW_ALL)
 from app.api.utils.resources_mixins import UserMixin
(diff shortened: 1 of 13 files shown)


simensma-fresh (approved):

services/core-api/app/api/report_error/resources/report_error_resource.py
```diff
@@ -12,6 +14,25 @@
...
+
+def _build_observe_logs_link(trace_id):
+    log_query = (
+        '{ kubernetes_namespace_name="' + Config.OPENSHIFT_NAMESPACE + '", '
+        'kubernetes_labels_app="core-api", kubernetes_container_name="app" }'
+    )
+    if trace_id:
+        log_query += f' |= "trace_id={trace_id}"'
+    log_query += ' | json'
+    return (f'{Config.OBSERVE_LOGS_BASE_URL}/dev-monitoring/ns/{Config.OPENSHIFT_NAMESPACE}/logs'
```
alazar-aot, line 32: Looks like this is hardcoding `dev-monitoring` in the path, should this be a variable in the same way that `OPENSHIFT_NAMESPACE` is, so the link resolves correctly in test/prod?

matbusby-fw: We could include it, but even though it's labeled with `dev` that's actually a fixed, constant path for the OpenShift console's logging-view-plugin, not something that varies by target namespace or environment.

alazar-aot: Ohh ok gotcha - in that case I don't think it needs to be a parameter, if it's not going to be changing. I'm fine to keep it as-is, I'll mark as resolved

alazar-aot (approved):

PR #3995, opened 2026-09-11
[MDS-7018] Include manually uploaded files in spatial validation + fix error in certain cases

simensma-fresh:
## Objective 

[MDS-7018](https://bcmines.atlassian.net/browse/MDS-7018)

- Included manually uploaded application files in NoW spatial validation. Docman doesn't know which files these are, so they're passed in from the core-api.

- Fixed an error that pops up in the docman worker if you're trying to update a record using the AuditMixin (e.g. happens if you re-validate a spatial file with a different user than the first attempt). It would try to get the update_user from the user request token, which is not present as it's an async task. In this case, fall back to the `mds` user, which is in use as the "system user" for docman

- Little cleanup: Removed the last use of the flower celery API in docman, instead use `task.delay` / `celery.control.revoke` to be consistent.


services/document-manager/backend/app/services/commands_helper.py
```diff
@@ -139,27 +139,30 @@ def start_zip_job(job_type, docs, task, zip_file_name=None):
...
+        import_job.celery_task_id = result.id
         import_job.save()
 
         # Create the response message
-        message = f'Added an Import Notice of Work Submission Documents job with ID: {import_now_submission_documents_job_id}, TaskID: {response["task-id"]} to the task queue: {len(import_job.import_now_submission_documents)} docs will be imported.'
+        message = f'Added an Import Notice of Work Submission Documents job with ID: {import_now_submission_documents_job_id}, TaskID: {result.id} to the task queue: {len(import_job.import_now_submission_documents)} docs will be imported.'
     except Exception as e:
         message = f'Failed to add an Import Notice of Work Submission Documents job to the task queue: {str(e)}'
-
+        raise
```
alazar-aot: The error message built on 164 won't be returned to the caller since we raise on 165, and the caller (`import_now_submission_documents_job.py`) has no try/except, so this becomes an unhandled 500 instead of a proper error response. Could we have the caller handle this explicitly via try/except that returns the correct status code w/the error message?

alazar-aot (changes requested):

services/document-manager/backend/app/services/commands_helper.py
simensma-fresh: Wops, intention was to log the message 😓 

Think the 500 is still appropriate in this case, but added a nicer message to it

alazar-aot: I figured we probably wanted the message, and I do think a 500 is the correct code - thanks for propagating the message up!

alazar-aot (approved):
Apprové!

PR #3996, opened 2026-09-11
[MDS-7019] - Fix replace documents in Major Projects on Core

matbusby-fw:
## Objective 

[MDS-7019](https://bcmines.atlassian.net/browse/MDS-7019)

Fix document replacement display issues in Major Projects where replacing an existing document caused version ordering in the accordion to invert (newest at bottom, oldest at top) and left the table row displaying stale details. Also migrates `DecisionPackageTab` from JavaScript to TypeScript (`.tsx`).

---

## Source of the Issue

1. **Version History Inversion in Constructor**:
    - In `ReplaceDocumentModal.tsx`, the list of existing versions was reversed before creating the new version entries. However, the `MineDocument` model constructor also performs its own sort/reversal handling for version history. This resulted in a double reversal: the newest replaced version was pushed to the bottom of the accordion list instead of remaining at the top.

2. **Date Format Inconsistency**:
    - The replacement timestamp was previously formatted without time precision (or in non-ISO format), causing date comparisons against ISO backend timestamps (`YYYY-MM-DDTHH:mm:ssZ`) in Ant Design table sorters to tie or sort inconsistently.

3. **Loss of Document Subclass / Metadata**:
    - When preparing the replaced document, it was being re-instantiated strictly as a generic `MineDocument`, dropping subclass-specific attributes and prototype methods associated with `MajorMineApplicationDocument` (e.g., category codes, labels, and document subtypes).

4. **Missing Callback Propagation (`onReplaceDocument`)**:
    - `DocumentTable` opened the replacement modal and triggered the upload action, but did not accept or invoke an `onReplaceDocument` callback upon completion. Because the parent tabs (`ProjectDocumentsTab`, `MajorMineApplicationTab`, `DecisionPackageTab`) were not notified of the replacement, the parent container never re-fetched the updated project/document list from the backend, leaving the table display out of sync until a manual page reload.

---

## Why This Fixes It

1. **Corrects Version Array Ordering & Subclass Preservation (`ReplaceDocumentModal.tsx`)**:
    - Retains the original document's constructor prototype (e.g., `MajorMineApplicationDocument`) so all custom getters and metadata remain intact.
    - Sets `update_timestamp` using `new Date().toISOString()` for exact chronological comparison against backend timestamps.
    - Ensures versions are chronologically ordered before passing to the document constructor, ensuring index 0 correctly represents the latest version at the top of the version accordion.
    - Awaits `props.handleSubmit(newDocument)` before closing the modal.

2. **Propagates Refresh Callbacks (`DocumentTable.tsx`, `ProjectDocumentsTab.tsx`, `ProjectDocumentsTabSection.tsx`)**:
    - Added `onReplaceDocument?: (document: T) => void | Promise<void>` prop to `DocumentTableProps`.
    - Wired `onReplaceDocument` in `ProjectDocumentsTab` to trigger `refreshData` / `refreshAmsApps` across all document sections (primary, appendix, supporting, ministry decision, authorization amendment, and AMS documents).
    - Wired `onReplaceDocument={fetchData}` in `MajorMineApplicationTab.tsx` and `DecisionPackageTab.tsx`.

3. **Migrated `DecisionPackageTab` to TypeScript (`DecisionPackageTab.tsx`)**:
    - Converted the legacy React class component to a TypeScript functional component (`FC`) using React hooks (`useDispatch`, `useSelector`, `useParams`, `useFeatureFlag`, `useState`, `useEffect`).
    - Defined `DecisionPackageTabProps` to remain backward-compatible with unit tests that pass props directly while cleanly falling back to Redux selectors and React Router hooks when mounted in `Project.tsx`.
    - Deleted the obsolete `DecisionPackageTab.js`.

---

## Automated Tests & Verification

- **Common Unit Tests**:
    - `DocumentTable.spec.tsx`: Added test verifying that `onReplaceDocument` is invoked with the updated document when a replacement is submitted.
    - `ReplaceDocumentModal.spec.tsx`: Added test verifying that document version history correctly sorts newest versions first.
  ```bash
  yarn workspace @mds/common run test DocumentTable.spec.tsx ReplaceDocumentModal.spec.tsx


services/common/src/components/documents/ReplaceDocumentModal.tsx
```diff
@@ -92,12 +93,16 @@ const ReplaceDocumentModal: FC<ReplaceDocumentModalProps> = (props) => {
...
+        const existingVersionsAsc = (document.versions ?? []).slice().reverse();
+        const newVersionsChronological = [...existingVersionsAsc, ConnectedVersion.data as IMineDocumentVersion];
+        const newDocument = new docConstructor({
           ...updatedDocument,
-          versions: [ConnectedVersion.data as IMineDocumentVersion, ...document.versions].reverse(),
+          versions: newVersionsChronological,
         });
 
-        props.handleSubmit(newDocument).then(() => dispatch(closeModal()));
+        await props.handleSubmit(newDocument);
```
alazar-aot: If `handleSubmit`/`onReplaceDocument` rejects here, it looks to me like the modal never closes and the table would show the replacement as done (even though it failed/was rejected). Can we wrap this in a try/catch and roll back the update (or show an error) on failure?

services/core-web/src/components/mine/Projects/DecisionPackageTab.tsx
```diff
@@ -0,0 +1,447 @@
...
+    return () => {
+      window.removeEventListener("scroll", handleScroll);
+    };
+  }, []);
+
+  useEffect(() => {
+    if (tab === "project-decision-package") {
+      handleFetchData();
+    }
+  }, [tab]);
```
alazar-aot, line 140: It looks like this will fire both on page mount and again via the [tab] effect (since the tab is already set on load), so we end up doing two fetches on mount. 

Can we have the tab-effect skip the initial mount to avoid that?

alazar-aot (changes requested):
Noticed 2 small things, but looks super solid overall! Should be easy fixes

services/core-web/src/components/mine/Projects/DecisionPackageTab.tsx
matbusby-fw: Ah, good call

services/common/src/components/documents/ReplaceDocumentModal.tsx
matbusby-fw: Indeed!

alazar-aot (approved):
LGTM!

PR #4002, opened 2026-09-18
[MDS-7025] - Refine file update process

matbusby-fw:
## Objective 

[MDS-7025](https://bcmines.atlassian.net/browse/MDS-7025)

### Summary

Replacing a file under a Major Project's "All Documents" tab already linked the new file correctly (fixed in #3996), but the metadata shown for the file and its version history was wrong in several ways:

- The **Category** and **File Type** columns showed `N/A` for every archived (previous) version.
- The **Last Modified** date collapsed to the same "now" timestamp for every version, not just the newest one.
- The **Created By** column always showed the *original* uploader, even for the current file and even after multiple replaces.

This PR fixes all three, plus two regressions found while verifying the fix live against a running instance (a crash on a second replace, and a case where `update_user` still resolved to `N/A` for some document types).

### Root causes

There were three independent bugs, not one:

**1. `MineDocumentVersion` never carried category/file type.**
The archived-version model (backend and frontend) only ever stored `document_name`, `upload_date`, and audit fields. Category and file type were never part of a "version" — they belong to the document as a whole and don't change when a file is replaced (replace is already type-locked to the same extension). Rather than add new columns to store a value that never varies, `MineDocumentVersion` on the frontend now inherits `category`/`category_code` from its parent `MineDocument` and derives `file_type` from its own filename, the same way `MineDocument` does.

**2. The archived version's audit fields were stamped with "who is replacing now", not "who touched this file before".**
`MineDocumentVersion.create_from_docman_version` let `AuditMixin`'s column defaults stamp `create_user`/`create_timestamp` with the *current* user and "now" — i.e. the person doing the replace, not the person who last touched the file being archived. It now explicitly snapshots `mine_document`'s own `update_user`/`update_timestamp` (its state *before* this replace) onto the new archived row.

That snapshot has to happen before `mine_document` itself gets updated to the new file, so `initializeFileVersionUploadWithDocumentManager` no longer eagerly overwrites `mine_document.document_name` at upload time (before the user even confirms the replace) — that update now happens inside `create_from_docman_version`, after the archive snapshot is taken, keeping both operations atomic and correctly ordered.

**3. `update_user` never reached the frontend at all, for most document types.**
Even once (1) and (2) were fixed, `update_user` was still missing or `null` in API responses for anything living under a Major Project (project summary, major mine application, decision package, IRT documents). These all serialize through per-context "xref" wrapper models (`ProjectSummaryDocumentXref`, `MajorMineApplicationDocumentXref`, etc.) that proxy a hand-picked subset of `MineDocument`'s fields via SQLAlchemy `association_proxy`. Every one of them proxied `create_user` but had simply never had `update_user` added.  Added the missing proxy to all four xref models (one of them via the shared `DocumentXrefMixin`).

With `update_user` available end-to-end, the "Created By" column now reads `update_user` instead of the immutable `create_user`, for both the top-level document and its archived versions.


simensma-fresh (dismissed):

services/common/src/components/documents/DocumentTable.tsx
```diff
@@ -262,7 +262,7 @@ export const DocumentTable: FC<DocumentTableProps> = ({
       ...additionalColumns,
       renderTextColumn("file_type", "File Type", !isMinimalView),
       uploadDateColumn("update_timestamp", "Last Modified"),
-      uploadedByColumn("create_user", "Created By"),
+      uploadedByColumn("update_user", "Created By"),
```
alazar-aot: Just want to check - is this supposed to be `update_user` instead of `create_user`? The "Created By" will now show the person who lasted edited it, not created it

services/core-api/app/api/services/document_manager_service.py
```diff
@@ -280,6 +276,16 @@ def get_document_version(cls, request, document_manager_guid, document_manager_v
...
 
+    @classmethod
+    def get_document(cls, request, document_manager_guid):
+        resp = requests.get(
+            url=f'{cls.document_manager_document_resource_url}/{document_manager_guid}',
+            headers={key: value
+                     for (key, value) in request.headers if key != 'Host'},
+        )
+
+        return resp.json()
```
alazar-aot, line 295: This should probably check the response status before parsing, if this errors a failed docman lookup will leave `document_name` stale and commit/return a success, even though it failed

services/core-api/app/api/mines/documents/models/mine_document_version.py
```diff
@@ -48,9 +52,25 @@ def create_from_docman_version(
             mine_document_guid=mine_document.mine_document_guid,
             document_manager_version_guid=document_manager_version_guid,
             document_name=docman_version.get('file_display_name'),
+            upload_date=mine_document.upload_date,
```
alazar-aot: Should this have the same fallback pattern as lines 57-60 below? I think if something is archived/replaced more than once, the `upload_date` will drift/be wrong if this doesn't

alazar-aot (changes requested):
Left 2 questions and a small request, but this looks great otherwise!

services/common/src/components/documents/DocumentTable.tsx
matbusby-fw: The problem is that create_user never changes.  It is always the user who uploaded the initial document.  update_user is the field that actually represents the person who uploaded that version (which is what was asked for in the ticket).  But, looking at it now I think I'm going to create a new `uploaded_by` column.  Probably better than using a column that could change based on other user manipulation of the document (like archiving).

matbusby-fw: After looking a bit more, I actually opted to update the create_user and create_timestamp to be inline with the new version upload rather than holding on to the original values.  I think this makes more sense.  So I reverted back to the create_user in the table.

alazar-aot: Sounds good, I can get with that. Thanks!

alazar-aot (dismissed):

alazar-aot (approved):

